PR #1, opened 2016-10-05
storage: fix up docstring comments

cyphar:
In order to make actually reading the interfaces easier (as well as
making godoc actually useful), we need to comment the interface methods
themselves rather than the whole interface.

Signed-off-by: Aleksa Sarai <asarai@suse.de>

### DIFF
--- a/storage/containers.go
+++ b/storage/containers.go
@@ -17,57 +17,57 @@ var (
 )
 
 // A Container is a reference to a read-write layer with a metadata string.
-// ID is either one specified at create-time or a randomly-generated value.
-// Names is an optional set of user-defined convenience values.
-// ImageID is the ID of the image which was used to create the container.
-// LayerID is the ID of the read-write layer for the container itself.
-// It is assumed that the image's top layer is the parent of the container's
-// read-write layer.
 type Container struct {
-	ID           string                 `json:"id"`
-	Names        []string               `json:"names,omitempty"`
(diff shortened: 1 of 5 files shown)


nalind:
Looks good, merging.

cyphar:
Uhh d4ee4f6f488fa2f026e136a95205daec035ee02b isn't a merge, you took my commit then stripped my `Signed-off-by` and my authorship (while also changing the patch slightly). I understand that it's "just" a comment change, but that's not how you merge external contributors' commits -- you're erasing copyright holder information from the repository.

nalind:
Doh.  That's what I get for using git merge instead of the web UI.  Alright if I rewrite the history to try to restore it?

nalind:
Oh, wait, those are two commits.  6a6198a should be exactly what was proposed.

cyphar:
Ah, my apologies. The short message was the same so I didn't think to go through the history (since there wasn't a merge commit). :P

nalind:
No worries.  I'll try to use the web UI to do those in the future.

cyphar:
I actually don't like the Web UI either, but personally I always make sure to use `--no-ff` when using `git merge` because when you fast-forward master then you lose the merge history (this is what the Web UI does by default). If the top commit had been a merge commit I wouldn't have been confused. :P

nalind:
Ah, so that's what I missed.  Thanks for the tip!

PR #3, opened 2016-11-01
Do not merge: additional support routines for CRI use cases

nalind:
These change introduces a new package which aims to make it easier for consumers of the library that want to support CRI use cases to interoperate with each other, by sharing a common metadata format that we can encode and store in the storage library's more basic locations.

The support for holding and tracking container/pod status in runtime.go might not be used, and if they go, fields which only they touch can be dropped, too.

We also take the vendoring script and move things around to better match the newer Go style of doing vendoring, which should probably be moved to a different PR for merging, but currently lives here because we make a minor change on top of it.


philips:
Having a library vendor another library isn't a good practice in Go. It makes it hard for the end application to re-vendor everything. re: 9413337

nalind:
Hmm, splitting out or dropping the CLI wrapper would probably help with that.

philips:
@nalind got it, understood. Probably not a huge deal, didn't notice the CLI wrapper.

nalind:
Going to see if moving the logic changes into `cri-o` via https://github.com/kubernetes-incubator/cri-o/pull/210 makes more sense.

runcom:
@nalind isn't this surpassed by https://github.com/kubernetes-incubator/cri-o/pull/210?

nalind:
@runcom I think so.  It certainly avoids some of the interdependency problems with `containers/image` that adding these routines to this library would create.

nalind:
Since https://github.com/kubernetes-incubator/cri-o/pull/210 was merged, these routines don't belong here any more.  Closing without merging.

PR #6, opened 2016-12-06
Add truncindex to stores, to allow use of truncated IDs

nalind:
Add a truncindex to the layer, image, and container stores, so that layers, images, and containers can also be referred to by truncated versions of their IDs in any place where we previously accepted names.
    
For containers, we don't add a truncated index for referring to containers by layer ID, because there doesn't appear to be a minimum length in truncindex, and we don't want to mistakenly treat a truncated but not sufficiently-unique container ID as a layer ID, which would likely resolve to some other different container.

### DIFF
--- a/storage/layers.go
+++ b/storage/layers.go
@@ -311,30 +311,32 @@ func newLayerStore(rundir string, layerdir string, driver drivers.Driver) (Layer
 	return &rlstore, nil
 }
 
-func (r *layerStore) ClearFlag(id string, flag string) error {
-	if layer, ok := r.byname[id]; ok {
-		id = layer.ID
+func (r *layerStore) lookup(id string) (*Layer, bool) {
+	if layer, ok := r.byid[id]; ok {
+		return layer, ok
+	} else if layer, ok := r.byname[id]; ok {
+		return layer, ok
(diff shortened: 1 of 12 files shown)


storage/layers.go
```diff
@@ -467,6 +481,8 @@ func (r *layerStore) Unmount(id string) error {
 	}
 	if layer, ok := r.byname[id]; ok {
 		id = layer.ID
+	} else if longid, err := r.idindex.Get(id); err == nil {
```
runcom: can you have a function which fisrt check the `r.byname[id]` and the the truncindex to not repeat this else if over? skip this if you think it's not worth.

nalind: Might as well; it should make a dent in the lint warnings.

nalind:
Okay, updated.  I think it got rid of a couple of lint warnings, at least.

nalind:
Rebased on top of #24.

rhatdan:
@runcom PTAL  We need to get this in.

rhatdan:
:+1:

runcom:
LGTM

rhatdan:
@nalind Can we merge this now?

nalind:
Sure, merging.

PR #9, opened 2017-01-02
Use Mtimespec instead of Mtim to compile on darwin

mfojtik:
Migrate the `syscall` to `golang.org/x/sys/unix` (see: https://golang.org/pkg/syscall).
Also use `Mtimespec` instead of `Mtim` so this library compile on darwin.

### DIFF
--- a/storage/lockfile.go
+++ b/storage/lockfile.go
@@ -4,8 +4,9 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
-	"syscall"
 	"time"
+
+	"golang.org/x/sys/unix"
 
 	"github.com/containers/storage/pkg/stringid"
 )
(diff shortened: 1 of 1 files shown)


mfojtik:
@nalind PTAL

nalind:
LGTM, merging.

mfojtik:
@nalind you should not trust me :-) i forgot to try to run this on linux and it breaks for some reason... Can you please revert this?

runcom:
This seems to be why PRs in c/image are failing https://github.com/containers/image/pull/195

mfojtik:
@runcom for some reason this works on OSX but not on Linux :-/ I wonder if there a way to make this work on both, but for now I think we should revert this commit

nalind:
Can I get you to look over #11 and see if it works for you?

nalind:
Okay, reverted this one and readded it as the first patch in #11.

PR #11, opened 2017-01-03
Revendor golang.org/x/sys

nalind:
Revendor "golang.org/x/sys" to pick up "golang.org/x/sys/unix", which we now use instead of syscall.  Handle Linux and BSD using different names for the mtime part of a Stat_t.  Add "darwin/amd64" to the list of OS/arch combinations that we preserve dependencies for when we're stripping out unused things in hack/vendor.sh.

### DIFF
--- a/hack/make/cross
+++ b/hack/make/cross
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-CROSSPLATFORMS="linux/amd64 linux/386 linux/arm"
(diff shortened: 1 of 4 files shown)


runcom:
This should be fixing https://travis-ci.org/containers/image/builds/188613572#L344, thx

nalind:
Rebased on top of #24; this should fix https://github.com/projectatomic/skopeo/issues/310.

rhatdan:
LGTM

PR #12, opened 2017-01-04
Fix handing modification time of lock file

mfojtik:
@runcom @nalind this should be more robust and I tested in on both OSX and Linux


runcom:
Is https://github.com/containers/storage/pull/11 fixing the same issue? If so, I'd avoid using reflection here and go with @nalind's patch.

mfojtik:
@runcom I believe so, closing this.

Issue #13, opened 2017-01-08
failure to compile in docker


runcom:
vendoring c/storage in docker now gives:
```
2017/01/08 13:38:01 duplicate symbol _cgoexp_57a08a852b47_DevmapperLogCallback (types 1 and 1) in github.com/docker/docker/pkg/devicemapper and $WORK/github.com/docker/docker/vendor/github.com/containers/storage/pkg/devicemapper.a(_go_.o)
cp: cannot stat 'bundles/latest/dynbinary-daemon/dockerd': No such file or directory
```

runcom:
maybe we just need an `IFDEF` in the embedded C here in c/storage

PR #14, opened 2017-01-09
Rename "DevmapperLogCallback" to avoid conflicts

nalind:
Rename the "DevmapperLogCallback" exported-to-C function to "StorageDevmapperLogCallback", to avoid tripping up anyone who vendors the library but already has a copy of the "pkg/devicemapper" pkg which defines the callback with its previous name.


nalind:
This should fix #13.

runcom:
we do not `IFDEF` from upstream docker because they (or we) may change it right? Otherwise, I'll try to test this out and see if it works. Thanks!

PR #17, opened 2017-01-12
Remove "docker" strings from container storage

rhatdan:
We want to make sure that content created out of container storage
does not refer to docker.

### DIFF
--- a/drivers/aufs/aufs.go
+++ b/drivers/aufs/aufs.go
@@ -279,7 +279,7 @@ func (a *Driver) Remove(id string) error {
 	}
 
 	// Atomically remove each directory in turn by first moving it out of the
-	// way (so that docker doesn't find it anymore) before doing removal of
+	// way (so that container runtimes don't find it anymore) before doing removal of
 	// the whole tree.
 	tmpMntPath := path.Join(a.mntPath(), fmt.Sprintf("%s-removing", id))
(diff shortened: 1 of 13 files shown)


nalind:
@rhatdan Looks like a good start - can you also update what we pass to `ioutil.TempDir`, `reexec.Register` and `reexec.Command` under the `pkg` subdirectory?

rhatdan:
I was not sure what those were doing.  Do they actually execute commands?

nalind:
They're used for setting argv[0] when the calling process runs a second copy of itself as a subprocess, in this case so that it can be chrooted, to extract a layer onto disk.

rhatdan:
@nalind PTAL

nalind:
I think the `pkg/archive` and `pkg/chrootarchive` package still need a look, otherwise this looks complete.

rhatdan:
I did not do any of the pkg packages since I thought they came from elsewhere.  I will take a quick look.

nalind:
They're not vendored from elsewhere, so they "live" here as much as anything else.  They're isolated enough from the rest of the code that they could conceivably be spun off into their own repositories, but I expect that's more about maintainability than any actual intent to spin them off.

rhatdan:
I modified the code that could be seen outside of test code.  I could change all of the references to the test code, if you want, but that would cause more hassle when trying to back port fixes.

nalind:
Looks good, merging.

PR #18, opened 2017-01-18
Add initial Travis configuration

nalind:
Add an initial Travis configuration file, and whatever fixups and tweaks we need to build in it.  This doesn't enable every test (`make lint` would fail, and integration tests are missing), but it's a start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: go
+go:
+    - tip
+    - 1.7
+    - 1.6
+dist: trusty
+sudo: required
+before_install:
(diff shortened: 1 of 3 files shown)


runcom:
looks fine, just weird that travis didn't kick in though, should it now that there's a .travis.yml file?

nalind:
@runcom chicken-egg problem: until we merged this, we didn't have a .travis.yml file.

runcom:
yeah, but I thought since this PR has a travis.yml then travis would have kicked in :confused: fair enough though :)

PR #19, opened 2017-01-25
Switch to GO15VENDOREXPERIMENT=1 vendoring

nalind:
Moved vendored packages from vendor/src to vendor, which should reduce conflicts with other packages when we're vendored.


nalind:
This is done better in #24, closing.

PR #22, opened 2017-01-27
Unbreak containers without images

nalind:
Make it possible to create a container without using an underlying image.

### DIFF
--- a/storage/store.go
+++ b/storage/store.go
@@ -749,19 +749,25 @@ func (s *store) CreateContainer(id string, names []string, image, layer, metadat
 		id = stringid.GenerateRandomID()
 	}
 
-	cimage, err := ristore.Get(image)
-	if err != nil {
-		return nil, err
+	imageTopLayer := ""
+	imageID := ""
+	if image != "" {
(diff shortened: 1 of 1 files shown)


storage/store.go
```diff
@@ -1369,7 +1375,9 @@ func (s *store) DeleteImage(id string, commit bool) (layers []string, err error)
 		}
 		aContainerByImage := make(map[string]string)
 		for _, container := range containers {
-			aContainerByImage[container.ImageID] = container.ID
+			if container.ImageID != "" {
+				aContainerByImage[container.ImageID] = container.ID
```
rhatdan: Is it ok if this is not set?  Or should it cause an error?

rhatdan: Otherwise LGTM

nalind: It should be fine if it's not set, but now that you mention it, this doesn't seem necessary to make this case work.  I'll drop it.

rhatdan:
:+1:

nalind:
Thanks, merging.

PR #23, opened 2017-01-28
Prefer overlay2/overlay over devicemapper

nalind:
It looks like overlay is going to be generally preferred over devicemapper once its SELinux support is figured out, so alter our preferences list for Linux to prefer it.

### DIFF
--- a/drivers/driver_linux.go
+++ b/drivers/driver_linux.go
@@ -56,8 +56,9 @@ var (
 		"aufs",
 		"btrfs",
 		"zfs",
-		"devicemapper",
+		"overlay2",
 		"overlay",
+		"devicemapper",
 		"vfs",
 	}
 


runcom:
:+1: @mrunalp @rhatdan are you OK with this?

nalind:
CC @runcom

rhatdan:
LGTM.

We need to get container-storage-selinux setup so this is easy to modify, but defaulting to Overlay2 is best.

jeremyeder:
Is there a way to test if the kernel is recent enough to have selinux enabled with overlay? If it's not recent enough, should the default still be changed?

mrunalp:
Sounds good. Can we check for SELinux support in overlay and adjust preference?

rhatdan:
I think we should do this in container-storage-setup rather then adding this to the code.

@rhvgoyal WDYT?

rhvgoyal:
@rhatdan, having some defaults for storage library is good anyway. (For the use case where users are using it without container-storage-setup).

And when container-storage-setup is in use, it will override the defaults anyway (like in case of docker).

container-storage-setup will continue to default to devicemapper for some time till we have gained confidence that most of the users are fine with overlay2. And fedora will enabled overlay2 by default based on configuration files. And these distros are carrying new kernels which support selinux anyway so there is no need to determine this dynamically yet.

IOW, I think container/storage should continue with its logic of whatever default they want to set. And determining on the fly whether selinux is supported with overlay2 or not makes sense to me.

BTW, I have a pending PR open with docker which determines if selinux is supported with overlay2 or not and warn. May be some of that code can be reused.

https://github.com/docker/docker/pull/27071

rhatdan:
This pull request has been replaced.  Should be closed.
https://github.com/containers/storage/pull/34

PR #24, opened 2017-01-30
Fix a number of testing & dependency issues

erikh:
Please see the individual commits. Until this was done, I was unable to test
locally.

The virtualbox flip in particular is important if you have the kvm module
running but no qemu or libvirt vagrant provider, which doesn't install on
modern vagrant without a rubygems error. So, I think it's better to keep the
safe, easy-to-install version preferred, especially because vagrant was made to
run virtualbox.

I also incorporated github.com/LK4D4/vndr to manage dependencies.

I think it might be a good idea to setup a monorepo for all the `containers/`
projects as a submodule to `vendor/`. This will help quite a bit with conflicts
I think. cc @runcom @mtrmac

The integration-cli target doesn't exist because the dir doesn't exist. I don't
know if that's an intentional thing or not, but as an open source project it's
pretty much impossible to contribute from the outside without this information.

go as installed from package (at least on fedora) does not work to compile
golint and a few other things, so I decided it was best to get official 1.7.4
in there -- it'll also help if the debian/fedora packages mismatch.

Again, I couldn't test at all until these patches were made. All of them except
the .gitignore changes were required.


nalind:
I love just about everything in here.
I'm a bit worried about bringing our own version of the compiler into VMs for running the tests, since this package gets vendored into other projects where the compiler version is not something we dictate. Fedora 23 had 1.5.4, while Fedora 25, if we tweaked the Vagrantfile to switch to it, and Debian testing both get 1.7.5.  Travis tests with multiple versions, but none of those are the packaged versions of the compiler, either, so we wouldn't be testing with the OS-provided compiler anywhere.

.travis.yml
```diff
@@ -9,10 +9,10 @@ before_install:
...
 script:
-    - AUTO_GOPATH=1 make install.tools
-    - AUTO_GOPATH=1 ./hack/make.sh validate-gofmt validate-pkg validate-lint validate-test validate-toml validate-vet validate-vendor
-    - AUTO_GOPATH=1 make .gitvalidation
-    - AUTO_GOPATH=1 make build-binary
-    - AUTO_GOPATH=1 ./hack/make.sh cross
-    - sudo env AUTO_GOPATH=1 PATH="$PATH" ./hack/make.sh test-unit
-    - AUTO_GOPATH=1 make docs
+    - make install.tools
+    - ./hack/make.sh validate-gofmt validate-pkg validate-lint validate-test validate-toml validate-vet validate-vendor
```
nalind: I think Travis may only be failing because this patch removes the 'validate-vendor' script, but leaves it in this list.

erikh:
the unfortunate problem with jumping between pre-1.7 and 1.7 is the context package. It typically works with one or the other, and most of the libs have already ported to the 1.7 context package (as opposed to the golang.org third-party one)

erikh:
I've refreshed the branch and will monitor it to ensure tests pass.

.travis.yml
```diff
@@ -9,10 +9,10 @@ before_install:
...
-    - AUTO_GOPATH=1 make build-binary
-    - AUTO_GOPATH=1 ./hack/make.sh cross
-    - sudo env AUTO_GOPATH=1 PATH="$PATH" ./hack/make.sh test-unit
-    - AUTO_GOPATH=1 make docs
+    - make install.tools
+    - ./hack/make.sh validate-gofmt validate-pkg validate-lint validate-test validate-toml validate-vet
+    - make .gitvalidation
+    - make build-binary
+    - ./hack/make.sh cross
+    - sudo ./hack/make.sh test-unit
```
nalind: Apparently the go compiler's not in the default $PATH in Travis.

erikh:
some pathing issue; I'll have to deal with this tomorrow, sorry

erikh:
This should be passing now; the fundamental issue was some GOPATH muckery I missed and travis's sudo config has `env_keep PATH` set I think, so the PATH needs to be injected directly.

All in the last commit. PTAL.

erikh:
Sorry to press but the lint branch is getting pretty expansive, I was hoping we could discuss or finalize this somehow.

nalind:
I'm still going to need to get builds going on non-1.7 versions of go, but that's not a good reason to hold this up, as the tests pass i CI and on my box.

erikh:
Ok, I will put in a ticket for that and try to address it this week.

nalind:
That should be sorted by #30, assuming nothing in there is too outlandish.

PR #25, opened 2017-02-16
Remove pkg/ioutils.cancelReadCloser

nalind:
Remove pkg/ioutils's cancelReadCloser type and the vendored copy of golang.org/x/net/context which it needed, neither of which was used anywhere else in here.  This should let us avoid having to deal with potential incompatibilities with the standard library's context package, which was introduced in golang 1.7.


erikh:
I hope it's ok if I jump in here. I don't think this is going to resolve any conflicts here. The library should work just fine on 1.7. It's that 1.6 and below will be unable to process the `"context"` import that will be in many libraries.

nalind:
Of course it's ok, and I appreciate the info.  The `cancelReadCloser` implementation was importing `golang.org/x/net/context`, so I thought any incompatibility would be with code that might try to pass in a `context.Context` from 1.7 where a `golang.org/x/net/context.Context` was expected, or vice-versa.  I must be missing something here.

erikh:
that would happen, yes. Sorry I guess I misunderstood.

nalind:
No worries.

PR #26, opened 2017-02-27
Rewrite build tags so they are no longer required

erikh:
This rewrites btrfs_noversion to btrfs_version and
libdm_no_deferred_remove to libdm_deferred_remove. This allows many
systems to compile without build tags; the current situation is that
most hosts require the build tags and very few do not.

This eliminates the dependency in containers/image and my own project (erikh/box). These tools should be able to `go build` now without specifying tags and get no compile errors on platforms that have the dependencies installed, even if they are outdated.

### DIFF
--- a/drivers/btrfs/version.go
+++ b/drivers/btrfs/version.go
@@ -1,4 +1,4 @@
-// +build linux,!btrfs_noversion
+// +build linux,btrfs_version
 
 package btrfs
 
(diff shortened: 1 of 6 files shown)


erikh:
Note this currently incorporates #24 to get tests to pass, so please resolve that ticket first presuming this is a worthwhile patch to you all. :)

erikh:
rebased

rhatdan:
@nalind isn't this what we talked about inverting the build tags?

vrothberg:
This issue looks a bit dusty. Is this still an issue or can we close it, @nalind, @rhatdan, @erikh?

PR #31, opened 2017-03-16
Modify default directories for storage to /var/lib/containers/storage

rhatdan:
Also modify work directory to /var/run/containers/storage to be consistent.
/var/lib/containers/atomic is currently being used by other tools including the
atomic cli for storing OSTree based containers.  Running tools like CRI-O
ends up dropping lots of files in /var/lib/containers directory causing an
admin to get perhaps confused.  Since this tool is containers/storage, I think
it is best we store it in this subdir, and cleaner for administrators.

### DIFF
--- a/storage/store.go
+++ b/storage/store.go
@@ -2171,8 +2171,8 @@ func makeBigDataBaseName(key string) string {
 }
 
 func init() {
-	DefaultStoreOptions.RunRoot = "/var/run/containers"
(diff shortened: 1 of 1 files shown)


rhatdan:
This pull request hinges on 
https://github.com/containers/storage/pull/31

nalind:
LGTM.

PR #32, opened 2017-03-17
Ensure directories are there before saving stores

nalind:
Ensure that the containing directories are there before we call `AtomicWriteFile` to save the lists of containers, images, or layers, each time we go to do so.  Samuel Ortiz noticed that we weren't making sure of this everywhere, and that this can trip up long-running processes.


rhatdan:
:+1:

mrunalp:
LGTM

nalind:
Thanks!  Merging.

PR #34, opened 2017-03-27
We need to remove overlay as an option in containers/storage

rhatdan:
Since we prefer our users to use overlay2, there is not reason to
support and provide overlay as an option, and potentially use a worse
driver by accident.

We also need to make overlay2 the prefered driver in containers/storage.


rhatdan:
@rhvgoyal @nalind PTAL

rhatdan:
https://bugzilla.redhat.com/show_bug.cgi?id=1435767

rhatdan:
@jeremyeder PTAL

drivers/driver_linux.go
```diff
@@ -53,11 +53,11 @@ const (
 var (
 	// Slice of drivers that should be used in an order
 	priority = []string{
+		"overlay",
```
jeremyeder: This means we default to overlay in containers/storage, and overlay2-behavior is now default. 

Do you foresee any possibility to misunderstand this, and people ask "where did overlay2 go in containers/storage?"

Should we have a similar proposal for docker in that regard?

rhatdan: There was no overlay2 in the list, and I am no expert on containers/storage.  My goal with this change is

nalind: This should be "overlay2", I think.

drivers/register/register_overlay.go
```diff
@@ -1,9 +0,0 @@
-// +build !exclude_graphdriver_overlay,linux
-
-package register
-
-import (
-	// register the overlay graphdriver
-	_ "github.com/containers/storage/drivers/overlay"
```
nalind: We need to drop this line, but not the rest of this file.

nalind:
LGTM.

rhatdan:
@nalind I can't merge.

PR #35, opened 2017-04-19
Fix {container,image}Store.Lookup()

nalind:
Fix the Lookup methods in container and image stores, which broke due to a bug in #6.


rhatdan:
LGTM

nalind:
Thanks, merging.

PR #36, opened 2017-04-19
Fix canonicalization of layer IDs

nalind:
Fix canonicalization of layer IDs when computing layer differences, which we also broke in #6.

### DIFF
--- a/storage/layers.go
+++ b/storage/layers.go
@@ -619,19 +619,35 @@ func (r *layerStore) Wipe() error {
 	return nil
 }
 
-func (r *layerStore) Changes(from, to string) ([]archive.Change, error) {
-	toLayer, ok := r.lookup(to)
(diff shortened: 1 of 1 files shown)


storage/layers.go
```diff
@@ -624,14 +624,21 @@ func (r *layerStore) Changes(from, to string) ([]archive.Change, error) {
 	if !ok {
 		return nil, ErrLayerUnknown
 	}
-	fromLayer, ok := r.lookup(from)
-	if !ok {
-		fromLayer, ok = r.lookup(toLayer.Parent)
+	to = toLayer.ID
```
rhatdan: It looks like this same block of code is copied three times, shouldn't we just add a new function to handle it?

nalind: Yeah, that's a better idea.  Fixing that.

nalind:
Okay, this gets buildah's self-tests working again.

rhatdan:
LGTM

Much better

nalind:
Okay, merging it.

PR #38, opened 2017-04-21
[WIP] Allow storage driver to use a r/o and r/w root directory

rhatdan:
Currently this is just for use with overlay.  We want to get to the point where we could
have multiple drivers for overlay, one for local storage and potentially multiple for
network storage.

The idead would be the read/writeable content would always be set on local storage while
images could also be pulled from network storage.


rhatdan:
@nalind @rhvgoyal PTAL

rhatdan:
After talking to Nalin, I realize I was going the wrong way.

PR #39, opened 2017-04-24
Remove Get prefix on method calls.

rhatdan:
Go validate suggests that you should not use functions that begin with Get,
unless you have to.


rhatdan:
@nalind PTAL, I would like to get this one so it becomes easier to review the others.

nalind:
Looks alright.  While we're making changes to the API that consumers of the library see, I'd like to go ahead and relocate the storage package up a level so that people can stop having to import ".../storage/storage".

rhatdan:
@nalind storage/storage is now just storage.
Sort of like pizza/pizza

nalind:
LGTM pending CI.

PR #40, opened 2017-04-24
Rename 'overlay2' to 'overlay'

nalind:
Finish the overlay->overlay2 changeover by:
* renaming overlay2 to overlay
  * teaching the driver to accept the overlay.override_kernel_check and
    overlay2.override_kernel_check options as having the same meaning
* registering the driver init function using both names

### DIFF
--- a/drivers/overlay/mount.go
+++ b/drivers/overlay/mount.go
@@ -1,6 +1,6 @@
 // +build linux
 
-package overlay2
+package overlay
 
 import (
 	"bytes"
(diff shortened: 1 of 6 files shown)


runcom:
:+1:

nalind:
Added a bit of logic to have driver.String() return the name that was used to look up its init function.

nalind:
I think this is where we're breaking https://github.com/kubernetes-incubator/cri-o/pull/450 ATM.

rhatdan:
👍

PR #41, opened 2017-04-24
[WIP] Handling Multiple storage backends.

rhatdan:


rhatdan:
@nalind @rhvgoyal 
I just threw out my first pass at handling multiple graphdrivers at once.  Totally untested. Just want to know if this is crazy or not.

rhatdan:
Doing a simplified redesign based on @nalind comments.

PR #42, opened 2017-04-25
Support additional image stores 

rhatdan:
We want to support additional read/only image stores available on
file systems.  Usually these images stores would be on network shares.
Currently the only driver that will support additional images is the
overlay file system.


rhatdan:
@nalind Second pass.

nalind:
We've got separate PRs for these two things now, so this one's redundant.  Closing.

PR #43, opened 2017-04-28
storage: add comments on the defaults

vbatts:
And how they are detected. A bit of these environment variable is a bit
magic and may need to change a bit, such that if the caller only
provides a driver name, but not a path, etc. that the right thing is
done.

But for now, at least document that the DefaultOptions will be used, and
there is another way to influence what driver is chosen.

### DIFF
--- a/store.go
+++ b/store.go
@@ -404,6 +404,12 @@ type store struct {
 // GetStore attempts to find an already-created Store object matching the
 // specified location and graph driver, and if it can't, it creates and
 // initializes a new Store object, and the underlying storage that it controls.
+//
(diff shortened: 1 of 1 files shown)


nalind:
[[spike] Kick off containers/storage redesign](https://trello.com/c/856BXeFS/459-spike-kick-off-containers-storage-redesign)

vbatts:
rebased

vbatts:
the can be reviewed and merged, right?

rhatdan:
LGTM

vbatts:
does this need anything else?

vbatts:
...

rhatdan:
I will merge it.  Certainly not going to break anything.

PR #45, opened 2017-05-12
Move locks to /run/storage/locks

rhatdan:
This could increase spead, and should allow us to have read only image
and layer stores

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -170,7 +170,7 @@ func newContainerStore(dir string) (ContainerStore, error) {
 	if err := os.MkdirAll(dir, 0700); err != nil {
 		return nil, err
 	}
-	lockfile, err := GetLockfile(filepath.Join(dir, "containers.lock"))
+	lockfile, err := GetLockfile(filepath.Join(DefaultStoreOptions.LockDir, dir, "containers.lock"))
 	if err != nil {
 		return nil, err
 	}
--- a/images.go
+++ b/images.go
@@ -161,7 +161,7 @@ func newImageStore(dir string) (ImageStore, error) {
 	if err := os.MkdirAll(dir, 0700); err != nil {
 		return nil, err
 	}
-	lockfile, err := GetLockfile(filepath.Join(dir, "images.lock"))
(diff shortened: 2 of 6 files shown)


rhatdan:
@nalind PTAL

rhatdan:
@rhvgoyal PTAL

nalind:
This creates breakage if a shared location is going to be used for read-write from more than one location.

storage/containers.go
```diff
@@ -170,7 +170,7 @@ func newContainerStore(dir string) (ContainerStore, error) {
 	if err := os.MkdirAll(dir, 0700); err != nil {
 		return nil, err
 	}
-	lockfile, err := GetLockfile(filepath.Join(dir, "containers.lock"))
+	lockfile, err := GetLockfile(filepath.Join(lockDir, dir, "containers.lock"))
```
nalind: `dir` is already an absolute path, so this creates a lock file named `/run/storage/locks/var/lib/containers/storage/containers/containers.lock`.  It looks weird, but it's consistent and unique among different graph roots, so I guess it's fine.

rhatdan:
> This creates breakage if a shared location is going to be used for read-write from more than one location.

Please explain?

rhvgoyal:
So you are moving all locks to a tmpfs (/run) ?

rhatdan:
Yes the main reason for this is so that alternate container/storage repos can be on read/only file systems. (NFS for example).  I would figure the locking would also be quicker on a tmpfs then on physical disk, especially if the disk is remote.  Only issue I am aware of would be locking form other systems, but I don't think we could do a good job with that anyways. 
Potentially someone could remove some nfs data while a tool on a host is using it, but this could happen in thousands of different ways.  We are only trying to handle local locking.  We need to tell users that they would need to be careful with updating container stores if they are in use.

rhatdan:
@nalind I now made this configurable so if you want to have locks on the nfs shares it will work.

rhatdan:
If we merge https://github.com/containers/storage/pull/52 we can close this one.

PR #47, opened 2017-05-17
Remove layer,image, container code that is not used

rhatdan:
Lots of cut and pasted code that ends up doing nothing, but takes time
to process.

### DIFF
--- a/store.go
+++ b/store.go
@@ -613,10 +613,6 @@ func (s *store) PutLayer(id, parent string, names []string, mountLabel string, w
 	if err != nil {
 		return nil, -1, err
 	}
-	ristore, err := s.ImageStore()
-	if err != nil {
-		return nil, -1, err
-	}
 	rcstore, err := s.ContainerStore()
 	if err != nil {
 		return nil, -1, err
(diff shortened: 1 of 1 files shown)


store.go
```diff
@@ -692,12 +677,6 @@ func (s *store) CreateImage(id string, names []string, layer, metadata string, o
 	if modified, err := ristore.Modified(); modified || err != nil {
 		ristore.Load()
 	}
-	rcstore.Lock()
-	defer rcstore.Unlock()
-	defer rcstore.Touch()
```
TomSweeneyRedHat: Ditto slight Touch removal concern here too.

TomSweeneyRedHat:
LGTM, but would like to get someone with more familiarity to eyeball, especially the touch removals.  Overall a nice cleanup.

store.go
```diff
@@ -628,12 +624,6 @@ func (s *store) PutLayer(id, parent string, names []string, mountLabel string, w
 	if modified, err := rlstore.Modified(); modified || err != nil {
 		rlstore.Load()
 	}
-	ristore.Lock()
-	defer ristore.Unlock()
-	defer ristore.Touch()
```
TomSweeneyRedHat: I think this removal is OK too, but don't know for sure.  I'm just not sure if the Touch should be removed here.

rhatdan:
@nalind PTAL

store.go
nalind: We're okay not calling Touch() if the set of data that the store object keeps isn't being changed, which appears to be true here.

nalind:
LGTM.

PR #48, opened 2017-05-19
Fix container storage to support additional image stores

rhatdan:
We want to support additional read/only image stores available on
file systems.  Usually these images stores would be on network shares.
Currently the only driver that will support additional images is the
overlay file system.

### DIFF
--- a/drivers/aufs/aufs.go
+++ b/drivers/aufs/aufs.go
@@ -372,6 +372,12 @@ func (a *Driver) Diff(id, parent string) (archive.Archive, error) {
 	})
 }
 
+// AdditionalImageStores returns additional image stores supported by the driver
+func (a *Driver) AdditionalImageStores() []string {
+	var imageStores []string
+	return imageStores
+}
+
 type fileGetNilCloser struct {
 	storage.FileGetter
 }
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"syscall"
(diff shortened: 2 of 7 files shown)


rhatdan:
@nalind PTAL

TomSweeneyRedHat:
Unless I'm mis-reading the error, it looks like validate-lint wants you to use (d *Driver) instead of (a *Driver)?  If that's true and someone could learn me why that's an issue, I'd appreciate it.
Otherwise LGTM, but I'd like @nalind to eyeball it too.

rhatdan:
@TomSweeneyRedHat Actually those were correct.  I cut and pasted the same code from aufs driver which used a *Driver into the others which all used d *Driver.  Anyways that is fixed.

drivers/aufs/aufs.go
```diff
@@ -372,6 +372,12 @@ func (a *Driver) Diff(id, parent string) (archive.Archive, error) {
 	})
 }
 
+// AdditionalImageStores returns additional image stores supported by the driver
+func (a *Driver) AdditionalImageStores() []string {
```
TomSweeneyRedHat: Should or does this one need to change to d *Driver too?

drivers/overlay/overlay.go
```diff
@@ -186,6 +189,15 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
 			}
+		case "overlay.imagestore":
+			// Additional read only image stores to use for lower paths
+			for _, store := range strings.Split(val, ",") {
+
+				if _, err := os.Stat(store); err != nil {
+					return nil, fmt.Errorf("overlay: Can't stat imageStore dir %s: %v", store, err)
+				}
+				o.imageStores = append(o.imageStores, store)
+			}
```
nalind: I think these should be run through `path/filepath/Clean()`, and there should be a check for cases where `path/filepath/IsAbs()` returns `false`, because we may have to retrieve these values later, and there's no guarantee that we'll still be in the same working directory when we do.

May also want to verify that the `Stat_t` returned thinks that the path points to a directory.

rhatdan:
@nalind PTAL

nalind:
LGTM

PR #49, opened 2017-05-19
Add support for a toml config file

rhatdan:
Container Storage is going to be shared by multiple tools.  Therefore
we don't want each one of these tools setting up its own container
storage.  This storage config can be shared between multiple users.

### DIFF
--- a/storage.conf
+++ b/storage.conf
@@ -1,0 +1,17 @@
+# The "container storage" table contains all of the server options.
+[storage]
+
+# Default Storage Driver
+driver = "overlay"
+
+# Temporary storage location
+runroot = "/var/run/containers/storage"
+
+# Primary Read/Write location of container storage
+graphroot = "/var/lib/containers/storage"
+
+[storage.options]
+# AdditionalImageStores is used to pass paths to additional Read/Only image stores
+# Must be comma separated list.
+additionalimagestores = [
+]
--- a/store.go
+++ b/store.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"encoding/base64"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -14,6 +15,7 @@ import (
 	// register all of the built-in drivers
 	_ "github.com/containers/storage/drivers/register"
 
+	"github.com/BurntSushi/toml"
 	drivers "github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/idtools"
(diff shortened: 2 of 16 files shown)


rhatdan:
@nalind Tested with buildah, seems to work well.

store.go
```diff
@@ -1958,11 +1960,64 @@ func stringSliceWithoutValue(slice []string, value string) []string {
...
+		DefaultStoreOptions.GraphDriverName = config.Storage.Driver
+	}
+	if config.Storage.RunRoot != "" {
+		DefaultStoreOptions.RunRoot = config.Storage.RunRoot
+	}
+	if config.Storage.GraphRoot != "" {
+		DefaultStoreOptions.GraphRoot = config.Storage.GraphRoot
+	}
+	for _, s := range config.Storage.Options.AdditionalImageStores {
+		DefaultStoreOptions.GraphDriverOptions = append(DefaultStoreOptions.GraphDriverOptions, fmt.Sprintf("%s.imageStore=%s", config.Storage.Driver, s))
```
nalind: This uses different capitalization than what #48 gave to that option.

nalind:
The case-folding behavior of the parser when assigning values to struct members means that we probably don't need any of the toml-specific tagging of structure types, but that's not a blocker.  Except for the capitalization of the `imagestore`/`imageStore` option that it synthesizes, this looks good to me.

rhatdan:
Ok, I changed all of them to be lower case in the config file, and added a man page.

storage.conf
```diff
@@ -0,0 +1,17 @@
+# The "crio" table contains all of the server options.
+[storage]
+
+# Default Storage Driver
+driver = "overlay"
+
+# Temporary storage location
+runroot = "/var/run/containers/storage"
+
+# Primary Read/Write location of containre storage
```
nalind: Typo: "containre"

storage.conf
```diff
@@ -0,0 +1,17 @@
+# The "crio" table contains all of the server options.
```
nalind: "crio" should be "storage"?

store.go
```diff
@@ -1958,11 +1960,64 @@ func stringSliceWithoutValue(slice []string, value string) []string {
...
+}
+
+// TOML-friendly explicit tables used for conversions.
+type tomlConfig struct {
+	Storage struct {
+		Driver    string                  `toml:"driver"`
+		RunRoot   string                  `toml:"runroot"`
+		GraphRoot string                  `toml:"graphroot"`
+		Options   struct{ OptionsConfig } `toml:"options"`
+	} `toml:"Storage"`
```
nalind: Was this meant to be changed to lower-case as well?

storage.conf
rhatdan: Fixed

storage.conf
rhatdan: Fixed

store.go
```diff
@@ -1958,11 +1960,64 @@ func stringSliceWithoutValue(slice []string, value string) []string {
 	return modified
 }
 
+const configFile = "/etc/containers/storage.conf"
+
+// OptionsConfig represents the "crio.api" TOML config table.
```
nalind: "crio" -> "storage.options"?

TomSweeneyRedHat:
LGTM, but I'm not real hip at all with TOML

store.go
rhatdan: Fixed,  missed that one.  Lots of cut and paste errors.

nalind:
LGTM, merging.

PR #50, opened 2017-05-24
Support for additional roLayers and roImages

rhatdan:
We are now allowing users to specify additional container storage directories in the
configuration.  We need to create a roLayer and roImage object for each additional
storage driver.  These stores are treated as reaonly.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -377,8 +377,18 @@ func (d *Driver) getLower(parent string) (string, error) {
 	return strings.Join(lowers, ":"), nil
 }
 
-func (d *Driver) dir(id string) string {
-	return path.Join(d.home, id)
+func (d *Driver) dir(val string) string {
+	newpath := path.Join(d.home, val)
+	if _, err := os.Stat(newpath); err != nil {
+		for _, p := range d.AdditionalImageStores() {
+			l := path.Join(p, d.name, val)
+			_, err = os.Stat(l)
+			if err == nil {
+				return l
+			}
+		}
+	}
+	return newpath
 }
 
 func (d *Driver) getLowerDirs(id string) ([]string, error) {
@@ -386,11 +396,12 @@ func (d *Driver) getLowerDirs(id string) ([]string, error) {
 	lowers, err := ioutil.ReadFile(path.Join(d.dir(id), lowerFile))
 	if err == nil {
 		for _, s := range strings.Split(string(lowers), ":") {
-			lp, err := os.Readlink(path.Join(d.home, s))
+			lower := d.dir(s)
+			lp, err := os.Readlink(lower)
 			if err != nil {
 				return nil, err
 			}
-			lowersArray = append(lowersArray, path.Clean(path.Join(d.home, "link", lp)))
+			lowersArray = append(lowersArray, path.Clean(d.dir(path.Join("link", lp))))
 		}
 	} else if !os.IsNotExist(err) {
 		return nil, err
@@ -431,6 +442,31 @@ func (d *Driver) Get(id string, mountLabel string) (s string, err error) {
 		return "", err
 	}
 
+	newlowers := ""
+	for _, l := range strings.Split(string(lowers), ":") {
+		lower := ""
+		newpath := path.Join(d.home, l)
+		if _, err := os.Stat(newpath); err != nil {
+			for _, p := range d.AdditionalImageStores() {
+				lower = path.Join(p, d.name, l)
+				if _, err2 := os.Stat(lower); err2 == nil {
+					break
+				}
+				lower = ""
+			}
+			if lower == "" {
+				return "", fmt.Errorf("Can't stat lower layer %q: %v", newpath, err)
+			}
+		} else {
+			lower = l
+		}
+		if newlowers == "" {
+			newlowers = lower
+		} else {
+			newlowers = newlowers + ":" + lower
+		}
+	}
+
 	mergedDir := path.Join(dir, "merged")
 	if count := d.ctr.Increment(mergedDir); count > 1 {
 		return mergedDir, nil
@@ -444,7 +480,7 @@ func (d *Driver) Get(id string, mountLabel string) (s string, err error) {
 	}()
 
 	workDir := path.Join(dir, "work")
-	opts := fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s", string(lowers), path.Join(id, "diff"), path.Join(id, "work"))
+	opts := fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s", newlowers, path.Join(id, "diff"), path.Join(id, "work"))
 	mountLabel = label.FormatMountLabel(opts, mountLabel)
 	if len(mountLabel) > syscall.Getpagesize() {
 		return "", fmt.Errorf("cannot mount layer, mount label too large %d", len(mountLabel))
--- a/store.go
+++ b/store.go
@@ -148,13 +148,21 @@ type Store interface {
 	// by the Store.
 	GraphDriver() (drivers.Driver, error)
 
-	// LayerStore obtains and returns a handle to the layer store object used by
+	// LayerStore obtains and returns a handle to the writeable layer store object used by
 	// the Store.
 	LayerStore() (LayerStore, error)
 
-	// ImageStore obtains and returns a handle to the image store object used by
+	// ROLayerStore obtains additional read/only layer store objects used by
+	// the Store.
+	ROLayerStores() ([]LayerStore, error)
+
+	// ImageStore obtains and returns a handle to the writable image store object used by
 	// the Store.
 	ImageStore() (ImageStore, error)
+
+	// ROImageStores obtains additional read/only image store objects used by
+	// the Store.
+	ROImageStores() ([]ImageStore, error)
 
 	// ContainerStore obtains and returns a handle to the container store object
 	// used by the Store.
@@ -397,7 +405,9 @@ type store struct {
 	gidMap          []idtools.IDMap
 	graphDriver     drivers.Driver
 	layerStore      LayerStore
+	roLayerStores   []LayerStore
 	imageStore      ImageStore
+	roImageStores   []ImageStore
 	containerStore  ContainerStore
 }
 
@@ -513,6 +523,9 @@ func (s *store) load() error {
 		return err
 	}
 	s.layerStore = rls
+	if _, err := s.ROLayerStores(); err != nil {
+		return err
+	}
 
 	gipath := filepath.Join(s.graphRoot, driverPrefix+"images")
 	if err := os.MkdirAll(gipath, 0700); err != nil {
@@ -523,6 +536,10 @@ func (s *store) load() error {
 		return err
 	}
 	s.imageStore = ris
+	if _, err := s.ROImageStores(); err != nil {
+		return err
+	}
+
 	gcpath := filepath.Join(s.graphRoot, driverPrefix+"containers")
 	if err := os.MkdirAll(gcpath, 0700); err != nil {
 		return err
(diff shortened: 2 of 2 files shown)


rhatdan:
@nalind @rhvgoyal PTAL
This is the patch that allows container/storage to support multiple nfs based images.

layers.go
```diff
@@ -527,7 +527,11 @@ func (r *layerStore) SetMetadata(id, metadata string) error {
 }
 
 func (r *layerStore) tspath(id string) string {
-	return filepath.Join(r.layerdir, id+tarSplitSuffix)
+	tmpPath := ""
+	if DefaultStoreOptions.GraphRoot == "" {
+		tmpPath = filepath.Join(DefaultStoreOptions.GraphRoot, "tmp")
+	}
+	return filepath.Join(tmpPath, r.layerdir, id+tarSplitSuffix)
 }
```
nalind: Why is this moving the tarsplit data to a directory named "tmp"?  We need that information when reading layers, so we don't want to suggest that it's okay to discard it.

rhatdan: I believe this is temporary data? But I could be mistaken.  The problem was that this was attempting to create the split on the readonly partition.

drivers/overlay/overlay.go
```diff
@@ -353,44 +353,54 @@ func (d *Driver) Create(id, parent, mountLabel string, storageOpt map[string]str
...
 	}
 	lowers := []string{path.Join(linkDir, string(parentLink))}
-
 	parentLower, err := ioutil.ReadFile(path.Join(parentDir, lowerFile))
 	if err == nil {
 		parentLowers := strings.Split(string(parentLower), ":")
-		lowers = append(lowers, parentLowers...)
+		for _, p := range parentLowers {
+			lowers = append(lowers, d.dir(p))
+		}
```
nalind: This causes the items in the parents list to have the driver's home path prepend to them each time the list is copied for a new layer, which isn't what we want.  I think you can drop this part.

layers.go
nalind: That file should only be written when we're applying a diff to a initialize a layer, though.  Were you attempting to create layers in a read-only location?

drivers/overlay/overlay.go
```diff
@@ -431,6 +441,20 @@ func (d *Driver) Get(id string, mountLabel string) (s string, err error) {
 		return "", err
 	}
 
+	if _, err = os.Stat(string(lowers)); err != nil {
```
nalind: If the file contains a colon-separated list of lower link partial paths, don't we need to be splitting this list?

layers.go
```diff
@@ -527,7 +527,7 @@ func (r *layerStore) SetMetadata(id, metadata string) error {
...
 
 func (r *layerStore) tspath(id string) string {
-	return filepath.Join(r.layerdir, id+tarSplitSuffix)
+	return filepath.Join(DefaultStoreOptions.GraphRoot, DefaultStoreOptions.GraphDriverName+"-layers", id+tarSplitSuffix)
 }
```
nalind: `r.layerdir` is already set to this value by `store.LayerStore()` and `newLayerStore()`.

nalind: ... if the store is using the default location, that is.  Otherwise this is some other unrelated location, which looks weird.

drivers/overlay/overlay.go
rhatdan: I though the parentLowers would be just an ID, this will prepend a path instead of the ID on the lowers.

layers.go
rhatdan: But r.layerdir could point to the read/only storage, and it needs to write to the read/write layer.

drivers/overlay/overlay.go
rhatdan: I fixed this

drivers/overlay/overlay.go
```diff
@@ -431,6 +441,24 @@ func (d *Driver) Get(id string, mountLabel string) (s string, err error) {
...
 
+	newlowers := ""
+	for _, lower := range strings.Split(string(lowers), ":") {
+		if _, err = os.Stat(string(lower)); err != nil {
+			for _, p := range d.AdditionalImageStores() {
+				lower = path.Join(p, d.name, lower)
+				if _, err := os.Stat(lower); err != nil {
+					return "", err
+				}
+				break
```
nalind: If AdditionalImageStores has more than one element, will this ever search past the first one?

drivers/overlay/overlay.go
nalind: The values returned from here are used to initialize the `lower` file in `Create()`.

drivers/overlay/overlay.go
rhatdan: Yup that is a bug.

drivers/overlay/overlay.go
rhatdan: So if the parent is on /var/lib/nfs/storage, it would be found, or are you saying we fix this else where in the code?

layers.go
nalind: We should only be writing to that file in `ApplyDiff()`, `Put()` (which calls `ApplyDiff()`) and `Delete()`, and I don't think we could do any of those to a layer in a read-only store.

rhatdan: If I don't do this conversion, I end up with it attempting to create the tarsplit in /var/lib/nfs  Unless we are calling it incorrectly,.

rhatdan:
We can look at these tomorrow.  A lot of it is easier to play with when you deal with buildah.

rhatdan:
bot retest

rhatdan:
Not sure what changed, but the tests are now passing.

layers.go
nalind: When is that happening?

TomSweeneyRedHat:
LGTM, but I'm leaning way heavily on Nalin for this one.

drivers/overlay/overlay.go
```diff
@@ -353,44 +353,54 @@ func (d *Driver) Create(id, parent, mountLabel string, storageOpt map[string]str
...
 	lowers := []string{path.Join(linkDir, string(parentLink))}
-
 	parentLower, err := ioutil.ReadFile(path.Join(parentDir, lowerFile))
 	if err == nil {
 		parentLowers := strings.Split(string(parentLower), ":")
-		lowers = append(lowers, parentLowers...)
+		for _, p := range parentLowers {
+			lowers = append(lowers, p)
+		}
 	}
```
nalind: Does this do anything now?

drivers/overlay/overlay.go
nalind: I don't see `lowers` being modified after it's read from the file.  Running 'go test' under strace is still showing me calls to `stat()` with inputs like `l/BQQTO2FOMZN5GFRI6JBQFXF2DP:l/TNXWF7RJCERTMLWZ75EI73J54H:l/64MPZFDTMQ623P5YJIZGZFSVU4:l/SEFLQHUOKLMFAOTKZJODYHGTTF`.

drivers/overlay/overlay.go
rhatdan: Oops that is a bug, maybe that is why it is succeeding.

layers.go
rhatdan: If you setup a layer/image on an nfs share it uses the path associated with the image to create the tarsplit.  It passes in /var/lib/nfs/storage which I mount as readonly.

drivers/overlay/overlay.go
rhatdan: I guess we don't need to do the d.dir(p) any longer, so I reverted this back to the original code.

drivers/overlay/overlay.go
rhatdan: Should be fixed now

drivers/overlay/overlay.go
```diff
@@ -414,7 +422,7 @@ func (d *Driver) Remove(id string) error {
 	return nil
 }
 
-// Get creates and mounts the required file system for the given id and returns the mount path.
+// Get creates and mounts teh required file system for the given id and returns the mount path.
```
nalind: Typo: "teh"

store.go
```diff
@@ -596,13 +618,64 @@ func (s *store) LayerStore() (LayerStore, error) {
 	return s.layerStore, nil
 }
 
+func (s *store) ROLayerStores() ([]LayerStore, error) {
+	s.graphLock.Lock()
+	defer s.graphLock.Unlock()
+	if s.graphLock.TouchedSince(s.lastLoaded) {
+		s.graphDriver = nil
+		s.layerStore = nil
```
nalind: I don't think you mean to be deleting the read-write store pointer here.

rhatdan: What about the s.graphDriver?

rhatdan:
Fixed both requests, removed the entire section.

store.go
```diff
@@ -523,6 +539,12 @@ func (s *store) load() error {
 	if err != nil {
 		return err
 	}
+	rois, err := s.ROImageStores()
+	if err != nil {
+		return err
+	}
+	s.roImageStores = rois
```
nalind: This assignment looks redundant, since `ROImageStores()` is already making sure `s.roImageStores` is set.

store.go
```diff
@@ -515,6 +525,12 @@ func (s *store) load() error {
 	}
 	s.layerStore = rls
 
+	rols, err := s.ROLayerStores()
+	if err != nil {
+		return err
+	}
+	s.roLayerStores = rols
```
nalind: This assignment looks redundant, since `ROLayerStores()` is already making sure `s.roLayerStores` is set.

store.go
```diff
@@ -1534,124 +1680,210 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
...
+	for _, rlstore := range stores {
+		rlstore.Lock()
+		defer rlstore.Unlock()
+		if modified, err := rlstore.Modified(); modified || err != nil {
+			rlstore.Load()
+		}
+		if rlstore.Exists(to) {
+			return rlstore.ApplyDiff(to, diff)
+		}
+	}
```
nalind: If I'm reading things right, if the layer to which the diff is being applied is in a read-only location, the overlay driver will create a layer with the same ID in the local read-write storage path (per the overlay driver's `getDiffPath()` method), but we won't have an entry for it in the read-write layerStore.

nalind:
I think the overlay driver's `getDiffPath()` method will need some work in order to allow extracting the contents of layers that are in the read-only locations.  We'll need that in order to be able to push images that use layers from those locations.

store.go
rhatdan: Are sure this is not working.  When I do a buildah commit on a container image that has a readonly share it seems to work?  Doesn't this do the equivalent of a push?

nalind:
Ah, right, the updated dir() method covers that.

store.go
nalind: Ah, the updated dir() covers this, too.

rhatdan:
@nalind Any Verdict?

drivers/overlay/overlay.go
```diff
@@ -431,6 +439,25 @@ func (d *Driver) Get(id string, mountLabel string) (s string, err error) {
...
+	for _, lower := range strings.Split(string(lowers), ":") {
+		newpath := path.Join(d.home, lower)
+		if _, err := os.Stat(newpath); err != nil {
+			for _, p := range d.AdditionalImageStores() {
+				lower := path.Join(p, d.name, lower)
+				_, err = os.Stat(lower)
+				if err == nil {
+					break
+				}
+			}
```
nalind: Is this `for` loop doing what it's supposed to?  Neither `p` defined on line 446 nor the new `lower` defined on line 447 seem to be affecting anything outside of the loop's scope.

rhatdan: lower is being appended to the newlowers list.

nalind: Yes, but that's happening outside of the scope of the variable with that name that's being defined on line 447, so unless I'm mistaken, the value is going to be the one assigned on line 443.

rhatdan: Ok I did not notice the := I fixed it to redefine the lower variable from above.

nalind: If the loop doesn't exit via `break` at line 450, what value is expected to be appended to the `newlowers` list?

rhatdan:
@nalind Now returns error if the link can not be found.

nalind:
LGTM.

rhatdan:
@nalind Your the only one allowed to merge containers/storage.  Well at least I am not allowed.

nalind:
Just realized that some of the methods now take locks on stores in differing orders, which can lead to deadlocks if one process calls `Exists()` while another calls, say `SetNames()`.  We've generally been taking the lock on the layers store, if it's being taken, before the lock for the images store, if it's being taken, and then the lock on the containers store, if it's being taken.
`CreateImage()` takes the images store lock before the layers store lock.
`CreateContainer()` takes the images lock, then the layers lock, then the containers lock.
`Exists()` takes the containers lock, then the images lock, then the layers lock.
`Mount()` and `Unmount()` take the containers lock, then the layers lock.
Those'll need to change the order in which they're acquiring locks in order to avoid creating deadlocks.

rhatdan:
I think I have this covered,  I always lockin the order of

Layer.Locks
ImageLocks
Container.Locks

If more then one lock is taken within a function.

rhatdan:
I broke this pull request up a little to get the smaller pull requests merged first.

nalind:
Okay, looks good.

nalind:
@TomSweeneyRedHat any concerns before I merge this?

TomSweeneyRedHat:
LGTM.  I double checked the lock ordering and it looks good to me as well.  Still leaning heavily on Nalin for this bad boy.

PR #52, opened 2017-05-31
If a lockfile is on a readonly directory then create a read lock

rhatdan:
We need to handle the situation where lockfiles are stored on readonly
image stores.  We will first attempt a read/write lock and then fall
back on a read/only lock.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -152,7 +152,6 @@ func (r *containerStore) Load() error {
 	r.bylayer = layers
 	r.byname = names
 	if needSave {
-		r.Touch()
 		return r.Save()
 	}
 	return nil
--- a/containers.go
+++ b/containers.go
@@ -158,6 +158,9 @@ func (r *containerStore) Load() error {
 }
 
 func (r *containerStore) Save() error {
+	if !r.lockfile.RWLock() {
+		return ErrLockReadOnly
+	}
 	rpath := r.containerspath()
 	if err := os.MkdirAll(filepath.Dir(rpath), 0700); err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


rhatdan:
@nalind Redesigned to add GetROLockFile.

nalind:
I'm really having trouble with changing the established behavior of `GetLockfile()`.  I'd like to try coming at it from another direction that looks roughly like:

* Add the parts of this PR that add read-only locks as a different type.  Basically, the `lockfile.go` bits, but renaming `RWLock()` to something starting with `Is`, a la `time.IsZero()`, so it doesn't keep confusing me into thinking it's some variation on the `Lock()` and `RLock()` methods of the `sync` package's mutex types.  
* Break out the methods of `LayerStore` and `ImageStore` that don't modify and try to `Save()` their contents out into separate read-only interfaces (`ROLayerStore` or something), and modify the existing interfaces to embed the read-only interfaces rather than listing each of the same methods.  This would also be done to `FileBasedStore`, `MetadataStore`, `BigDataStore`, and `FlaggableStore`.
* Add new variants of `newLayerStore` and `newImageStore` that obtain read-only locks instead of read-write locks, use them to create underlying objects of the same `layerStore` and `imageStore` types that the current functions do, and return the objects using the read-only interface types.
* Switch to using the read-only interface types for handling additional read-only layer and image stores in the main `Store` object.  The writable stores should mix right in when we're building slices of stores.

This should let us use build-time type-checking to avoid inadvertently calling code paths that try to modify things, letting them change in-memory contents as they do, and then failing to `Save()` them to disk.  As you've noted out of band, reverting the in-memory changes when that happens is still something we need to handle, but this should help us avoid inflicting that on ourselves.

What do you think?

rhatdan:
This all makes sense to me.

nalind:
Closing by way of #63.

PR #53, opened 2017-06-01
Mark stores as modified by SetNames()

nalind:
Use the lockfile to mark that a store's contents need to be reloaded by other consumers whenever we call SetNames() to update a by-name index.

### DIFF
--- a/store.go
+++ b/store.go
@@ -1043,12 +1043,15 @@ func (s *store) SetNames(id string, names []string) error {
 	}
 
 	if rlstore.Exists(id) {
+		defer rlstore.Touch()
 		return rlstore.SetNames(id, deduped)
 	}
 	if ristore.Exists(id) {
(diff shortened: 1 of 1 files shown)


store.go
```diff
@@ -1044,12 +1044,15 @@ func (s *store) SetNames(id string, names []string) error {
 	}
 
 	if rlstore.Exists(id) {
+		defer rlstore.Touch()
```
rhatdan: Why not put this into the SetNames function, and only call it on success?

nalind: Hmm, it's a pretty reasonable idea to move `Touch()` calls into the item store's `Save()` method, and out of the `Store`'s methods in general.

rhatdan:
LGTM

nalind:
Thanks, merging.

PR #54, opened 2017-06-01
Fix consistency errors after adding/removing items

nalind:
Fix consistency errors we'd hit after creating or deleting a layer, image, or container, by replacing the slice of items in their respective stores with a slice of pointers to items, so that pointers in name- and ID-based indexes don't become invalid when the slice is resized.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -93,7 +93,7 @@ type ContainerStore interface {
 type containerStore struct {
 	lockfile   Locker
 	dir        string
-	containers []Container
+	containers []*Container
 	idindex    *truncindex.TruncIndex
 	byid       map[string]*Container
 	bylayer    map[string]*Container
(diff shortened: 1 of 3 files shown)


rhatdan:
Not quite sure what this does, but the code LGTM

TomSweeneyRedHat:
LGTM

containers.go
```diff
@@ -123,22 +127,22 @@ func (r *containerStore) Load() error {
...
 	layers := make(map[string]*Container)
 	idlist := []string{}
 	ids := make(map[string]*Container)
 	names := make(map[string]*Container)
 	if err = json.Unmarshal(data, &containers); len(data) == 0 || err == nil {
 		for n, container := range containers {
 			idlist = append(idlist, container.ID)
-			ids[container.ID] = &containers[n]
-			layers[container.LayerID] = &containers[n]
+			ids[container.ID] = containers[n]
```
nalind: They will be now, as `container` goes from being a `Container` struct (which for each iteration has the contents of one of the elements in the slice copied to it) to being a pointer.

rhatdan:
LGTM

nalind:
Thanks, merging.

PR #55, opened 2017-06-02
Only touch when images, containers, layers save function called

rhatdan:


rhatdan:
@nalind lets get this one merged in first before the ROLayers Patch.  Since it should make the other patch a little cleaner.

nalind:
LGTM, merging.

PR #56, opened 2017-06-03
Add new function to return ContainerLayerID

rhatdan:
In order to get ROLayers to merge, we need to break
some functions into smaller subsets to not hold the locks long.

### DIFF
--- a/store.go
+++ b/store.go
@@ -1508,9 +1508,8 @@ func (s *store) Version() ([][2]string, error) {
 }
 
 func (s *store) Mount(id, mountLabel string) (string, error) {
-	rcstore, err := s.ContainerStore()
-	if err != nil {
-		return "", err
+	if layerID, err := s.ContainerLayerID(id); err == nil {
+		id = layerID
 	}
 	rlstore, err := s.LayerStore()
 	if err != nil {
@@ -1522,23 +1521,13 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
 	defer rlstore.Touch()
 	if modified, err := rlstore.Modified(); modified || err != nil {
 		rlstore.Load()
-	}
-	rcstore.Lock()
-	defer rcstore.Unlock()
-	if modified, err := rcstore.Modified(); modified || err != nil {
-		rcstore.Load()
-	}
-
-	if c, err := rcstore.Get(id); c != nil && err == nil {
-		id = c.LayerID
 	}
 	return rlstore.Mount(id, mountLabel)
 }
 
 func (s *store) Unmount(id string) error {
-	rcstore, err := s.ContainerStore()
-	if err != nil {
-		return err
+	if layerID, err := s.ContainerLayerID(id); err == nil {
+		id = layerID
 	}
 	rlstore, err := s.LayerStore()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


store.go
```diff
@@ -1749,6 +1729,23 @@ func (s *store) Container(id string) (*Container, error) {
...
+	if err != nil {
+		return "", err
+	}
+	rcstore.Lock()
+	defer rcstore.Unlock()
+	if modified, err := rcstore.Modified(); modified || err != nil {
+		rcstore.Load()
+	}
+	container, err := rcstore.Get(id)
+	if err != nil {
```
TomSweeneyRedHat: Do you have to verify that the container != nil and handle that too?  Or will the return line 1746 be happy if the container is nil?

rhatdan: I see no code path where it would ever be successful and not return a container, so I think we are all set here.

We call store.Get functions all over and never check the first return value.

TomSweeneyRedHat: Thought I saw a case where we could get a nil container w/o an error.  Just ran through the code again and you're right, the container shouldn't be nil if there's not an error.  Sorry about the fire drill.

TomSweeneyRedHat:
LGTM, but please let @nalind take a look see.  Also could you touch up the pr title please?  (returne to return)

rhatdan:
Fixed, also already fixed the pull request.  Always wait for @nalind...

store.go
```diff
@@ -1508,9 +1508,8 @@ func (s *store) Version() ([][2]string, error) {
 }
 
 func (s *store) Mount(id, mountLabel string) (string, error) {
-	rcstore, err := s.ContainerStore()
-	if err != nil {
-		return "", err
+	if layerID, err := s.ContainerLayerID(id); err == nil {
+		id = layerID
 	}
```
nalind: Shouldn't an error be returned here if `ContainerLayerID` returns an error, instead of falling through and letting the layer store's `Mount()` method eventually return an `ErrLayerUnknown` error?

store.go
```diff
@@ -1523,22 +1522,12 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
...
 	return rlstore.Mount(id, mountLabel)
 }
 
 func (s *store) Unmount(id string) error {
-	rcstore, err := s.ContainerStore()
-	if err != nil {
-		return err
+	if layerID, err := s.ContainerLayerID(id); err == nil {
+		id = layerID
 	}
```
nalind: Shouldn't an error be returned here if `ContainerLayerID` returns an error, instead of falling through and letting the layer store's `Unmount()` method eventually return an `ErrLayerUnknown` error?

store.go
rhatdan: I believe this matches the current code.  Where we only replaced the id field if it was a container, otherwise we just used it as a layer.

store.go
rhatdan: Same answer, above I think current code only changes the id to container.LayerID if the id was a container id,

rhatdan: Current code
```
-       rcstore.Lock()
-       defer rcstore.Unlock()
-       if modified, err := rcstore.Modified(); modified || err != nil {
-               rcstore.Load()
-       }
-
-       if c, err := rcstore.Get(id); c != nil && err == nil {
-               id = c.LayerID
-       }
        return rlstore.Unmount(id)
 }
```

rhatdan: We don't report error if rcstore.Get(id) fails.

store.go
nalind: Ah, right.

nalind:
LGTM, merging.

PR #58, opened 2017-06-05
Add better error reporting when layers are in use

rhatdan:
We need to inform user what object is using a layer when returning and
error.

### DIFF
--- a/store.go
+++ b/store.go
@@ -2,7 +2,6 @@ package storage
 
 import (
 	"encoding/base64"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -21,6 +20,7 @@ import (
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/stringid"
+	"github.com/pkg/errors"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


rhatdan:
This patch is to help diagnose issues like https://github.com/projectatomic/buildah/issues/134

store.go
```diff
@@ -1255,8 +1255,8 @@ func (s *store) DeleteImage(id string, commit bool) (layers []string, err error)
 		for _, container := range containers {
 			aContainerByImage[container.ImageID] = container.ID
 		}
-		if _, ok := aContainerByImage[id]; ok {
-			return nil, ErrImageUsedByContainer
+		if container, ok := aContainerByImage[id]; ok {
+			return nil, errors.Wrap(ErrImageUsedByContainer, "Image used by %v", container.ID)
```
TomSweeneyRedHat: Should this be WrapF?

rhatdan: yes

store.go
```diff
@@ -1916,7 +1916,7 @@ func (s *store) Shutdown(force bool) ([]string, error) {
 		}
 	}
 	if len(mounted) > 0 && err == nil {
-		err = ErrLayerUsedByContainer
+		err = errors.Wrapf(ErrLayerUsedByContainer, "A layer is mounted")
```
nalind: This can be `errors.Wrap()`.

rhatdan:
@nalind Fixed to use Wrap instead of Wrapf.

nalind:
LGTM, merging.

PR #59, opened 2017-06-06
Lockfile is leaking file descriptors,  Needs to closexec

rhatdan:
SELinux is complaining about the leaked file descriptors.

### DIFF
--- a/lockfile.go
+++ b/lockfile.go
@@ -56,6 +56,7 @@ func GetLockfile(path string) (Locker, error) {
 	if err != nil {
 		return nil, err
 	}
+	unix.CloseOnExec(fd)
 	locker := &lockfile{file: path, fd: uintptr(fd), lw: stringid.GenerateRandomID()}
 	lockfiles[filepath.Clean(path)] = locker
 	return locker, nil


lockfile.go
```diff
@@ -56,6 +57,10 @@ func GetLockfile(path string) (Locker, error) {
 	if err != nil {
 		return nil, err
 	}
+	_, _, err = syscall.Syscall(syscall.SYS_FCNTL, uintptr(fd), syscall.F_SETFD, 0)
+	if err != syscall.Errno(0x0) {
+		return nil, err
+	}
```
nalind: We don't get as much in the way of diagnostic feedback from it, but it looks like `unix.CloseOnExec()` is meant to be used for this.

rhatdan: Ok I did not see this.

rhatdan: Looks like 	syscall.CloseOnExec(fd)
 also exists.

rhatdan:
@nalind Fixed.

nalind:
LGTM, pending CI.

nalind:
Merging.

PR #60, opened 2017-06-07
Uneccessary Touch functions.

rhatdan:
We don't need these Touch calls, since the Save function will handle it.


nalind:
LGTM pending CI results.

PR #62, opened 2017-06-09
Some error return codes were accidently changed

rhatdan:
Changes from ErrImageUnknown to  ErrIsNotImage is breaking protocol and has changed the
behaviour of buildah, Breaking previous behaviour.


rhatdan:
@nalind This is what is causing buildah from --pull=False to fail.

nalind:
LGTM, merging.

PR #63, opened 2017-06-09
Implement read-only layer and image stores

nalind:
Building off of #50 and subsuming the changes from #52, this patch set refactors the existing LayerStore and ImageStore interfaces and introduces subsets of them that aren't expected to modify the underlying data.  It also implements versions of the implementations that take read-only locks instead of read-write locks, and uses those to handle additional storage locations.

### DIFF
--- a/images.go
+++ b/images.go
@@ -49,11 +49,32 @@ type Image struct {
 	Flags map[string]interface{} `json:"flags,omitempty"`
 }
 
+// ROImageStore provides bookkeeping for information about Images.
+type ROImageStore interface {
+	ROFileBasedStore
+	ROMetadataStore
+	ROBigDataStore
+
+	// Exists checks if there is an image with the given ID or name.
+	Exists(id string) bool
+
+	// Get retrieves information about an image given an ID or name.
+	Get(id string) (*Image, error)
+
+	// Lookup attempts to translate a name to an ID.  Most methods do this
+	// implicitly.
+	Lookup(name string) (string, error)
+
+	// Images returns a slice enumerating the known images.
+	Images() ([]Image, error)
+}
+
 // ImageStore provides bookkeeping for information about Images.
 type ImageStore interface {
-	FileBasedStore
-	MetadataStore
-	BigDataStore
+	ROImageStore
+	RWFileBasedStore
+	RWMetadataStore
+	RWBigDataStore
 	FlaggableStore
 
 	// Create creates an image that has a specified ID (or a random one) and
@@ -65,24 +86,11 @@ type ImageStore interface {
 	// supplied values.
 	SetNames(id string, names []string) error
 
-	// Exists checks if there is an image with the given ID or name.
-	Exists(id string) bool
-
-	// Get retrieves information about an image given an ID or name.
-	Get(id string) (*Image, error)
-
 	// Delete removes the record of the image.
 	Delete(id string) error
 
 	// Wipe removes records of all images.
 	Wipe() error
-
-	// Lookup attempts to translate a name to an ID.  Most methods do this
-	// implicitly.
-	Lookup(name string) (string, error)
-
-	// Images returns a slice enumerating the known images.
-	Images() ([]Image, error)
 }
 
 type imageStore struct {
--- a/images.go
+++ b/images.go
@@ -2,7 +2,6 @@ package storage
 
 import (
 	"encoding/json"
-	"errors"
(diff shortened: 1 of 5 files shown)


rhatdan:
Lets merge the first two pulls together.  I don't care about getting credit, and It would read a lot better.  Or I can update my pull request with your changes.

images.go
```diff
@@ -150,13 +150,16 @@ func (r *imageStore) Load() error {
 	r.idindex = truncindex.NewTruncIndex(idlist)
 	r.byid = ids
 	r.byname = names
-	if needSave {
+	if r.IsReadWrite() && shouldSave {
```
rhatdan: Should we return an  error if r.IsReadWrite == False and shouldSave == True?

rhatdan: Since Save returns an error, no need to check here?

images.go
```diff
@@ -205,6 +228,9 @@ func (r *imageStore) lookup(id string) (*Image, bool) {
 }
 
 func (r *imageStore) ClearFlag(id string, flag string) error {
+	if !r.IsReadWrite() {
+		return errors.New("called a write method on a read-only store")
```
rhatdan: Lets create a CONSTANT for this.

store.go
```diff
@@ -1669,20 +1669,13 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
...
-	for _, rlstore := range stores {
-		rlstore.Lock()
-		defer rlstore.Unlock()
-		if modified, err := rlstore.Modified(); modified || err != nil {
-			rlstore.Load()
-		}
-		if rlstore.Exists(id) {
-			return rlstore.Mount(id, mountLabel)
-		}
+	if rlstore.Exists(id) {
```
rhatdan: I guess this means you can only mount read/write shares. which makes sense.

rhatdan:
Overall this looks great...

images.go
nalind: I suppose we can return an error - the only case where we should want to immediately save things is when the data we loaded has name collisions, and someone opening it for read-write would be able to fix that.

images.go
nalind: Changing it.

nalind:
Okay, I squashed the lockfile support patches together, added more detail to errors that we return when someone tries to call a writing method on a read-only object, and updated Load() to return an error if the data looks malformed in a way that we'd try to correct if the store was read-write, but the store is read-only.

images.go
```diff
@@ -140,23 +140,26 @@ func (r *imageStore) Load() error {
...
+	}
 	r.images = images
 	r.idindex = truncindex.NewTruncIndex(idlist)
 	r.byid = ids
 	r.byname = names
-	if needSave {
-		return r.Save()
-	}
-	return nil
+	return r.Save()
```
rhatdan: This looks wrong.  Shouldn't you return nil if shouldSave is not set?

nalind: Eek, you're right.

nalind:
Updated.

images.go
```diff
@@ -49,11 +49,32 @@ type Image struct {
...
 }
 
+// ROImageStore provides bookkeeping for information about Images.
+type ROImageStore interface {
+	ROFileBasedStore
+	ROMetadataStore
+	ROBigDataStore
+
+	// Exists checks if there is an image with the given ID or name.
+	Exists(id string) bool
```
nalind: The interfaces limit which methods are available through a value of that interface type.  While a value of type `FlaggableStore` doesn't have these methods, an `ImageStore` value, which is of a type that embeds both `ROImageStore` and `FlaggableStore`, includes both (and more), and that's what we actually use.
They're sliced up like this to avoid having any given method appear more than once in an interface, whether directly or through embedding.  The compiler rejected that kind of duplication as an error when I first tried it.
The method implementations receive the underlying object type, so they're free to call any method that the object implements.
If we implemented read-only and read-write stores as different object types in addition to different interface types, we wouldn't need to check for read-only status in several of the write methods.  We're adding those checks just to be safe in case someone decides to try using a type assertion to convert a read-only interface type to a read-write interface.  The object that underlies a read-only interface value also provides all of the methods of the read-write interface, so it wouldn't cause an error.

rhatdan:
LGTM

rhatdan:
@nalind Now we need to get this vendored into Buildah and the buildah commit/push merged, and we can release new versions of buildah.

PR #65, opened 2017-06-16
Make Diff() methods take an optional *DiffOptions

nalind:
Add an optional *DiffOptions parameter to Diff() methods (which can be nil), to allow overriding of default behaviors.

At this time, that's just what type of compression is applied, if we want something other than what was recorded when the diff was applied, but we can add more later if needed.


rhatdan:
:+1:

nalind:

PR #66, opened 2017-06-17
layers|containers: do not allocate slices at every delete

runcom:
Was playing around with CRI-O profiler (https://github.com/kubernetes-incubator/cri-o/pull/612) and ran into this:

When Delete:ing a layer or a container the code was always allocating a
new slice just to remove an element from the original slice.
Profiling cri-o with c/storage showed that doing it at every delete is
pretty expensive:

```
         .          .    309:   newContainers := []Container{}
         .          .    310:   for _, candidate := range r.containers
{
         .          .    311:           if candidate.ID != id {
  528.17kB   528.17kB    312:                   newContainers =
append(newContainers, candidate)
         .          .    313:           }
         .          .    314:   }

         .          .    552:           newLayers := []Layer{}
         .          .    553:           for _, candidate := range
r.layers {
         .          .    554:                   if candidate.ID != id {
    1.03MB     1.03MB    555:                           newLayers =
append(newLayers, candidate)
         .          .    556:                   }
         .          .    557:           }
         .          .    558:           r.layers = newLayers
```

This patch just filters out the element to remove from the original
slice w/o allocating a new slice. After this patch, no memory overhead
anymore is shown in the profiler.

@mrunalp @rhatdan @nalind PTAL

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -309,10 +309,11 @@ func (r *containerStore) Delete(id string) error {
 		return ErrContainerUnknown
 	}
 	id = container.ID
-	newContainers := []*Container{}
-	for _, candidate := range r.containers {
-		if candidate.ID != id {
-			newContainers = append(newContainers, candidate)
+	toDeleteIndex := -1
+	for i, candidate := range r.containers {
+		if candidate.ID == id {
+			toDeleteIndex = i
+			break
 		}
 	}
 	delete(r.byid, id)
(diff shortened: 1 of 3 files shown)


containers.go
```diff
@@ -309,10 +309,10 @@ func (r *containerStore) Delete(id string) error {
 		return ErrContainerUnknown
 	}
 	id = container.ID
-	newContainers := []*Container{}
-	for _, candidate := range r.containers {
-		if candidate.ID != id {
-			newContainers = append(newContainers, candidate)
+	toDeleteIndex := -1
+	for i, candidate := range r.containers {
```
runcom: I guess, question for @nalind , I suppose `r.containers` and `r.layers` can't contain duplicates, or do they? I'll edit this patch if they do cause here we're just filtering out 1 element.

nalind: They shouldn't contain duplicates, give or take bugs of course.

nalind:
LGTM, but there's similar logic in `image.go` which I suspect you'll want to make a similar change to.

runcom:
@nalind updated `images.go` as well.

nalind:
LGTM.

rhatdan:
LGTM

PR #67, opened 2017-06-19
Remember digests

nalind:
This change adds fields to `Layer` structures that track compressed and uncompressed digests and diff sizes on behalf of callers, and populates them when a layer has `ApplyDiff()` called against it, or is created with `Put()`.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -15,6 +15,7 @@ import (
 	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/containers/storage/pkg/truncindex"
+	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/vbatts/tar-split/tar/asm"
 	"github.com/vbatts/tar-split/tar/storage"
@@ -72,6 +73,32 @@ type Layer struct {
 	// is set before using it.
 	Created time.Time `json:"created,omitempty"`
 
+	// CompressedDigest is the digest of the blob that was last passed to
+	// ApplyDiff() or Put(), as it was presented to us.
+	CompressedDigest digest.Digest `json:"compressed-diff-digest,omitempty"`
+
+	// CompressedSize is the length of the blob that was last passed to
+	// ApplyDiff() or Put(), as it was presented to us.  If
+	// CompressedDigest is not set, this should be treated as if it were an
+	// uninitialized value.
+	CompressedSize int64 `json:"compressed-size,omitempty"`
+
+	// UncompressedDigest is the digest of the blob that was last passed to
+	// ApplyDiff() or Put(), after we decompressed it.  Often referred to
+	// as a DiffID.
+	UncompressedDigest digest.Digest `json:"diff-digest,omitempty"`
+
+	// UncompressedSize is the length of the blob that was last passed to
+	// ApplyDiff() or Put(), after we decompressed it.  If
+	// UncompressedDigest is not set, this should be treated as if it were
+	// an uninitialized value.
+	UncompressedSize int64 `json:"diff-size,omitempty"`
+
+	// CompressionType is the type of compression which we detected on the blob
+	// that was last passed to ApplyDiff() or Put().
+	CompressionType archive.Compression `json:"compression,omitempty"`
+
+	// Flags is arbitrary data about the layer.
 	Flags map[string]interface{} `json:"flags,omitempty"`
 }
 
@@ -121,9 +148,22 @@ type ROLayerStore interface {
 	// produced by Diff.
 	DiffSize(from, to string) (int64, error)
 
+	// Size produces a cached value for the uncompressed size of the layer,
+	// if one is known, or -1 if it is not known.  If the layer can not be
+	// found, it returns an error.
+	Size(name string) (int64, error)
+
 	// Lookup attempts to translate a name to an ID.  Most methods do this
 	// implicitly.
 	Lookup(name string) (string, error)
+
+	// LayersByCompressedDigest returns a slice of the layers with the
+	// specified compressed digest value recorded for them.
+	LayersByCompressedDigest(d digest.Digest) ([]Layer, error)
+
+	// LayersByUncompressedDigest returns a slice of the layers with the
+	// specified uncompressed digest value recorded for them.
+	LayersByUncompressedDigest(d digest.Digest) ([]Layer, error)
 
 	// Layers returns a slice of the known layers.
 	Layers() ([]Layer, error)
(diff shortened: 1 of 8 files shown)


layers.go
```diff
@@ -72,6 +73,28 @@ type Layer struct {
 	// is set before using it.
 	Created time.Time `json:"created,omitempty"`
 
+	// CompressedDigest is the digest of the blob that was last passed to
+	// ApplyDiff() or Put(), as it was presented to us.
+	CompressedDigest digest.Digest `json:"compressed-diff-digest,omitempty"`
+
+	// CompressedSize is the length of the blob that was last passed to
```
rhatdan: Why are you storing this, if you are never using it?

rhatdan:
@mtrmac PTAL

layers.go
nalind: This'll be useful in the image library.  It'll be able to easily determine if there's some layer already in local storage that was initialized using a compressed diff that had this digest.

rhatdan:
LGTM, but I think @mtrmac and @runcom would understands these parts better then me.

layers.go
```diff
@@ -228,10 +267,11 @@ func (r *layerStore) Load() error {
 				}
 				names[name] = layers[n]
 			}
-			if pslice, ok := parents[layer.Parent]; ok {
-				parents[layer.Parent] = append(pslice, layers[n])
-			} else {
-				parents[layer.Parent] = []*Layer{layers[n]}
+			if layer.UncompressedDigest != "" {
```
TomSweeneyRedHat: Nit of a nit.  Everywhere else in the code you reference compressed first, then uncompressed.  If/when you make changes here, could you flip the order of these if statements.  (I know, I know, it's my OCD kicking in).

store.go
```diff
@@ -1786,6 +1799,65 @@ func (s *store) ApplyDiff(to string, diff archive.Reader) (int64, error) {
 	return -1, ErrLayerUnknown
 }
 
+func (s *store) layersByMappedDigest(m func(ROLayerStore, digest.Digest) ([]Layer, error), d digest.Digest) ([]Layer, error) {
+	var layers []Layer
+	rlstore, err := s.LayerStore()
+	if err != nil {
+		return nil, err
```
TomSweeneyRedHat: You've a number of no-op's on errors scattered about. Would it be worthwhile (if possible) to at least add unique debug messages to them?

TomSweeneyRedHat:
LGTM too, but I'll mirror Dan's comments for others to take a look.

layers.go
nalind: Easily changed.

layers.go
```diff
@@ -228,10 +267,11 @@ func (r *layerStore) Load() error {
...
 			}
-			if pslice, ok := parents[layer.Parent]; ok {
-				parents[layer.Parent] = append(pslice, layers[n])
-			} else {
-				parents[layer.Parent] = []*Layer{layers[n]}
+			if layer.CompressedDigest != "" {
+				compressedsums[layer.CompressedDigest] = append(compressedsums[layer.CompressedDigest], layer.ID)
+			}
+			if layer.UncompressedDigest != "" {
+				uncompressedsums[layer.UncompressedDigest] = append(uncompressedsums[layer.UncompressedDigest], layer.ID)
```
mtrmac: This suggests that several different objects representing an exactly identical layer can exist in the layer store. Is that so?  (I have no idea, it may well be, e.g. if the store only contained filesystems resulting from applying the layer, i.e. the result of applying the layer may be different depending on parents.)

layers.go
```diff
@@ -382,6 +424,17 @@ func (r *layerStore) lookup(id string) (*Layer, bool) {
 	return nil, false
 }
 
+func (r *layerStore) Size(name string) (int64, error) {
+	layer, ok := r.lookup(name)
+	if !ok {
+		return -1, ErrLayerUnknown
+	}
+	if layer.UncompressedDigest != "" {
+		return layer.UncompressedSize, nil
```
mtrmac: Personal-style nitpick, this relationship between `UncompressedDigest != ""` <=> `UncompressedSize` is undocumented.

layers.go
```diff
@@ -895,15 +949,59 @@ func (r *layerStore) ApplyDiff(to string, diff archive.Reader) (size int64, err
...
+	}
+	updateDigestMap(&r.bycompressedsum, layer.CompressedDigest, compressedDigest.Digest(), layer.ID)
+	layer.CompressedDigest = compressedDigest.Digest()
+	layer.CompressedSize = compressedCounter.Count
+	updateDigestMap(&r.byuncompressedsum, layer.UncompressedDigest, uncompressedDigest.Digest(), layer.ID)
+	layer.UncompressedDigest = uncompressedDigest.Digest()
+	layer.UncompressedSize = uncompressedCounter.Count
+	layer.CompressionType = compression
+
+	if err == nil {
```
mtrmac: Does it really make sense to do all of these metadata updates if `err != nil`, i.e. `ApplyDiff` failed? (I have absolutely no idea.)

layers.go
```diff
@@ -72,6 +73,28 @@ type Layer struct {
 	// is set before using it.
 	Created time.Time `json:"created,omitempty"`
 
+	// CompressedDigest is the digest of the blob that was last passed to
+	// ApplyDiff() or Put(), as it was presented to us.
+	CompressedDigest digest.Digest `json:"compressed-diff-digest,omitempty"`
```
mtrmac: Does any layer deduplication for layers with the same `UncompressedDigest` happen? If it does, it could be useful to support storing several `CompressedDigest+CompressedSize` values for such a deduplicated layer.

(Either way, storing at least one will handle most of the cases and is clearly better than nothing.)

Longer-term, it could be useful to record some kind of “here is where the compressed version of the layer can be downloaded from” pointer for each `CompressedDigest` value, to allow reproducing the original exactly (and therefore to allow “pushing” an image without breaking the manifest+signatures, or to allow pushing without copying the layer back to the registry it came from). That’s not happening soon, I’m only mentioning this to make sure something like that will be possible—looking at this PR, it seems that extending the metadata in the future will work just fine.

layers.go
nalind: Hmm, probably not.  Going to change that.

layers.go
nalind: Deduplication would be possible for the overlay driver, which stores layers as diffs, but not for the devicemapper driver, which stores layers as the filesystems that are produced by applying multiple layers on top of each other.  I think deduplication would have to be done on, or at least determined on, a per-driver basis.

layers.go
nalind: If the underlying driver stores the layer purely as a diff, there can conceivably be multiple copies of the same diff in the driver.  I think we can fix that by changing how the image library creates layers when it's committing an image, but I'd want to add supporting logic to the storage library in a different PR.

rhatdan:
@nalind I think you can merge.

nalind:
Thanks!  Merging.

layers.go
mtrmac: Thanks for the explanation. containers/image does not require layers to be deduplicated , this is perfectly fine.

nalind:

PR #68, opened 2017-06-20
Allow setting the creation-date on images

nalind:
Add a field to ImageOptions that allows a caller to specify a date of creation when calling CreateImage(), if there's a value in the image metadata that would be more useful than the default (which is "now" at
the time CreateImage() is called).


rhatdan:
LGTM

nalind:

PR #69, opened 2017-06-23
When creating images, options are optional

nalind:
Since ImageOptions are optional when calling a store's CreateImage() method, we can't assume that there's a value there.


rhatdan:
LGTM

PR #72, opened 2017-07-07
Fix compression handling for Diff() methods

nalind:
The `Diff()` method didn't handle recompression correctly when generating a diff between a layer and its parent when there was no tar-split data present.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -823,33 +823,49 @@ func (r *layerStore) Diff(from, to string, options *DiffOptions) (io.ReadCloser,
 	if options != nil && options.Compression != nil {
 		compression = *options.Compression
 	}
+	maybeCompressReadCloser := func(rc io.ReadCloser) (io.ReadCloser, error) {
+		// Depending on whether or not compression is desired, return either the
+		// passed-in ReadCloser, or a new one that provides its readers with a
+		// compressed version of the data that the original would have provided
+		// to its readers.
+		if compression == archive.Uncompressed {
+			return rc, nil
+		}
+		preader, pwriter := io.Pipe()
+		compressor, err := archive.CompressStream(pwriter, compression)
+		if err != nil {
+			rc.Close()
+			pwriter.Close()
+			preader.Close()
+			return nil, err
+		}
+		go func() {
+			defer pwriter.Close()
+			defer compressor.Close()
+			defer rc.Close()
+			io.Copy(compressor, rc)
+		}()
+		return preader, nil
+	}
+
 	if from != toLayer.Parent {
 		diff, err := r.driver.Diff(to, from)
-		if err == nil && (compression != archive.Uncompressed) {
-			preader, pwriter := io.Pipe()
-			compressor, err := archive.CompressStream(pwriter, compression)
-			if err != nil {
-				diff.Close()
-				pwriter.Close()
-				return nil, err
-			}
-			go func() {
-				io.Copy(compressor, diff)
-				diff.Close()
-				compressor.Close()
-				pwriter.Close()
-			}()
-			diff = preader
+		if err != nil {
+			return nil, err
 		}
-		return diff, err
+		return maybeCompressReadCloser(diff)
 	}
 
 	tsfile, err := os.Open(r.tspath(to))
 	if err != nil {
-		if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


layers.go
```diff
@@ -823,33 +823,45 @@ func (r *layerStore) Diff(from, to string, options *DiffOptions) (io.ReadCloser,
...
+			return nil, err
 		}
-		return diff, err
+		return maybeCompressReadCloser(diff)
 	}
 
 	tsfile, err := os.Open(r.tspath(to))
 	if err != nil {
-		if os.IsNotExist(err) {
-			return r.driver.Diff(to, from)
```
nalind: This is where we were ignoring the compression settings.

layers.go
```diff
@@ -823,33 +823,45 @@ func (r *layerStore) Diff(from, to string, options *DiffOptions) (io.ReadCloser,
...
 	}
+	maybeCompressReadCloser := func(rc io.ReadCloser) (io.ReadCloser, error) {
+		if compression == archive.Uncompressed {
+			return rc, nil
+		}
+		preader, pwriter := io.Pipe()
+		compressor, err := archive.CompressStream(pwriter, compression)
+		if err != nil {
+			rc.Close()
+			pwriter.Close()
```
rhatdan: Why not just add this as a defer?

layers.go
```diff
@@ -823,33 +823,45 @@ func (r *layerStore) Diff(from, to string, options *DiffOptions) (io.ReadCloser,
...
+			preader.Close()
+			return nil, err
+		}
+		go func() {
+			io.Copy(compressor, rc)
+			rc.Close()
+			compressor.Close()
```
rhatdan: Should be a defer?

layers.go
```diff
@@ -823,33 +823,45 @@ func (r *layerStore) Diff(from, to string, options *DiffOptions) (io.ReadCloser,
 	if options != nil && options.Compression != nil {
 		compression = *options.Compression
 	}
+	maybeCompressReadCloser := func(rc io.ReadCloser) (io.ReadCloser, error) {
+		if compression == archive.Uncompressed {
```
rhatdan: This name is meaningless to me.  Could you add a comment explaining what this function is doing?

layers.go
```diff
@@ -823,33 +823,45 @@ func (r *layerStore) Diff(from, to string, options *DiffOptions) (io.ReadCloser,
...
+			return rc, nil
+		}
+		preader, pwriter := io.Pipe()
+		compressor, err := archive.CompressStream(pwriter, compression)
+		if err != nil {
+			rc.Close()
```
TomSweeneyRedHat: Can (should) you do a defer for the rc.Close and pwriter.Close?

TomSweeneyRedHat:
LGTM, one minor question.

layers.go
nalind: Depending on whether or not we're asked to compress the diff, this function either returns the passed-in io.ReadCloser, or an io.ReadCloser that returns a compressed version of what the passed-in one would supply to readers.  Adding that in a comment.

layers.go
nalind: If we don't encounter an error, we want to close these only after the goroutine that we're spinning up finishes its work, which will be some time after Diff() has returned.

layers.go
nalind: Maybe?  Are we worried about cleaning up if Copy() panics?  If not, then deferring them inside the anonymous goroutine doesn't buy us anything.

layers.go
nalind: We shouldn't here - if we end up starting the goroutine, we need to let it close them.

layers.go
nalind: Eh, might as well.

layers.go
rhatdan: ok

layers.go
rhatdan: Perfect add that comment...

nalind:
Updated to add a comment to what `maybeCompressReadCloser` is doing, and to make sure we close the pipe in case something goes awry with its call to io.Copy().

rhatdan:
LGTM

nalind:

PR #77, opened 2017-07-11
Default to overlay as "overlay"

nalind:
When selecting overlay as a possible use-by-default driver, use it under the name "overlay".


rhatdan:
LGTM

PR #78, opened 2017-07-18
Refuse to start loopback devicemapper on AUFS

nalind:
An AUFS filesystem doesn't give us the "real" device and inode numbers of an underlying file when we stat() it, so we'll hit errors trying to resume use of a pool when that pool is built on loopback devices using
files that live on an AUFS filesystem.  Refuse to let ourselves be put into that situation.


rhatdan:
Strange bug, but LGTM

PR #79, opened 2017-07-18
Check for loopback device/inode mismatches

nalind:
After attaching a loopback device to a backing file, compare the loopback driver's idea of the backing file's device and inode numbers against the numbers we get back from stat().  If they differ, log an
error, but continue.

This should help diagnose some problems that can pop up, for example, when we try to resume use of a thin pool built on top of loopback files which live on a union filesystem that returns synthesized values in
stat() results.


rhatdan:
LGTM

PR #83, opened 2017-08-02
Do not default to overlay

rhatdan:
We should default to "" and then allow the library to pick the
best driver for the particular distribution.  Picking overlay is
currently breaking cri-o on Centos and RHEL, since they don't
support overlay driver until 7.4 release.


nalind:
LGTM pending CI results.

nalind:
Merging.

TomSweeneyRedHat:
LGTM

PR #85, opened 2017-08-16
RunRoot directory should be mounted private

rhatdan:
We are leaking files into other mount namespaces, making the
runroot private might also improve performance.

### DIFF
--- a/pkg/mount/sharedsubtree_notlinux.go
+++ b/pkg/mount/sharedsubtree_notlinux.go
@@ -1,0 +1,9 @@
+// +build !linux
+
+package mount
+
+// MakePrivate ensures a mounted filesystem has the PRIVATE mount option enabled.
+// See the supported options in flags.go for further reference.
+func MakePrivate(mountPoint string) error {
+	return nil
+}
--- a/store.go
+++ b/store.go
@@ -19,6 +19,7 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/ioutils"
+	"github.com/containers/storage/pkg/mount"
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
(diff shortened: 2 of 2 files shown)


rhatdan:
@nalind PTAL

nalind:
Nice catch!  It needs a corresponding call to `mount.Unmount` in `store.Shutdown()`, right around where it calls `s.graphdriver.Cleanup()`.

rhatdan:
I am not sure we want to umount this, this will only ensure that it is mounted private, will not mount it again, if it is not mounted.

pkg/mount/sharedsubtree_notlinux.go
```diff
@@ -0,0 +1,51 @@
...
+// See the supported options in flags.go for further reference.
+func MakePrivate(mountPoint string) error {
+	return nil
+}
+
+// MakeRPrivate ensures a mounted filesystem has the RPRIVATE mount option
+// enabled. See the supported options in flags.go for further reference.
+func MakeRPrivate(mountPoint string) error {
+	return nil
+}
```
nalind: Do we need anything other than `MakePrivate()` in this file?

store.go
```diff
@@ -2251,6 +2255,7 @@ func (s *store) Shutdown(force bool) ([]string, error) {
 		s.graphLock.Touch()
 		modified = true
 	}
+	mount.Unmount(s.RunRoot())
```
nalind: Move this so that it's only called if `err` is `nil`, along with `s.graphdriver.Cleanup()`.

store.go
```diff
@@ -529,6 +530,9 @@ func GetStore(options StoreOptions) (Store, error) {
 	if err := os.MkdirAll(options.RunRoot, 0700); err != nil && !os.IsExist(err) {
 		return nil, err
 	}
+
+	mount.MakePrivate(options.RunRoot)
```
nalind: Check the result for an error here.

pkg/mount/sharedsubtree_notlinux.go
rhatdan: No but I thought this would keep us future proof.

store.go
nalind: Use `s.runRoot` instead of `s.RunRoot()` here, read the result from `Unmount()`, and set `err` to its value if `err` is not already set.

rhatdan: Fixed

store.go
rhatdan: Fixed

pkg/mount/sharedsubtree_notlinux.go
nalind: Go ahead and remove them.  We can add them later if/when we find we need them.

nalind:
LGTM

PR #86, opened 2017-08-18
Switch all calls to MakePrivate to MakeUnbindable

rhatdan:
This should eliminate the need for oci-umount, and stop leaking of file paths.
According to the man pages Unbindable meaks the same as Private except that
the mount point will not be bound on a rbind mount.

Running containers with -v /:/host has caused issues with this for a long time.
Only issue with this, is if you attempt to mount the mount point like /run/containers/storage
into a container it will fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 go:
     - tip
+    - 1.8
     - 1.7
-    - 1.6
 dist: trusty
 sudo: required
 before_install:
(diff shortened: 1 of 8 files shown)


rhatdan:
@rhvgoyal PTAL

rhvgoyal:
I think runc still need to be sorted out as it can reset the property of unbindable mounts. Especially with shared volumes etc. So making it unbindable might not hurt. But before we can get rid of oci-umount, we need to make sure runc is working well in all configurations. It is still an unresolved problem.

rhatdan:
I am closing this, I don't think it helps and will reopen if it comes up in the future.

PR #87, opened 2017-08-20
Rename storage.conf(5) to containers-storage.conf(5)

rhatdan:
We have a conflict on man pages storage.conf man page is owned
by the inn package, so we need to pick a less generic name.


rhatdan:
retest

TomSweeneyRedHat:
The change LGTM, but something wonky is going on in dockerCmd_utils_test.go:270.

rhatdan:
Tests are now passing.

@nalind PTAL

storage.conf
```diff
@@ -1,8 +1,9 @@
+# See man 5 containers-storage.conf for mor information
```
nalind: Typo: `mor` should be `more`.

nalind:
Did the bits that change the name of the file fall out of this, or is the description no longer correct?

storage.conf
rhatdan: Fixed

rhatdan:
@nalind PTAL
The rename of the man page still exists.

nalind:
LGTM, merging.

PR #88, opened 2017-08-21
We should not force a default storage driver

rhatdan:
This is causing cri-o to break in RHEL7.3 environments.

Changing to "" allows the containers/storage to pick a storage
driver that will work on a particular platform.  It will still
default to overlay by default, if the platform supports it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 go:
     - tip
+    - 1.8
     - 1.7
-    - 1.6
 dist: trusty
 sudo: required
 before_install:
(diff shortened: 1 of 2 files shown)


rhatdan:
It looks like tip has changed and is breaking all tests.

rhatdan:
Tests are now passing.

@nalind @rhvgoyal PTAL

nalind:
Everything in here is also part of #87.  Do we need two PRs for it?

rhatdan:
No lets merge this and then I will update that pull request.

PR #89, opened 2017-08-28
Remove no-longer-used pkg/integration

nalind:
We're not currently using this package, and removing it should avoid the CI failures we're seeing in #88.


rhatdan:
LGTM

PR #90, opened 2017-09-01
Create errors using "errors"

nalind:
Use the standard library's "errors" package to create errors so that backtraces in wrapped errors terminate at the point where the error was first wrapped, and not at the line where we created the error, which isn't as useful for troubleshooting.


rhatdan:
LGTM

PR #91, opened 2017-09-01
Accept "overlay2.imagestore" as a synonym for the "overlay.imagestore" option

nalind:
Accept "overlay2.imagestore" as a synonym for the "overlay.imagestore" option, and be a bit more consistent about using the driver name we were invoked by in error messages.


rhatdan:
LGTM

PR #92, opened 2017-09-01
Make sure you get the realpath when checking Mounted

rhatdan:
Do not umount when exiting, because you are not sure if you mounted
in the first place and you could accidently umount other peoples mounts.


rhatdan:
@nalind PTAL

nalind:
LGTM, merging.

PR #93, opened 2017-09-01
Remove direct access to substores from Store()

nalind:
Remove methods that fetch the underlying layer, image, and container store objects from the public Store interface.  They're still accessible via type assertions, but ideally that's never going to be necessary.

### DIFF
--- a/store.go
+++ b/store.go
@@ -184,31 +184,6 @@ type Store interface {
 	// by the Store.
 	GraphDriver() (drivers.Driver, error)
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM, this will prevent people from using it without locks, correct?

nalind:
Yes, the goal is to make it harder to bypass locking, because it's trivial to do so otherwise, and that's been problematic at times.

rhatdan:
I agree.

PR #94, opened 2017-09-03
[WIP] Add support for lazy unmount using MNT_DETACH

rhatdan:
Mount points are not being cleaned up because of overmounts by
others. this patch will allow for mount points to eventually get
cleaned up.


rhatdan:
Dropping this request. Since we have moved passed it.

PR #95, opened 2017-09-05
     Revert "RunRoot directory should be mounted private"

rhatdan:
We instead need to move the shm content into a different directory.
This patch basically reverts 05e45e0b084152a4eff88db0d4ea308fd326a3d4.
The reason we need this patch is to allow oci-umount to remove leaked
mount points into the container.  We are going to attempt to move the
shm mount point to a different directory.

### DIFF
--- a/pkg/mount/sharedsubtree_notlinux.go
+++ b/pkg/mount/sharedsubtree_notlinux.go
@@ -1,9 +1,0 @@
-// +build !linux
-
-package mount
-
-// MakePrivate ensures a mounted filesystem has the PRIVATE mount option enabled.
-// See the supported options in flags.go for further reference.
-func MakePrivate(mountPoint string) error {
-	return nil
-}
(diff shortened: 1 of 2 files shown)


nalind:
#85 also added `pkg/mount/sharedsubtree_notlinux.go`, which we might as well drop here, too.  Or replace this with a change generated with `git revert`, which would look the same.

rhatdan:
I don't see a reason to drop the MakePrivate definition in there.  We might end up using it elsewhere and does not cost us anything.  I say just merge this, and then we can look at how we can umount the shm stuff.

nalind:
The code's in source control if we ever need to recall it, and in its current form it creates a build conflict when GOOS=solaris (granted, not the only problem with that setting, but still).

rhatdan:
@nalind Happy now?

nalind:
Patch needs a signed-off-by, otherwise LGTM.

rhatdan:
Merging.

Issue #97, opened 2017-09-06
Support quota for overlay2


mrunalp:
We need to support quota support to limit containers in overlay2 with xfs similar to docker.

https://github.com/moby/moby/blob/8af4db6f002ac907b6ef8610b237879dfcaa5b7a/daemon/graphdriver/overlay2/overlay.go#L202

@nalind @rhatdan

mrunalp:
cc: @runcom

runcom:
:+1:

nalind:
Closing by #98.

PR #73, opened 2017-07-07
Wrap some driver-level errors

nalind:
Wrap `graphdriver.{ErrNotSupported,ErrPrerequisites,ErrIncompatibleFS}` errors in contexts using `github.com/pkg/errors`, and dig them out for comparison using `errors.Cause()`.

PR #98, opened 2017-09-07
Grab changes form moby/moby/graphdriver for overlay quota

rhatdan:
This led to lots of packages being updated from moby
I tried not to pull in the snap driver stuff.


rhatdan:
The goal of this patch set is to satisfy https://github.com/containers/storage/issues/97

rhatdan:
@nalind @rhvgoyal @mrunalp @runcom PTAL

runcom:
looks good (assuming travis will pass) and the integration with CRI-O is smooth :) (and as long as overlay quota works)

drivers/aufs/aufs_test.go
```diff
@@ -148,7 +149,10 @@ func TestRemoveImage(t *testing.T) {
 
 	for _, p := range paths {
 		if _, err := os.Stat(path.Join(tmp, p, "1")); err == nil {
-			t.Fatalf("Error should not be nil because dirs with id 1 should be delted: %s", p)
+			t.Fatalf("Error should not be nil because dirs with id 1 should be deleted: %s", p)
+		}
+		if _, err := os.Stat(path.Join(tmp, p, "1-removing")); err == nil {
+			t.Fatalf("Error should not be nil because dirs with id 1-removing should be deleted: %s", p)
```
TomSweeneyRedHat: I'm back and forth.  Should 1-removing be '1-removing'?  Just to set it off more quickly?

rhatdan:
@runcom a lot of "if's"

runcom:
> @runcom a lot of "if's"

well, it's still marked WIP :) I'll take a better look once it looks ready for you for sure :)

pkg/idtools/idtools_windows.go
```diff
@@ -5,14 +5,21 @@ package idtools
 import (
 	"os"
 
-	"github.com/containers/storage/pkg/system"
+	"github.com/docker/docker/pkg/system"
```
TomSweeneyRedHat: Is this change right?

rhatdan: NOPE

rhatdan: Fixed

nalind:
The first patch in the series also removes a couple of packages from the `vendor` directory, so it should mention that, and might as well remove them from `vendor.conf`.

drivers/aufs/aufs_test.go
rhatdan: We don't support aufs anyways, This code is just copied from Moby

rhatdan:
@nalind @rhvgoyal @mrunalp @runcom Tests are passing now, PTAL

nalind:
[Backport overlayxfs quota support from moby to cri-o](https://trello.com/c/wonuFj8Z/666-backport-overlayxfs-quota-support-from-moby-to-cri-o)

drivers/overlay/overlay.go
```diff
@@ -146,38 +178,45 @@ func InitWithName(name, home string, options []string, uidMaps, gidMaps []idtool
...
-		uidMaps: uidMaps,
-		gidMaps: gidMaps,
-		ctr:     graphdriver.NewRefCounter(graphdriver.NewFsChecker(graphdriver.FsMagicOverlay)),
-		opts:    opts,
+	supportsDType, err := fsutils.SupportsDType(home)
+	if err != nil {
+		return nil, err
+	}
+	if !supportsDType {
+		// not a fatal error until v17.12 (#27443)
```
nalind: When do we intend to make this a fatal error?

drivers/overlay/overlay.go
```diff
@@ -191,24 +230,30 @@ func parseOptions(name string, options []string) (*overlayOptions, error) {
 			if err != nil {
 				return nil, err
 			}
+		case "overlay.size":
```
nalind: Alias this as "overlay2.size"?

drivers/overlay/overlay.go
```diff
@@ -475,19 +591,37 @@ func (d *Driver) Get(id string, mountLabel string) (s string, err error) {
...
 		if err != nil {
 			if c := d.ctr.Decrement(mergedDir); c <= 0 {
-				syscall.Unmount(mergedDir, 0)
+				unix.Unmount(mergedDir, 0)
 			}
 		}
 	}()
 
 	workDir := path.Join(dir, "work")
 	opts := fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s", newlowers, path.Join(id, "diff"), path.Join(id, "work"))
```
nalind: Since we're trying to avoid the fork/exec when possible, and that means there's no child process to change its working directory to a place where relative paths resolve correctly, the last two parameters here need to be absolute paths (`diffDir` and `workDir`, respectively).  Likewise, at line 578, `lower` needs to become an absolute path, so it needs to be set to `newpath` instead of `l`.

drivers/overlay/overlay.go
```diff
@@ -475,19 +591,37 @@ func (d *Driver) Get(id string, mountLabel string) (s string, err error) {
...
+	mountTarget := mergedDir
+
+	pageSize := unix.Getpagesize()
+
+	// Use relative paths and mountFrom when the mount data has exceeded
+	// the page size. The mount syscall fails if the mount data cannot
+	// fit within a page and relative links make the mount data much
+	// smaller at the expense of requiring a fork exec to chroot.
+	if len(mountData) > pageSize {
+		opts = fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s", string(lowers), path.Join(id, "diff"), path.Join(id, "work"))
```
nalind: Given that `lowers` was constructed, when the layer was created, by concatenating the contents of the parent layer's "link" file with its own "lowers" list, some of these symbolic links are in the additional stores, so this section some work.

drivers/overlay/overlay.go
```diff
@@ -475,19 +591,37 @@ func (d *Driver) Get(id string, mountLabel string) (s string, err error) {
...
+
+		mount = func(source string, target string, mType string, flags uintptr, label string) error {
+			return mountFrom(d.home, source, target, mType, flags, label)
+		}
+		mountTarget = path.Join(id, "merged")
 	}
 
-	if err := mountFrom(d.home, "overlay", path.Join(id, "merged"), "overlay", mountLabel); err != nil {
+	if err := mount("overlay", mountTarget, "overlay", 0, mountData); err != nil {
 		return "", fmt.Errorf("error creating overlay mount to %s: %v", mergedDir, err)
```
nalind: I think it would be easier to troubleshoot if we used `mountTarget` when formatting this error message.

drivers/overlay/overlay.go
rhatdan: I would figure when moby does?  Or should we do it now?

rhatdan: Does xfs support DType?  If we turn this on now, will it break on RHEL.  Man pages do not mention DT_TYPE for xfs. @rhvgoyal WDYT?

drivers/overlay/overlay.go
rhatdan: Fixed

drivers/overlay/overlay.go
rhatdan: Fixed

drivers/overlay/overlay.go
rhatdan: Fixed.

rhatdan:
@nalind @rhvgoyal 
Finally passing tests.  Added some storage configuration and am about to start testing with this on CRI-O to make sure the overaly quota actually works.
We know we have issues on systems with mount options larger then 4096 bytes but everything else seems to be good to Review.
@mrunalp @runcom FYI

rhatdan:
I now have this code working with CRI-O will a couple of small changes to it.
https://github.com/rhatdan/cri-o/tree/quota

It would probably be easier to review this patch by looking at directory differences from Moby.

rhvgoyal:
I quickly browsed through the patches and I did not spot anything bad.

mrunalp:
@nalind @rhvgoyal any chance this could be merged soon? :)

drivers/aufs/aufs.go
```diff
@@ -84,6 +86,7 @@ func Init(root string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 	// Try to load the aufs kernel module
 	if err := supportsAufs(); err != nil {
 		return nil, errors.Wrap(graphdriver.ErrNotSupported, "kernel does not support aufs")
+
```
nalind: No need for this line.

drivers/aufs/aufs.go
```diff
@@ -201,17 +231,23 @@ func (a *Driver) Exists(id string) bool {
 	return true
 }
 
+// AdditionalImageStores returns additional image stores supported by the driver
+func (a *Driver) AdditionalImageStores() []string {
+	var imageStores []string
+	return imageStores
+}
```
nalind: This initializes `imageStores` to the zero value for a slice, which is `nil`, so the method could as easily return `nil` directly.

drivers/aufs/aufs_test.go
```diff
@@ -312,16 +317,16 @@ func TestCreateWithInvalidParent(t *testing.T) {
 	d := newDriver(t)
 	defer os.RemoveAll(tmp)
 
-	if err := d.Create("1", "storage", "", nil); err == nil {
-		t.Fatalf("Error should not be nil with parent does not exist")
+	if err := d.Create("1", "docker", nil); err == nil {
```
nalind: Lost a rename here?  It's test code, though.

drivers/devmapper/deviceset.go
```diff
@@ -1464,17 +1478,14 @@ func (devices *DeviceSet) closeTransaction() error {
...
+	// Kernel driver version >= 4.27.0 support deferred removal
 
-	logrus.Debugf("devicemapper: driver version is %s", version)
+	logrus.Debugf("devicemapper: kernel dm driver version is %s", version)
 
 	versionSplit := strings.Split(version, ".")
 	major, err := strconv.Atoi(versionSplit[0])
 	if err != nil {
-		return errors.Wrapf(graphdriver.ErrNotSupported, "unable to parse driver major version %q as a number", versionSplit[0])
+		return graphdriver.ErrNotSupported
```
nalind: Preferable to keep wrapping this error.

drivers/devmapper/deviceset.go
```diff
@@ -1488,7 +1499,7 @@ func determineDriverCapabilities(version string) error {
 
 	minor, err := strconv.Atoi(versionSplit[1])
 	if err != nil {
-		return errors.Wrapf(graphdriver.ErrNotSupported, "unable to parse driver minor version %q as a number", versionSplit[1])
+		return graphdriver.ErrNotSupported
```
nalind: Preferable to keep wrapping this error.

drivers/devmapper/deviceset.go
```diff
@@ -1648,36 +1660,19 @@ func (devices *DeviceSet) enableDeferredRemovalDeletion() error {
...
-	// 		logrus.Errorf("devmapper: Udev sync is not supported. This will lead to data loss and unexpected behavior. Install a dynamic binary to use devicemapper or select a different storage driver. For more information, see https://docs.docker.com/engine/reference/commandline/daemon/#daemon-storage-driver-option")
-	// 	} else {
-	// 		logrus.Errorf("devmapper: Udev sync is not supported. This will lead to data loss and unexpected behavior. Install a more recent version of libdevmapper or select a different storage driver. For more information, see https://docs.docker.com/engine/reference/commandline/daemon/#daemon-storage-driver-option")
-	// 	}
-	//
-	// 	if !devices.overrideUdevSyncCheck {
-	// 		return graphdriver.ErrNotSupported
-	// 	}
-	// }
+	// https://github.com/containers/storage/issues/4036
```
nalind: This URL needs to not be rewritten.

drivers/devmapper/deviceset.go
```diff
@@ -1692,20 +1687,47 @@ func (devices *DeviceSet) initDevmapper(doInit bool) error {
...
 	// "reg-" stands for "regular file".
 	// In the future we might use "dev-" for "device file", etc.
-	// container-maj,min[-inode] stands for:
-	//	- Managed by container storage
+	// storage-maj,min[-inode] stands for:
+	//	- Managed by storage
 	//	- The target of this device is at major <maj> and minor <min>
 	//	- If <inode> is defined, use that file inside the device as a loopback image. Otherwise use the device itself.
-	devices.devicePrefix = fmt.Sprintf("container-%d:%d-%d", major(sysSt.Dev), minor(sysSt.Dev), sysSt.Ino)
+	devices.devicePrefix = fmt.Sprintf("storage-%d:%d-%d", major(st.Dev), minor(st.Dev), st.Ino)
```
nalind: This change to the name may create problems if tools using different versions of the library, and thus different names, want to use a loopback thinpool built on the same files - if a pool with the expected name is not found, a (potentially second) pair of loopback devices will be created on the files, and a (potentially second) thinpool will be created using them, and I don't know if the kernel's going to keep everything working.

drivers/devmapper/deviceset.go
```diff
@@ -1733,22 +1755,13 @@ func (devices *DeviceSet) initDevmapper(doInit bool) error {
...
 
-		fsMagic, err := graphdriver.GetFSMagic(devices.loopbackDir())
-		if err != nil {
-			return err
-		}
-		switch fsMagic {
-		case graphdriver.FsMagicAufs:
-			return errors.Errorf("devmapper: Loopback devices can not be created on AUFS filesystems")
-		}
-
```
nalind: This reverts #78.

drivers/overlay/mount.go
```diff
@@ -31,12 +31,12 @@ type mountOptions struct {
...
-func mountFrom(dir, device, target, mType, label string) error {
+func mountFrom(dir, device, target, mType string, flags uintptr, label string) error {
 	options := &mountOptions{
 		Device: device,
 		Target: target,
 		Type:   mType,
-		Flag:   0,
+		Flag:   uint32(flags),
 		Label:  label,
 	}
```
nalind: This seems to be missing recent changes that ensure that `w` is always closed, even on error paths.

drivers/overlay/overlay.go
```diff
@@ -146,38 +178,44 @@ func InitWithName(name, home string, options []string, uidMaps, gidMaps []idtool
...
-	return InitWithName("overlay", home, options, uidMaps, gidMaps)
-}
+	d := &Driver{
+		home:          home,
+		uidMaps:       uidMaps,
+		gidMaps:       gidMaps,
+		ctr:           graphdriver.NewRefCounter(graphdriver.NewFsChecker(graphdriver.FsMagicOverlay)),
+		supportsDType: supportsDType,
+		locker:        locker.New(),
+		options:       *opts,
```
nalind: The "name" field is not being initialized.  It is used to construct names for files and directories both here and in the parent `Store` object.

drivers/overlayutils/overlayutils.go
```diff
@@ -0,0 +1,18 @@
...
+	"fmt"
+)
+
+// ErrDTypeNotSupported denotes that the backing filesystem doesn't support d_type.
+func ErrDTypeNotSupported(driver, backingFs string) error {
+	msg := fmt.Sprintf("%s: the backing %s filesystem is formatted without d_type support, which leads to incorrect behavior.", driver, backingFs)
+	if backingFs == "xfs" {
+		msg += " Reformat the filesystem with ftype=1 to enable d_type support."
+	}
+	msg += " Running without d_type support will no longer be supported in Docker 17.12."
```
nalind: This error message probably needs to be updated.

drivers/quota/projectquota.go
```diff
@@ -0,0 +1,337 @@
...
+// Test to make sure that quota can be set on a test dir and find
+// the first project id to be used for the next container create.
+//
+// Returns nil (and error) if project quota is not supported.
+//
+// First get the project id of the home directory.
+// This test will fail if the backing fs is not xfs.
+//
+// xfs_quota tool can be used to assign a project id to the driver home directory, e.g.:
+//    echo 999:/var/lib/container/storage/overlay >> /etc/projects
```
nalind: Missing an `s` at the end of `container` in this pathname.

drivers/quota/projectquota.go
```diff
@@ -0,0 +1,337 @@
...
+//    echo storage:999 >> /etc/projid
+//    xfs_quota -x -c 'project -s storage' /<xfs mount point>
+//
+// In that case, the home directory project id will be used as a "start offset"
+// and all containers will be assigned larger project ids (e.g. >= 1000).
+// This is a way to prevent xfs_quota management from conflicting with container/storage.
```
nalind: Missing an `s` at the end of `container`.

drivers/aufs/aufs.go
rhatdan: Fixed

drivers/aufs/aufs_test.go
rhatdan: Fixed

drivers/devmapper/deviceset.go
rhatdan: Fixed

drivers/devmapper/deviceset.go
rhatdan: FIxed

drivers/devmapper/deviceset.go
rhatdan: Fixed

nalind:
* Dropping `drivers/plugin.go` means that `drivers/proxy.go` no longer compiles with the `experimental` tag, so this should drop that file, too.
* The devicemapper driver appears to be missing updates to `xfsSupported()` and its callers, and `setCloseOnExec()`.

drivers/devmapper/deviceset.go
rhatdan: Change back to container.

drivers/devmapper/deviceset.go
rhatdan: Fixed

drivers/overlayutils/overlayutils.go
rhatdan: msg += " Running without d_type is not supported."
Fixed with this ^^

drivers/quota/projectquota.go
rhatdan: Fixed

drivers/quota/projectquota.go
rhatdan: Fixed

drivers/overlay/overlay.go
rhatdan: Fixed

rhatdan: Initialized to "overlay"

rhatdan:
Removed drivers/proxy.go

drivers/overlay/overlay.go
```diff
@@ -473,22 +592,42 @@ func (d *Driver) Get(id string, mountLabel string) (s string, err error) {
 		return mergedDir, nil
 	}
 	defer func() {
-		if err != nil {
-			if c := d.ctr.Decrement(mergedDir); c <= 0 {
-				syscall.Unmount(mergedDir, 0)
+		if retErr != nil {
+			if err != nil {
```
nalind: We're switching from checking `err` to checking `retErr` here, so this line and its matching closing brace shouldn't be here.

rhatdan: Yup I screwed this patch up.  Will push the correct change from upstream.

drivers/aufs/aufs_test.go
```diff
@@ -31,7 +33,7 @@ func init() {
 func testInit(dir string, t testing.TB) graphdriver.Driver {
 	d, err := Init(dir, nil, nil, nil)
 	if err != nil {
-		if errors.Cause(err) == graphdriver.ErrNotSupported {
+		if err == graphdriver.ErrNotSupported {
```
nalind: Should keep unwrapping here, per #73.

drivers/aufs/aufs.go
```diff
@@ -567,14 +664,14 @@ func (a *Driver) aufsMount(ro []string, rw, target, mountLabel string) (err erro
 // version of aufs.
 func useDirperm() bool {
 	enableDirpermLock.Do(func() {
-		base, err := ioutil.TempDir("", "storage-aufs-base")
+		base, err := ioutil.TempDir("", "docker-aufs-base")
```
nalind: Did we lose a rename here?

drivers/aufs/aufs.go
```diff
@@ -567,14 +664,14 @@ func (a *Driver) aufsMount(ro []string, rw, target, mountLabel string) (err erro
...
 		if err != nil {
 			logrus.Errorf("error checking dirperm1: %v", err)
 			return
 		}
 		defer os.RemoveAll(base)
 
-		union, err := ioutil.TempDir("", "storage-aufs-union")
+		union, err := ioutil.TempDir("", "docker-aufs-union")
```
nalind: Did we lose a rename here?

drivers/devmapper/deviceset.go
```diff
@@ -1692,20 +1699,47 @@ func (devices *DeviceSet) initDevmapper(doInit bool) error {
...
+	var st unix.Stat_t
+	if err := unix.Stat(devices.root, &st); err != nil {
 		return fmt.Errorf("devmapper: Error looking up dir %s: %s", devices.root, err)
 	}
-	sysSt := st.Sys().(*syscall.Stat_t)
 	// "reg-" stands for "regular file".
 	// In the future we might use "dev-" for "device file", etc.
-	// container-maj,min[-inode] stands for:
-	//	- Managed by container storage
+	// storage-maj,min[-inode] stands for:
```
nalind: Nit: comment doesn't match the string below.

drivers/overlay/mount.go
rhatdan: Fixed

drivers/aufs/aufs_test.go
rhatdan: Fixed

drivers/aufs/aufs.go
rhatdan: Fixed

drivers/aufs/aufs.go
rhatdan: Fixed

drivers/devmapper/deviceset.go
rhatdan: Changes to container

drivers/overlay/overlay.go
nalind: This should be equivalent to whether or not the XFS filesystem was formatted using `ftype=1`.

nalind:
Okay, I think the current state of things answers all of my comments, so LGTM.

rhatdan:
Ok I will squash the commits and push again.

nalind:
[(8) Containers/storage needs to back port moby/moby support for quota support.](https://trello.com/c/5E3mkECi/679-8-containers-storage-needs-to-back-port-moby-moby-support-for-quota-support)

Issue #102, opened 2017-09-12
Locking bug in store.go


umohnani8:
There is an issue with the store lock that came up while refactoring libkpod to libpod.
https://github.com/containers/storage/blob/master/store.go#L1261 uses `stores` instead of `ristores`, thus it is trying to lock a store that has already been locked.

umohnani8:
@nalind PTAL

rhatdan:
@umohnani8 Open a pull request.

PR #103, opened 2017-09-12
Fix a locking error in Store.Names()

nalind:
@umohnani8 noticed that `Store.Names()` was hitting a deadlock - when attempting to iterate through the concatenated list of read-write and read-only image stores, the method was actually walking the list of read-write and read-only layer stores a second time, so when it attempted to obtain a lock on the first layer store, it would deadlock.

This PR also teaches `Store.Lookup()` about read-only locations, and updates sections of the code that scan read-only layer and image stores to be more consistent in variable naming in an effort to avoid bugs like this one.

### DIFF
--- a/store.go
+++ b/store.go
@@ -769,23 +769,22 @@ func (s *store) CreateImage(id string, names []string, layer, metadata string, o
 		id = stringid.GenerateRandomID()
 	}
 
-	rlstore, err := s.LayerStore()
+	lstore, err := s.LayerStore()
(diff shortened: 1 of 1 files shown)


nalind:
This should fix #102.

rhatdan:
LGTM
@umohnani8  PTAL

umohnani8:
LGTM 👍

nalind:
Okay, merging.

PR #106, opened 2017-09-29
Move no Dtype on Overlay from Error to warning.

rhatdan:
Move lack of DTYpe on Overlay from Error down to Warning, temporarily until infrastructure CI has updated VMs with correct support.

Moby currently  has the same compromise.

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -42,7 +42,7 @@ The `storage` table supports the following options:
   Default Copy On Write (COW) container storage driver
 
 **additionalimagestores**=[]
-  Paths to additional congtainer image stores. Usually these are read/only and stored on remote network shares.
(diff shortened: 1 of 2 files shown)


rhatdan:
@mrunalp @runcom @nalind @rhvgoyal PTAL

TomSweeneyRedHat:
LGTM.  In general, do we mark a change like this with a TODO in the comments?  Also should a Trello card be created to go turn this back on so we don't forget to later?

rhatdan:
Yes a trello card would be good, also need a card to add tests to make sure overlayfs with quota is actually working.

nalind:
LGTM.

TomSweeneyRedHat:
kk, I'll spin the Trello cards up then.

mrunalp:
Let us get this in for now while we get the amis sorted out. Thanks!

PR #107, opened 2017-09-29
Preallocate some slices that we build up

nalind:
Take a guess at the final size of some slices that we build up item by item, and try to allocate enough capacity for them before starting to build them.  It's probably not a big speedup, though.


rhatdan:
Is this some kind of golang performance thing,where append knows to use the next empty space in the preallocated array?

nalind:
Yes - allocating them with `make(type, length, capacity)` makes append()ing to the slice faster, since a new slice doesn't need to be allocated if there's capacity for it.

rhatdan:
LGTM, will merge after I cut release for override_kernel_check package.

PR #108, opened 2017-09-29
Stop rejecting creation of images with no layers

nalind:
We need to be able to create images which consist of just a list of manifests, and those don't contain layers, so relax CreateImage()'s requirement that a layer be specified.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

nalind:
Thanks!  Merging.

PR #109, opened 2017-09-29
Track the digests of big data items

nalind:
Teach image and container store types to also track the digests of "big data" items that we have them store.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -10,7 +10,7 @@ import (
 	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/containers/storage/pkg/truncindex"
-
+	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 )
 
@@ -45,6 +45,10 @@ type Container struct {
 	// BigDataSizes maps the names in BigDataNames to the sizes of the data
 	// that has been stored, if they're known.
 	BigDataSizes map[string]int64 `json:"big-data-sizes,omitempty"`
+
+	// BigDataDigests maps the names in BigDataNames to the digests of the
+	// data that has been stored, if they're known.
+	BigDataDigests map[string]digest.Digest `json:"big-data-digests,omitempty"`
 
 	// Created is the datestamp for when this container was created.  Older
 	// versions of the library did not track this information, so callers
@@ -253,15 +257,16 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
 	}
 	if err == nil {
 		container = &Container{
-			ID:           id,
-			Names:        names,
-			ImageID:      image,
-			LayerID:      layer,
-			Metadata:     metadata,
-			BigDataNames: []string{},
-			BigDataSizes: make(map[string]int64),
-			Created:      time.Now().UTC(),
-			Flags:        make(map[string]interface{}),
+			ID:             id,
+			Names:          names,
+			ImageID:        image,
+			LayerID:        layer,
+			Metadata:       metadata,
+			BigDataNames:   []string{},
+			BigDataSizes:   make(map[string]int64),
+			BigDataDigests: make(map[string]digest.Digest),
+			Created:        time.Now().UTC(),
+			Flags:          make(map[string]interface{}),
 		}
 		r.containers = append(r.containers, container)
 		r.byid[id] = container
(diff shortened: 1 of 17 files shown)


containers.go
```diff
@@ -368,16 +381,61 @@ func (r *containerStore) BigData(id, key string) ([]byte, error) {
...
+				return size, nil
+			}
+		}
+	}
 	return -1, ErrSizeUnknown
 }
 
+func (r *containerStore) BigDataDigest(id, key string) (digest.Digest, error) {
+	if key == "" {
+		return "", errors.Wrapf(ErrInvalidBigDataName, "data name %q can not be used as a filename", key)
```
TomSweeneyRedHat: This error message should be changed.  key will always be blank here.

containers.go
```diff
@@ -387,6 +445,9 @@ func (r *containerStore) BigDataNames(id string) ([]string, error) {
 }
 
 func (r *containerStore) SetBigData(id, key string, data []byte) error {
+	if key == "" {
+		return errors.Wrapf(ErrInvalidBigDataName, "data name %q can not be used as a filename", key)
```
TomSweeneyRedHat: Another error message needing a touch up.

containers.go
```diff
@@ -368,16 +381,61 @@ func (r *containerStore) BigData(id, key string) ([]byte, error) {
 }
 
 func (r *containerStore) BigDataSize(id, key string) (int64, error) {
+	if key == "" {
+		return -1, errors.Wrapf(ErrInvalidBigDataName, "data name %q can not be used as a filename", key)
```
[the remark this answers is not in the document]
TomSweeneyRedHat: Ok, after going through the whole review , I see this error in a number of places.  I think it's misleading to tell them "data name  can not be used as a file name."  Perhaps "data key was empty and is invalid for a filename" or something along those lines?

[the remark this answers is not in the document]
nalind: Well, I want to keep the detail that ultimately this is used to construct a filename private, so that we don't implicitly promise that it'll always stay that way.  But I agree that we could be more specific when rejecting an empty string, since the error code itself already says that it's a bad name in general, which made the detail the wrapping step was adding not very helpful.

nalind:
Tweaked that error message, using different wording for create vs retrieve cases, and rebased it.

containers.go
```diff
@@ -405,7 +449,7 @@ func (r *containerStore) BigDataNames(id string) ([]string, error) {
 
 func (r *containerStore) SetBigData(id, key string, data []byte) error {
 	if key == "" {
-		return errors.Wrapf(ErrInvalidBigDataName, "data name %q can not be used as a filename", key)
+		return errors.Wrapf(ErrInvalidBigDataName, "data name can not be empty")
```
rhatdan: Should we put the name of the function into the error.
Can't SetBigData, data name can not be empty.

containers.go
```diff
@@ -392,9 +397,48 @@ func (r *containerStore) BigDataSize(id, key string) (int64, error) {
...
+				return size, nil
+			}
+		}
+	}
 	return -1, ErrSizeUnknown
 }
 
+func (r *containerStore) BigDataDigest(id, key string) (digest.Digest, error) {
+	if key == "" {
+		return "", errors.Wrapf(ErrInvalidBigDataName, "data name is empty")
```
rhatdan: Should we put the name of the function into the error.
Can't BigDataDigest, data name can not be empty.

containers.go
```diff
@@ -380,7 +385,7 @@ func (r *containerStore) BigData(id, key string) ([]byte, error) {
 
 func (r *containerStore) BigDataSize(id, key string) (int64, error) {
 	if key == "" {
-		return -1, errors.Wrapf(ErrInvalidBigDataName, "data name %q can not be used as a filename", key)
+		return -1, errors.Wrapf(ErrInvalidBigDataName, "data name is empty")
```
rhatdan: Should we put the name of the function into the error.
Can't return BigDataSize, data name can not be empty.

images.go
```diff
@@ -442,6 +487,9 @@ func (r *imageStore) BigDataNames(id string) ([]string, error) {
 }
 
 func (r *imageStore) SetBigData(id, key string, data []byte) error {
+	if key == "" {
+		return errors.Wrapf(ErrInvalidBigDataName, "data name can not be empty")
```
rhatdan: Should we put the name of the function into the error.
Can't SetBigData, data name can not be empty.

containers.go
nalind: Hmm, we don't usually do that.  We're using `Wrapf` to get backtraces in error objects.

nalind:
Okay, tweaked all of the error messages to be unique.

rhatdan:
LGTM

PR #110, opened 2017-09-29
Be more thorough about deduplicating names

nalind:
We already deduplicated names in Store.SetNames(), but we weren't also doing that when creating layers, images, and containers, or in the individual store SetNames() methods.


rhatdan:
LGTM

PR #113, opened 2017-09-29
Fix builds with go 1.7

nalind:
The skopeo package builds on Darwin use Go 1.7, so make sure we can build on it, by replacing a call to `time.Until(t)` with `t.Sub(time.Now())`, as suggested in the standard library documentation.

This should get https://github.com/projectatomic/skopeo/pull/427 to pass its tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: go
 go:
     - tip
     - 1.8
+    - 1.7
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM, although I would prefer to get skopeo to build on 1.8.

PR #114, opened 2017-09-29
Initialize Flags and BigDataSizes maps

nalind:
When we read itms from disk, if maps in the structures are empty, they won't be allocated as part of the decoding process.  When we subsequently go to read or write something from such a map, make sure
it's been initialized.

Add some validation of names that we convert to file names, and of digest values, so that we can be more precise about the error code we return when there's a problem with the values.


rhatdan:
LGTM

PR #115, opened 2017-10-03
Check read-only stores when Putting new layers

nalind:
When we try to Store.Put() a new layer, also check read-only layer stores as part of verifying that the new layer's parent is known.


rhatdan:
LGTM

PR #116, opened 2017-10-03
Make the CLI use either all defaults, or none

nalind:
Make the CLI use all of the default settings if none of them are overridden on the command line, or none of them, if any are overridden.  Any other behavior is hard to use in integration tests, and confusing.


rhatdan:
I am fine with this, since this is only used for testing.  I think we can control the behaviour.

PR #117, opened 2017-10-03
Add some integration tests

nalind:
This adds some integration tests for exercising the higher-level APIs by way of using the CLI helper.  We add read-only store support to the "vfs" driver to let us always test that option, too.

### DIFF
--- a/tests/stores.bats
+++ b/tests/stores.bats
@@ -1,0 +1,90 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "additional-stores" {
+	storage --graph ${STORAGE_DRIVER}.imagestore=${TESTDIR}/ro-root --run ${TESTDIR}/ro-runroot layers
+	storage --storage-opt ${STORAGE_DRIVER}.imagestore=${TESTDIR}/ro-root layers
+	if ! storage --storage-opt ${STORAGE_DRIVER}.imagestore=${TESTDIR}/ro-root layers ; then
+		skip
+	fi
+	# Create a layer.
+	run storage --graph ${TESTDIR}/ro-root --run ${TESTDIR}/ro-runroot --debug=false create-layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	lowerlayer="$output"
+	# Mount the layer.
+	run storage --graph ${TESTDIR}/ro-root --run ${TESTDIR}/ro-runroot --debug=false mount $lowerlayer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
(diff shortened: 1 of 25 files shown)


tests/stores.bats
```diff
@@ -0,0 +1,90 @@
...
+	storage --graph ${TESTDIR}/ro-root --run ${TESTDIR}/ro-runroot unmount $lowerlayer
+
+	# Create an image using this second layer.
+	run storage --graph ${TESTDIR}/ro-root --run ${TESTDIR}/ro-runroot --debug=false create-image $midlayer
+        [ "$status" -eq 0 ]
+        [ "$output" != "" ]
+        image=${output%%  *}
+
+	# Create a third layer based on the second one.
+	run storage --storage-opt ${STORAGE_DRIVER}.imagestore=${TESTDIR}/ro-root create-layer "$midlayer"
```
rhatdan: Shouldn't you check status on this run?

TomSweeneyRedHat:
Other than Dan's question and just a generic can you add more status checks?  LGTM.

tests/stores.bats
nalind: Whoops, that's a leftover from when it was being worked out.  There's no need to call it that first time.

nalind:
I think every call to the `run` helper is followed by a check on the `$status` that it would have set.  Can you point me at any that are missing?

TomSweeneyRedHat:
I didn't spot any 'run' calls without checks, just a number of places that didn't have run and did some call without a check.  For instance tests/apply-diff.bats you do "checkchanges" and "checkdiffs" at the end.  Can those be checked?  They seem to me like they could, but your call.

I wouldn't hold due to those, just something to consider for later revisions.

nalind:
Ah, there's no need to do that.  Functions like `checkchanges` and `checkdiffs` check `$status` after using `run`, so they fail if that comparison fails.  Those functions aren't called using `run`, so if they fail, the tests or other functions that called them will fail.

Thanks for catching places where `$status` was being checked after it wouldn't have been set; those were also leftovers from when earlier iterations, and they should be gone now.

nalind:
Updated to skip tests that require 'tar' when it isn't available.

nalind:
[(5) [containers/storage] Add tests for layer diff contents](https://trello.com/c/VV1Q408f/656-5-containers-storage-add-tests-for-layer-diff-contents)

nalind:
Updated the test harness to test with all available drivers.  It'll take a bit longer to run, though.

nalind:
Undoing that last change, for now.

rhatdan:
LGTM

PR #118, opened 2017-10-05
Fallback to use naive diff driver if enable CONFIG_OVERLAY_FS_REDIREC…

rhatdan:
…T_DIR

Grab Lei Jitang <leijitang@huawei.com> patches from
github.com/Moby/Moby/49c3a7c4bac2877265ef8c4eaf210159560f08b4

    When use overlay2 as the graphdriver and the kernel enable
    `CONFIG_OVERLAY_FS_REDIRECT_DIR=y`, rename a dir in lower layer
    will has a xattr to redirct its dir to source dir. This make the
    image layer unportable. This patch fallback to use naive diff driver
    when kernel enable CONFIG_OVERLAY_FS_REDIRECT_DIR

### DIFF
--- a/drivers/overlay/mount.go
+++ b/drivers/overlay/mount.go
@@ -49,7 +49,6 @@ func mountFrom(dir, device, target, mType string, flags uintptr, label string) e
 	output := bytes.NewBuffer(nil)
 	cmd.Stdout = output
 	cmd.Stderr = output
-
 	if err := cmd.Start(); err != nil {
 		w.Close()
 		return fmt.Errorf("mountfrom error on re-exec cmd: %v", err)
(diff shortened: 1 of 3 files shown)


rhatdan:
@nalind @rhvgoyal PTAL
@coolljt0725 FYI

nalind:
LGTM.

drivers/overlay/mount.go
```diff
@@ -49,7 +49,6 @@ func mountFrom(dir, device, target, mType string, flags uintptr, label string) e
 	output := bytes.NewBuffer(nil)
 	cmd.Stdout = output
 	cmd.Stderr = output
-
```
rhvgoyal: Removing this line seems to be not needed?

rhvgoyal:
Except that minor nit, looks good to me. Looks same as upstream except that file naming thing. overlay2/overlay.

LGTM.

PR #120, opened 2017-10-09
[WIP] Set the permissions on the upper diff directory based on the lower dir

rhatdan:

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -410,7 +410,8 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 		}
 	}
 
-	if err := idtools.MkdirAs(path.Join(dir, "diff"), 0755, rootUID, rootGID); err != nil {
+	diffDir := path.Join(dir, "diff")
+	if err := idtools.MkdirAs(diffDir, 0755, rootUID, rootGID); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


rhatdan:
We have a bug where a user sets the Image / to chmod 555, but when you run a container based on Overlay on top, it ends up running as 755.  This is an attempt to fix.

TomSweeneyRedHat:
Looks like some tests will need adjusting, Travis isn't happy.

rhatdan:
@TomSweeneyRedHat Tests finally pass, now I got to merge this into buildah to see if it actually works.

drivers/overlay/overlay.go
```diff
@@ -441,6 +442,15 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 		return err
 	}
 	if lower != "" {
+		lowers := strings.Split(lower, ":")
+		newpath := path.Join(d.home, lowers[0])
```
nalind: This will fail if the parent layer is in a read-only store.  Try `newpath := d.dir(lowers[0])` instead.

rhatdan:
Have we basically decided on this that this needs to be fixed somewhere else?

nalind:
Layer blobs omit the root directory, so I don't think this can fix all of the cases.

rhatdan:
Closing since this is not an easily solved problem

PR #121, opened 2017-10-13
use ffjson to encode/decode layer structs

runcom:
Use [`ffjson`](https://github.com/pquerna/ffjson) instead of golang's json implementation. This is 2/3x faster.
I ran the go profiler on a CRI-O stress test and most of the time was spent in `layers.Save` which calls marshal. This patch fixes that and the profiler is no longer reporting that much time spent in layers's saving.

@mrunalp @rhatdan @nalind PTAL we might want this soon to be vendored in CRI-O.

Signed-off-by: Antonio Murdaca <runcom@redhat.com>


runcom:
@nalind could you take a look at why `go tip` is failing with those weird errors?

mrunalp:
:+1: It will be good to include this for cri-o 1.0

nalind:
@runcom can you rebase and try again?  I don't the problems the tests against tip were hitting were the result of anything you're changing, and the just-merged #124 should take it out of the picture.

rhatdan:
Replaced by 125

abitrolly:
#125

Issue #84, opened 2017-08-11
"Does not support overlay" intermittent error


aweiteka:
When running cri-o integration tests on F25 I sometimes get an error that overlay is not enabled in the kernel:

```
"'overlay' not found as a supported filesystem on this host. Please ensure kernel is new enough and has overlay support loaded."
```

I haven't reproduced this when overlay is manually enabled before running. [More details here](https://paste.fedoraproject.org/paste/T3bF-chLII7ujSsk9VeopA).

Speaking to @rhatdan he suspects the 'supportsoverlay' function is racing through too quickly: https://github.com/containers/storage/blob/master/drivers/overlay/overlay.go#L219

rhatdan:
@nalind @aweiteka and I just looked at this and I think it might be a race condition in the test to see if overlay is supported.

I believe below that the /proc/filesystem can be read before the modprobe overlay completes?
```
func supportsOverlay() error {
	// We can try to modprobe overlay first before looking at
	// proc/filesystems for when overlay is supported
	exec.Command("modprobe", "overlay").Run()

	f, err := os.Open("/proc/filesystems")
	if err != nil {
		return err
	}
	defer f.Close()
```

nalind:
`Run()` is documented to wait for the child process to exit before it returns (compare to using `Start()` and `Wait()`), so I don't think that there's a race condition in here.

rhatdan:
Well I will close https://github.com/containers/storage/pull/123 and see if this happens to anyone again.

PR #123, opened 2017-10-13
Wait for the kernel module to load

rhatdan:
There is a race condition possible, where we load the overlay kernel module
but it does not complete before we check to see if it loaded.  This patch
causes to code to check three times with a 1 second sleep between checks.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/containers/storage/drivers"
 	"github.com/containers/storage/drivers/overlayutils"
@@ -264,11 +265,7 @@ func parseOptions(options []string) (*overlayOptions, error) {
 	return o, nil
 }
 
-func supportsOverlay() error {
-	// We can try to modprobe overlay first before looking at
-	// proc/filesystems for when overlay is supported
-	exec.Command("modprobe", "overlay").Run()
-
+func readOverlayFilesystem() error {
 	f, err := os.Open("/proc/filesystems")
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


rhatdan:
The goal of this pull request is to fix this issue.  https://github.com/containers/storage/issues/84

It is a sucky fix, but not sure what we could do better.

drivers/overlay/overlay.go
```diff
@@ -281,10 +278,26 @@ func supportsOverlay() error {
...
+	var err error
+	// We can try to modprobe overlay first before looking at
+	// proc/filesystems for when overlay is supported
+	exec.Command("modprobe", "overlay").Run()
+	for i := 0; i < 3; i++ {
+		if err = readOverlayFilesystem(); err == nil {
+			return nil
+		}
+		if i < 2 {
+			time.Sleep(1 * time.Second)
```
TomSweeneyRedHat: should we retry if we get the graphdriver.ErrNoSupported error from line 281?

drivers/overlay/overlay.go
```diff
@@ -281,10 +278,26 @@ func supportsOverlay() error {
...
-	logrus.Error("'overlay' not found as a supported filesystem on this host. Please ensure kernel is new enough and has overlay support loaded.")
 	return errors.Wrap(graphdriver.ErrNotSupported, "'overlay' not found as a supported filesystem on this host. Please ensure kernel is new enough and has overlay support loaded.")
 }
 
+func supportsOverlay() error {
+	var err error
+	// We can try to modprobe overlay first before looking at
+	// proc/filesystems for when overlay is supported
+	exec.Command("modprobe", "overlay").Run()
+	for i := 0; i < 3; i++ {
```
TomSweeneyRedHat: Might be good to leave a comment here on the race condition and the thinking for the retry.

TomSweeneyRedHat:
couple comments/questions.  Travis isn't happy.

rhatdan:
@nalind questions if this is actually a race condition, so I will close along with the issue, as a Works For Me.  Probably would only happen one time on a fresh boot, so tough to create.

PR #126, opened 2017-10-13
Make sure layers_ffjson.go is always current

nalind:
Add a rule to make sure that `layers_ffjson.go` is regenerated whenever we've updated `layers.go`, and do the same for `images.go` and `containers.go`.


rhatdan:
LGTM

PR #128, opened 2017-10-18
overlay: avoid bogus error unmounting base layers

nalind:
When an overlay layer has no parent, "mounting" it is just a no-op, since we're using its "diff" directory directly, and we didn't bother to create a "merged" directory for the layer.  Check for this case when we
go to unmount the layer's "merged" directory, to avoid a spurious "not mounted" error.


rhatdan:
LGTM

PR #129, opened 2017-10-18
Fixups for read-only store tests

nalind:
* After we're done initializing a read-only store for a test, shut it down so that we'll be able to delete it when cleaning up later.
* When mounting a layer that inherits from read-only layers, don't forget to pass the read-only store's location to the CLI helper.


rhatdan:
LGTM

PR #132, opened 2017-10-30
Fix LayersBy*Digest with multiple layer stores

nalind:
`LayersByCompressedDigest` and `LayersByUncompressedDigest` shouldn't fail if one of the stores being consulted doesn't have any layers that match the specified digest, since that's an expected possibility.

### DIFF
--- a/store.go
+++ b/store.go
@@ -1888,9 +1888,15 @@ func (s *store) layersByMappedDigest(m func(ROLayerStore, digest.Digest) ([]Laye
 		}
 		storeLayers, err := m(store, d)
 		if err != nil {
-			return nil, err
+			if errors.Cause(err) != ErrLayerUnknown {
+				return nil, err
+			}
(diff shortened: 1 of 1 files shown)


store.go
```diff
@@ -1887,11 +1887,14 @@ func (s *store) layersByMappedDigest(m func(ROLayerStore, digest.Digest) ([]Laye
 			store.Load()
 		}
 		storeLayers, err := m(store, d)
-		if err != nil {
+		if err != nil && errors.Cause(err) != ErrLayerUnknown {
 			return nil, err
 		}
 		layers = append(layers, storeLayers...)
```
rhatdan: Should you still be appending if you get ErrLayernUnknown?

nalind: Hmm, probably not.  Revised.

rhatdan:
LGTM

PR #134, opened 2017-11-08
Don't run supportsOverlay check by default

rhatdan:
This will fail on a people using containers/storage inside of a container.
For example if you run buildah inside of a docker container on a system
with devicemapper backend.  The overlay module will not be loaded.  The
exec of "modprobe overlay" fails because there are no overlay modules inside
of the containers mount namespace.  If we ignore this check, when container/storage
eventually attempts the mount command, the kernel will be smart enough to load
the overlay module and things will succeed.  If the kernel actually does not support
overlay, the mount will fail and the user will get his answer.

If a container runtime daemon like CRI-O or Docker wants to verify at start up that
overlay file system is supported, I have made SupportsOverlay a public function and the
daemons can call this at start time.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -124,10 +124,6 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 		return nil, err
 	}
 
-	if err := supportsOverlay(); err != nil {
-		return nil, errors.Wrap(graphdriver.ErrNotSupported, "kernel does not support overlay fs")
-	}
-
 	// require kernel 4.0.0 to ensure multiple lower dirs are supported
 	v, err := kernel.GetKernelVersion()
 	if err != nil {
@@ -264,7 +260,7 @@ func parseOptions(options []string) (*overlayOptions, error) {
 	return o, nil
 }
 
-func supportsOverlay() error {
+func SupportsOverlay() error {
 	// We can try to modprobe overlay first before looking at
 	// proc/filesystems for when overlay is supported
 	exec.Command("modprobe", "overlay").Run()


rhatdan:
@nalind @rhvgoyal PTAL

nalind:
Rebased, added a debug log message, and factored the ReadLine() calling into its own function.

TomSweeneyRedHat:
@nalind, did you comment to the wrong PR?

drivers/overlay/overlay.go
```diff
@@ -264,7 +260,7 @@ func parseOptions(options []string) (*overlayOptions, error) {
 	return o, nil
 }
 
-func supportsOverlay() error {
+func SupportsOverlay() error {
```
TomSweeneyRedHat: Is this actually called anywhere else?  I'm not seeing references to it elsewhere.  Just wondering if we should drop it entirely or if we can do the check when not in a container.

nalind:
@TomSweeneyRedHat Yes, yes I did.

nalind:
With overlay being the first driver we try when we're not explicitly configured, this can break the "select a working driver" logic in `graphdriver.New`.

The current code doesn't check the exit status of "modprobe", so I'm not sure I understand the problem we're trying to solve.

rhatdan:
Well the check, I believe is racy and in the case of running inside of the container, the answer is wrong. 
Turns out we do support Overlay, and the check is failing because its mount namespace does not have the overlay driver.

rhatdan:
The goal of this patch is to fix https://github.com/projectatomic/buildah/issues/158

rhatdan:
I left SupportsOverlay so that tools like Nalin is talking about can call this semi broken function to attempt to see if it is supported.  (Of course you would be hard pressed to find a kernel that does not support this) RHEL 6 I guess.

rhatdan:
The more I think about this the less I like the library picking out random storage.  I think we should force the user to specify the storage, and fail if their is not storage specified.

nalind:
Maybe something like the tip of https://github.com/nalind/storage/commits/overlay-mount-test, despite producing more churn, would be a more dependable test?

rhatdan:
I think we can get away with a lot fewer directories there.
Overlay only requires a lower and a merge directory I think.  @rhvgoyal This is something I think you like.

nalind:
Okay, opened #135 to have a conversation about it.

rhatdan:
Closing since we seem to be moving towards https://github.com/containers/storage/pull/135

PR #135, opened 2017-11-08
overlay: when checking for support, actually try a mount

nalind:
When checking if we can use the overlay driver at initialization time, actually try a mount using temporary directories instead of assuming that the kernel listing it as a known filesystem is enough, since it
might not be listed if we're in a container, and we can't load the module from inside of a container.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -3,7 +3,6 @@
 package overlay
 
 import (
-	"bufio"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -26,7 +25,6 @@ import (
 	"github.com/containers/storage/pkg/locker"
 	"github.com/containers/storage/pkg/mount"
 	"github.com/containers/storage/pkg/parsers"
-	"github.com/containers/storage/pkg/parsers/kernel"
 	"github.com/containers/storage/pkg/system"
 	units "github.com/docker/go-units"
 	"github.com/opencontainers/selinux/go-selinux/label"
@@ -124,22 +122,6 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 		return nil, err
 	}
 
-	if err := supportsOverlay(); err != nil {
-		return nil, errors.Wrap(graphdriver.ErrNotSupported, "kernel does not support overlay fs")
-	}
-
-	// require kernel 4.0.0 to ensure multiple lower dirs are supported
-	v, err := kernel.GetKernelVersion()
-	if err != nil {
-		return nil, err
-	}
-	if kernel.CompareKernelVersion(*v, kernel.VersionInfo{Kernel: 4, Major: 0, Minor: 0}) < 0 {
-		if !opts.overrideKernelCheck {
-			return nil, errors.Wrap(graphdriver.ErrNotSupported, "kernel too old to provide multiple lowers feature for overlay")
-		}
-		logrus.Warn("Using pre-4.0.0 kernel for overlay, mount failures may require kernel update")
-	}
-
 	fsMagic, err := graphdriver.GetFSMagic(home)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


drivers/overlay/overlay.go
```diff
@@ -264,23 +264,39 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+		upperDir := filepath.Join(layerDir, "diff")
+		workDir := filepath.Join(layerDir, "work")
+		lowerDir := filepath.Join(layerDir, "lower")
+		defer func() {
+			// Permitted to fail, since the various subdirectories
+			// can be empty or not even there, and the home might
+			// legitimately be not empty
+			_ = unix.Unmount(mergedDir, unix.MNT_DETACH)
+			_ = os.RemoveAll(layerDir)
+			_ = os.Remove(home)
```
rhatdan: Why are you removing home?  That was passed in,  I don't thing the function should remove this.

drivers/overlay/overlay.go
```diff
@@ -264,23 +264,39 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+		}()
+		_ = idtools.MkdirAs(mergedDir, 0700, rootUID, rootGID)
```
rhatdan: @rhvgoyal Can we get away with fewer directories.  Do we actually need an upper directory?

drivers/overlay/overlay.go
nalind: This bit runs before we "commit" to using the driver, so if the directory's empty except for the things we've just created, we try remove it, but not recursively.  If it fails, fine, but if it succeeds, then we avoid confusing `scanPriorDrivers()`.

drivers/overlay/overlay.go
nalind: If we wanted to mount the overlay for read-only use, I think we could.  The overlay filesystem has requirements for the upper directory that it doesn't have for lower directories, though, so I think it's more useful to mimic the eventual use case, to avoid false "yeah, it's fine" results.
This function call was moved after the checks on the kernel version and of the underlying filesystem's type, so that when we'll reject the driver based on one of those checks, we won't bother with this check.

rhatdan:
LGTM but I would like @rhvgoyal to appove.

drivers/overlay/overlay.go
```diff
@@ -124,10 +123,6 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 		return nil, err
 	}
 
-	if err := supportsOverlay(); err != nil {
-		return nil, errors.Wrap(graphdriver.ErrNotSupported, "kernel does not support overlay fs")
-	}
-
 	// require kernel 4.0.0 to ensure multiple lower dirs are supported
 	v, err := kernel.GetKernelVersion()
```
rhvgoyal: Shold we get rid of this kernel version check also. This will not work with rhel kernels. In fact, while you are doing overlay tests, you can try to mount multiple lower directories and if that works, you know overlay is supported at the same time kernel is new enough to support multiple lower and there should not be any need for version check.

drivers/overlay/overlay.go
rhvgoyal: Upper dir is not must. If upper is not being passed in, then workdir is not required as well. All we need is multiple lower dirs and one merged dir.

drivers/overlay/overlay.go
nalind: That's a good point.  I'll drop this bit.

drivers/overlay/overlay.go
nalind: In the general case you're right, but all of the mounts we end up making in the driver use an upper, and I think the kernel's check on whether or not the upper directory supports d_type is a useful thing to notice at this point, so I'm inclined to leave that detail unchanged unless someone insists that it be changed.

rhatdan:
I love the fact that we got rid of the kernel check.  My last concern is overhead.  For CRI-O this check will only happen once.  But for every buildah and kpod command we will not be doing 5 mkdir/rmdir and one mount before we do anything useful.  Should we add a flag that just says only do this when the user does not specify the driver in the config?

drivers/overlay/overlay.go
rhvgoyal: Kernel d_type check just emits  a warning. So it does not give you any extra information whether upper supports d_type or not. I think that will have to be a separate test. If filesystem is xfs, run xfs_info and check ftype=1.

nalind: Hmm, that's a good point: if the kernel's not going to reject that mount attempt, we learn nothing.  Moving the existing d_type check to be here so that it can use the temporary directory that we're already creating.

drivers/overlay/overlay.go
```diff
@@ -264,25 +232,59 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+	if err == nil {
+		// Check if reading the directory's contents populates the d_type field, which is required
+		// for proper operation of the overlay filesystem.
+		supportsDType, err = fsutils.SupportsDType(layerDir)
+		if err != nil {
+			return false, err
+		}
+		if !supportsDType {
+			logrus.Warn(overlayutils.ErrDTypeNotSupported("overlay", backingFs))
+			// TODO: Will make fatal when CRI-O Has AMI built on RHEL7.4
```
rhatdan: I believe we have this?  @mrunalp @runcom Please confirm.

rhatdan: Or should this comment be removed, since we can allow the caller to decide if it cares?

rhatdan:
@nalind @rhvgoyal Any comment on speed.  Should we just add a flag to storage.conf saying don't bother doing the check.

drivers/overlay/overlay.go
mrunalp: RHEL AMIs are based on 7.4 now. However Fedora AMI is still ext4 based.

rhvgoyal:
So when is this check done? Docker used to do it only when driver is initialized. Do we have similar notion here. I mean can we store the current driver in a file in internal metadata and do this check only if driver changes.

nalind:
@rhvgoyal This would be done at startup-time.  Knowledge of which driver we're using is either explicitly set in the configuration file, or deduced based on the presence of a subdirectory of the root of the storage area that matches the name of a driver.

rhatdan:
My concern is for single use tools like kpod and buildah versus cri-o, If crio takes the hit on startup, noone cares. if every kpod command takes a hit, that kind of stinks.

rhatdan:
Well I will merge and if we find performance to be an issue, we can reexamine this.

PR #136, opened 2017-11-09
overlay, vfs: Accept options without the driver name prefix

nalind:
In the overlay and vfs drivers, accept option names which we'd recognize, if they had the driver name as a prefix, as aliases for their proper names.


rhatdan:
LGTM

PR #137, opened 2017-11-14
ostree: new option for the overlay driver

giuseppe:
if `ostree_repo` is specified, then any layer is first committed to the specified ostree repository and then checkout out to the same location so that files are deduplicated using OSTree hardlinks.

The `exclude_ostree` BUILDTAG can be used to disable ostree support.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 EPOCH_TEST_COMMIT := 0418ebf59f9e1f564831c0ba9378b7f8e40a1c73
 NATIVETAGS := exclude_graphdriver_devicemapper exclude_graphdriver_btrfs exclude_graphdriver_overlay
-AUTOTAGS := $(shell ./hack/btrfs_tag.sh) $(shell ./hack/libdm_tag.sh)
+AUTOTAGS := $(shell ./hack/btrfs_tag.sh) $(shell ./hack/libdm_tag.sh) $(shell ./hack/ostree_tag.sh)
 BUILDFLAGS := -tags "$(AUTOTAGS) $(TAGS)" $(FLAGS)
 GO := go
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,13 +1,13 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 #
-#  The fedora/25-cloud-base and debian/jessie64 boxes are also available for
+#  The fedora/28-cloud-base and debian/jessie64 boxes are also available for
 #  the "virtualbox" provider.  Set the VAGRANT_PROVIDER environment variable to
 #  "virtualbox" to use them instead.
 #
 Vagrant.configure("2") do |config|
   config.vm.define "fedora" do |c|
-    c.vm.box = "fedora/25-cloud-base"
+    c.vm.box = "fedora/28-cloud-base"
     c.vm.synced_folder ".", "/vagrant", type: "rsync",
       rsync__exclude: "bundles", rsync__args: ["-vadz", "--delete"]
     c.vm.provision "shell", inline: <<-SHELL
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -129,6 +129,9 @@ Marks thinpool device for deferred deletion. If the thinpool is in use when the 
 
 Specifies the maximum number of retries XFS should attempt to complete IO when ENOSPC (no space) error is returned by underlying storage device. (default: 0, which means to try continuously.)
 
+**ostree_repo=""**
+  Tell storage drivers to use the specified OSTree repository.  Some storage drivers, such as overlay, might use
+
 # HISTORY
 May 2017, Originally compiled by Dan Walsh <dwalsh@redhat.com>
 Format copied from crio.conf man page created by Aleksa Sarai <asarai@suse.de>
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -132,6 +132,9 @@ Specifies the maximum number of retries XFS should attempt to complete IO when E
 **ostree_repo=""**
   Tell storage drivers to use the specified OSTree repository.  Some storage drivers, such as overlay, might use
 
+**skip_mount_home=""**
+  Tell storage drivers to not create a PRIVATE bind mount on their home directory.
+
 # HISTORY
 May 2017, Originally compiled by Dan Walsh <dwalsh@redhat.com>
 Format copied from crio.conf man page created by Aleksa Sarai <asarai@suse.de>
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -24,6 +24,7 @@ import (
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/locker"
 	"github.com/containers/storage/pkg/mount"
+	"github.com/containers/storage/pkg/ostree"
 	"github.com/containers/storage/pkg/parsers"
 	"github.com/containers/storage/pkg/system"
 	units "github.com/docker/go-units"
@@ -85,6 +86,7 @@ type overlayOptions struct {
 	imageStores         []string
 	quota               quota.Quota
 	fuseProgram         string
+	ostreeRepo          string
 }
 
 // Driver contains information about the home directory and the list of active mounts that are created using this driver.
@@ -99,6 +101,7 @@ type Driver struct {
 	naiveDiff     graphdriver.DiffDriver
 	supportsDType bool
 	locker        *locker.Locker
+	convert       map[string]bool
 }
 
 var (
@@ -164,6 +167,12 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 		return nil, err
 	}
 
+	if opts.ostreeRepo != "" {
+		if err := ostree.CreateOSTreeRepository(opts.ostreeRepo, rootUID, rootGID); err != nil {
+			return nil, err
+		}
+	}
+
 	d := &Driver{
 		name:          "overlay",
 		home:          home,
@@ -173,6 +182,7 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 		supportsDType: supportsDType,
 		locker:        locker.New(),
 		options:       *opts,
+		convert:       make(map[string]bool),
 	}
 
 	d.naiveDiff = graphdriver.NewNaiveDiffDriver(d, d)
@@ -240,6 +250,12 @@ func parseOptions(options []string) (*overlayOptions, error) {
 				return nil, fmt.Errorf("overlay: can't stat FUSE program %s: %v", val, err)
 			}
 			o.fuseProgram = val
+		case "overlay2.ostree_repo", "overlay.ostree_repo", ".ostree_repo":
+			logrus.Debugf("overlay: ostree_repo=%s", val)
+			if !ostree.OstreeSupport() {
+				return nil, fmt.Errorf("overlay: ostree_repo specified but support for ostree is missing")
+			}
+			o.ostreeRepo = val
 		default:
 			return nil, fmt.Errorf("overlay: Unknown option %s", key)
 		}
@@ -394,6 +410,11 @@ func (d *Driver) Create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 			return fmt.Errorf("--storage-opt size is only supported for ReadWrite Layers")
 		}
 	}
+
+	if d.options.ostreeRepo != "" {
+		d.convert[id] = true
+	}
+
 	return d.create(id, parent, opts)
 }
 
@@ -561,6 +582,12 @@ func (d *Driver) getLowerDirs(id string) ([]string, error) {
 func (d *Driver) Remove(id string) error {
 	d.locker.Lock(id)
 	defer d.locker.Unlock(id)
+
+	// Ignore errors, we don't want to fail if the ostree branch doesn't exist,
+	if d.options.ostreeRepo != "" {
+		ostree.DeleteOSTree(d.options.ostreeRepo, id)
+	}
+
 	dir := d.dir(id)
 	lid, err := ioutil.ReadFile(path.Join(dir, "link"))
 	if err == nil {
@@ -784,6 +811,13 @@ func (d *Driver) ApplyDiff(id string, idMappings *idtools.IDMappings, parent str
 		WhiteoutFormat: archive.OverlayWhiteoutFormat,
 	}); err != nil {
 		return 0, err
+	}
+
+	_, convert := d.convert[id]
+	if convert {
+		if err := ostree.ConvertToOSTree(d.options.ostreeRepo, applyDir, id); err != nil {
+			return 0, err
+		}
 	}
 
 	return directory.Size(applyDir)
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -87,6 +87,7 @@ type overlayOptions struct {
 	quota               quota.Quota
 	fuseProgram         string
 	ostreeRepo          string
+	skipMountHome       bool
 }
 
 // Driver contains information about the home directory and the list of active mounts that are created using this driver.
@@ -163,8 +164,10 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 		}
 	}
 
-	if err := mount.MakePrivate(home); err != nil {
-		return nil, err
+	if !opts.skipMountHome {
+		if err := mount.MakePrivate(home); err != nil {
+			return nil, err
+		}
 	}
 
 	if opts.ostreeRepo != "" {
@@ -256,6 +259,12 @@ func parseOptions(options []string) (*overlayOptions, error) {
 				return nil, fmt.Errorf("overlay: ostree_repo specified but support for ostree is missing")
 			}
 			o.ostreeRepo = val
+		case "overlay2.skip_mount_home", "overlay.skip_mount_home", ".skip_mount_home":
+			logrus.Debugf("overlay: skip_mount_home=%s", val)
+			o.skipMountHome, err = strconv.ParseBool(val)
+			if err != nil {
+				return nil, err
+			}
 		default:
 			return nil, fmt.Errorf("overlay: Unknown option %s", key)
 		}
--- a/drivers/vfs/driver.go
+++ b/drivers/vfs/driver.go
@@ -9,6 +9,7 @@ import (
 	"github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/chrootarchive"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/ostree"
 	"github.com/containers/storage/pkg/system"
 	"github.com/opencontainers/selinux/go-selinux/label"
 )
@@ -42,6 +43,27 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 			d.homes = append(d.homes, strings.Split(option[12:], ",")...)
 			continue
 		}
+		if strings.HasPrefix(option, "vfs.ostree_repo=") {
+			if !ostree.OstreeSupport() {
+				return nil, fmt.Errorf("vfs: ostree_repo specified but support for ostree is missing")
+			}
+			d.ostreeRepo = option[16:]
+		}
+		if strings.HasPrefix(option, ".ostree_repo=") {
+			if !ostree.OstreeSupport() {
+				return nil, fmt.Errorf("vfs: ostree_repo specified but support for ostree is missing")
+			}
+			d.ostreeRepo = option[13:]
+		}
+	}
+	if d.ostreeRepo != "" {
+		rootUID, rootGID, err := idtools.GetRootUIDGID(uidMaps, gidMaps)
+		if err != nil {
+			return nil, err
+		}
+		if err := ostree.CreateOSTreeRepository(d.ostreeRepo, rootUID, rootGID); err != nil {
+			return nil, err
+		}
 	}
 	return graphdriver.NewNaiveDiffDriver(d, graphdriver.NewNaiveLayerIDMapUpdater(d)), nil
 }
@@ -53,6 +75,7 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 type Driver struct {
 	homes      []string
 	idMappings *idtools.IDMappings
+	ostreeRepo string
 }
 
 func (d *Driver) String() string {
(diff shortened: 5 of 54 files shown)


giuseppe:
@rhatdan, I've pushed another version with all the comments addressed ⬆️

drivers/ostree/driver.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// Create prepares the filesystem for the OSTREE driver and copies the directory for the given id under the parent.
+func (d *Driver) Create(id, parent string, opts *graphdriver.CreateOpts) error {
+	d.convert[id] = true
+	return d.overlay.Create(id, parent, opts)
+}
+
+func getDir(home, id string) string {
+	return fmt.Sprintf("%s/%s/diff", home, id)
```
rhatdan: We tend to use filepath.Join() for this type of thing.

drivers/ostree/driver.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// Create prepares the filesystem for the OSTREE driver and copies the directory for the given id under the parent.
+func (d *Driver) convertToOSTree(root, id string) error {
+	if _, err := d.repo.PrepareTransaction(); err != nil {
+		return fmt.Errorf("Could not prepare the OSTree transaction: %v", err)
```
rhatdan: Use 
errors.Wrapf(err, "could not prepare the OSTree transaction")

Standard is to have errors start with lowercase. I believe.

drivers/ostree/driver.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	commitOpts := otbuiltin.NewCommitOptions()
+	commitOpts.Timestamp = time.Now()
+	commitOpts.Parent = "0000000000000000000000000000000000000000000000000000000000000000"
+	branch := fmt.Sprintf("ocilayer/%s", id)
+
+	if _, err := d.repo.Commit(root, branch, commitOpts); err != nil {
+		return fmt.Errorf("Could not commit the layer: %v", err)
```
rhatdan: errors.Wrapf.

drivers/ostree/driver.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if _, err := d.repo.CommitTransaction(); err != nil {
+		return fmt.Errorf("Could not complete the OSTree transaction: %v", err)
```
rhatdan: Ditto

drivers/ostree/driver.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if err := system.EnsureRemoveAll(root); err != nil {
+		return err
```
rhatdan: errors.Wrapf(err, "failed to remove %q", root)

drivers/ostree/driver.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	checkoutOpts := otbuiltin.NewCheckoutOptions()
+	if err := otbuiltin.Checkout(d.repoLocation, root, branch, checkoutOpts); err != nil {
+		return fmt.Errorf("Could not checkout from OSTree: %v", err)
```
rhatdan: errors.Wrapf

drivers/ostree/driver.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return nil
+}
+
+// Remove deletes the content from the directory for a given id.
+func (d *Driver) Remove(id string) error {
+	branch := fmt.Sprintf("ocilayer/%s", id)
+	if err := exec.Command("ostree", fmt.Sprintf("--repo=%s", d.repoLocation), "refs", "--delete", branch).Run(); err != nil {
+		return fmt.Errorf("Could not delete OSTree branch: %v", err)
```
rhatdan: errors.Wrapf("Could not delete OSTree branch")

drivers/ostree/driver.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return nil
+}
+
+// Get returns the directory for the given id.
+func (d *Driver) Get(id, mountLabel string) (string, error) {
+	path, err := d.overlay.Get(id, mountLabel)
```
rhatdan: return d.overlay.Get(id, mountLabel)

drivers/ostree/driver_stub.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/containers/storage/drivers"
+	"github.com/containers/storage/pkg/idtools"
+)
+
+func init() {
+	graphdriver.Register("ostree", Init)
+}
+
+func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (graphdriver.Driver, error) {
+	return nil, fmt.Errorf("ostree driver not supported")
```
rhatdan: Use errors.Errorf()

drivers/ostree/driver.go
rhatdan: lowercase

giuseppe:
thanks for the quick review.  I've addressed the new comments

rhatdan:
@nalind Should we add a new driver or enhance the Overlay driver to use OSTREE via an option?  
overlay.

drivers/ostree/driver.go
giuseppe: I looked at the code and it already gives a similar error message, so I think it would be duplicated `return errors.Wrapf(e, "error while removing %s", dir)`.  In some other cases it returns directly the error from `os.RemoveAll()`

drivers/ostree/driver.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (d *Driver) Put(id string) error {
+	root := getDir(d.home, id)
+
+	_, convert := d.convert[id]
+	if convert {
+		err := d.convertToOSTree(root, id)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
giuseppe: yes, it is defined in this module and it already gives useful errors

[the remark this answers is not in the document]
rhatdan: Thanks for checking.

giuseppe:
I've changed it to be an option for overlay instead of a new driver

rhatdan:
I find your description of the PR to be very confusing, you talk about skopeo requiring a patch that is different then this patch set.  Can you fix up the commit message for the patch.

Also you are using a different BUILDTAG for elimination of ostree then containers/image uses. Wouldn't we need to use both BUILDTAGS to to not require OSTREE?

giuseppe:
sure, I forgot to update the PR after I changed the implementation to be an option instead of a new driver.

rhatdan:
What about `exclude_ostree` versus `containers_image_ostree_stub`

giuseppe:
should I use `containers_image_ostree_stub`?  That name sounds specific to containers/image, that is why I've used exclude_ostree, but the end result is the same

rhatdan:
I hate the name containers_image_ostree_stub and much prefer exclude_ostree. but I think to get this to work we would need to specify both?  @mitrmac WDYT
Could we change the boolean in containers/image to exclude_ostree or are we too late?  I guess we could add it to containers/image and support both?

nalind:
How much space are we typically saving when enabling this?

giuseppe:
I've done only a quick test:

```
## with overlay
# for i in gscrivano/etcd gscrivano/flannel; do \
  skopeo copy docker://$i containers-storage:\[overlay@/var/lib/containers/storage+/var/run/containers/storage]$i
  done
# du --summarize -x --si /var/lib/containers/storage/
406M        /var/lib/containers/storage/

## with ostree
# for i in gscrivano/etcd gscrivano/flannel; do \
  skopeo copy docker://$i containers-storage:\[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay2.ostree_repo=/var/lib/containers/storage/overlay2/ostree/.repo]$i
  done
# du --summarize -x --si /var/lib/containers/storage/
340M        /var/lib/containers/storage/
```

rhatdan:
16% savings, but the real question would be if you had 100 images.

giuseppe:
yes, it works better with more images, the more images you have the bigger chance to have the same files coming from different layers.

I've tried with a bunch of images, this is the script I've used:

https://gist.github.com/giuseppe/95f917fa271bcdc6555851cc4774d83a

Please be careful that it wipes the entire overlay storage, edit the path to something else if you care to keep it.

This is what I got:

```
3.2G	/var/lib/containers/storage
2.1G	/var/lib/containers/storage-ostree/
```

rhatdan:
@mtrmac Would this cause verification to fail since timestamps will end up different?

rhatdan:
@vbatts ^^

mtrmac:
> @mtrmac Would this cause verification to fail since timestamps will end up different?

(Disclaimer: I really don’t know all that much about c/storage.) Looking at https://github.com/containers/storage/blob/master/layers.go , `layerStore.ApplyDiff` and `layerStore.Diff` use tar-split at the general level, before invoking the underlying drivers to record the actual files. IIUC tar-split stores the raw tar headers, i.e. including the timestamps, so the filesystem timestamps should not matter — at least as far as c/image and extracting the image is concerned, at least. If the on-filesystem timestamps did not match the original layers, this would probably be visible, and detectable, from within a container.

rhatdan:
@vbatts Could you confirm whether this is going to be a problem?

vbatts:
@rhatdan as long as it is an opt-in basis. The C linkage to libostree makes this difficult for minikube, and I'm likely openshift too

vbatts:
oh, for tar-split ... lemme think

rhatdan:
Yes you will be able to compile it out.

vbatts:
for the re-assembly of a layer it would not use inode attributes of the files, only needing the path to the file and the contents of the file to be intact.
If you build an entirely new layer and export it, then the initial tar pushed would use the mtimes of the filesystem at the time of push (producing a tar object).

I don't see any technical issues, though users expectation of mtimes may be subject to confusion. 

If there were a thin copy-up for attribute changes, then this could be a case for mtree (https://github.com/vbatts/go-mtree) so the prepared rootfs could have it's attributes like mtime restored from what the tar archive described.

giuseppe:
I've started working on OSTree to allow deduplication of files with different SELinux labels using reflinks where possible, such as XFS on new kernels:

With XFS reflinks we will have different inodes, so if needed we can setup different attributes for a file even if it is sharing the payload with a different file.

https://github.com/ostreedev/ostree/pull/1443

This is only a first step, I will also need to:

- change the lookup in OSTree to find the same files in parent repositories
- drop the bind mount on `/var/lib/containers/storage/[driver]`

This will allow to share files with OSTree on the host, and system containers.

This is a local optimization, it doesn't change the way images are pulled or stored on the server/registry.

cgwalters:
There are some upsides and downsides to this that I see.  Space savings is good, but OTOH right now libostree's locking story isn't fully fleshed out; see https://github.com/ostreedev/ostree/pull/1292 

And even if we had the locking fully done, this would mean installing/removing containers could block/slow host updates.  Of course we could prioritize host updates.  Another issue related to this is that personally right now I treat my container storage as fully disposable.  If it's entangled with the host storage, that gets harder.

I feel like we should be focusing on the system container storage path first; those usually change with the host, and are critical for bootstrap (i.e. they're distinct from app containers).

That all said this is a really interesting space to explore for sure.

giuseppe:
I agree with you.  I think we should still have a separate ostree repository for the app containers.  It is important to be able to easily wipe the entire containers storage if required without affecting the rest of the system.

alexlarsson:
Could we use ostree for the store, and then rofiles-fuse --copyup instead of overlayfs for deduplicated COW non-root support?

giuseppe:
@alexlarsson I guess so, but wouldn't that be a completely different storage driver?

After the discussion with running Buildah as non-root user, I agree that an `ostree` storage could be a very useful one, and much better than my first version of this patchset where I was adding a new `ostree` storage driver that was a clone of overlay+ostree storage.

alexlarsson:
Yeah, a pure ostree storage driver seem useful. There are some possible issues with it not preserving mtimes though.

rhatdan:
Sounds good, my only question would be do we want both.  This one for deduplication of resources, so less storage is used with overlay and the other for the non-privileged use case.  Not sure preservice mtimes is a big deal when building container images.

giuseppe:
I think we want both.  The pure ostree storage driver is useful in the rootless case, while having deduplication with overlay is another feature.

Is there anything blocking this PR?

rhatdan:
Did we ever fix up the ostree buildtag issues?

giuseppe:
I changed it to `exclude_ostree`, is that fine or should I use a different buildtag?

rhatdan:
No that is fine, but I want to get the same buildatag into containers/image  So we have consistency.

mtrmac:
c/image currently uses `containers_image_ostree_stub`.  Consistency in Go tags is just another way to say “namespace collision” ;-), though I do see how it is attractive for consumers of multiple libraries.

rhatdan:
Right can't we add exclude_ostree also to containers/image, so if the user of the library sets that flag it will not get compiled in.  Although I would think we want a positive instead of a negative. Like just ostree, This is what we do with SELinux and Apparmor.  If you specify a buildahtag og ostree you get the bindings, if you don't you don't get them.

mtrmac:
\*shrug\* I can live with that.

rhatdan:
@giuseppe Could you reverse this so ostree and !ostree are the build tags to match selinux, apparmor etc...

giuseppe:
I've changed "exclude_ostree" to "!ostree" :arrow_up:

giuseppe:
I've added a new option for skipping the bind mount on `/var/lib/containers/storage/overlay*`, so that the ostree repository can be placed outside of that directory, and more important: we can use an ostree parent repository for deduplicating files in two different storages.

.travis.yml
```diff
@@ -5,10 +5,12 @@ go:
     - 1.7
 dist: trusty
 sudo: required
+env:
+  - TAGS='!ostree'
```
rhatdan: Shouldn't just not having TAGS defined make this work?

rhatdan:
You need to add code to Makefile ALLTAGS to discover if it should use BUILDTAG=ostree.

Something like
ls -l hack/*tag.sh
-rwxrwxr-x. 1 dwalsh dwalsh 132 Sep 29 15:32 hack/btrfs_tag.sh
-rwxrwxr-x. 1 dwalsh dwalsh 332 Sep 29 15:32 hack/libdm_tag.sh

.travis.yml
giuseppe: yes true, I'll drop that

giuseppe:
changes done and tests pass again :arrow_up:

rhatdan:
LGTM

drivers/driver_linux.go
```diff
@@ -59,6 +59,7 @@ var (
 		"btrfs",
 		"zfs",
 		"vfs",
+		"ostree",
```
nalind: There's no call to the `Register()` function with this name, so I don't think this does anything.

giuseppe: this chunk must be deleted as now I am not adding a new driver but only an option for the overlay driver.

Thanks for spotting it!

drivers/overlay/ostree.go
```diff
@@ -0,0 +1,122 @@
...
+		return errors.Wrap(err, "could not complete the OSTree transaction")
+	}
+
+	if err := system.EnsureRemoveAll(root); err != nil {
+		return err
+	}
+
+	checkoutOpts := otbuiltin.NewCheckoutOptions()
+	checkoutOpts.RequireHardlinks = true
+	checkoutOpts.Whiteouts = true
```
nalind: Whiteout information is stored differently for overlay than it is by aufs, and the format is converted when creating and extracting archives using pkg/archives.  What does this flag cause ostree to do?

nalind:
Is there a particular rationale for calling out to the ostree CLI in `createOSTreeRepository()` and `deleteOSTree()`, but using the `otbuiltins` package in `convertToOSTree()`?

drivers/overlay/ostree.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+		} else if usermode && (info.Mode().IsRegular()) {
+			if err := os.Chmod(fullpath, info.Mode()|0600); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
```
nalind: This function might be simpler if it used `path/filepath.Walk()`.

drivers/overlay/ostree.go
giuseppe: it tells ostree to handle files with a prefix ".wh." (i.e. delete the existing file/directory) in the checkout directory.

giuseppe:
> Is there a particular rationale for calling out to the ostree CLI in createOSTreeRepository() and deleteOSTree(), but using the otbuiltins package in convertToOSTree()?

these functionalities are not exposed by the Golang bindings yet :-)

drivers/overlay/ostree.go
giuseppe: done in the new version

drivers/overlay/ostree.go
```diff
@@ -0,0 +1,110 @@
...
+	return true
+}
+
+func fixFiles(dir string, usermode bool) error {
+	return filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if info.Mode()&(os.ModeNamedPipe|os.ModeSocket|os.ModeDevice) != 0 {
+			if err := os.Remove(path); err != nil {
+				return err
+			}
+			if info.IsDir() {
```
nalind: This looks like it's nested at the wrong level - a directory is unlikely to be a pipe/socket/device.

drivers/overlay/ostree.go
nalind: That's how aufs stores whiteouts (and I'm not sure that what you're describing handles opaque directories), but in overlay whiteouts are character device nodes with major/minor of 0/0.  (Under [whiteouts and opaque directories](https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt) in the overlayfs docs.)

drivers/overlay/ostree.go
```diff
@@ -0,0 +1,110 @@
...
+				if usermode {
+					if err := os.Chmod(path, info.Mode()|0700); err != nil {
+						return err
+					}
+				}
+			} else if usermode && (info.Mode().IsRegular()) {
```
nalind: This also looks like it's nested at the wrong level.

drivers/overlay/ostree.go
nalind: @cgwalters Does ostree handle opaque directory whiteouts?  I took a quick look at `ostree-repo-checkout.c` and I couldn't find logic specific to handling entries named `.wh..wh..opq`.

nalind: @cgwalters It appears not, then.

giuseppe: I've opened a PR to support this: https://github.com/ostreedev/ostree/pull/1486

nalind: Thanks for that.
If I'm reading it right, `fixFiles` removes file whiteouts from diff directories, and that's a problem.

giuseppe: ah thanks to have found this issue.  I thought there were no whiteouts files, I've changed the code so that there is no special handling of these files.

rhatdan:
@giuseppe Could you setup a WIP vendor of this into Buildah or Podman and attempt to remove some files and rename a directory and see if everything works correctly.

giuseppe:
I've tried a custom build for podman, I needed to manually workaround some build issues caused by the revendoring (not caused by containers/storage, in facts I can see the same just by running `vndr` on origin master).

This change is required in podman to start using ostree:
```
--- a/hack/ostree_tag.sh
+++ b/hack/ostree_tag.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
-if ! pkg-config ostree-1 2> /dev/null ; then
+if pkg-config ostree-1 2> /dev/null ; then
+       echo ostree
+else
        echo containers_image_ostree_stub
 fi
```

and this is what I have in my /etc/containers/storage.conf:
```

ostree_repo = "/ostree/repo"

skip_mount_home = "true"
```

after that everything in my limited testing works as expected in the container:

```
/usr # cd /
/ # ls -l
total 56
drwxr-xr-x    2 root     root         12288 Jan  1  1970 bin
drwxr-xr-x    5 root     root           360 Mar 14 17:50 dev
drwxr-xr-x    1 root     root          4096 Mar 14 17:50 etc
drwxr-xr-x    2 root     root          4096 Mar 14 17:51 foo
drwxr-xr-x    2 nobody   nogroup       4096 Jan  1  1970 home
dr-xr-xr-x  340 root     root             0 Mar 14 17:50 proc
drwx------    1 root     root          4096 Mar 14 17:50 root
dr-xr-xr-x   13 root     root             0 Mar 14 11:23 sys
drwxrwxrwt    2 root     root          4096 Jan  1  1970 tmp
drwxr-xr-x    1 root     root          4096 Mar 14 17:51 usr
drwxr-xr-x    4 root     root          4096 Jan  1  1970 var
/ # touch /usr/foo
/ # rm /usr/foo
/usr # mkdir bar
/usr # touch bar/file
/usr # mv bar baz
/usr # rm -rf baz
```

nalind:
The main thing I'd want to be sure of is that we're handling whiteout information correctly, since overlay uses device nodes for whiteout, and extended attributes to mark opacity.  Specifically, I'd want to be sure that if an image creates files or directories in one layer, then removes them in a second layer that is applied on top of that first layer, that we're not losing that information when we extract the image into local storage.  Likewise, if a directory is marked as opaque, that the commit/checkout step doesn't discard that information.

One way to verify this would be to find a multi-layered image whose history shows it creating files in one layer, then using "rm" to remove those files in a different layer, creating whiteout information in that layer, or an image where a directory was marked as opaque, signifying that it had been removed and then recreated within a single layer.  (That an image does this could be verified for certain by copying the image to a "dir:" location with skopeo, and examining the layer blobs directly using "tar tvf".  Whiteout entries for files and directories will have `.wh.` prefixed to their basenames, and opaque directories contain a `.wh..opq.` "file".)

For example, `docker.io/ubuntu@sha256:d3fdf5b1f8e8a155c17d5786280af1f5a04c10e95145a515279cf17abdf0191f` contains a layer with blobsum `3009ec50c8871f7a3f19c248194270a1a3aa28705934d619acbd96c663ce5c89` which removes the files in `/var/lib/apt/lists/`, while `docker.io/redis@sha256:26c93c5b06eaa323bb1089500f42b0dd158138772348b865e364127f1d554982` contains a layer with blobsum `67b22db27e5144c14efdf55d9cef9c3dadffb43593790ec99663fb1b783dbb6b` which marks `/var/lib/apt/lists/partial` as opaque.

The root filesystem of a working container created by versions of buildah built with and without the patch, if both versions are forced to download the image, should have the same set of files, with the same contents.  In particular, deleted files for which whiteout exists in the image's blobs, and the contents of directories from layers before they were marked as opaque, should not be present in the container.

giuseppe:
@nalind I had  to do one change in the `fixFiles` function.  When we encounter a file that ostree does not know how to manage (as a block device) we skip the ostree deduplication for the layer.  I needed to do this to support the 0/0 character devices whiteout.

We could add support for these files later on, but first it requires some changes in the ostree golang bindings otherwise we need to handle the entire logic of pruning and recreating these files.

With the new version I've pushed the whiteout layers are handled correctly and the deleted files are not visible from the container.

giuseppe:
@nalind does this look fine now?

rhatdan:
@giuseppe Lets rebase and get this merged.

giuseppe:
bot, retest this please

rhatdan:
LGTM

rhatdan:
@nalind PTAL

giuseppe:
I've rebased it again.  Is it useful to have this feature or should I close this PR?

rhatdan:
I believe this is useful.  As long as we make it easy to compile out.

giuseppe:
@rhatdan @nalind it is possible to control if OSTree is built or not.  I've verified that whiteout device files are kept

giuseppe:
@nalind @rhatdan I've added a test for the ostree deduplication part.  I've added it to vfs as well so that can be tested on Travis as well.  I've tried it on overlay locally using a Vagrant VM

rhatdan:
@giuseppe Tests are failing.

giuseppe:
@rhatdan fixed.  I needed to set the mtime to the EPOCH so it doesn't trigger a difference when OSTree is used.

giuseppe:
the only missing part that I would like to work on as a follow-up is dropping the `ostree` CLI usage and add the missing features to the go bindings (delete branch and init repository).

rhatdan:
LGTM
@nalind PTAL

giuseppe:
rebased

giuseppe:
@rhatdan fine to merge this?

rhatdan:
Going to finally merge this one.

PR #138, opened 2017-11-16
Add a notion of image "digests" based on their "manifest" data

nalind:
Add a "digest" of an image that is equal to the digest of its big data item named "manifest", if it has one, that we can index and use for locating images.  We'll be able to use this in the image library to help it locate images using canonical (digest-based) references.

### DIFF
--- a/images.go
+++ b/images.go
@@ -14,6 +14,13 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// ImageDigestBigDataKey is the name of the big data item whose
+	// contents we consider useful for computing a "digest" of the
+	// image, by which we can locate the image later.
+	ImageDigestBigDataKey = "manifest"
+)
+
 // An Image is a reference to a layer and an associated metadata string.
 type Image struct {
 	// ID is either one which was specified at create-time, or a random
@@ -28,7 +35,7 @@ type Image struct {
 	// TopLayer is the ID of the topmost layer of the image itself, if the
 	// image contains one or more layers.  Multiple images can refer to the
 	// same top layer.
-	TopLayer string `json:"layer"`
+	TopLayer string `json:"layer,omitempty"`
 
 	// Metadata is data we keep for the convenience of the caller.  It is not
 	// expected to be large, since it is kept in memory.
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
The code LGTM, but would like others to review.  FWIW, weird Travis error:
```
$ make local-binary docs local-cross local-validate
ffjson images.go
make: ffjson: Command not found
make: *** [images_ffjson.go] Error 127
```

nalind:
Right, that's a hazard of patching `images.go` in a way that doesn't end up generating changes in `images_ffjson.go` (maybe #126 wasn't the right thing to do there).  I did find a missing piece of #108 that lets us work around it.

images.go
```diff
@@ -486,6 +523,17 @@ func (r *imageStore) BigDataNames(id string) ([]string, error) {
 	return image.BigDataNames, nil
 }
 
+func imageSliceWithoutValue(slice []*Image, value *Image) []*Image {
```
rhatdan: Pretty horrible name.
removeImage()?
pruneImage()?

images.go
```diff
@@ -528,6 +576,28 @@ func (r *imageStore) SetBigData(id, key string, data []byte) error {
...
 			save = true
 		}
+		if key == ImageDigestBigDataKey {
+			if oldDigest != "" && oldDigest != newDigest {
+				// remove the image from the list of images in the digest-based
+				// index which corresponds to the old digest for this item
+				if list, ok := r.bydigest[oldDigest]; ok {
+					prunedList := imageSliceWithoutValue(list, image)
+					if len(prunedList) == 0 {
+						delete(r.bydigest, oldDigest)
```
rhatdan: If you delete thie r.bydigest, will the 

 +			list := r.bydigest[newDigest] 

Still work?

nalind: Yes, it returns the zero value, which is `nil`.

images.go
nalind: It'd probably be `pruneImageFromSlice()` or something like that, since we're not changing anything else.  This name is intentionally similar to `stringSliceWithoutValue` from `store.go`.

rhatdan: Ok, not crazy about the name, but if it consistently sucks...

rhatdan:
LGTM

PR #139, opened 2017-11-28
Print warning when shutdown fails

rhatdan:


rhatdan:
@nalind Right now we end up with storage still being mounted in kpod and their is no indication what is going on.

store.go
```diff
@@ -2304,6 +2304,7 @@ func (s *store) Shutdown(force bool) ([]string, error) {
 		}
 	}
 	if len(mounted) > 0 && err == nil {
+		logrus.Warn("a container/storage layer is still in use, storage shutdown will not complete")
```
TomSweeneyRedHat: Need to include logrus.

TomSweeneyRedHat: Verbiage is OK, just wondering if OpenShift might complain about log spam with this and if this should be debug level to protect against that.  I think .Warn is fine, but thought I'd mention it.

rhatdan: This is an error/rare situation, Well not rare since it  happens all the time in kpod run, but we need to figure out why.  But it should not be happening in cri-o.

rhatdan:
@nalind PTAL

store.go
nalind: If there's still a container running, it's an expected result.  We pretty much only bother with trying to shut down storage for testing cases, where not shutting things down means the test leaves temporary storage locations mounted, which the post-test cleanup scripts don't handle.

rhatdan: Well sure it might be expected by containers/storage but the users do not know that they screwed up somewhere, the images are left mounted and their is nothing logged.  As in the case of kpod run (Which has been mostly fixed)  

I think this situation should be logged.

nalind: At this level, we don't know that a mistake has been made, and I'm leery of adding a log message that we'll end up telling people to ignore.

What's the problem case that we're trying to diagnose here?

rhatdan: Basically I had used the library in kpod run and it was failing silently.  Eventually I fixed kpod to clean up after itself. but there was no way to know that this was failing other then noticing that thee was a huge amount of mounts not being umounted.

rhatdan:
Well since @nalind does not want it, I will close.

PR #141, opened 2017-12-05
Check for overlay merging optimization

nalind:
Check if the overlay filesystem implements an optimization that landed in 97c684cc911060ba7f97c0925eaf842f159a39e8, and in the mainline kernel in 4.10:  directories created in merged directories are marked as opaque to let the kernel know that it needn't bother looking at lower layers when reading the contents of that directory.

This means that, when generating a layer diff for an upper directory, we can't treat the presence of an opaque whiteout entry as an indication that a layer diff needs to include whiteout for a directory of the same name from a lower layer.  If we do, when that diff is later applied, the directory gets created, and then removed, and then only recreated (possibly with different permissions) if the diff also places contents in the directory.

### DIFF
--- a/drivers/btrfs/btrfs_test.go
+++ b/drivers/btrfs/btrfs_test.go
@@ -58,6 +58,10 @@ func TestBtrfsSubvolDelete(t *testing.T) {
 	}
 }
 
+func TestBtrfsEcho(t *testing.T) {
+	graphtest.DriverTestEcho(t, "btrfs")
+}
+
 func TestBtrfsTeardown(t *testing.T) {
 	graphtest.PutDriver(t)
 }
--- a/drivers/devmapper/devmapper_test.go
+++ b/drivers/devmapper/devmapper_test.go
@@ -86,6 +86,10 @@ func TestDevmapperTeardown(t *testing.T) {
 	graphtest.PutDriver(t)
 }
 
+func TestDevmapperEcho(t *testing.T) {
+	graphtest.DriverTestEcho(t, "devicemapper")
+}
+
 func TestDevmapperReduceLoopBackSize(t *testing.T) {
 	tenMB := int64(10 * 1024 * 1024)
 	testChangeLoopBackSize(t, -tenMB, defaultDataLoopbackSize, defaultMetaDataLoopbackSize)
--- a/drivers/graphtest/graphtest_unix.go
+++ b/drivers/graphtest/graphtest_unix.go
@@ -4,15 +4,18 @@ package graphtest
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"math/rand"
 	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"testing"
 	"unsafe"
 
 	"github.com/containers/storage/drivers"
+	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/docker/go-units"
 	"github.com/pkg/errors"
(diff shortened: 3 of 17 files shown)


rhatdan:
LGTM

rhatdan:
@rhvgoyal PTAL

nalind:
[(5) [containers/storage] Add tests for layer diff contents](https://trello.com/c/VV1Q408f/656-5-containers-storage-add-tests-for-layer-diff-contents)

rhvgoyal:
@nalind I understand the kernel change but don't understand the user space descrption. What is "opaque whiteout". I know about opaque attribute and I know about "whiteout" but don't know what is opaque whiteout and hence don't understand what problem "opaque" xattr is causing.

Secondly, why not make it same as upstream overlay2 check. Which checks for REDIRECT_DIR also and falls back to naivediff interface.

rhvgoyal:
@nalind Also upstream docker does not fall back to using naivediff interface if this optimization is present. In fact it will be present for newer kernels. So solution for this is not to fall back to naivediff interface instead be able to deal with opaque interface. naivediff is slow. 

looks like archive package of docker can deal with opaque xattr. can't we do same thing?

rhatdan:
@nalind Any movement on this one?

nalind:
Whiteouts are entries in a layer tar archive which are used to indicate that items, presumably added in previous layers, should be removed when applying the layer's contents to a directory tree as part of the iterative process of creating a root filesystem for a container.  The conventions used in layer tar archives are modeled after the ones used by aufs.

When generating change lists or diffs, the current logic generates whiteout for a directory (indicated, in the style of aufs, by placing a file named `..wh..wh..opq` in the directory) if the `trusted.overlay.opaque` extended attribute on the directory as set to `y`.  Now that this extended attribute is set more often, an archive that adds a new subdirectory, which should contain entries for the parent directory and the new subdirectory, now also includes a whiteout entry.

For example, building an image named "foo" using the Dockerfile 
```
FROM alpine
RUN mkdir /root/subdirectory && touch /root/subdirectory/regularfile
```
and then using `skopeo copy` to copy its layers to a temporary subdirectory using `skopeo copy docker-daemon:foo:latest dir:/path/to/subdir`, and then using `tar tf` to list the contents of the archive, should produce
```
root/
root/subdirectory/
root/subdirectory/regularfile
```

Instead, we get
```
root/
root/subdirectory/
root/subdirectory/.wh..wh..opq
root/subdirectory/regularfile
```

When applying that layer, the `.wh..wh..opq` is treated as an instruction to delete the directory, which is incorrect.  If the directory has contents, it will be recreated automatically (which makes this hard to spot in the wild), but not necessarily with the same permissions.  If the directory doesn't have contents, it's just deleted.

A different but similar code path is used for `docker diff`, and can be observed by running `docker run --name foo alpine sh -c 'mkdir /root/subdirectory && touch /root/subdirectory/regularfile'` and using `docker diff foo` to list the changes (**C**hanges, **A**dditions, **D**eletions) made to the container.  Instead of
```
C /root
A /root/subdirectory
A /root/subdirectory/regularfile
```
we get
```
C /root
D /root/subdirectory
A /root/subdirectory/regularfile
```

I'm revising this patch to check for the presence of similarly-named items in lower layers when deciding whether or not to consider an opaque directory as having been removed and recreated, rather than as simply having been created.

rhatdan:
@rhvgoyal How does this look to you?

rhvgoyal:
Ok, so presence of opaque xattr was being taken as if file has been deleted in previous layer and it was listed as "D" in changes list. But now opaque xattr can be present even if there is no such directory in lower layer. (As on optimization for overlay).

So this does not really break a working setup? It is just little semantics thing whether an opque directory should be listed as "D" or not in changes list or not (when nothing has been removed to begin with).

rhvgoyal:
@nalind Does it slow down the build (or commit operation), because now you are looking in lower layers to figure out if directory is present or not. (I am assuming previously you did not have to).

rhvgoyal:
@nalind Also does this change mean that we take the optimization away from overlay when we pull that this image from repos. Will opaque xattr be present in pulled image layer? If not, that means we are basically undoing the ovl_lookup() optimization overlay put it.

If yes, then we need to decide is this problem worth fixing. All opaque suggests that if there is a directory underneath layers, delete it. If it is not there, nobody should care. I understand there is little semantic difference when it comes to "Changes" list.

Is this issue big enough which needs fixing and take potential performance hit at other places?

nalind:
The output of the overlay driver's "changes" logic (which backends "docker diff") is not used by the overlay driver's "diff" logic (which is used to produce layer diffs), but they both get tripped up by the way the opaque attribute is used and produce different results than other storage drivers do.

This incorrect handling of the opaque attribute breaks things if the directory which is marked as opaque either has no content (in which case we produce a layer that includes only opaque whiteout for that directory, and extracting that layer doesn't create the directory, even though it should exist) or needs permissions or ownership other than an unspecified default (since the directory will be created so that its contents can be put down on the filesystem, but the correct permission and ownership information will have been lost).  If, for example, such a directory needs ownership and permissions set on it so that a process running as a non-root user in the container can access the directory's contents, that's a critical problem.

The added logic for catching this case is all but certain to slow down a build, but keep in mind that the baseline for comparison is still NaiveDiff.

The layer diffs themselves don't include opaqueness encoded as an extended attribute, since they use aufs-style formats for representing whiteout information.  Opaque directories are marked as such by including a "file" named `.wh..opq.` under the corresponding directory in the archive, along with the directory itself.  On extraction, the presence of the `.wh..opq.` entry in a directory is interpreted as an instruction to add the `trusted.overlay.opaque`=`y` attribute to the directory.  Because this change is directly applied to the `diff` tree as part of extracting it, rather than being applied to a directory on a mounted overlay filesystem, the new kernel-side optimization is not, and would not have been, triggered when extracting downloaded layers.

rhvgoyal:
nalin mentioned that normally we do restore opaque xattr if image is pulled on a system where overlay graphdriver is being used (if opaque was packed in layer diff to being with).

That means that after this change, we will not pack overlay xattr in layer diff if there is no such directory in lower layers and that means we will not restore it when pulled on a different system and that means this little kernel optimzation will be lost for pure upper directory case. 

I guess we don't have a choice for the sake of keep Changes() output consistent.

rhatdan:
Ok I am going to merge.

PR #142, opened 2017-12-08
Add explicitly-settable digest values for images

nalind:
Add explicitly-settable digest values for images that are treated mostly like the implicit digests that we track for manifests, for the sake of v1 images which have manifests that need to be preprocessed before being digested to produce an image's digest value.  We'll still have the digest of the unprocessed manifest, but that shouldn't hurt.

We add a test to ensure that an image with no implicit value can be found, and a test that an image with the same implicit and explicit value is still only returned once in a ByDigest() list.


images.go
```diff
@@ -165,9 +168,17 @@ func (r *imageStore) Load() error {
...
 			if digest, ok := image.BigDataDigests[ImageDigestBigDataKey]; ok {
 				digests[digest] = append(digests[digest], images[n])
 			}
+			// Explicit digest
+			if image.Digest == "" {
+				image.Digest = image.BigDataDigests[ImageDigestBigDataKey]
+			}
+			if image.Digest != "" && image.Digest != image.BigDataDigests[ImageDigestBigDataKey] {
+				digests[image.Digest] = append(digests[image.Digest], images[n])
+			}
```
rhatdan: Looks like these lines can be combined
```
if image.Digest == "" {
				image.Digest = image.BigDataDigests[ImageDigestBigDataKey]
} else if image.Digest != image.BigDataDigests[ImageDigestBigDataKey] {
				digests[image.Digest] = append(digests[image.Digest], images[n])
}
```

rhatdan: Do we need to do anything with digests[image.Digest] if image.Digest was == ""?

nalind: You're right, we don't.  Changing it as you've suggested.

rhatdan:
LGTM

PR #144, opened 2018-01-03
Turn DType testing back on

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

Reverting #106 to error on DType rather than just warn as support should have been added to the CI for this.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -245,9 +245,7 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 			return false, err
 		}
 		if !supportsDType {
-			logrus.Warn(overlayutils.ErrDTypeNotSupported("overlay", backingFs))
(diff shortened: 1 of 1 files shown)


drivers/overlay/overlay.go
```diff
@@ -245,9 +245,7 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 			return false, err
 		}
 		if !supportsDType {
-			logrus.Warn(overlayutils.ErrDTypeNotSupported("overlay", backingFs))
-			// TODO: Will make fatal when CRI-O Has AMI built on RHEL7.4
-			// return nil, overlayutils.ErrDTypeNotSupported("overlay", backingFs)
+			return nil, overlayutils.ErrDTypeNotSupported("overlay", backingFs)
```
rhatdan: Looks like this should be false not nil

nalind:
LGTM.

rhatdan:
LGTM

PR #147, opened 2018-01-19
Pass MountLabel down to diff drivers

rhatdan:
Currently when we do a commmit, we are mounting the container without using
the mountlabel.  In certain situations we can leak mount points where the
image is already mounted with a label.  If you then attempt to commit the
image, the kernel will attempt to mount the image without a label.  The
kernel will reject this mount since SELinux does not allow the same image
to be mounted with different labels.

Passing down the label to the diff drivers, fixes this issue.


rhatdan:
@nalind @rhvgoyal PTAL

drivers/fsdiff.go
```diff
@@ -34,7 +34,7 @@ type NaiveDiffDriver struct {
 //     Diff(id, parent string) (io.ReadCloser, error)
 //     Changes(id, parent string) ([]archive.Change, error)
 //     ApplyDiff(id, parent string, diff io.Reader) (size int64, err error)
-//     DiffSize(id, parent string) (size int64, err error)
+//     DiffSize(id, parent, string) (size int64, err error)
```
nalind: Best to name that third argument, for consistency.

layers_ffjson.go
```diff
@@ -2,1712 +2,3 @@
...
-	}
-	err = fs.Error.ToError()
-	if err != nil {
-		return fs.WrapErr(err)
-	}
-	panic("ffjson-generated: unreachable, please report bug.")
-done:
-
-	return nil
-}
```
nalind: Something went very wrong here.

rhvgoyal:
I had a quick look and this patch looks reasonable. Assuming this has been tested and it works fine (with commit being done while mount point has been leaked).

LGTM

rhatdan:
Would a test even work?  Since we have the other fixes in place.

rhvgoyal:
Which other fixes? We can mount run a container and then try to mount its rootfs on a different mount point and see if it works or not.

rhvgoyal:
Sorry, rather leak mount point of a container and then force commit (That will do Diff).

rhatdan:
Ok I created a version of buildah that vendors this patch.
```
ctr=$(buildah from fedora)
mnt=$(buildah mount $ctr)
touch $mnt/dan
```
In a separate terminal
```
unshare -m /bin/sh
```
Back in the original terminal
```
buildah commit $ctr dan1
buildah umount $ctr
```
Commit succeeds and mount point is no longer visible. In second terminal mount point is still mounted, but no errors when saving the image.  So I believe this patch fixes the problem.

rhatdan:
@nalind PTAL

TomSweeneyRedHat:
LGTM

drivers/windows/windows.go
```diff
@@ -188,7 +188,7 @@ func (d *Driver) CreateReadWrite(id, parent string, opts *graphdriver.CreateOpts
 }
 
 // Create creates a new read-only layer with the given id.
-func (d *Driver) Create(id, parent string, opts *graphdriver.CreateOpts) error {
+func (d *Driver) Create(id, parent, mountLabel string, opts *graphdriver.CreateOpts) error {
```
nalind: Is this left over from an earlier iteration of the patch?

rhatdan: Fixed

nalind:
LGTM

jwhonce:
[(3) podman needs to support sd_notify](https://trello.com/c/w9AdJtHT/754-3-podman-needs-to-support-sdnotify)

PR #149, opened 2018-01-24
Update to match latest code in moby

rhatdan:


rhatdan:
@nalind We can either merge this or just start using it directly from Moby,  WDYT?

rhatdan:
No that was a bug, Hopefully tests will pass now.

nalind:
Looks fine, but since we're not using the index's `Iterate()` method, can we list the motivation for the update in the commit log?

rhatdan:
Merging.

PR #150, opened 2018-01-25
lockfile: reshape for multiplatform

vbatts:
this stubs out for having platform-independent file locking

This still functions the same on `GOOS=linux`, and now succeeds for
`GOOS=windows`. But still fails on `freebsd`, `darwin`, `solaris`, etc


rhatdan:
@nalind PTAL

nalind:
This breaks compilation when GOOS=darwin, which is at least expected to work by skopeo.  Moving the version of `TouchedSince()` that's in `lockfile_unix.go` to a file that's built only on `linux` and `solaris`, and changing the build tags on the original file so that it, too, is built for `darwin` and `freebsd` cases, should fix that.  (We'll still have a couple of type mismatches breaking the build for `freebsd`, but it's progress.)

vbatts:
Ok I'll update this. When I tried darwin first it failed, so I wasn't sure
that it was supposed to work in the first place.

vbatts:
updated PTAL

nalind:
LGTM

rhatdan:
LGTM

Issue #151, opened 2018-02-16
backport recent copy_file_range changes from moby


redbaron:
Refs:
 - https://github.com/moby/moby/pull/34670
 - https://github.com/moby/moby/pull/35537

giuseppe:
it is being addressed with: https://github.com/containers/storage/pull/246

rhatdan:
Since this was merged, I can close this issue.

PR #152, opened 2018-02-16
update vfs and zfs from Moby

rhatdan:
This patch brings in quota support for vfs layers. Also uses new copy tool from storage.

### DIFF
--- a/drivers/copy/copy.go
+++ b/drivers/copy/copy.go
@@ -1,0 +1,277 @@
+// +build linux
+
+package copy
+
+/*
+#include <linux/fs.h>
+
+#ifndef FICLONE
+#define FICLONE		_IOW(0x94, 9, int)
+#endif
+*/
+import "C"
+import (
+	"container/list"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	"github.com/containers/storage/pkg/pools"
+	"github.com/containers/storage/pkg/system"
+	rsystem "github.com/opencontainers/runc/libcontainer/system"
+	"golang.org/x/sys/unix"
+)
+
+// Mode indicates whether to use hardlink or copy content
+type Mode int
+
+const (
+	// Content creates a new file, and copies the content of the file
+	Content Mode = iota
+	// Hardlink creates a new hardlink to the existing file
+	Hardlink
+)
+
+func copyRegular(srcPath, dstPath string, fileinfo os.FileInfo, copyWithFileRange, copyWithFileClone *bool) error {
+	srcFile, err := os.Open(srcPath)
+	if err != nil {
+		return err
+	}
+	defer srcFile.Close()
+
+	// If the destination file already exists, we shouldn't blow it away
+	dstFile, err := os.OpenFile(dstPath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, fileinfo.Mode())
+	if err != nil {
+		return err
+	}
+	defer dstFile.Close()
+
+	if *copyWithFileClone {
+		_, _, err = unix.Syscall(unix.SYS_IOCTL, dstFile.Fd(), C.FICLONE, srcFile.Fd())
+		if err == nil {
+			return nil
+		}
+
+		*copyWithFileClone = false
+		if err == unix.EXDEV {
+			*copyWithFileRange = false
+		}
+	}
+	if *copyWithFileRange {
+		err = doCopyWithFileRange(srcFile, dstFile, fileinfo)
+		// Trying the file_clone may not have caught the exdev case
+		// as the ioctl may not have been available (therefore EINVAL)
+		if err == unix.EXDEV || err == unix.ENOSYS {
+			*copyWithFileRange = false
+		} else {
+			return err
+		}
+	}
+	return legacyCopy(srcFile, dstFile)
+}
+
+func doCopyWithFileRange(srcFile, dstFile *os.File, fileinfo os.FileInfo) error {
+	amountLeftToCopy := fileinfo.Size()
+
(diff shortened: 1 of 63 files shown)


rhatdan:
I think this addresses https://github.com/containers/storage/issues/151

redbaron:
from the look of it, with this change VFS is best driver if you run on a recent kernel:
 - great simplicity of Go code, all heavy lifting is done in linux kernel
 - no mount operations, therefore avoiding all those nasty bugs which keep coming up with mounts leaking into other namespaces, etc, devices busy, usernamespaces etc
 - no exotic FS, XFS is as mature as FS can be. Full compatibility with POSIX (afaik there were some problems with tar on overlay fs, don't know how it isnow).
 - comparing to ZFS, BTRFS or LVM, XFS COW shares pagecache, so it becomes as memory efficient as aufs & overlayfs

rhatdan:
@nalind @TomSweeneyRedHat PTAL

drivers/copy/copy.go
```diff
@@ -0,0 +1,277 @@
...
+
+			if err := os.Symlink(link, dstPath); err != nil {
+				return err
+			}
+
+		case os.ModeNamedPipe:
+			fallthrough
+		case os.ModeSocket:
+			if rsystem.RunningInUserNS() {
+				// cannot create a device if running in user namespace
```
nalind: How does this comment relate to the `Mknod()` call a few lines later?  Are we expected to silently discard sockets and named pipes, but fail on device nodes?

drivers/copy/copy_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	copyWithFileRange := true
+	copyWithFileClone := false
+	doCopyTest(t, &copyWithFileRange, &copyWithFileClone)
+
+	if kernel.CompareKernelVersion(*v, kernel.VersionInfo{Kernel: 4, Major: 5, Minor: 0}) < 0 {
+		assert.False(t, copyWithFileRange)
+	} else {
+		assert.True(t, copyWithFileRange)
+	}
```
nalind: Since I expect we're going to want this feature to be used when the `FICLONE` ioctl has been backported, can we make this test key off of something other than the reported kernel version?

drivers/graphtest/graphtest_unix.go
```diff
@@ -312,27 +313,42 @@ func writeRandomFile(path string, size uint64) error {
 }
 
 // DriverTestSetQuota Create a driver and test setting quota.
-func DriverTestSetQuota(t *testing.T, drivername string) {
+func DriverTestSetQuota(t *testing.T, drivername string, required bool) {
```
nalind: This is only being called with `required` set to `false`.  Is it supposed to be called with `required` set to `true` somewhere?

drivers/quota/projectquota.go
```diff
@@ -329,9 +350,35 @@ func makeBackingFsDev(home string) (string, error) {
...
+	defer free(cs)
+	var qstat C.fs_quota_stat_t
+
+	_, _, errno := unix.Syscall6(unix.SYS_QUOTACTL, uintptr(C.Q_XGETQSTAT_PRJQUOTA), uintptr(unsafe.Pointer(cs)), 0, uintptr(unsafe.Pointer(&qstat)), 0, 0)
+	if errno == 0 && qstat.qs_flags&C.FS_QUOTA_PDQ_ENFD > 0 && qstat.qs_flags&C.FS_QUOTA_PDQ_ACCT > 0 {
+		return true, nil
+	}
+
+	switch errno {
+	// These are the known fatal errors, consider all other errors (ENOTTY, etc.. not supporting quota)
```
nalind: Should that be "non-fatal", since we suppress them?

drivers/copy/copy.go
rhatdan: I have no idea, this code comes directly from Moby.

drivers/copy/copy_test.go
rhatdan: Any idea how?

drivers/graphtest/graphtest_unix.go
rhatdan: Moby has an update to zfs that calls it with true, I updated the zfs driver to show the use case.

drivers/quota/projectquota.go
rhatdan: I think these errors indicate that the system does NOT support quota as opposed to the other errors would indicate that the caller can not determine whether the system supports quota.

drivers/copy/copy_test.go
rhatdan: @rhvgoyal Any suggestions?

drivers/copy/copy.go
rhatdan: I opened an issue on this in Moby to see if anyone there has an idea.
https://github.com/moby/moby/issues/36622

drivers/copy/copy_test.go
redbaron: This check has just been removed in https://github.com/moby/moby/pull/36736

drivers/copy/copy.go
rhatdan: Upstream fixed this by moving the check to the MkNod section.

drivers/copy/copy_test.go
rhatdan: @redbaron Thanks for pointing this out, I have removed it from this PR also.

rhatdan:
@nalind Looks like this patch is a little more ready to merge now.

redbaron:
@rhatdan , is it done in another PR?

rhatdan:
No I am going to ask @nalind To look at this.  I think there are a lot of gotchas in the merge, that I need him to concentrate on.

PR #153, opened 2018-02-19
overlay: perform the support check after creating homedir

nalind:
The supports-dtype check tries to create a subdirectory of the driver's home directory to use for its checks, so it needs to be called after that directory has been created.

If it fails, go ahead and try to rmdir() the directories that we've created to avoid creating a false positive indicator that this driver works, but let the attempt to remove them fail in case we've had previous successful startups and there's actually data here.


nalind:
@rhatdan @TomSweeneyRedHat PTAL

rhatdan:
LGTM

PR #154, opened 2018-02-19
overlay: always mount, even if we have no parent layer

nalind:
Always mount an overlay filesystem, even when there's no parent layer, so that we can set an SELinux context for the mountpoints.  We create an "empty" directory to use as the lower directory for the overlay mount to satisfy the kernel, but as it's not a real layer, we don't track it in the lists of lower directories.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -181,7 +181,7 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 		return nil, fmt.Errorf("Storage option overlay.size only supported for backingFS XFS. Found %v", backingFs)
 	}
 
-	logrus.Debugf("backingFs=%s, projectQuotaSupported=%v", backingFs, projectQuotaSupported)
+	logrus.Debugf("backingFs=%s, projectQuotaSupported=%v, useNativeDiff=%v", backingFs, projectQuotaSupported, !useNaiveDiff(home))
 
 	return d, nil
 }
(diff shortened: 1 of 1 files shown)


nalind:
@rhatdan @TomSweeneyRedHat PTAL

drivers/overlay/overlay.go
```diff
@@ -181,7 +181,7 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 		return nil, fmt.Errorf("Storage option overlay.size only supported for backingFS XFS. Found %v", backingFs)
 	}
 
-	logrus.Debugf("backingFs=%s, projectQuotaSupported=%v", backingFs, projectQuotaSupported)
+	logrus.Debugf("backingFs=%s, projectQuotaSupported=%v, useNativeDiff=%v", backingFs, projectQuotaSupported, !useNaiveDiff(home))
```
TomSweeneyRedHat: I know not part of this change, but curious why the variable useNaiveDiff and useNaiveLock are Naive and not Native?

nalind: `NaiveDiffDriver` is the name of the wrapper object that several of the storage drivers use to implement their `Changes()`/`Diff()`/`DiffSize()`/`ApplyDiff()` methods, so I expect that variables related to deciding whether or not its use is necessary are named after it.  The "native" implementation in the overlay driver uses knowledge of how the overlay filesystem works to gain speed advantages, but it can't cope with certain kernel features, so that decision has to be made at startup-time.

TomSweeneyRedHat: Thanks nalind!

TomSweeneyRedHat:
LGTM, would like to get another pair of eyeballs on it.

rhatdan:
LGTM and a little simpler.

PR #156, opened 2018-03-01
Add support for per-layer/per-container ID mapping

nalind:
Extend support for specifying ID mappings to use so that they can be specified on a per-layer or per-container basis, with defaults for layers where mappings are not specified being to inherit the mappings of the parent layer, or from the configuration in storage.conf.

We extend the overlay driver's "diff" subdirectory with peers to hold files and directories that we force the kernel to copy up.


nalind:
[(5) [containers/storage] Support ID mapping in the storage library](https://trello.com/c/tzOu5Rh5/858-5-containers-storage-support-id-mapping-in-the-storage-library)

drivers/chown.go
```diff
@@ -0,0 +1,171 @@
...
+	chownByMapsCmd = "storage-chown-by-maps"
+)
+
+func init() {
+	reexec.Register(chownByMapsCmd, chownByMapsMain)
+}
+
+func chownByMapsMain() {
+	if len(os.Args) < 1 {
+		fmt.Printf("requires mapping configuration on stdin and directory path")
```
rhatdan: Should this be logrus?  Is the library going to exec chown?

rhatdan:
@mtrmac @giuseppe PTAL

drivers/overlay/overlay.go
```diff
@@ -562,8 +562,35 @@ func (d *Driver) Get(id, mountLabel string) (_ string, retErr error) {
...
+	}
+	diffN := 1
+	_, err = os.Stat(filepath.Join(dir, nameWithSuffix("diff", diffN)))
+	for err == nil {
+		if absLowers == "" {
+			absLowers = filepath.Join(dir, nameWithSuffix("diff", diffN))
+			relLowers = dumbJoin(string(link), "..", nameWithSuffix("diff", diffN))
+		} else {
+			absLowers = absLowers + ":" + filepath.Join(dir, nameWithSuffix("diff", diffN))
+			relLowers = relLowers + ":" + dumbJoin(string(link), "..", nameWithSuffix("diff", diffN))
```
mtrmac: Nit: this would probably be a bit more convenient, and easier to follow, with all the `*lowers` variables a `[]string`, only joining them with colons at the very end.

drivers/overlay/overlay.go
```diff
@@ -580,15 +607,36 @@ func (d *Driver) Get(id, mountLabel string) (_ string, retErr error) {
...
 		}
-		if newlowers == "" {
-			newlowers = lower
+		diffN = 1
+		_, err = os.Stat(dumbJoin(lower, "..", nameWithSuffix("diff", diffN)))
+		for err == nil {
+			lower = lower + ":" + dumbJoin(lower, "..", nameWithSuffix("diff", diffN))
+			l = l + ":" + dumbJoin(l, "..", nameWithSuffix("diff", diffN))
+			diffN++
+			_, err = os.Stat(dumbJoin(lower, "..", nameWithSuffix("diff", diffN)))
```
mtrmac: How does this work after `lower = lower + ":" + …` above? Is that really a valid path?

(And does this need to the separate temporary variable `l`, or could it directly append to `relLowers`?)

drivers/overlay/overlay.go
```diff
@@ -771,3 +827,80 @@ func (d *Driver) Changes(id, parent, mountLabel string) ([]archive.Change, error
...
+		return string(os.PathSeparator)
+	}
+	result := ""
+	for _, n := range names {
+		if result != "" {
+			result = result + string(os.PathSeparator)
+		}
+		result = result + n
+	}
+	return result
```
mtrmac: Isn’t this just a `strings.Join(names, os.PathSeparator)`?

pkg/archive/archive.go
```diff
@@ -1019,14 +1022,27 @@ func untarHandler(tarArchive io.Reader, dest string, options *TarOptions, decomp
...
-	archive, err := TarWithOptions(src, &TarOptions{Compression: Uncompressed})
+	tarMappings := archiver.TarIDMappings
+	if tarMappings == nil {
+		tarMappings = &idtools.IDMappings{}
+	}
+	options := &TarOptions{
+		UIDMaps: tarMappings.UIDs(),
+		GIDMaps: tarMappings.GIDs(),
+	}
+	archive, err := TarWithOptions(src, options)
```
mtrmac: This drops `Compression: Uncompressed`. Is that intentional?

cmd/containers-storage/create.go
```diff
@@ -22,14 +25,124 @@ var (
...
+			mapping := idtools.IDMap{
+				ContainerID: int(cid),
+				HostID:      int(hid),
+				Size:        int(size),
+			}
+			options.UIDMap = append(options.UIDMap, mapping)
+		}
+	}
+	if len(paramGIDMap) > 0 {
+		gidSpec := strings.Fields(strings.Map(nonDigitsToWhitespace, paramGIDMap))
```
mtrmac: (The triple parsing code might be big enough to extract to a function shared for `paramUIDMap` and `paramGIDMap`.  The way `nonDigitsToWhitespace` throws away information worries me a bit.)

store.go
```diff
@@ -2412,6 +2544,89 @@ func init() {
...
+			return 0, 0, 0, fmt.Errorf("error parsing id map value %q: %v", spec[1], err)
+		}
+		sz, err := strconv.ParseUint(spec[2], 10, 32)
+		if err != nil {
+			return 0, 0, 0, fmt.Errorf("error parsing id map value %q: %v", spec[2], err)
+		}
+		return uint32(cid), uint32(hid), uint32(sz), nil
+	}
+	if len(config.Storage.Options.RemapUIDs) > 0 {
+		uidSpec := strings.Fields(strings.Map(nonDigitsToWhitespace, config.Storage.Options.RemapUIDs))
```
mtrmac: (Two more copies of the same fairly big parser.)

drivers/chown.go
nalind: Yes, this is invoked via `reexec`.  I want to let the caller use the output text to create an error, so it's preferable that it not include logrus timestamp/level information at this point.

pkg/archive/archive.go
mtrmac: > `Uncompressed` is the default value

Ah, never mind them; ISTR some other part of the package used to default to compression, and I thought this code change changed behavior.

If the behavior doesn’t change, I don’t have any preference WRT making the decision explicit vs. making the code shorter.

adelton:
Attempting to build CRI-O with `vendor/github.com/containers/storage` from this pull request results in

```
# make
go build -i -ldflags '-s -w -X main.gitCommit="01b118116d3f188869a368d40561a2095aa57998-dirty" -X main.buildInfo=1520335360' -tags "seccomp btrfs_noversion exclude_graphdriver_devicemapper libdm_no_deferred_remove exclude_graphdriver_btrfs containers_image_ostree_stub  containers_image_ostree_stub" -o bin/crio github.com/kubernetes-incubator/cri-o/cmd/crio
# github.com/kubernetes-incubator/cri-o/vendor/github.com/containers/image/storage
_output/src/github.com/kubernetes-incubator/cri-o/vendor/github.com/containers/image/storage/storage_image.go:564:55: not enough arguments in call to s.imageRef.transport.store.PutLayer
	have (string, string, nil, string, bool, io.ReadCloser)
	want (string, string, []string, string, bool, *storage.LayerOptions, io.Reader)
make: *** [Makefile:88: crio] Error 2
```

It's expected that it fails as new parameters are being added.

@nalind, do you have some branch for CRI-O which would mirror the changes from the pull request (at least on the level of making CRI-O compile)? Or should I give it a stab? Looking at https://trello.com/c/Ak2yMcpf/714-epic-cri-o-support-for-user-namespaces it does not show any pull request yet.

adelton:
For start,

```
diff --git a/vendor/github.com/containers/image/storage/storage_image.go b/vendor/github.com/containers/image/storage/storage_image.go
index 038195c1..54970097 100644
--- a/vendor/github.com/containers/image/storage/storage_image.go
+++ b/vendor/github.com/containers/image/storage/storage_image.go
@@ -561,7 +561,8 @@ func (s *storageImageDestination) Commit() error {
                        return errors.Errorf("error applying blob %q: content not found", blob.Digest)
                }
                // Build the new layer using the diff, regardless of where it came from.
-               layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, diff)
+               options := (*storage.LayerOptions)(nil)
+               layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, options, diff)
                if err != nil {
                        return errors.Wrapf(err, "error adding layer with blob %q", blob.Digest)
                }
```

would be needed, so the change would also need to go to https://github.com/containers/image/.

nalind:
@adelton I haven't worked up a patch for the image library yet, but passing `nil` for new `LayerOptions` parameters is expected to be what the image library ends up doing -- it's not in the business of caring about this stuff, so sticking with the storage library's defaults by passing `nil` is fine.

A runtime engine could do the same (if it even calls any of the methods that grew a new argument) and start checking the maps that are now fields in `Container` structs that the storage library keeps track of and returns from `CreateContainer()`, and decide whether or not to set up a user namespace when running processes based on whether or not those map fields are empty.

As a next step, the runtime engine would start setting new `IDMappingOptions` fields in the `ContainerOptions` argument to `storage.Store.CreateContainer()`, to control what values come back in those fields.

I haven't worked up a branch for CRI-O that does that, so if you've got some cycles to rough out the patch, I'm happy to look at what you put together.  Otherwise I can start on that after I finish making the other changes that @mtrmac suggested.  Thanks!

drivers/overlay/overlay.go
nalind: Yeah, I was aiming for keeping the list of changes wrt moby's version smaller in case we rebase again, but we have tests for that, and that cleanup makes sense.

drivers/overlay/overlay.go
nalind: No, it wouldn't have.

drivers/overlay/overlay.go
nalind: Yup, that'll work better.

cmd/containers-storage/create.go
nalind: Factoring that out helped with readability.  I'm cheating with `nonDigestToWhitespace()` so that we don't have to enforce use of particular separators for the list of mappings if there's more than one mapping triplet.

nalind:
Okay, cleanups added.  I'm roughing out the changes to consumers using buildah, so permissions tweaks may be need to be added here for some of the parent directories that the storage library manages.

rhatdan:
@nalind are you still working on this, or is this ready for review/merge?

nalind:
@rhatdan Still working on it.  I think that some of the parent directories under the storage top-level directory will have to have their permissions relaxed in order to be able to launch containers in the directories below them that will be the root of filesystems.

adelton:
In https://github.com/kubernetes/kubernetes/pull/55707/files I propose `0751` instead of the original `0750`.

nalind:
Repushed with some tweaks to permissions setting that I _think_ will work for the newer use cases.

nalind:
Rebased and updated to account for changes made in #159.

cmd/containers-storage/create.go
```diff
@@ -22,14 +25,110 @@ var (
...
+	paramSubGIDMap    = ""
 )
 
+func paramIDMapping() (*storage.IDMappingOptions, error) {
+	options := storage.IDMappingOptions{
+		HostUIDMapping: paramHostUIDMap,
+		HostGIDMapping: paramHostGIDMap,
+	}
+	if paramHostUIDMap && (len(paramUIDMap) > 0 || paramSubUIDMap != "") {
+		return nil, fmt.Errorf("host uid map specified along with UID mapping")
```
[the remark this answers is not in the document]
nalind: This is the CLI wrapper.  `CreateLayer()` and `CreateContainer()` explicitly check for a conflict now, and the docs for `IDMappingOptions` now notes that the map is ignored if the corresponding `Host` flag is true.

cmd/containers-storage/create.go
```diff
@@ -22,14 +25,110 @@ var (
...
+	}
+	if paramHostGIDMap && (len(paramGIDMap) > 0 || paramSubGIDMap != "") {
+		return nil, fmt.Errorf("host gid map specified along with GID mapping")
+	}
+	if paramSubUIDMap != "" && paramSubGIDMap == "" {
+		return nil, fmt.Errorf("subuid map specified but no subgid map")
```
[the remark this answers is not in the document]
nalind: Changed this so that if you specify one but not the other, we assume the same name for both.  It slightly reduces the amount of work configuring this requires, and `useradd` already adds entries to both files.

nalind:
Rebased, adding `UIDMap()` and `GIDMap()` methods to `Store` objects to retrieve copies of the maps they were initialized with.

nalind:
Added a patch to aid troubleshooting by keeping track of which UIDs and GIDs were included in a layer diff when applying one.

nalind:
Dropped that patch because it broke one of our tests, but added more tests and some post-mount logic to ensure that parent directories are searchable.

nalind:
Okay, I think this is ready for review and merging, if it passes.

drivers/chown.go
rhatdan: Should this be stderr then?

drivers/chown.go
```diff
@@ -0,0 +1,167 @@
...
+		os.Exit(1)
+	}
+	// Read and decode our configuration.
+	discreteMaps := [4][]idtools.IDMap{}
+	config := bytes.Buffer{}
+	if _, err := config.ReadFrom(os.Stdin); err != nil {
+		fmt.Printf("error reading configuration: %v", err)
```
rhatdan: Should this be stderr?

drivers/chown.go
```diff
@@ -0,0 +1,167 @@
...
+		os.Exit(1)
+	}
+	if err := json.Unmarshal(config.Bytes(), &discreteMaps); err != nil {
+		fmt.Printf("error decoding configuration: %v", err)
```
rhatdan: Should this be stderr?

drivers/chown.go
```diff
@@ -0,0 +1,167 @@
...
+		os.Exit(1)
+	}
+	// Try to chroot.  This may not be possible, and on some systems that
+	// means we just Chdir() to the directory, so from here on we should be
+	// using relative paths.
+	if err := chrootOrChdir(os.Args[0]); err != nil {
+		fmt.Printf("error chrooting to %q: %v", os.Args[0], err)
```
rhatdan: ditto

drivers/chown.go
```diff
@@ -0,0 +1,167 @@
...
+// UIDs and GIDs for "host" values from the second map which correspond to the
+// same "container" IDs.
+func (n *naiveLayerIDMapUpdater) UpdateLayerIDMap(id string, toContainer, toHost *idtools.IDMappings, mountLabel string) error {
+	driver := n.ProtoDriver
+	layerFs, err := driver.Get(id, mountLabel)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		driver.Put(id)
```
rhatdan: Should we be checking "err" here, before Put?

drivers/chroot_unix.go
```diff
@@ -0,0 +1,23 @@
...
+
+// chrootOrChdir() is either a chdir() to the specified path, or a chroot() to the
+// specified path followed by chdir() to the new root directory
+func chrootOrChdir(path string) error {
+	if err := syscall.Chroot(os.Args[1]); err != nil {
+		fmt.Printf("error chrooting to %q: %v", os.Args[1], err)
+		os.Exit(1)
+	}
+	if err := syscall.Chdir(string(os.PathSeparator)); err != nil {
+		fmt.Printf("error changing to %q: %v", os.Args[1], err)
```
rhatdan: stderr?

drivers/chroot_unix.go
```diff
@@ -0,0 +1,23 @@
...
+	"fmt"
+	"os"
+	"syscall"
+)
+
+// chrootOrChdir() is either a chdir() to the specified path, or a chroot() to the
+// specified path followed by chdir() to the new root directory
+func chrootOrChdir(path string) error {
+	if err := syscall.Chroot(os.Args[1]); err != nil {
+		fmt.Printf("error chrooting to %q: %v", os.Args[1], err)
```
rhatdan: stderr?

drivers/chroot_windows.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"os"
+	"syscall"
+)
+
+// chrootOrChdir() is either a chdir() to the specified path, or a chroot() to the
+// specified path followed by chdir() to the new root directory
+func chrootOrChdir(path string) error {
+	if err := syscall.Chdir(path); err != nil {
+		fmt.Printf("error changing to %q: %v", os.Args[1], err)
```
rhatdan: stderr?

storage.conf
```diff
@@ -23,3 +23,24 @@ size = ""
...
+
+# Remap-User/Group is a name which can be used to look up one or more UID/GID
+# ranges in the /etc/subuid or /etc/subgid file.  Mappings are set up starting
+# with an in-container ID of 0 and the a host-level ID taken from the lowest
+# range that matches the specified name, and using the length of that range.
+# Additional ranges are then assigned, using the ranges which specify the
+# lowest host-level IDs first, to the lowest not-yet-mapped container-level ID,
+# until all of the entries have been used for maps.
+#
+# remap-user = "crio"
```
rhatdan: Should we use the name of a project?  Is buildah/podman/crio are sharing this config?

drivers/aufs/aufs.go
```diff
@@ -118,14 +118,10 @@ func Init(root string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
...
 
-	rootUID, rootGID, err := idtools.GetRootUIDGID(uidMaps, gidMaps)
-	if err != nil {
-		return nil, err
-	}
 	// Create the root aufs driver dir and return
 	// if it already exists
 	// If not populate the dir structure
-	if err := idtools.MkdirAllAs(root, 0700, rootUID, rootGID); err != nil {
+	if err := os.MkdirAll(root, 0711); err != nil {
```
rhatdan: I take it that this is a directory which will just be mounted on?  SO you just need search perms in the UserNamespace?

rhatdan:
@nalind Did you get this to work with Buildah?

rhatdan:
@adelton @mitrmac @mheon PTAL

rhatdan:
@mtrmac  PTAL

nalind:
@rhatdan Yes, it's happy to set up user namespaces, but it still needs to learn how to set things up for a network namespace, since we can't usefully use the host's network namespace for those cases.

drivers/aufs/aufs.go
nalind: It's the top level for the driver (`/var/lib/containers/storage/aufs`), while the filesystems that get mounted are at `mnt/$layerid` under it.  We've seen permissions-related problems preventing us from launching containers using user namespaces when they were using root directories which had some parent directory wasn't at least searchable by the UID/GID that we were mapping to 0/0 for the container.

drivers/chown.go
nalind: It can be, if we start checking for output using `CombinedOutput()` rather than `Output()`.  Changed it.

drivers/chown.go
nalind: Changed it.

drivers/chown.go
nalind: Changed it.

drivers/chown.go
nalind: Preferably not - we want the `Put()` (unmount) to happen if `Get()` (mount) succeeded, whether the chowning child process succeeds or not.

drivers/chroot_unix.go
nalind: Changed it.

drivers/chroot_unix.go
nalind: Changed it.

drivers/chroot_windows.go
nalind: Changed it.

storage.conf
nalind: It'll usually be set to the name of a user account that usermod added.  We don't create any unprivileged accounts in any of the tools (and yes, they all share this one), but I wanted to have a placeholder value in there.  Do you have any suggestions for what this should be set to?

drivers/chroot_windows.go
```diff
@@ -0,0 +1,16 @@
+package graphdriver
```
mheon: Shouldn't this be +build windows?

layers.go
```diff
@@ -178,13 +185,13 @@ type LayerStore interface {
 	// underlying drivers can accept a "size" option.  At this time, most
 	// underlying drivers do not themselves distinguish between writeable
 	// and read-only layers.
-	Create(id, parent string, names []string, mountLabel string, options map[string]string, writeable bool) (*Layer, error)
+	Create(id string, parent *Layer, names []string, mountLabel string, options map[string]string, writeable bool, moreOptions *LayerOptions) (*Layer, error)
```
mheon: Argument ordering nit: Wouldn't it make sense to have moreOptions immediately after options, instead of after writeable?

mheon:
A couple small comments, but overall LGTM.

drivers/chroot_windows.go
nalind: [For a while now](https://golang.org/doc/go1.4#filenames), that's been implicit when the source file's base name ends with `_` followed by an architecture or operating system name.

layers.go
nalind: Sure, why not?  Changed it.

storage.conf
rhatdan: storage?

drivers/aufs/aufs.go
rhatdan: Ok. SGTM

drivers/chroot_windows.go
rhatdan: Right _windows.go indicates only build this on Windows
_linux.go only on Linux.

drivers/chown.go
rhatdan: I hate the term put, I always translate that in my mind as "Write", but it means unmount in containers/storage world.  I agree you need it.

rhatdan:
LGTM once we change  CRI-O line to something like "storage" or "storageuser"

storage.conf
nalind: "storage" it is.  Changed it.

adelton:
For the record: I'm attempting to run Kubernetes with CRI-O from https://github.com/nalind/cri-o/tree/idmapping to test these changes, without specifying any `--userns*` options, to check that without any configuration changes, the existing functionality is preserved. Errors that I currently see are

```
E0323 21:06:11.818297   17786 remote_runtime.go:187] CreateContainer in sandbox "e6f030f6c9c82e1a6f2ecdd6be2f4012b68b05b383b20a72df38b152adeecbc3" from runtime service failed: rpc error: code = Unknown desc = container create failed: nsenter: failed to setns to /proc/18624/ns/user: Invalid argument
container_linux.go:348: starting container process caused "process_linux.go:301: running exec setns process for init caused \"exit status 22\""
E0323 21:06:11.818523   17786 kuberuntime_manager.go:733] container start failed: CreateContainerError: container create failed: nsenter: failed to setns to /proc/18624/ns/user: Invalid argument
container_linux.go:348: starting container process caused "process_linux.go:301: running exec setns process for init caused \"exit status 22\""
E0323 21:06:11.818647   17786 pod_workers.go:186] Error syncing pod 73bf9d52-2eaf-11e8-af99-525400461e23 ("kube-dns-5c6c5b55b-m9kpx_kube-system(73bf9d52-2eaf-11e8-af99-525400461e23)"), skipping: [failed to "StartContainer" for "kubedns" with CreateContainerError: "container create failed: nsenter: failed to setns to /proc/18624/ns/user: Invalid argument\ncontainer_linux.go:348: starting container process caused \"process_linux.go:301: running exec setns process for init caused \\\"exit status 22\\\"\"\n"
, failed to "StartContainer" for "dnsmasq" with CreateContainerError: "container create failed: nsenter: failed to setns to /proc/18624/ns/user: Invalid argument\ncontainer_linux.go:348: starting container process caused \"process_linux.go:301: running exec setns process for init caused \\\"exit status 22\\\"\"\n"
, failed to "StartContainer" for "sidecar" with CreateContainerError: "container create failed: nsenter: failed to setns to /proc/18624/ns/user: Invalid argument\ncontainer_linux.go:348: starting container process caused \"process_linux.go:301: running exec setns process for init caused \\\"exit status 22\\\"\"\n"
```

drivers/chroot_unix.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+// chrootOrChdir() is either a chdir() to the specified path, or a chroot() to the
+// specified path followed by chdir() to the new root directory
+func chrootOrChdir(path string) error {
```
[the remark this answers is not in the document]
nalind: Eek, that was a bad refactor.  That should be fixed now.

drivers/devmapper/deviceset.go
```diff
@@ -1688,11 +1684,7 @@ func (devices *DeviceSet) initDevmapper(doInit bool) (retErr error) {
 
 	//create the root dir of the devmapper driver ownership to match this
 	//daemon's remapped root uid/gid so containers can start properly
-	uid, gid, err := idtools.GetRootUIDGID(devices.uidMaps, devices.gidMaps)
-	if err != nil {
-		return err
-	}
-	if err := idtools.MkdirAs(devices.root, 0700, uid, gid); err != nil && !os.IsExist(err) {
+	if err := os.MkdirAll(devices.root, 0711); err != nil && !os.IsExist(err) {
```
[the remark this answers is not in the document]
nalind: Yeah, reverted this part.

rhatdan:
@mtrmac PTAL

nalind:
Rebased.

mtrmac:
> @mtrmac PTAL

Weak ACK, I guess?

> The permissions problem (`EACCES`) was encountered during a `mount()` with `MS_BIND` and `MS_REC` flags attempting to recursively bind mount the container's root filesystem in place

I can’t quite see how that happens in the kernel, at least for the recursive part (I can see possible places for permission checks at the top level of the mount in `fs/namespace.c`, but not for the recursive submounts—but then I don’t understand the `copy_tree` logic one bit, so that’s most likely my mistake).

> in a process which had been started in the new mount and user namespaces, after its ID mappings had already been set in /proc uid_map and gid_map special files.

This is `runc`’s `prepareRootFs`? Intuitively it surprises very much that this runs inside a restricted namespace; I would expect it to run as the host-wide `CAP_ADMIN`/`root`, with access to all namespaces and filesystems. But then I know absolutely nothing about `runc`, this may well have been done that way for a major reason, and moving the FS setup from the `nsenter` subprocess into a parent looks like a major surgery.

Ultimately, this PR now removes all permission changes and adds the ~harmless `LayerParentOwners`/`ContainerParentOwners` API, with actual use of them outside of this repo (?), so I have nothing left to question 😉

cmd/containers-storage/container.go
```diff
@@ -179,6 +179,37 @@ func getContainerRunDir(flags *mflag.FlagSet, action string, m storage.Store, ar
...
 
+func containerParentOwners(flags *mflag.FlagSet, action string, m storage.Store, args []string) int {
+	matched := []*storage.Container{}
+	for _, arg := range args {
+		if container, err := m.Container(arg); err == nil {
+			matched = append(matched, container)
+		}
+	}
+	if jsonOutput {
+		json.NewEncoder(os.Stdout).Encode(matched)
```
mtrmac: The JSON output is a list of (details of?) containers, but that seems to be different from the `ContainerParentOwners` computation in the non-JSON case.

cmd/containers-storage/layer.go
```diff
@@ -35,6 +35,37 @@ func layer(flags *mflag.FlagSet, action string, m storage.Store, args []string)
...
 
+func layerParentOwners(flags *mflag.FlagSet, action string, m storage.Store, args []string) int {
+	matched := []*storage.Layer{}
+	for _, arg := range args {
+		if layer, err := m.Layer(arg); err == nil {
+			matched = append(matched, layer)
+		}
+	}
+	if jsonOutput {
+		json.NewEncoder(os.Stdout).Encode(matched)
```
mtrmac: The JSON output is a list of (details of?) layers, but that seems to be different from the `LayerParentOwners` computation in the non-JSON case.

nalind:
@adelton The `chmod o+x` did not appear to be needed once the owner's host ID was at least mapped, and it was more expedient to drop it rather than to try to keep doing it, in a safe way.

adelton:
@nalind, I'm all for not doing the `chmod o+x` if not necessary.

What I'm missing is how it can work -- assume I want to run root (uid 0) in container as host uid 808080, and the parent directories are owned by (host) root with mode `0750`. Even if we map that host uid 0 to the container to say 808081, how does it help the process running as 808080 to access through the directories?

giuseppe:
@adelton I am having the same issue as I don't want root to be mapped, and working on: https://github.com/containers/storage/pull/171

Could that be helpful?

mtrmac:
> Even if we map that host uid 0 to the container to say 808081, how does it help the process running as 808080 to access through the directories?

That shouldn’t be necessary in the first place AFAICT: root on the host sets up the namespace at $topDir, with root of the filesystem hierarchy and cwd under $topDir.  Why should anything running in this namespace ever access any of the parent directories, or need any access to them at all?

adelton:
> @adelton I am having the same issue as I don't want root to be mapped, and working on: #171
> 
> Could that be helpful?

Nod, I do the similar change `s/0750/0751/` in https://github.com/kubernetes/kubernetes/pull/55707.

adelton:
Hmm, but when I try it now on master, you seem right -- that 036c8de763f95b53769cba027796bdfa9c3af196 does not seem to be needed, 8a5fa2e36e73d0c110426b31ddaad44c141749c1 seems enough.

mtrmac:
> Which seems to suggest that there's access to the parent directories happening from non-root uid.

Yes, because `runc` first switches to the container’s UID namespace, and _then_ starts mounting filesystems. I can’t see how that makes sense, but then I have no idea how `runc` works.

But I was commenting on a more narrow question:
> Even if we map that host uid 0 to the container to say 808081, how does it help the process running as 808080 to access through the directories?

And the answer to that, IMHO, should be “N/A the process as 808080 should never need to do that nor be ever able to do that”.

giuseppe:
> Yes, because runc first switches to the container’s UID namespace, and then starts mounting filesystems. I can’t see how that makes sense, but then I have no idea how runc works.

the same path is used also by the rootless use case, where an unprivileged user in the host gains privileges within the userNS and is able to use `mount`

mtrmac:
>>Yes, because runc first switches to the container’s UID namespace, and then starts mounting filesystems. I can’t see how that makes sense, but then I have no idea how runc works.
> the same path is used also by the rootless use case, where an unprivileged user in the host gains privileges within the userNS and is able to use mount

PR #157, opened 2018-03-05
Fix a lock ordering error in Shutdown()

nalind:
Acquire the big storage lock before the lock on the layer store, because they should always be acquired in that order if we're picking up both.


nalind:
CC @baude

baude:
@nalind this fixed the lock contention in libpod/podman ... i tested it with simultaneous podman operations pounding on the system and the lock issue never showed up.  Thanks!

rhatdan:
LGTM

PR #159, opened 2018-03-07
Always return deep copies of Layers, Images, and Containers

nalind:
Always copy slices and maps in Layer, Image, and Container structures before handing them back to callers so that, even if they modify them directly, they won't accidentally mess with our in-memory copies of those fields in the copies of the structures that we're using.


rhatdan:
LGTM
@mtrmac PTAL

rhatdan:
@mheon @bbaude PTAL

mheon:
LGTM

baude:
lgtm with two comments:

* when doing performance analysis in libpod, we see that deepcopies are expensive.
* would it make any sense to not export them so they cannot be accidently used?

mtrmac:
Makes sense to me, subject to the very limited extent I understand c/storage.

nalind:
We're already doing shallow copies of these items when returning them to callers, and I'd rather avoid making API changes at this time.

rhatdan:
Ok I am going to merge, and users will have to make sure they don't hit performance issues.

PR #162, opened 2018-03-22
Enable Stable + build/test on Fedora and CentOS

cevich:
* Added the few small changes needed to support these other distros.

* Enable building GO_VERSION "stable" on all distros, but ignore the
result.  The failures are seemingly similar, but beyond my ability to
debug.  For now, ignore the result to provide exposure and runtime
without any requirement to pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,15 +24,43 @@ env:
                       -e TRAVIS_JOB_ID=$TRAVIS_JOB_ID
                       -e TRAVIS_BUILD_DIR=$TRAVIS_BUILD_DIR"
     matrix:
-        # TODO: broken?
-        #- GO_VERSION="stable"
-        #  DISTRO="ubuntu"
-
-        - GO_VERSION="1.9"
+        # Ubuntu
+        - GO_VERSION="stable"
           DISTRO="ubuntu"
 
         - GO_VERSION="1.8"
           DISTRO="ubuntu"
+
+        - GO_VERSION="1.9"
+          DISTRO="ubuntu"
+
+        # Fedora
+        - GO_VERSION="stable"
+          DISTRO="fedora"
+
+        - GO_VERSION="1.9"
+          DISTRO="fedora"
+
+        - GO_VERSION="1.8"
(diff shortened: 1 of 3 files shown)


cevich:
W00t! Build & Test successful under Fedora and CentOS, and the results from "stable" are simply ignored (see #164) for now.  @nalind or @rhatdan PTAL when you get a chance (not urgent).

rhatdan:
LGTM

rhatdan:
BTW Can we change these tests to use podman to run instead of Docker?

cevich:
> BTW Can we change these tests to use podman to run instead of Docker?

Maybe I'm wrong, but I don't think podman can be installed and run on ubuntu Trusty.  
@lsm5 please correct me, can this be done, if so, HOWTO?

cevich:
@rhatdan I think let's make the `podman` idea a separate PR.  As above, it's possible to do, the travis-part is easy, only the make-it-work on ubuntu-Trusty could be challenging.  Either way, no need to hold up this multi-distro PR for it IMHO.

lsm5:
On Tue, Mar 27, 2018 at 11:21:34AM -0700, Chris Evich wrote:
> > BTW Can we change these tests to use podman to run instead of Docker?
> 
> Maybe I'm wrong, but I don't think podman can be installed and run on ubuntu Trusty.  
> @lsm5 please correct me, can this be done, if so, HOWTO?

Right, the PPA builds are currently xenial only.

Last time I checked, there were some dep issues for trusty, but maybe I can
take another look if we really need this. LMK..


-- 
Lokesh
IRC: lsm5
GPG: 0xC7C3A0DD
https://keybase.io/lsm5

cevich:
Thanks @lsm5 so as I suspected.  Let's leave that discussion up to upstream podman.  If it's needed (eventually, Trusty will go away) and solved, it's easy to add a new matrix axis here.

rhatdan:
SGTM

rhatdan:
Ok I will merge this and talk more about replacing Docker with podman.

PR #166, opened 2018-04-02
Running all CI tests using podman or docker

cevich:
Use podman by default unless ``$CONTAINER=docker``.
Same for local testing with ``hack/run_ci_tests.sh``.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,8 @@ env:
                       -e TRAVIS_PULL_REQUEST_SLUG=$TRAVIS_PULL_REQUEST_SLUG
                       -e TRAVIS_BRANCH=$TRAVIS_BRANCH
                       -e TRAVIS_JOB_ID=$TRAVIS_JOB_ID
-                      -e TRAVIS_BUILD_DIR=$TRAVIS_BUILD_DIR"
+                      -e TRAVIS_BUILD_DIR=$TRAVIS_BUILD_DIR
+                      -e CONTAINER=podman"
     matrix:
         # Ubuntu
         - GO_VERSION="stable"
--- a/hack/run_ci_tests.sh
+++ b/hack/run_ci_tests.sh
@@ -3,15 +3,18 @@
 set -e
 
 # This script can be run manually - assuming password-less sudo access
-# and a docker-daemon running.  To play in the SPC, just
-# `export SPCCMD=bash` beforehand.
+# to docker or podman.  Export `CONTAINER=docker` for the former, since
+# podman is the default.  To play in the SPC, export `SPCCMD=bash`
+# and it'll drop you in, right before `./hack/spc_ci_test.sh` normally
+# would be called.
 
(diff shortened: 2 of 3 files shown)


cevich:
@rhatdan FYI follow-up to #162. I need @lsm5 to help figure how to get podman working under Trusty, if/when this becomes a priority.  Besides that, all the other necessary changes should be complete (though ***completely untested***)

rhatdan:
How about we build podman from master and run it in travis?

cevich:
@rhatdan IIUC, the issue is the travis CI host is too old and not easily upgraded.  We'd have to play games with building in a docker SPC, then running a podman SPC from there, to do more building and testing in the podman containers :confounded: @lsm5 can correct me if I'm wrong, maybe it is possible to build on 'Trusty'.

It could be made to work, but would also greatly complicate the overall CI situation (and slow it down).  It would be easier to do it via Papr where we have more control over the host environment.  Unless there's another CI alternative, I should look into.

rhatdan:
@cevich Is there any chance of travis ever being updated?  I Like travis way of reporting issues better then papr, since it highlights the errors.  I think Ed has a way of doing this by modifying firefox.

I would be fine if we could get .papr to work with Ubuntu.

cevich:
@rhatdan Ya, greasemonkey scripts...they tend to break every new firefox version.  Eventually Travis should update, but it's a free services, so... :man_shrugging:  Ya, similar roadblocks with PAPR, you're right.  There are other CI/CD services available, likely with different limitations.  I don't see many answers here, that don't involve committing additional time and/or money :persevere: The whole situation reminds me way too much of voting for politicians.  Pick which flavor of poison you dislike the least :see_no_evil:

rhatdan:
@cevich Any update on this script?

cevich:
@rhatdan this one's stuck in the "old travis-image" and "no ubuntu in papr" rut.  There's just not many options here that don't involve spending oodles of time[0] :confused:

Here are some other thoughts:

* My understanding is building podman on xenial (duct-tape included) before every CI test, would seriously frustrate developers (waiting for the same build, over and over).  But I could be wrong, it's happened a few times before :smile:   if this is possible/cheap, and/or could be cached in a tarball somewhere, it could be an option for working with travis.

* Doing it from Origin-CI is another option, but that has a maintainer bottleneck and excessive complexity.

* Rolling our own (using whatever cloud), would also stick us in a maintainer bottleneck.  Otherwise, I have the expertise to pull it off, but can't guarantee it would be any simpler than Origin-CI.

* With these free CI systems, there's a limit to available features, since cloud-resources are most certainly not free[0].  For the low-low price of (maybe) $75/month, this repo. too be covered by a commercial system, in a dedicated cloud.  Spending a few bucks buys us simplicity, image/VM control, and no maintainer-bottleneck.

[0] I'm assuming lsm5's "word" still holds, that installing podman on ubuntu xenial is an effort in futility.

PR #167, opened 2018-04-03
Add debug message on mountFrom error

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

While debugging an issue, the error message led you to believe that the error was with the kernel settings when in reality it was a port contention issue.  The error message from mountFrom that showed the true error was being swallowed.  Adding the debug will at least let us root cause it.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -267,16 +267,22 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 		_ = idtools.MkdirAs(lower2Dir, 0700, rootUID, rootGID)
 		flags := fmt.Sprintf("lowerdir=%s:%s", lower1Dir, lower2Dir)
 		if len(flags) < unix.Getpagesize() {
-			if mountFrom(filepath.Dir(home), "overlay", mergedDir, "overlay", 0, flags) == nil {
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
The second mountFrom fails too. I thought about adding a second debug there too, but wasn't sure it would be helpful or if we'd just get the same error again.

What I've done is stub together a basic RestAPI server for libpod.  It comes up just fine, but when I ask it to list the libpod images via curl, the server tries to get the libpod runtime again (I think that's a mistake in my code, still chasing it) and I get the below error within the brackets:

TOM: mountFrom err [mountfrom re-exec error: exit status 1: output: LibPod Rest API Hello
2018/04/03 10:12:06 listen tcp :8000: bind: address already in use
]TOM: overlay 4

But the main thing was the original error was:
ERRO[0009] 'overlay' is not supported over xfs at
"/var/lib/containers/storage/overlay"

Granted this is probably a one-off situation, but the original error was not at all enlightening.

nalind:
It sounds like we need a similar message when the single-lower case fails, then.

TomSweeneyRedHat:
bot, retest this please

TomSweeneyRedHat:
bot's ignoring me, going to close and reopen.

TomSweeneyRedHat:
Take that bot!

PR #169, opened 2018-04-10
CLI: fix JSON output for layer/container-parent-owners

nalind:
Provide the information that was actually requested when we produce JSON output for the layer-parent-owners and container-parent-owners commands.  Carried over from #156.


rhatdan:
LGTM

PR #170, opened 2018-04-10
GetStore: add a note about needing to reexec.Init()

nalind:
Add a note about needing to call `reexec.Init()` to the doc for the `GetStore()` function.  Also correct incorrect usage elsewhere.


rhatdan:
LGTM

PR #171, opened 2018-04-11
overlay: make all parent dirs accessible with userNS

giuseppe:
ensure all the parent directories are readable when using an userNS
otherwise the container rootfs won't be accessible.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -882,6 +882,30 @@ func (d *Driver) UpdateLayerIDMap(id string, toContainer, toHost *idtools.IDMapp
 	if err := idtools.MkdirAs(diffDir, 0755, rootUID, rootGID); err != nil {
 		return err
 	}
+
+	if rootUID != 0 {
+		if err = os.Chown(dir, rootUID, rootGID); err != nil {
+			return errors.Wrapf(err, "cannot chown '%s'", dir)
+		}
+		if err = os.Chmod(dir, 0700); err != nil {
+			return errors.Wrapf(err, "cannot chmod '%s'", dir)
+		}
+		parent := filepath.Dir(dir)
+		for parent != "/" {
+			fi, err := os.Stat(parent)
+			if err != nil {
+				return errors.Wrapf(err, "cannot stat '%s'", parent)
+			}
+
+			const desiredMask = 0711
+			if fi.Mode()&desiredMask != desiredMask {
+				if err = os.Chmod(parent, fi.Mode()|desiredMask); err != nil {
+					return errors.Wrapf(err, "cannot chmod '%s'", parent)
+				}
+			}
+			parent = filepath.Dir(parent)
+		}
+	}
 	return nil
 }
 


nalind:
Making all parent directories up to the root world-readable seems like something best avoided if possible.  Are the owners of those directories (user and group) mapped into the user namespace?

drivers/overlay/overlay.go
```diff
@@ -882,6 +882,17 @@ func (d *Driver) UpdateLayerIDMap(id string, toContainer, toHost *idtools.IDMapp
 	if err := idtools.MkdirAs(diffDir, 0755, rootUID, rootGID); err != nil {
 		return err
 	}
+
+	if rootUID != 0 {
+		parent := filepath.Dir(diffDir)
+		for parent != "/" {
+			err = os.Chmod(parent, 0755)
```
rhatdan: Wouldn't 711 be enough.  Don't directories have to be searchable not listable?

rhatdan:
Is there potential security issues here where we could allow non priv users access to data mounted into container?

drivers/overlay/overlay.go
giuseppe: yes that is enough, I have changed the patch to use 0711.

Also, I have added another check to not override the mode if it is already permissive

giuseppe:
the last component of the path is owned by the "root in the container" user, so leaking information to other users should not be possible.

drivers/overlay/overlay.go
```diff
@@ -882,6 +882,24 @@ func (d *Driver) UpdateLayerIDMap(id string, toContainer, toHost *idtools.IDMapp
...
+		for parent != "/" {
+			fi, err := os.Stat(parent)
+			if err != nil {
+				return err
+			}
+
+			const desiredMask = 0711
+			if fi.Mode()&desiredMask != desiredMask {
+				if err = os.Chmod(parent, fi.Mode()|desiredMask); err != nil {
+					return err
```
TomSweeneyRedHat: Will this err give us enough information on which file/directory had the issue?  Could/should you wrap it with a little verbiage saying what parent was set to when the failure occurred.

drivers/overlay/overlay.go
```diff
@@ -882,6 +882,24 @@ func (d *Driver) UpdateLayerIDMap(id string, toContainer, toHost *idtools.IDMapp
...
+
+	if rootUID != 0 {
+		parent := filepath.Dir(diffDir)
+		for parent != "/" {
+			fi, err := os.Stat(parent)
+			if err != nil {
+				return err
+			}
+
+			const desiredMask = 0711
```
TomSweeneyRedHat: I'm still not sure about 711 here.  But if Dan an Nalin think it's OK, then I'm good with it.

drivers/overlay/overlay.go
rhatdan: Yes that would make sense to wrap the err.

giuseppe: I've pushed a new version with the Wrapf

nalind:
We dropped patches that did this from #156 because it turned out that mapping the owners of parent directories into the user namespace was enough to avoid permissions problems that we were encountering when testing the functionality of that PR in CRI-O.  If that's not being done in https://github.com/projectatomic/libpod/pull/597, it would be worth knowing why adding logic to do that isn't sufficient to overcome the problem.

giuseppe:
should we require root from the host to be always mapped into the userNS?  That doesn't seem intuitive when using the podman CLI `--uidmap X1:Y1:Z1[:Xn:Yn:Zn]*` that the root user will always be mapped even if not specified.  Also, isn't that worse than opening some directories to be world accessible?

nalind:
I think mapping an ID (and doing it automatically -- `ContainerParentOwners()` is of some help in figuring out which ones would be mapped) is preferable to loosening filesystem permissions.
With looser permissions, a local unprivileged user can find where an overlay filesystem is mounted, change to it, and then has access to a container's filesystem, which includes some world-writeable locations.

nalind:
What's the downside of mapping the ID into the namespace?  The last revision appears to break containers whose images specify a user to run as that isn't UID 0.

giuseppe:
how can I reproduce that error?  I've tried:
```
# podman run --user 2 --security-opt="label=disable" --net none -t --rm  --uidmap 0:20000:60000 --gidmap 0:20000:60000 alpine echo hello
hello
```

The downside of mapping root inside the userNS when not requested is that you can `setuid` to that user and if you can escape the container somehow you are mapped to root in the host

nalind:
Oh, that's not good.

To test things manually, use `podman mount` or `buildah mount` to mount a container's root filesystem, change the tree's owner and permissions manually, use `runc spec` to generate a default `config.json` runtime configuration file, add an entry of `type` `"user"` to its `/linux/namespaces` array, add ID mappings such as `"uidMappings": [
{ "containerID":0, "hostID":1000, "size":500 } ], "gidMappings": [ { "containerID":0, "hostID":1000, "size":500 } ]` to its `/linux` section, and try starting it with `runc run`.  You'll need to add `CAP_SETUID` and `CAP_SETGID` to the set of capabilities in order for the `su` attempt to succeed.

rhatdan:
I'm a little confused. @nalind Are you saying you want UID==0 of the host to be mapped into the container?

giuseppe:
Thanks, I see the issue now.  I've pushed another version that addresses it and at the same time doesn't allow unprivileged users to read from the container rootfs

drivers/overlay/overlay.go
```diff
@@ -882,6 +882,32 @@ func (d *Driver) UpdateLayerIDMap(id string, toContainer, toHost *idtools.IDMapp
...
+		parent := filepath.Dir(dir)
+		for parent != "/" {
+			fi, err := os.Stat(parent)
+			if err != nil {
+				return errors.Wrapf(err, "cannot stat '%s'", parent)
+			}
+
+			// An unprivileged user cannot still reach this directory as the parent
+			// has mode=0700
+			const desiredMask = 0755
```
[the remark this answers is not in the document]
giuseppe: I had to change that as otherwise it would break the use case @nalind showed.

At the same time I made the parent directory `0700` so that will avoid any other user to look/read into this directory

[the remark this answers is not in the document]
giuseppe: double checking it...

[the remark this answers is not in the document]
giuseppe: true, I don't need `0755` here.  What I needed was the diffDir to have `0755`.

Pushed another patch to restore `0711` for the parent directories

rhatdan:
Ok I have been playing around with this, and basically the the container needs to be able to search through the 

/var/lib/containers/storage & /var/lib/containers/storage/overlay

These directories are currently 0700 root/root. Switching them to 701 solves this issue.  As Miroslav in another PR says it looks like runc is mounting the file systems after it has switched on the user namespace.
Since it can't search through these directories it fails to mount the container.

I think a better fix would be to just make these directories searchable, when they get created rather then the hack of walking the chain.

giuseppe:
should we ensure existing `/var/lib/containers/storage` directories can still be used with the new userNS features?

giuseppe:
@rhatdan still debugging it but `0701` seems not enough for me, I still need `0711` to get `-t` working with podman.

giuseppe:
closed in favor of https://github.com/containers/storage/pull/172

PR #173, opened 2018-04-18
store: bind mount rootfs to a different path when using userNS

giuseppe:
we need the directory to be accessible by the user that is mapped to
root in the userNS.  Make it accessible under
/var/run/containers/storage/$ID/ that is owned by the user mapped to
root in the userNS.

### DIFF
--- a/store.go
+++ b/store.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	// register all of the built-in drivers
@@ -23,6 +24,7 @@ import (
 	"github.com/containers/storage/pkg/stringid"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
 )
 
 var (
@@ -548,7 +550,7 @@ func GetStore(options StoreOptions) (Store, error) {
 		return nil, ErrIncompleteOptions
 	}
 
-	if err := os.MkdirAll(options.RunRoot, 0700); err != nil && !os.IsExist(err) {
+	if err := os.MkdirAll(options.RunRoot, 0711); err != nil && !os.IsExist(err) {
 		return nil, err
 	}
 	if err := os.MkdirAll(options.GraphRoot, 0700); err != nil && !os.IsExist(err) {
@@ -1783,6 +1785,31 @@ func (s *store) DeleteContainer(id string) error {
 	if rcstore.Exists(id) {
 		if container, err := rcstore.Get(id); err == nil {
 			if rlstore.Exists(container.LayerID) {
+				root := filepath.Join(s.RunRoot(), container.LayerID, "root")
+				if _, err := os.Stat(root); err == nil {
+					if err = unix.Unmount(root, 0); err != nil && err != syscall.EINVAL {
+						return errors.Wrapf(err, "cannot unmount %s", root)
+					}
+				}
+				userdataMount := filepath.Join(s.RunRoot(), container.ID, "userdata")
+				if _, err := os.Stat(userdataMount); err == nil {
+					if err = unix.Unmount(userdataMount, 0); err != nil && err != syscall.EINVAL {
+						return errors.Wrapf(err, "cannot unmount %s", userdataMount)
+					}
+				}
+
+				if err = os.RemoveAll(filepath.Join(s.RunRoot(), container.ID)); err != nil {
+					if !os.IsNotExist(err) {
+						return err
+					}
+				}
+				rlpath := filepath.Join(s.RunRoot(), container.LayerID)
+				if err = os.RemoveAll(rlpath); err != nil {
+					if !os.IsNotExist(err) {
+						return err
+					}
+				}
+
 				if err = rlstore.Delete(container.LayerID); err != nil {
 					return err
 				}
@@ -1839,6 +1866,18 @@ func (s *store) Delete(id string) error {
 	if rcstore.Exists(id) {
 		if container, err := rcstore.Get(id); err == nil {
 			if rlstore.Exists(container.LayerID) {
+				root := filepath.Join(s.RunRoot(), container.LayerID, "root")
+				if _, err := os.Stat(root); err == nil {
+					if err = unix.Unmount(root, 0); err != nil && err != syscall.EINVAL {
+						return errors.Wrapf(err, "cannot unmount %s", root)
+					}
+				}
+				userdataMount := filepath.Join(s.RunRoot(), container.ID, "userdata")
+				if _, err := os.Stat(userdataMount); err == nil {
+					if err = unix.Unmount(userdataMount, 0); err != nil && err != syscall.EINVAL {
+						return errors.Wrapf(err, "cannot unmount %s", userdataMount)
+					}
+				}
 				if err = rlstore.Delete(container.LayerID); err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


giuseppe:
alternative solution to https://github.com/containers/storage/pull/172

store.go
```diff
@@ -1932,10 +1945,46 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
...
 	}
-	return "", ErrLayerUnknown
+
+	rootUID, rootGID, err := idtools.GetRootUIDGID(s.uidMap, s.gidMap)
+	if err != nil {
+		return "", err
+	}
+
+	merged := ""
+	if rootUID != 0 {
```
rhatdan: Why have two if rootUID blocks?  Can't this just be done later after the rlstore.Mount()

store.go
```diff
@@ -1798,6 +1799,18 @@ func (s *store) DeleteContainer(id string) error {
...
 				}
+				rupath := filepath.Join(s.RunRoot(), container.ID)
+				if err = os.RemoveAll(rupath); err != nil {
+					if !os.IsNotExist(err) {
+						return err
+					}
+				}
+				rlpath := filepath.Join(s.RunRoot(), container.LayerID)
+				if err = os.RemoveAll(rlpath); err != nil {
+					if !os.IsNotExist(err) {
```
rhatdan: You can combine this code to line above.

store.go
```diff
@@ -1798,6 +1799,18 @@ func (s *store) DeleteContainer(id string) error {
 				if err = os.RemoveAll(rcpath); err != nil {
 					return err
 				}
+				rupath := filepath.Join(s.RunRoot(), container.ID)
+				if err = os.RemoveAll(rupath); err != nil {
```
rhatdan: You can combine this code to line above.

store.go
```diff
@@ -548,7 +549,7 @@ func GetStore(options StoreOptions) (Store, error) {
 		return nil, ErrIncompleteOptions
 	}
 
-	if err := os.MkdirAll(options.RunRoot, 0700); err != nil && !os.IsExist(err) {
+	if err := os.MkdirAll(options.RunRoot, 0711); err != nil && !os.IsExist(err) {
```
rhatdan: Yuck.

store.go
giuseppe: I preferred to keep the Mount operation as late as possible, so if we fail earlier we don't need to Mount and Unmount for the cleanup

store.go
```diff
@@ -2427,6 +2503,18 @@ func (s *store) ContainerRunDirectory(id string) (string, error) {
 		return "", err
 	}
 
+	rootUID, rootGID, err := idtools.GetRootUIDGID(s.uidMap, s.gidMap)
```
rhatdan: Will this mean that the user data ends up being written to /run?   Won't this cause issues on reboot?  I think this data needs to be bind mounted from the host onto /run, just like /var/lib/containers/storage/overlay

giuseppe: ah true, I need to take care of that.

I've pushed another version that implements this part.  I will test it better tomorrow though

store.go
```diff
@@ -1932,13 +1944,50 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
...
-	return "", ErrLayerUnknown
+
+	rootUID, rootGID, err := idtools.GetRootUIDGID(s.uidMap, s.gidMap)
+	if err != nil {
+		return "", err
+	}
+
+	merged := ""
+	if rootUID != 0 {
+		merged = filepath.Join(s.RunRoot(), id, "merged")
```
nalind: What's the rationale for naming this "merged"?  Only the overlay driver names its mountpoints that.

store.go
```diff
@@ -1932,13 +1944,50 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
...
+			return "", errors.Wrapf(err, "cannot resolve symlink %s", merged)
+		}
+	}
+
+	mount, err := rlstore.Mount(id, mountLabel)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		if err != nil {
```
nalind: This won't read the result of the `unix.Mount()` call below, since the scope of that variable is the `if` statement that defines it.

store.go
```diff
@@ -2402,11 +2470,28 @@ func (s *store) ContainerDirectory(id string) (string, error) {
...
+
 	middleDir := s.graphDriverName + "-containers"
 	gcpath := filepath.Join(s.GraphRoot(), middleDir, id, "userdata")
-	if err := os.MkdirAll(gcpath, 0700); err != nil {
+	if err := idtools.MkdirAllAs(gcpath, 0700, rootUID, rootGID); err != nil {
 		return "", err
 	}
+
+	if rootUID != 0 {
+		userdata := filepath.Join(s.RunRoot(), id, "userdata")
```
nalind: This seems to break the implicit promise that the contents of the `ContainerDirectory()` will survive a reboot.

store.go
giuseppe: I was just copying the names from overlay.  But I agree, it is better to be more neutral, renamed to "root"

store.go
giuseppe: fixed

giuseppe:
working on the CI failure right now

tests/idmaps.bats
```diff
@@ -332,4 +333,5 @@ load helpers
 	if ! fgrep -q 'GIDs: [0]' <<< "$output" ; then
 		fgrep -q 'GIDs: [0, '$(id -g)']' <<< "$output"
 	fi
+	run storage unmount $mount
```
nalind: This should be done implicitly by the `storage wipe` in the `teardown` function.  Does not doing this create test failures?

nalind: Ah, the bind mount isn't being cleared in the Store's Wipe() method.

giuseppe: yes, I must drop this change.  I've added only to check if it could help, but I am working on the real fix right now, as it is a regression in the PR code

giuseppe: ok nevermind my last comment, I've commented only on your first one :-)

Still it is not enough, as the issue is in the code (unmount needs to read the uid/gid mapping used for the container)

nalind:
One big concern I have here is that this adds a requirement that the caller have the necessary privileges to create bind mounts, which is more than users of the vfs driver previously needed to have, and how that impacts consumers of the library that attempt to run without those privileges.

giuseppe:
True but still needs to be root to access `/var/lib/containers/storage`.  I personally still prefer more the much simpler approach in https://github.com/containers/storage/pull/172

giuseppe:
tests are happy

rhatdan:
Well the third option that @nalind and I talked about this week would be to fix runc to be able to handle this situation.  We were thinking that it might be possible to just enter the mnt namespace without entering the usernamspace to mount the file systems.  This would mean host-root should be able to search through these paths, and then once mounted the container-root would be able to use the bind mounted directory.

giuseppe:
run_oci, the OCI runtime used by ChromeOS has an OCI extension `performInIntermediateNamespace`:

nalind:
The storage library doesn't know anything about namespaces, though, and after it's mounted the rootfs, it's not called when starting/running/stopping containers.

rhatdan:
I agree this is not a containers/storage problem.  This should either be done in runc or in the container runtimes that are launching runc.

rhatdan:
@giuseppe Now that we have you new solution, I think we can close this. Reopen if you disagree.

PR #174, opened 2018-04-30
Fix handling of Copy[File]WithTar with ChownOptions

nalind:
When extracting archives with non-`nil` `ChownOptions`, we weren't mapping the UID and GID in the `ChownOptions` struct to layer-level IDs before fixing up permissions on files and directories that we were creating, and that would cause us to create files with unmapped owners if `ChownOptions` was being used.  This fixes that, breaks the ID mapping in `CopyFileWithTar` evenly between the tar and untar portions of its logic, and adds integration tests.


rhatdan:
LGTM
Looks like tests can't find ffjson?

TomSweeneyRedHat:
LGTM but tests aren't happy.

nalind:
Pushed an additional hunk to update the generated files.

nalind:
... and a new file that I overlooked adding to the commit series.

PR #175, opened 2018-05-07
Only try to copy new layers/images/containers if we succeed in creating them

nalind:
Methods that created new layer, image, or container records could attempt to create copies of `nil` records if an error occurred.  This is mainly an issue if the lower-level storage driver fails during its `Create` method, but we also make a similar change for images and containers to keep the logic similar.


rhatdan:
LGTM

PR #176, opened 2018-05-14
Add default mount options to pass to drivers

rhatdan:
I believe we should be running container images mounted with nodev by default.
This would eliminate the disk of a device sneaking into the container without
being on the approved list.  This would give us the same or potentially additional
security over the device cgroup.

It would be nice if this could be passed in on an image by image basis.  So users
could also specify if they want nosuid images.

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -58,6 +58,10 @@ The `storage.options` table supports the following options:
 **mount_program**=""
   Specifies the path to a custom program to use instead for mounting the file system.
 
+**mountopt**=""
+
+  Comma separated list of default options to be used to mount container images.  Suggested value "nodev".
+
 [storage.options.thinpool]
 
 Storage Options for thinpool
@@ -111,10 +115,6 @@ Specifies the min free space percent in a thin pool required for new device crea
 **mkfsarg**=""
 
 Specifies extra mkfs arguments to be used when creating the base device.
-
-**mountopt**=""
-
-Specifies extra mount options used when mounting the thin devices.
 
 **use_deferred_removal**=""
 
--- a/drivers/devmapper/deviceset.go
+++ b/drivers/devmapper/deviceset.go
@@ -2690,7 +2690,7 @@ func NewDeviceSet(root string, doInit bool, options []string, uidMaps, gidMaps [
 			devices.filesystem = val
 		case "dm.mkfsarg":
 			devices.mkfsArgs = append(devices.mkfsArgs, val)
-		case "dm.mountopt":
+		case "dm.mountopt", "devicemapper.mountopt":
 			devices.mountOptions = joinMountOptions(devices.mountOptions, val)
 		case "dm.metadatadev":
 			devices.metadataDevice = val
--- a/drivers/zfs/zfs.go
+++ b/drivers/zfs/zfs.go
@@ -24,8 +24,9 @@ import (
 )
 
 type zfsOptions struct {
-	fsName    string
-	mountPath string
+	fsName       string
+	mountPath    string
+	mountOptions string
 }
 
 func init() {
@@ -134,6 +135,8 @@ func parseOptions(opt []string) (zfsOptions, error) {
 		switch key {
 		case "zfs.fsname":
 			options.fsName = val
+		case "zfs.mountopt":
+			options.mountOptions = val
 		default:
 			return options, fmt.Errorf("Unknown option %s", key)
 		}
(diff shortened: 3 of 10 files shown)


rhatdan:
@nalind PTAL

nalind:
Looks reasonable, but we'll want to make sure we cover the other drivers, too.

rhatdan:
@nalind Rebased PTAL
@rhvgoyal PTAL

docs/containers-storage.conf.5.md
```diff
@@ -115,7 +115,13 @@ Specifies extra mkfs arguments to be used when creating the base device.
 
 **mountopt**=""
 
-Specifies extra mount options used when mounting the thin devices.
+Default options to be used to mount container images.  Suggested value "nodev".
+
+**ostree_repo=""**
+  Tell storage drivers to use the specified OSTree repository.  Some storage drivers, such as overlay, might use
```
nalind: Looks like we dropped some text here.

drivers/devmapper/deviceset.go
```diff
@@ -2690,8 +2692,6 @@ func NewDeviceSet(root string, doInit bool, options []string, uidMaps, gidMaps [
 			devices.filesystem = val
 		case "dm.mkfsarg":
 			devices.mkfsArgs = append(devices.mkfsArgs, val)
-		case "dm.mountopt":
-			devices.mountOptions = joinMountOptions(devices.mountOptions, val)
```
nalind: I get that the `devicemapper.mountopt` name comes from the driver name and how the option is parsed from the config file, but removing recognition of `dm.mountopt` seems unnecessary.  Can we treat the two names as values for the same case?

drivers/zfs/zfs.go
```diff
@@ -300,6 +303,9 @@ func (d *Driver) create(id, parent string, storageOpt map[string]string) error {
 	}
 	if parent == "" {
 		mountoptions := map[string]string{"mountpoint": "legacy"}
+		if d.options.mountOptions != "" {
+			mountoptions[d.options.mountOptions] = ""
+		}
```
nalind: This looks like it wouldn't handle options of the `key=value` form (for example, `uid=1000`) correctly.

docs/containers-storage.conf.5.md
nalind: This shouldn't be in the `storage.options.thinpool` section.

rhatdan: Removed from my patch set.

drivers/devmapper/deviceset.go
rhatdan: Wouldn't this just confuse people?  I don't see a reason to have two options to do the same thing.

drivers/zfs/zfs.go
rhatdan: FIxed

rhatdan:
@nalind @giuseppe @rhvgoyal PTAL

drivers/devmapper/deviceset.go
nalind: They do the same thing.  Otherwise we're renaming it, and configurations that used one name for the option will start returning an error in anything that uses the newer version of this library.

rhatdan: Added back with no documentation.

drivers/zfs/zfs.go
```diff
@@ -300,6 +303,10 @@ func (d *Driver) create(id, parent string, storageOpt map[string]string) error {
 	}
 	if parent == "" {
 		mountoptions := map[string]string{"mountpoint": "legacy"}
+		for _, opt := range strings.Split(d.options.mountOptions, ",") {
+			val := strings.SplitN(opt, "=", 2)
+			mountoptions[val[0]] = val[1]
```
nalind: If `opt` doesn't contain `=`, wouldn't `val[1]` be out of bounds?

drivers/zfs/zfs.go
```diff
@@ -300,6 +303,14 @@ func (d *Driver) create(id, parent string, storageOpt map[string]string) error {
...
 	if parent == "" {
 		mountoptions := map[string]string{"mountpoint": "legacy"}
+		for _, opt := range strings.Split(d.options.mountOptions, ",") {
+			val := strings.SplitN(opt, "=", 2)
+			if len(val) > 1 {
+				mountoptions[val[0]] = val[1]
+			} else {
+				mountoptions[val[0]] = ""
+			}
+		}
```
nalind: You're sure this goes here instead of in `Get()` below?

rhatdan:
I would say a majority do support it.
Supported:
zfs, devicemapper, overlay, aufs 

Not supported
Btrfs, Windows, VFS

drivers/zfs/zfs.go
rhatdan: Nope, Fixed.

rhatdan:
@nalind Now passes tests, Ready to Merge?
@giuseppe @mrunalp PTAL

layers_ffjson.go
```diff
@@ -1,5 +1,5 @@
 // Code generated by ffjson <https://github.com/pquerna/ffjson>. DO NOT EDIT.
-// source: layers.go
+// source: ./layers.go
```
nalind: This PR doesn't need this change.

rhatdan: I think it does, but I will try without it.

nalind:
LGTM

PR #178, opened 2018-05-20
If storage.conf graphdriver not specified, devmapper should return Er…

rhatdan:
…rNotSupported

This change will return setup is ErrNotSupported if user specifies "" for
driver or "devicemapper" without specifying a device to assign.  We don't
want to support the use of loopback devices as was previoulsy done.

This change is necessary since tests in tools that use containers storage are
being run when driver is not being specified. In this situation the test fails
because devmapper returns and error other then ErrNotSupported.   With this
change a useful message will go to the logs if the user actually specifies
"devicemapper" in config.


TomSweeneyRedHat:
LGTM, but would like someone else to give it a head nod too.

rhatdan:
@nalind PTAL

nalind:
LGTM

PR #179, opened 2018-05-21
Decide how to chroot based on capabilities, not uid_map contents

nalind:
When deciding whether to `pivot_root` or `chroot` in `pkg/chrootarchive`, check if we have `CAP_SYS_CHROOT` and `CAP_SYS_ADMIN`, instead of checking if we have non-default UID mappings, which don't directly affect whether or not we'll be able to do one or the other of those things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ go:
     - master
 
 services:
-  - docker
+    - docker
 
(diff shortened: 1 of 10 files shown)


rhatdan:
Nice LGTM

PR #180, opened 2018-05-29
pkg/archive: use subsecond timestamps in Copy...WithTar

nalind:
When we know we're going to extract the contents of a tar archive as part of a copying process, set the desired format in the tar header to `FormatPAX` so that we get subsecond precision in timestamps, so that the timestamps on the copies of files that we create aren't rounded off.


nalind:
Making this change based on [the docs](https://tip.golang.org/pkg/archive/tar/#Format) in the standard library.

rhatdan:
LGTM

PR #183, opened 2018-05-31
pkg/system: exposed functions need a windows equivalent

vbatts:
Windows does not have a concept of UID/GID, but rather ACLs and SIDs
which do not map cleanly at all.
Since we've exposed `UID()` and `GID()` for this `StatT`, then we need
something equivalent.


rhatdan:
LGTM
@nalind PTAL

rhatdan:
@giuseppe PTAL

giuseppe:
LGTM

PR #184, opened 2018-05-31
driver/windows: duplicated code?

vbatts:
maybe a copy-paste error? seems like some simple test or compile check
would've caught this.


rhatdan:
LGTM
Merging since it is obvios.

PR #186, opened 2018-06-04
Fix build on macOS

mtrmac:
Without this build fails on macOS with
> ./store.go:1561:12: undefined: directory.Size
> ./store.go:1566:11: undefined: directory.Size

because `directory.Size` has no implementation.


TomSweeneyRedHat:
LGTM and happy green test buttons.

rhatdan:
LGTM

PR #187, opened 2018-06-05
containers/storage new driver dummy

giuseppe:
It allows to use the storage for containers but without managing
images.

It is used with https://github.com/projectatomic/libpod/pull/871

### DIFF
--- a/drivers/dummy/driver.go
+++ b/drivers/dummy/driver.go
@@ -1,0 +1,75 @@
+// The dummy driver allows the use of storage for containers but without the need of managing images.
+package dummy
+
+import (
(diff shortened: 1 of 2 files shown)


drivers/dummy/driver.go
```diff
@@ -0,0 +1,74 @@
+package dummy
+
```
TomSweeneyRedHat: Suggestion.  It might be nice to have a top level comment here.  Perhaps something like:

The dummy driver allows the use of storage for containers but without the need of managing
images.

giuseppe: added the comment and pushed a new version

Issue #189, opened 2018-06-19
improve error messages


vrothberg:
This issue is based on a discussion from https://github.com/projectatomic/libpod/issues/956. The underlying problem is that `Podman` doesn't print a user-friendly error message when, for instance, trying to run a container with the name of another one.

You may reproduce with the following commands:
```bash
$ podman run --name foo alpine echo "foo" > /dev/null
$ podman run --name foo alpine echo "foo" > /dev/null
error creating container storage: that name is already in use
```

Docker prints a more user-friendly message guiding the user to a potential conflict resolution:

`docker: Error response from daemon: Conflict. The container name "/foo" is already in use by container "c6c908895cb656718e15d5313a8e94fc3743e5ed3aa7ceb16ff0c865e8bc6f5c". You have to remove (or rename) that container to be able to reuse that name.`

I had a look at https://github.com/containers/storage/blob/master/errors.go and believe that more conflict-related error messages can be improved to make tools using containers/stoarge a bit more user friendly.

The upper exemplary issue, for instance, could be improved by giving a Docker-like error and printing the name of the other container in https://github.com/containers/storage/blob/f382d39d7a6caa855bb1f6243967b630d7389781/containers.go#L278-L283

rhatdan:
@vrothberg Open a PR with an errors.Wrapf(ErrDuplicateName, fmt.Sprintf("the container name %s is already in use by %s.  You have to remove that container to be able to reuse that name." name, nameInUse)

PR #190, opened 2018-06-19
Improve duplicate name error message on container create

marcov:
Make the error message more informative by specifying the duplicate name
and the existing container ID. Implements #189.

Opening this PR as asked by @rhatdan and agreed with @vrothberg


rhatdan:
LGTM

rhatdan:
@marcov I think you need to update the ffjson files.  If you run make does it recompile them?

vrothberg:
LGTM (given the CI passes)

Note that this PR does not yet fully implement https://github.com/containers/storage/issues/189 as there are other error cases that could be improved in a similar fashion (see https://github.com/containers/storage/blob/master/errors.go). Such cases can be duplicate ID, or layer in use, etc. I suggest looking at the code using the errors and see what can be done (in follow-up PRs).

nalind:
@marcov could you `go install github.com/pquerna/ffjson`, run `ffjson containers.go`, and add the updated `containers_ffjson.go` to the patch?  That should keep CI from trying to do the same at build-time.

marcov:
@rhatdan, @nalind: I tried to run ```ffjson``` locally and it just regenerate exactly the same ```containers_ffjson.go```. 

The CI logs make this look more like an issue in the CI env:
```
rm -f containers_ffjson.go
ffjson containers.go
make: ffjson: Command not found
Makefile:33: recipe for target 'containers_ffjson.go' failed
make: *** [containers_ffjson.go] Error 127
```

rhatdan:
I think the makefile wants to generate an updated version. If you run make locally does it execute the ffjson command?

marcov:
OK did like you said.

rhatdan:
Thanks  Lets see if it passes tests.

rhatdan:
Woo Hoo, the hack worked.
@nalind PTAL

nalind:
LGTM, merging.

PR #192, opened 2018-07-04
Size can not be larger then a unint32

rhatdan:
Since the kernel does not use sizes larger then uint32 we should setup
the structure to reflect this fact.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -247,7 +247,7 @@ func parseOptions(options []string) (*overlayOptions, error) {
 	return o, nil
 }
 
-func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGID int) (supportsDType bool, err error) {
+func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGID uint32) (supportsDType bool, err error) {
 	// We can try to modprobe overlay first
 
 	exec.Command("modprobe", "overlay").Run()
@@ -412,8 +412,8 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 		if err != nil {
 			return err
 		}
-		rootUID = int(st.UID())
-		rootGID = int(st.GID())
+		rootUID = st.UID()
+		rootGID = st.GID()
 	}
 	if err := idtools.MkdirAs(dir, 0700, rootUID, rootGID); err != nil {
 		return err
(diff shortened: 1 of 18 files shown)


rhatdan:
@nalind @giuseppe PTAL

drivers/overlay/overlay.go
```diff
@@ -709,7 +709,7 @@ func (d *Driver) Get(id, mountLabel string) (_ string, retErr error) {
 		return "", err
 	}
 
-	if err := os.Chown(path.Join(workDir, "work"), rootUID, rootGID); err != nil {
+	if err := os.Chown(path.Join(workDir, "work"), int(rootUID), int(rootGID)); err != nil {
```
giuseppe: should we raise an error when rootUID is bigger than MaxInt.  On 32 bits systems this will be otherwise casted without any indication to the user

pkg/archive/archive.go
```diff
@@ -486,16 +486,17 @@ func (ta *tarAppender) addTarFile(path, name string) error {
 		if err != nil {
 			return err
 		}
-		hdr.Uid, hdr.Gid, err = ta.IDMappings.ToContainer(fileIDPair)
+		uid, gid, err := ta.IDMappings.ToContainer(fileIDPair)
 		if err != nil {
 			return err
 		}
+		hdr.Uid, hdr.Gid = int(uid), int(gid)
```
TomSweeneyRedHat: We seem to do a lot of conversion from int to uint32 for hdr.Uid and hdr.Gid.  Would it be easier/better to convert those two fields to uint32 and then change them to convert them to int if necessary?  I may have missed a use point that makes that problematic.

drivers/overlay/overlay.go
TomSweeneyRedHat: probably wouldn't hurt to at least do a log.debug.

pkg/archive/archive.go
giuseppe: one problem I see is that `os.Chown` uses `int`, and on i386 an `int` cannot represent an `uint32` as they are both 32 bits.

While `uint32` is more correctly mapping what the kernel does, converting to `uint32` won't still solve this problem so we are back to square one to the initial problem.  We cannot use the full 0-4294967295 range on i386.

rhatdan: So, we should just close this and go with the fix on CRI-O?

giuseppe: I think so, let's not add more logic to handle uids>=2147483647 on 386.  Fixing it here could enable to have these uids mapped in the userNS but these uids could not be used to chown files but I don't think it is worth.

PR #195, opened 2018-07-11
fuse-overlayfs: enable shifting

giuseppe:
[FUSE overlayfs](https://github.com/giuseppe/fuse-overlayfs) has builtin support for UID/GID shifting so that we don't need to re-create a new image when using different mappings for the user namespaces.

When the FUSE backend is used, don't create a new image and chown it, but let the FUSE backend do the remapping on the fly.

Starting with an empty `/var/lib/containers/storage`:
```console
# podman pull fedora
# for i in $(seq 10)
> do
>     bin/podman --storage-opt=overlay.fuse_program=/usr/local/bin/fuse-overlayfs run  \
       --uidmap=0:$i:10000 --gidmap=0:$i:10000 --rm fedora \
        stat -c "owner is: %U" /usr/bin/ls
>     du --si --summarize /var/lib/containers/storage/
>     echo
> done
```

```
owner is: root
286M	/var/lib/containers/storage/

owner is: root
286M	/var/lib/containers/storage/

owner is: root
286M	/var/lib/containers/storage/

owner is: root
286M	/var/lib/containers/storage/

owner is: root
286M	/var/lib/containers/storage/

owner is: root
286M	/var/lib/containers/storage/

owner is: root
286M	/var/lib/containers/storage/

owner is: root
286M	/var/lib/containers/storage/

owner is: root
286M	/var/lib/containers/storage/

owner is: root
286M	/var/lib/containers/storage/
```

/cc @rhatdan @nalind @rhvgoyal @AkihiroSuda


rhatdan:
LGTM
@nalind PTAL

giuseppe:
I'll probably need to rebase this PR on top of https://github.com/containers/storage/pull/199

giuseppe:
let's merge the other one first, as it addresses some real issues :-)

giuseppe:
rebased :arrow_up:

rhatdan:
Should supports_shifting be a flag in the storage.conf?  Rather then force any mount driver to support shifting?

giuseppe:
I am personally in favor of making it configurable, but I understood @nalind would prefer to keep this simpler for now.  @nalind what do you think?

nalind:
I don't think that making assumptions based on whether or not the `mount_program` option is set is going to work if we need to keep adding variations on how overlay works.

nalind:
This needs tests.

How does a caller "know" which ID mappings to use when launching a container using a layer with this setting enabled as its rootfs?  Previously, they'd consult the information recorded along with the layer, but we appear to be setting that information to indicate that host mappings should be used, while quietly using the global default mappings for the actual mounting.

What happens when we create a new layer that uses the new feature, using a parent layer that didn't, but which used a non-default set of ID mappings?

giuseppe:
> How does a caller "know" which ID mappings to use when launching a container using a layer with this setting enabled as its rootfs? Previously, they'd consult the information recorded along with the layer, but we appear to be setting that information to indicate that host mappings should be used, while quietly using the global default mappings for the actual mounting.

I thought I could propagate this information from podman, for example when starting a container that was previously created.  I tried to do it, but it doesn't look nice.  Where is the best place to store this information?  When we mount directly the layer we don't know what mappings were used when it was created

nalind:
We keep track of the mappings that a layer is created for using its `UIDMap` and `GIDMap` fields.  We use that information when creating layers to set permissions in the layer, possibly reverting mappings inherited from the parent layer as part of the permissions-changing process, and then make it available to higher-level tools so that they can, in turn, pass them on to runtime tools.
`CreateContainer()` passes mappings that it's asked to use down to `PutLayer()`, and then keeps a copy of what it ends up using in the container's `IDMappingOptions` for the convenience of callers that don't want to deal with individual layers.

giuseppe:
should I store that we are using shifting or we should allow only one mode to be used (either shifting enabled or not)?

giuseppe:
test added :arrow_up:

rhatdan:
@nalind @giuseppe Where are we with this now?  Did we agree that if you launch a mount_program, it needs to take care of shifting?

giuseppe:
that was my understanding but I am fine to change it.  There probably won't be many others implementations of overlay in FUSE and we can address to disable the shifting when the problem exists, but on the other hand it would still be useful to have a switch for shifting, if the user requires so.

giuseppe:
is the new option blocking this PR?

giuseppe:
I'd like to get this moving, should I add an option for enabling shifting or is it fine as the current PR does to assume it is supported by the FUSE implementation?

rhatdan:
I think we should just assume the mount_program does the shifting.

vagrant/provision.sh
```diff
@@ -4,20 +4,6 @@ set -xe
 source /etc/os-release
 
 case "${ID_LIKE:-${ID:-unknown}}" in
-  debian)
```
rhatdan: Why are you removing debian testing?

tests/helpers.bash
```diff
@@ -2,7 +2,7 @@
 
 STORAGE_BINARY=${STORAGE_BINARY:-$(dirname ${BASH_SOURCE})/../containers-storage}
 TESTSDIR=${TESTSDIR:-$(dirname ${BASH_SOURCE})}
-STORAGE_DRIVER=${STORAGE_DRIVER:-vfs}
+STORAGE_DRIVER=${STORAGE_DRIVER:-overlay}
```
rhatdan: @nalind is this something that will work everywhere?

nalind: I'm not okay with changing this testing default -- it trades running tests with the driver that always has to work for running them with another.

tests/idmaps.bats
```diff
@@ -753,3 +753,45 @@ load helpers
 	echo ntops:$ntops
 	[ $ntops -eq 1 ]
 }
+
+@test "idmaps-create-mapped-container-shifting" {
+	case "$STORAGE_DRIVER" in
+	overlay*)
+		;;
+		*)
+	skip "not supported by driver $STORAGE_DRIVER"
```
nalind: Indentation for these two lines is backwards.

nalind:
If the mount program takes care of shifting, then we basically can never not use the option on a given layer once we've set up a layer that needs it, even if the configuration changes to no longer include using the mount program that provides the feature, because the layer essentially becomes "broken".  We need to be keep track of this somehow to avoid creating a problem for callers when this happens.

giuseppe:
@nalind the layer is stored without any mapping done.  The translation happens only at runtime.  For example, from a podman container with `--uidmap 0:100000:100000`:

```console
# touch foobar foobar2
# chmod 998:999 foobar2
```

and from the host:
```
# find /var/lib/containers/storage/ -name 'foobar*' -exec stat --format="%n uid=%u gid=%g" \{\} \;
/var/lib/containers/storage/overlay/50119762b2a2366b0fa3ddaa79c7c506799a5e66d961cf53b59ac4bb78552afe/merged/foobar uid=100000 gid=100000
/var/lib/containers/storage/overlay/50119762b2a2366b0fa3ddaa79c7c506799a5e66d961cf53b59ac4bb78552afe/merged/foobar2 uid=100998 gid=100999
/var/lib/containers/storage/overlay/50119762b2a2366b0fa3ddaa79c7c506799a5e66d961cf53b59ac4bb78552afe/diff/foobar2 uid=998 gid=999
/var/lib/containers/storage/overlay/50119762b2a2366b0fa3ddaa79c7c506799a5e66d961cf53b59ac4bb78552afe/diff/foobar uid=0 gid=0
```

so the shifting has the opposite effect of running with a mapping enabled, are there cases when this is not enough?

nalind:
We're changing what's stored in the `Layer` structure when we store a layer this way, in that it may no longer agree with the copy we store in the `Container` structure.  That copy was originally added there as a convenience, to save callers from having to look up the mappings by looking up the `Layer` structure using the ID stored in the `Container` structure.  I don't know how that'll impact current callers.
Leaving that aside, if the configuration is changed so that the mount helper is no longer used, we'll quietly discard the maps stored in the `Container` structure and mount the layer with no mappings, but if the caller is using the mapping information in the `Container` structure, things will break, so there needs to be an error check there.

giuseppe:
yes good point, we should error out when we try to mount the same layer/container and not using shifting.  I've added a patch for doing it, now I've:

```console
# podman --storage-opt overlay.mount_program=/usr/local/bin/fuse-overlayfs run --uidmap 0:100000:100000 --name test1 fedora echo hello
hello
# podman start -a test1
unable to start container b991456592f83b96b7675b54bdde84d0249686f1eec676261051f39f74b831f8: error mounting storage for container b991456592f83b96b7675b54bdde84d0249686f1eec676261051f39f74b831f8: cannot mount layer 3160cc1b237be611c580ddb2a61076491506caf1b8f9ff0080e5a2dbaa353d1a: shifting not enabled

```

tests/idmaps.bats
```diff
@@ -753,3 +753,45 @@ load helpers
...
+	# Mount the layer.
+	run storage --debug=false mount $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	lowermount="$output"
+	# Put a file in the layer.
+	createrandom "$lowermount"/file
+	storage unmount $layer
+
+        imagename=idmappedimage-shifting
```
nalind: Mixing tabs and spaces makes this diff look weird.

nalind:
LGTM

giuseppe:
@rhatdan tests are passing

PR #198, opened 2018-07-13
Add force to umount to force the umount of a container image

rhatdan:


rhatdan:
@nalind PTAL

layers.go
```diff
@@ -211,7 +211,10 @@ type LayerStore interface {
 	Mount(id, mountLabel string) (string, error)
 
 	// Unmount unmounts a layer when it is no longer in use.
-	Unmount(id string) error
+	Unmount(id string, force bool) (bool, error)
+
+	// Mounted returns whether or not the layer is mounted
```
nalind: Missing a `.` at the end of the godoc.

layers.go
```diff
@@ -677,7 +691,7 @@ func (r *layerStore) Unmount(id string) error {
 		layer.MountPoint = ""
 		err = r.Save()
 	}
-	return err
+	return false, err
```
nalind: If `err` is not `nil`, this probably needs to be `true`, since we're leaving the recorded `MountCount` unchanged.

store.go
```diff
@@ -262,8 +262,8 @@ type Store interface {
 	Mount(id, mountLabel string) (string, error)
 
 	// Unmount attempts to unmount a layer, image, or container, given an ID, a
-	// name, or a mount path.
-	Unmount(id string) error
+	// name, or a mount path. Returns whether or not the layer is still mounted
```
nalind: Missing a `.` at the end of the godoc.

nalind:
The new value that `Unmount()` returns doesn't seem to be consulted anywhere, and this needs a test.

rhatdan:
Added test and handling of the umount return value.

rhatdan:
@nalind PTAL

cmd/containers-storage/mount.go
```diff
@@ -92,6 +125,17 @@ func init() {
...
+		},
+	})
+	commands = append(commands, command{
+		names:       []string{"mounted"},
+		optionsHelp: "LayerOrContainerNameOrID",
+		usage:       "Check if a file system is mounted",
+		minArgs:     1,
+		action:      mounted,
 		addFlags: func(flags *mflag.FlagSet, cmd *command) {
 			flags.BoolVar(&jsonOutput, []string{"-json", "j"}, jsonOutput, "Prefer JSON output")
```
nalind: The error message at line 100 needs to be updated.

layers.go
```diff
@@ -677,7 +691,7 @@ func (r *layerStore) Unmount(id string) error {
 		layer.MountPoint = ""
 		err = r.Save()
 	}
-	return err
+	return err != nil, err
```
nalind: If the `Put()` succeeds but we fail to save the updated mount count, we return `true` here.  Is that the intention?

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false create-layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	layer="$output"
+
+	# Mount the layer.
+	run storage --debug=false mount $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	# Check it layer is mounted
```
nalind: s/it/if, add a period at the end.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false mounted $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
```
nalind: Can we compare to `"$layer mounted"` here?

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	# Unmount the layer.
+	run storage --debug=false unmount $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	# Check it layer is mounted
```
nalind: s/it/if, add a period at the end.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	[ "$output" == "" ]
+
+	# Mount the layer twice.
+	run storage --debug=false mount $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	run storage --debug=false mount $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	# Check it layer is mounted
```
nalind: s/it/if, add a period at the end.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false mounted $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	# Unmount the third layer.
```
nalind: This comment looks wrong.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false unmount $layer
+	[ "$status" -eq 0 ]
+	[ "$output" == "" ]
+	# Check it layer is mounted
```
nalind: s/it/if, add a period at the end.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false mounted $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	# Unmount the third layer.
+	run storage --debug=false unmount $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	# Check it layer is mounted
```
nalind: s/it/if, add a period at the end.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false mounted $layer
+	[ "$status" -eq 0 ]
+	[ "$output" == "" ]
+
+
+	# Mount the layer twice and force umount
```
nalind: Add a period at the end.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false mount $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	run storage --debug=false mount $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	# Check it layer is mounted
```
nalind: s/it/if, add a period at the end.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false mounted $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	# Unmount the third layer.
```
nalind: This comment looks wrong.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false unmount --force $layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	# Check it layer is mounted
```
nalind: s/it/if, add a period at the end.

tests/mount-layer.bats
```diff
@@ -0,0 +1,81 @@
...
+	run storage --debug=false mounted $layer
+	[ "$status" -eq 0 ]
+	[ "$output" == "" ]
+
+	# Try to delete the first layer, which should fail because it has children.
```
nalind: This comment doesn't agree with the lines below it.

nalind:
This includes changes to `layers.go`, but updates `images_ffjson.go`'s timestamp, so CI still tries to rebuild `layers_ffjson.go`.

layers.go
rhatdan: you tell me.  The storage has unmounted the file system, but the database does not have the correct data.  Does this mean it is mounted or not?

tests/mount-layer.bats
rhatdan: Fixed

tests/mount-layer.bats
rhatdan: Fixed

tests/mount-layer.bats
rhatdan: Fixed

tests/mount-layer.bats
rhatdan: Fixed

tests/mount-layer.bats
rhatdan: Fixed

tests/mount-layer.bats
rhatdan: Fixed

tests/mount-layer.bats
rhatdan: FIxed

tests/mount-layer.bats
rhatdan: Fixed

tests/mount-layer.bats
rhatdan: Fixed

tests/mount-layer.bats
rhatdan: Fixed

tests/mount-layer.bats
rhatdan: Fixed

rhatdan:
WooHoo tests pass.
@nalind PTAL, we still have the one question on what to return on partial error.

cmd/containers-storage/mount.go
```diff
@@ -51,12 +51,15 @@ func unmount(flags *mflag.FlagSet, action string, m storage.Store, args []string
...
-		err := m.Unmount(arg)
+		mounted, err := m.Unmount(arg, force)
 		errText := ""
 		if err != nil {
 			errText = fmt.Sprintf("%v", err)
 			errors = true
 		}
+		if !mounted {
+			fmt.Printf("%s mountpoint unmounted\n", arg)
+		}
```
nalind: Non-blocker: in principle it seems... off to treat `mounted` as useful if `err` is set.

cmd/containers-storage/mount.go
```diff
@@ -74,6 +77,36 @@ func unmount(flags *mflag.FlagSet, action string, m storage.Store, args []string
...
+	for _, arg := range args {
+		mounted, err := m.Mounted(arg)
+		errText := ""
+		if err != nil {
+			errText = fmt.Sprintf("%v", err)
+			errors = true
+		}
+		if mounted {
+			fmt.Printf("%s mounted\n", arg)
+		}
```
nalind: Non-blocker: in principle it seems... off to treat `mounted` as useful if `err` is set.

nalind:
A couple of nits about the CLI's interpretation of results from `Unmount()`, but otherwise it LGTM.

PR #199, opened 2018-07-16
storage: allow to use a different configuration file

giuseppe:
Move the configuration file parsing to a new function `ReloadConfigurationFile` so that it is possible to load a differentconfiguration file and override the current settings.  This is useful with rootless containers so a configuration file specific to the user can be loaded.

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -55,8 +55,8 @@ The `storage.options` table supports the following options:
   old, the driver is not supported.  But for kernels that have had the drivers backported, this flag
   allows users to override the checks
 
-**fuse_program**=""
-  Specifies the path to a custom FUSE program to use instead for mounting the file system.
+**mount_program**=""
+  Specifies the path to a custom program to use instead for mounting the file system.
 
 [storage.options.thinpool]
 
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -55,6 +55,9 @@ The `storage.options` table supports the following options:
   old, the driver is not supported.  But for kernels that have had the drivers backported, this flag
   allows users to override the checks
 
+**fuse_program**=""
+  Specifies the path to a custom FUSE program to use instead for mounting the file system.
+
 [storage.options.thinpool]
 
 Storage Options for thinpool
@@ -88,10 +91,6 @@ Tells driver to wipe device (directlvm_device) even if device already has a file
 **fs**="xfs"
 
 Specifies the filesystem type to use for the base device. (default: xfs)
-
-**fuse_program**=""
-
-Specifies the path to a custom FUSE program to use instead for mounting the file system.
 
 **log_level**=""
 
--- a/storage.conf
+++ b/storage.conf
@@ -25,9 +25,9 @@ additionalimagestores = [
 # certain container storage drivers.
 size = ""
 
-# Path to a FUSE helper to use for mounting the file system instead of mounting it
+# Path to an helper program to use for mounting the file system instead of mounting it
 # directly.
-#fuse_program = "/usr/bin/fuse-overlayfs"
+#mount_program = "/usr/bin/fuse-overlayfs"
 
 # OverrideKernelCheck tells the driver to ignore kernel checks based on kernel version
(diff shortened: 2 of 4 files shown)


store.go
```diff
@@ -3004,11 +3004,7 @@ type tomlConfig struct {
 	} `toml:"storage"`
 }
 
-func init() {
-	DefaultStoreOptions.RunRoot = "/var/run/containers/storage"
-	DefaultStoreOptions.GraphRoot = "/var/lib/containers/storage"
-	DefaultStoreOptions.GraphDriverName = ""
-
+func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
```
rhatdan: Add a comment defining this function.

storage.conf
```diff
@@ -25,6 +25,11 @@ additionalimagestores = [
 # certain container storage drivers.
 size = ""
 
+# Path to a FUSE helper to use for mounting the file system instead of mounting it
+# directly.
+#fuse_program = "/usr/bin/fuse-overlayfs"
```
rhatdan: Should we call this mount_program rather then fuse.  I see no reason to only do this for fuse executables.
```
#mount_program = "/usr/bin/fuse-overlayfs"
```

store.go
```diff
@@ -2992,6 +2992,9 @@ type OptionsConfig struct {
 
 	// Do not create a bind mount on the storage home
 	SkipMountHome string `toml:"skip_mount_home"`
+
+	// Alternative program to use for the mount of the file system
+	FuseProgram string `toml:"fuse_program"`
```
rhatdan: MountProgram ...

store.go
giuseppe: done

storage.conf
giuseppe: this was already present in containers/storage.  So I've added a new commit to rename the option name everywhere

store.go
giuseppe: done, I've renamed Fuse -> Mount everywhere

rhatdan:
LGTM
@nalind PTAL

docs/containers-storage.conf.5.md
```diff
@@ -89,9 +89,9 @@ Tells driver to wipe device (directlvm_device) even if device already has a file
 
 Specifies the filesystem type to use for the base device. (default: xfs)
 
-**fuse_program**=""
+**mount_program**=""
 
-Specifies the path to a custom FUSE program to use instead for mounting the file system.
+Specifies the path to a custom program to use instead for mounting the file system.
```
nalind: This shouldn't be in the `storage.options.thinpool` section.

store.go
```diff
@@ -3024,72 +3025,78 @@ func init() {
...
 	if config.Storage.Options.SkipMountHome != "" {
-		DefaultStoreOptions.GraphDriverOptions = append(DefaultStoreOptions.GraphDriverOptions, fmt.Sprintf("%s.skip_mount_home=%s", config.Storage.Driver, config.Storage.Options.SkipMountHome))
+		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.skip_mount_home=%s", config.Storage.Driver, config.Storage.Options.SkipMountHome))
+	}
+	if config.Storage.Options.MountProgram != "" {
+		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.mount_program=%s", config.Storage.Driver, config.Storage.Options.MountProgram))
+	}
+	if config.Storage.Options.OverrideKernelCheck != "" {
+		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.override_kernel_check=%s", config.Storage.Driver, config.Storage.Options.OverrideKernelCheck))
 	}
```
nalind: This block duplicates the block that follows it.

nalind:
A couple of comments, but mostly fine.

docs/containers-storage.conf.5.md
giuseppe: thanks, I've moved it

store.go
giuseppe: dropped

PR #200, opened 2018-07-17
Change Mounted to return the number of times mounted

rhatdan:
podman unmount wants to know if the image is only mounted 1 time
and refuse to unmount if the container state expects it to be mounted.


mheon:
Turns out to be a lot more useful for us than just a bool - LGTM

TomSweeneyRedHat:
LGTM and happy green tests.  I smell a quick vendor for Podman

rhatdan:
@nalind PTAL

nalind:
LGTM

PR #201, opened 2018-07-25
Fix overlay to handle mountopt properly

rhatdan:
We need to translate the mount options into flags or
data, so this PR makes the parse code public so we can
use it in containers/storage.


rhatdan:
@nalind @giuseppe PTAL

nalind:
LGTM pending CI.

giuseppe:
LGTM

PR #204, opened 2018-07-30
idtools: improve no ranges err message

giuseppe:
Improve the error message telling the user where to look for.

Closes: https://github.com/projectatomic/libpod/issues/1182


giuseppe:
I think #205 looks better than this, let's merge it

PR #206, opened 2018-07-30
Set environment variable to indicate you are in rootless mode

rhatdan:


nalind:
Setting variables in the value returned by `Command` which aren't set by the similarly-named function in `os/exec` creates surprises, and the variable being set indicates something that the reexec package doesn't guarantee.  NAK.

rhatdan:
CLosed since I can fix this in buildah.

PR #208, opened 2018-08-20
storage.conf: Don't specify nodev by default

cgwalters:
This breaks previously valid configurations, such as using
`mock` inside a privileged container.  While it's easy for
such containers to learn to `mount -o remount,dev /`, it's a needless
compatibility break.

Rather, e.g. podman should learn to inject it only if it's known
safe to do so (e.g. the target container doesn't have CAP_MKNOD anyways).

### DIFF
--- a/storage.conf
+++ b/storage.conf
@@ -33,7 +33,7 @@ size = ""
 override_kernel_check = "false"
 
 # mountopt specifies comma separated list of extra mount options
-mountopt = "nodev"
+mountopt = ""
 
 # Remap-UIDs/GIDs is the mapping from UIDs/GIDs as they should appear inside of
 # a container, to UIDs/GIDs as they should appear outside of the container, and


rhatdan:
I actually want this for CRI-O more then podman, since running mock and having it create devices is not something we would do in mock.  Why doesn't mock create a tmpfs for /dev?
Currently we give out MKNOD by default and only protection is the Device CGROUP which I believe is going away.

cgwalters:
> only protection is the Device CGROUP which I believe is going away.

Reference?

> since running mock and having it create devices is not something we would do in mock. 

You mean do in cri-o/Kubernetes?  I think scheduling traditional build systems like `mock` in Kubernetes is in fact a totally sane thing to do; again today it requires privileges.  That said of course there's a really powerful tension to just drop mock and have RPM builds just be `yum install` from the base container.

Anyways, the point here is this broke compatibility, and today doesn't add any security.

rhatdan:
Running mock in Kubernetes should be done in a volume in the container not on the container image.
The NODEV is just going to block any devices that come in the image.  Not going to block devices being created in writable directories.

cgwalters:
> Running mock in Kubernetes should be done in a volume in the container not on the container image.

I agree.  But this change still broke my container, and doesn't really help anything today.

rhatdan:
It is my understanding that the Device Cgroup was not implemented in V2 cgroups.  You are supposed to use some Seccomp/eBPF Filtering, I believe.

giuseppe:
tests are failing because the commit is missing `Signed-off-by:`

rhatdan:
I am closing this PR since @cgwalters can get what he wants from https://github.com/containers/storage/pull/226

PR #210, opened 2018-08-25
Add lock sanity checks to Save() methods

zmedico:
I have experienced "layer not known" corruption triggered by concurrent
buildah/skopeo processes, and hopefully lock sanity checks will help to
prevent this kind of problem.

Signed-off-by: Zac Medico <zmedico@gmail.com>

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -192,6 +192,9 @@ func (r *containerStore) Load() error {
 }
 
 func (r *containerStore) Save() error {
+	if !r.Locked() {
+		return errors.New("container store is not locked")
+	}
 	rpath := r.containerspath()
 	if err := os.MkdirAll(filepath.Dir(rpath), 0700); err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


rhatdan:
Thanks @zmedico.
@nalind WDYT?

TomSweeneyRedHat:
Me thinks @zmedico was reading my mind, or I was reading his.  I was going to chat with @nalind today to see if we should throw a lock of some kind around the save() function....   I think yes, we probably should.

nalind:
LGTM for finding bugs -- we shouldn't be saving when we don't have the lock.

rhatdan:
LGTM

mtrmac:
FWIW this broke cross-compilation to Windows.

zmedico:
> FWIW this broke cross-compilation to Windows.

Sorry about that, I hope https://github.com/containers/storage/pull/212 will fix it.

zmedico:
I can confirm that c7ba5749d44a65fde2daf114c16fb0272d82d73b has solved my concurrent buildah/skopeo "layer not known" corruption issue. There must be one or more `Save()` calls that this patch causes to fail non-fatally.

PR #211, opened 2018-08-26
 Add MountOpts to stop adding fields to Get Interface 

rhatdan:
This patch adds a MountOpts field to the drivers so we can simplify
the interface to Get and allow additional options to be passed in the future.

### DIFF
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -42,6 +42,15 @@ type CreateOpts struct {
 	StorageOpt map[string]string
 }
 
+// MountOpts contains optional arguments for LayerStope.Mount() methods.
+type MountOpts struct {
+	// Mount label is the MAC Labels to assign to mount point (SELINUX)
+	MountLabel string
+	// UidMaps & GidMaps are the User Namespace mappings to be assigned to content in the mount point
+	UidMaps []idtools.IDMap
+	GidMaps []idtools.IDMap
+}
+
 // InitFunc initializes the storage driver.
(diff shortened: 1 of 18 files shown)


TomSweeneyRedHat:
Tests aren't hip @rhatdan

rhatdan:
Finally green

@nalind PTAL

vrothberg:
LGTM

TomSweeneyRedHat:
LGTM but would like a head nod from @nalind or @mtrmac

drivers/driver.go
```diff
@@ -42,6 +42,13 @@ type CreateOpts struct {
 	StorageOpt map[string]string
 }
 
+// MountOpts contains optional arguments for Get() methods.
```
mtrmac: _Mount_ options for _Get_ methods? The name seems right (external callers call `LayerStore.Mount`, `ProtoDriver.Get` is an internal implementation detail of the driver interface), but for the same reason the documentation should probably focus on `LayerStore.Mount` rather than `Get`.

drivers/driver.go
```diff
@@ -42,6 +42,13 @@ type CreateOpts struct {
...
+type MountOpts struct {
+	MountLabel string
```
mtrmac: All of the members could do with some documentation as well (e.g. it’s not locally obvious that `MountLabel` is a SELinux label.

(OTOH the documentation of the `Get`/`Mount` methods that now use `MountOpts` could be trimmed and rely on the member documentation here.)

mtrmac:
ACK

PR #213, opened 2018-09-12
Fix a lock inversion

nalind:
In CreateContainer(), don't use ROLayerStores() to get a list of the read-only layer stores after we've acquired the lock on the writeable layer store.  ROLayerStores() acquires the graph lock, which we should never try to acquire while we're holding the layer store lock.


nalind:
@mheon PTAL

TomSweeneyRedHat:
LGTM
That's a subtle little bugger.

rhatdan:
LGTM

PR #214, opened 2018-09-13
Hold the layer store lock while diffing

nalind:
While generating a Diff, hold the lock on the layer store until after we've completely finished building the diff.

There's an internal Mount/Unmount being done so that we can read the layer's contents, and we don't update the mount counts properly if we're not still holding the lock when the layer store's Unmount() method is called, and that doesn't happen until the ReadCloser that Diff() returns gets closed.

This should fix https://github.com/projectatomic/buildah/issues/981.


TomSweeneyRedHat:
LGTM assuming happy tests.

rhatdan:
LGTM

kbaegis:
Thanks everyone!

PR #215, opened 2018-09-21
layerStore.Put(): always check for Create() errors

nalind:
If we needed to try to update the ID mappings on a just-created layer, we were inadvertently failing to check that the layer had been successfully created.


rhatdan:
LGTM

PR #216, opened 2018-09-25
chown: restore SUID and SGID bits

giuseppe:
be sure the SUID and SGID bits are not lost when we do
a chown.

Closes: https://github.com/containers/libpod/issues/1526

### DIFF
--- a/drivers/chown_unix.go
+++ b/drivers/chown_unix.go
@@ -45,9 +45,19 @@ func platformLChown(path string, info os.FileInfo, toHost, toContainer *idtools.
 			uid, gid = mappedPair.UID, mappedPair.GID
 		}
 		if uid != int(st.Uid) || gid != int(st.Gid) {
+			stat, err := os.Lstat(path)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

nalind:
LGTM

drivers/chown_unix.go
```diff
@@ -45,10 +45,20 @@ func platformLChown(path string, info os.FileInfo, toHost, toContainer *idtools.
 			uid, gid = mappedPair.UID, mappedPair.GID
 		}
 		if uid != int(st.Uid) || gid != int(st.Gid) {
+			stat, err := os.Lstat(path)
```
kolyshkin: Maybe I'm missing something here, but I don't understand the reason for calling `os.Lstat()` here -- this function is already called with the `info` parameter, which is result of `os.Lstat(path)`.

giuseppe: good catch, I agree we can avoid `Lstat` here

PR #217, opened 2018-09-29
Need to preserve the Xattr security.capability when chowning.

rhatdan:
When we chown we not only loose the setuid flags but we also loose
the file capabilties.  This means in a user namespace when we chown
ping, it looses its filecap, which means it will not longer work.

This fix will check for the security capabilies and retain them if they
exist when chowning.

### DIFF
--- a/drivers/chown_unix.go
+++ b/drivers/chown_unix.go
@@ -8,6 +8,7 @@ import (
 	"syscall"
 
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/system"
 )
 
 func platformLChown(path string, info os.FileInfo, toHost, toContainer *idtools.IDMappings) error {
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind @giuseppe PTAL

drivers/chown_unix.go
```diff
@@ -49,6 +50,11 @@ func platformLChown(path string, info os.FileInfo, toHost, toContainer *idtools.
 			if err != nil {
 				return fmt.Errorf("%s: lstat(%q): %v", os.Args[0], path, err)
 			}
+			cap, err := system.Lgetxattr(path, "security.capablity")
```
vrothberg: typo s/capablity/capability/ ("i" missing)

TomSweeneyRedHat:
Ditto what @vrothberg said

drivers/chown_unix.go
```diff
@@ -49,6 +50,11 @@ func platformLChown(path string, info os.FileInfo, toHost, toContainer *idtools.
 			if err != nil {
 				return fmt.Errorf("%s: lstat(%q): %v", os.Args[0], path, err)
 			}
+			cap, err := system.Lgetxattr(path, "security.capability")
+			if err != nil {
+				return fmt.Errorf("%s: Lgetxattr(%q): %v", os.Args[0], path, err)
```
nalind: AIUI this will always return an error on non-Linux hosts.

nalind: ... so we either need to check for and ignore `system.ErrNotSupportedPlatform`, or add a note that we're going to have to fix this later.

rhatdan:
@vrothberg Yes I will add a test to podman for this once it gets merged.

giuseppe:
LGTM

nalind:
LGTM

PR #220, opened 2018-10-12
Start to store SELinux labels in layer store

rhatdan:


rhatdan:
Nalin is this what you had in mind?

nalind:
It's a big part of it.  I was expecting us to handle allocating a default, too, for the sake of applications that don't want to have to care about any of this.
It looks as though, when we have multiple writers, calls to `Load()` that attempt to update our in-memory state to match what another writer saved to disk will currently trigger errors from `label.ReserveLabel()`.

store.go
```diff
@@ -1175,7 +1178,19 @@ func (s *store) CreateContainer(id string, names []string, image, layer, metadat
 			},
 		}
 	}
-	clayer, err := rlstore.Create(layer, imageTopLayer, nil, "", nil, layerOptions, true)
+	if options.Flags == nil {
+		options.Flags = make(map[string]interface{})
+	}
+	if options.Flags["ProcessLabel"].(string) == "" {
```
nalind: What happens, or should happen, if the "MountLabel" flag is set but the "ProcessLabel" flag is not?  If this type assertion fails, what happens?

containers.go
```diff
@@ -297,7 +311,7 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
 			BigDataSizes:   make(map[string]int64),
 			BigDataDigests: make(map[string]digest.Digest),
 			Created:        time.Now().UTC(),
-			Flags:          make(map[string]interface{}),
+			Flags:          options.Flags,
```
nalind: Use `copyStringInterfaceMap(options.Flags)` here.  It attempts to copy the map to avoid possible shenanigans if the passed-in map changes later.

store.go
rhatdan: I changed it to thrown an error.

containers.go
rhatdan: Fixed

store.go
```diff
@@ -2256,7 +2278,7 @@ func (s *store) Version() ([][2]string, error) {
 	return [][2]string{}, nil
 }
 
-func (s *store) Mount(id, mountLabel string) (string, error) {
+func (s *store) Mount(id string) (string, error) {
```
nalind: I'd still like to let the caller override our default, so please revert this bit.

store.go
```diff
@@ -1175,7 +1180,26 @@ func (s *store) CreateContainer(id string, names []string, image, layer, metadat
...
 		}
 	}
-	clayer, err := rlstore.Create(layer, imageTopLayer, nil, "", nil, layerOptions, true)
+	if options.Flags == nil {
+		options.Flags = make(map[string]interface{})
+	}
+	if (options.Flags["ProcessLabel"].(string) != "" &&
+		options.Flags["MountLabel"].(string) == "") ||
+		(options.Flags["ProcessLabel"].(string) == "" &&
+			options.Flags["MountLabel"].(string) != "") {
```
nalind: These type assertions need to be checked, in case any of these are unset.  When they're not set, we'll panic otherwise: https://play.golang.org/p/s_9RhYGbc8M.

nalind:
I'm reluctant to ever reset the essentially global list of MCS categories that the selinux package owns, since we could eventually find ourselves interacting badly with another library that wants to do something similar, but for now this will probably work for our consumers.

nalind:
LGTM

PR #224, opened 2018-10-25
Setting storage options could break overlay commit.

rhatdan:
If in storage.conf I set the mount options to:
mountopt = "nodev,metacopy=on,redirect_dir=on"
This will turn on redirect_dir but the useNaiveDiff does not currently check
this option.  This will cause container images on commit to not be saved
correctly.

### DIFF
--- a/drivers/overlay/check.go
+++ b/drivers/overlay/check.go
@@ -20,7 +20,7 @@ import (
 // which copies up the opaque flag when copying up an opaque
 // directory or the kernel enable CONFIG_OVERLAY_FS_REDIRECT_DIR.
 // When these exist naive diff should be used.
-func doesSupportNativeDiff(d string) error {
+func doesSupportNativeDiff(d, mountOpts string) error {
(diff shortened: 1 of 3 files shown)


rhatdan:
@rhvgoyal @nalind PTAL

rhatdan:
I tested this in podman and it worked correctly with the flags turned on.

```
# ./bin/podman run -ti --name test1 fedora mv /etc/yum.repos.d/ /etc/yum.repos.d.1/
```

Without Patch
```
# podman diff test1
C /etc
D /etc/yum.repos.d
A /etc/yum.repos.d.1
```
With Patch
```
# ./bin/podman diff test1
C /etc
A /etc/yum.repos.d.1
A /etc/yum.repos.d.1/fedora.repo
A /etc/yum.repos.d.1/fedora-cisco-openh264.repo
A /etc/yum.repos.d.1/fedora-updates-testing.repo
A /etc/yum.repos.d.1/fedora-updates.repo
D /etc/yum.repos.d
```

rhvgoyal:
LGTM.

PR #225, opened 2018-10-30
overlay: force naiveDiff when using a mount program

giuseppe:


rhatdan:
LGTM, but I would prefer a little more description in your PR or why this is necessary.

Something like:
```
Overlay diff relies on the kernel's Overlay driver, if a user specifies an alternative driver, then we need to default back to the naivediff.
```

giuseppe:
Thanks, I've amended your suggestion in the commit message

rhatdan:
LGTM
@nalind @TomSweeneyRedHat @vrothberg @umohnani8 PTAL

vrothberg:
LGTM
Had to wrap my head around the surrounding anonymous function `useNaiveDiffLock.Do(func() {..` first but it looks good to me.

PR #226, opened 2018-10-30
Allow the passing in and retrieval of mount options

rhatdan:
We want to allow tools like podman/buildah to override default storage
container mount options on a container by container basis.

For example if the default mount options for containers/storage include
nodev or nosuid, we want to allow podman to turn these off if the user
specifies --privileged.

We also might want to turn off certain user namespace flags that will cause
buildah and podman build to work slower when creating container images.


rhatdan:
@nalind @giuseppe @vrothberg @cgwalters PTAL

cgwalters:
> For example if the default mount options for containers/storage include
nodev or nosuid, we want to allow podman to turn these off if the user
specifies --privileged.

Yes.  https://github.com/containers/storage/pull/208

giuseppe:
it seems to introduce some differences in TestChrootUntarPath

rhatdan:
Yes I am seeing crashes in the storage test program.  Will continue investigating later.

rhatdan:
@nalind @giuseppe @vrothberg @umohnani8 PTAL

containers.go
```diff
@@ -293,6 +300,9 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
 	if _, idInUse := r.byid[id]; idInUse {
 		return nil, ErrDuplicateID
 	}
+	if options.MountOpts != nil {
+		options.Flags["MountOpts"] = options.MountOpts
```
nalind: Please use `append([]string{}, options.MountOpts...)` here, so that the value we set in `Flags` is a copy of the passed-in `options.MountOpts` slice, and any changes made to the slice in `options.MountOpts` after we're called won't be unexpectedly made in our map.

drivers/overlay/overlay.go
```diff
@@ -743,7 +743,11 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 
 	workDir := path.Join(dir, "work")
 	opts := fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s", strings.Join(absLowers, ":"), diffDir, workDir)
-	if d.options.mountOptions != "" {
+	if options.Options != nil {
+		if len(options.Options) > 0 {
```
nalind: The length of a `nil` slice is defined as 0, so we don't need to check for `nil` before checking the length.

containers.go
rhatdan: Fixed

rhatdan:
@nalind I think this is ready to merge?

TomSweeneyRedHat:
LGTM and happy green test buttons.  Would like a head nod from @nalind

giuseppe:
comments were addressed.  LGTM

nalind:
Does the new `MountOpts.Options` field need to be wired in to any drivers, or is it just for `overlay`?

rhatdan:
@nalind Options is now wired into all drivers.

rhatdan:
@nalind I think this is ready to merge...

nalind:
LGTM

PR #227, opened 2018-10-31
make: clean up .PHONY targets list

marcov:
Make the list of .PHONY targets consistent with the actual Makefile
targets. Update The README with the correct make target for the
CLI wrapper.

Signed-off-by: Marco Vedovati <mvedovati@suse.com>


rhatdan:
LGTM
@TomSweeneyRedHat @vrothberg @nalind @giuseppe PTAL

PR #228, opened 2018-11-02
Update the versions of golang to test with

rhatdan:
We should be testing with the latest golang versions.
Change golang 8,9 to 10 and 11 for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,30 +28,30 @@ env:
         - GO_VERSION="stable"
           DISTRO="ubuntu"
 
-        - GO_VERSION="1.8"
+        - GO_VERSION="1.10"
           DISTRO="ubuntu"
 
-        - GO_VERSION="1.9"
+        - GO_VERSION="1.11"
           DISTRO="ubuntu"
 
         # Fedora
         - GO_VERSION="stable"
           DISTRO="fedora"
 
-        - GO_VERSION="1.9"
+        - GO_VERSION="1.11"
           DISTRO="fedora"
 
-        - GO_VERSION="1.8"
+        - GO_VERSION="1.10"
           DISTRO="fedora"
 
         # CentOS
         - GO_VERSION="stable"
           DISTRO="centos"
 
-        - GO_VERSION="1.9"
+        - GO_VERSION="1.11"
           DISTRO="centos"
 
-        - GO_VERSION="1.8"
+        - GO_VERSION="1.10"
           DISTRO="centos"
 
 # GO_VERSION="stable" builds successfully, but tests fail on all platforms.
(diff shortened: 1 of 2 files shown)


giuseppe:
LGTM

giuseppe:
could you rebase this PR?

vrothberg:
> there were changes in the tar archive format. I think go1.11 broke cache.
see openSUSE/umoci#269 and moby/moby#37498

I had a similar thought but the CI here seems to break on go 1.10 as well.

mtrmac:
Ah, `sameFsTime`/`sameFsTimeSpec` is already supposed to account for loss of time precision.

In 1.9.5, and apparently in 1.10 WIP branches ( https://go.googlesource.com/go/+/ead6255ce3bd39e56d66f39471ab1854fd4f67f2%5E%21/ ), the time was truncated _down_; since actual 1.10 ( https://go.googlesource.com/go/+/577aab0c595825ba113b63f4ef1460e8471c803e%5E%21/ ), time is _rounded_ to the whole second, and `sameFsTime{,Spec` don’t account for that.

Let’s reuse #238 to check this hypothesis…

---

**NOTE**: At a first glance seems, at the very best, _very unclear_ that unpredictably rounding times down/up is safe WRT change detection (remember all the `make` failures and corner cases before nanosecond time stamps). Has anyone checked that this is not a problem?

- From a quick check, `NaiveDiffDriver.Diff` (and related methods), through `addChanges`, seems to use `sameFsTimeSpec` - which seems both unnecessary (both source and destination are on a local filesystem) and risky (~2E-9 chance of not noticing a change). Am I missing something?
- Now that tar+untar can forward `ModTime` up to 0.5 seconds into the future, can this mean that a real change on the filesystem can happen to set `ModTime` to the same value, and remain unnoticed?

(Sure, maybe we can’t actually manage to create + need to diff a layer in 0.5 seconds right now, but we certainly want to make it as fast as possible…)

---

Meanwhile, the higher versions of Go have also uncovered two trivial real bugs: https://github.com/containers/storage/pull/239 .

mtrmac:
> Let’s reuse #238 to check this hypothesis…

OK, #238 confirms that all the outstanding failures (i.e. apart from #239) are caused by the timestamp rounding.

**DO NOT JUST MERGE THE WORKAROUND IN #238**.  That only shows where the failure manifests; as detailed above, AFAICS in at least some important code paths it seems that this relaxed check should not be used at all.  (Maybe the relaxed check should only exists in tests, even?)

**DO NOT JUST MERGE THE WORKAROUND IN #238**. I can’t tell that it is safe. I wash my hands of it.

**DO NOT JUST MERGE THE WORKAROUND IN #238**. If you ended up merging something similar, only ever do so based on your own knowledge or analysis of how the code works and why.

rhatdan:
Version updated in a different PR.

PR #230, opened 2018-11-06
overlay: use AUFS whiteout format when using mount_program

giuseppe:
an unprivileged user doesn't have the permission to use mknod (except
in file systems owned by the user itself, e.g. tmpfs), so the unpack
would always fail with EPERM.  Use the AUFS whiteout format that
doesn't require CAP_MKNOD.

Closes: https://github.com/containers/buildah/issues/1160

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -845,6 +845,17 @@ func (d *Driver) isParent(id, parent string) bool {
 	return ld == parentDir
 }
 
+func (d *Driver) getWhiteoutFormat() archive.WhiteoutFormat {
+	whiteoutFormat := archive.OverlayWhiteoutFormat
+	if d.options.mountProgram != "" {
+		// If we are using a mount program, we are most likely running
+		// as an unprivileged user that cannot use mknod, so fallback to the
+		// AUFS whiteout format.
+		whiteoutFormat = archive.AUFSWhiteoutFormat
+	}
+	return whiteoutFormat
+}
+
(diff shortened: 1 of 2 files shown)


giuseppe:
I've moved the CI fix here so that it unblocks other PRs as well: https://github.com/containers/storage/pull/235

rhatdan:
LGTM

rhatdan:
@nalind @vrothberg PTAL

drivers/overlay/overlay.go
```diff
@@ -909,13 +914,18 @@ func (d *Driver) Diff(id string, idMappings *idtools.IDMappings, parent string,
 		return nil, err
 	}
 
+	whiteoutFormat := archive.OverlayWhiteoutFormat
+	if d.options.mountProgram != "" {
```
vrothberg: Could you add a comment here? It seems like there's some implicit condition when `mountProgram != ""`.

giuseppe: I've pushed a new version with a comment (and the common code block refactored in a function)

vrothberg: Thanks a lot, @giuseppe !

vrothberg:
LGTM

giuseppe:
I'd expect more issues to come when using fuse-overlayfs, but for now this PR solves an EPERM and I've verified an image built with buildah and this patch correctly generates whiteout files.  I've inspected the generated .tar file and also tried to run it with Podman and Docker.

PR #237, opened 2018-11-13
archive: propagate InUserNS option to storage-untar

giuseppe:
we cannot check if we are running in an user namespace after the
chroot, as /proc is not accessible.  So let's pass this option down
before re-execing and doing a chroot.

Closes: https://github.com/containers/libpod/issues/1775

### DIFF
--- a/drivers/fsdiff.go
+++ b/drivers/fsdiff.go
@@ -8,6 +8,7 @@ import (
 	"github.com/containers/storage/pkg/chrootarchive"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/ioutils"
+	rsystem "github.com/opencontainers/runc/libcontainer/system"
 	"github.com/sirupsen/logrus"
 )
 
@@ -167,7 +168,9 @@ func (gdw *NaiveDiffDriver) ApplyDiff(id string, applyMappings *idtools.IDMappin
 	}
 	defer driver.Put(id)
 
-	options := &archive.TarOptions{}
+	options := &archive.TarOptions{
+		InUserNS: rsystem.RunningInUserNS(),
+	}
 	if applyMappings != nil {
 		options.UIDMaps = applyMappings.UIDs()
 		options.GIDMaps = applyMappings.GIDs()
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -29,6 +29,7 @@ import (
 	"github.com/containers/storage/pkg/parsers"
 	"github.com/containers/storage/pkg/system"
 	units "github.com/docker/go-units"
+	rsystem "github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -874,6 +875,7 @@ func (d *Driver) ApplyDiff(id string, idMappings *idtools.IDMappings, parent str
 		UIDMaps:        idMappings.UIDs(),
 		GIDMaps:        idMappings.GIDs(),
 		WhiteoutFormat: d.getWhiteoutFormat(),
+		InUserNS:       rsystem.RunningInUserNS(),
 	}); err != nil {
 		return 0, err
 	}
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -22,6 +22,7 @@ import (
 	"github.com/containers/storage/pkg/pools"
 	"github.com/containers/storage/pkg/promise"
 	"github.com/containers/storage/pkg/system"
+	rsystem "github.com/opencontainers/runc/libcontainer/system"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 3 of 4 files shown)


mtrmac:
I don’t understand how the “is running in user NS” checks are _supposed_ to work.

On one hand, this adds an automatic check fairly deep inside the call stack.

On the other hand, there is https://github.com/containers/image/pull/532/files which sets the `InUserNS` variable at the top level.

Also, compare https://github.com/containers/storage/blob/5235deddcce0b2b809ee75c637f7e7d7b644f9e1/pkg/archive/archive.go#L583 in the caller with the immediately following https://github.com/containers/storage/blob/5235deddcce0b2b809ee75c637f7e7d7b644f9e1/pkg/archive/archive_unix.go#L91 in the callee.

Who is _responsible_ for determining the behavior?

And, this may be related or mostly independent, who is _responsible_ for setting `TarOptions.InUserNS`?

---

I’m not really happy with just sprinkling checks close to test failures (… but then I don’t understand most of the c/storage code, and I don’t have, and don’t deserve, write access to this repo — so me being happy or not has no relevance :) ).

giuseppe:
I am not really familar with all the c/storage code as well, and took me a while to figure out how these checks fit together :-)
From what I understood:

The check in `handleTarTypeBlockCharFifo` will always fail when used by `storage-untar`.  After the `chroot` there is no `/proc` mounted and `/proc/self/uid_map` doesn't exist.  It believes to not be running in a user namespace even if it does.  I guess this check is there only for callers that don't setup a chroot and there is `/proc` mounted in.

Differently the check in `archive.go` uses the setting passed down by `Unpack` and it is higher in the stack, `handleTarTypeBlockCharFifo` won't be called at all.  This patch only makes sure the `options *TarOptions` argument for `Unpack` contains the correct value for `InUserNS` which is already expected to be set in the options.

rhatdan:
@nalind PTAL

mtrmac:
> I am not really familar with all the c/storage code as well, and took me a while to figure out how these checks fit together :-)

Arguably, this is not making anything about the design worse, so I should just shut up :)

> The check in `handleTarTypeBlockCharFifo` will always fail when used by `storage-untar`.

So, it should be either fixed (if possible), replaced, or removed; I don’t feel happy about leaving known-broken code around, (nor about leaving known-broken-in-some-situations code around without any explanation). (Again, fully acknowledging that making me feel happy is not the purpose of c/storage.)

>I guess this check is there only for callers that don't setup a chroot and there is `/proc` mounted in.

Ultimately, this leaks into the public API. So are public API callers expected to know which particular backends do or don’t internally use a chroot, and hard-code the relationship between chroots and `RunningInUserNS`?

---

It seems to me, ideally, that the behavior depends on whether `RunningIsUserNS` returns correct values when called at the API boundary of c/storage:

- If it does, `InUserNS` should be private (or technically public but documented as internal-use-only) to c/storage, and set at the API boundary of c/storage; then callers don’t need to care, and we make exactly one `RunningInUserNS` call (which is cheap but not quite negligible) per operation.
- If it does not, `InUserNS` needs to be public, the responsibility of the caller, nothing inside c/storage should call `RunningInUserNS` (and the public API of c/storage may need extending so that the `RunningInUserNS` option can always be conveniently passed, which is not quite the case in https://github.com/containers/image/pull/532 .)
- If “it depends”, then I don’t know.

giuseppe:
> > 
> 
> So, it should be either fixed (if possible), replaced, or removed; I don’t feel happy about leaving known-broken code around, (nor about leaving known-broken-in-some-situations code around without any explanation). (Again, fully acknowledging that making me feel happy is not the purpose of c/storage.)

I don't think it can be fixed without mounting `/proc` so the answer is "maybe", it returns correct values only in certain cases.  I don't think though we can just drop the check though, as any caller that is not going through `Unpack` is going to hit it.  It was not added to support rootless containers, AFAICS, it was already present in graphdriver (https://github.com/moby/moby/blob/master/pkg/archive/archive_unix.go#L83-L86)

giuseppe:
@nalind are you fine with this change?

giuseppe:
can someone please take a look at this PR?

TomSweeneyRedHat:
LGTM, but want a head nod from @nalind or @mtrmac

nalind:
My preference here is to not overwrite that field if it's part of a structure that's passed to us by an external caller that might be in a better position to know what value that field should have.  In places where we currently create a `TarOptions` structure (at least the code paths that extract archives), we should make sure that we're specifying a value for the field.

giuseppe:
> My preference here is to not overwrite that field if it's part of a structure that's passed to us by an external caller that might be in a better position to know what value that field should have. In places where we currently create a `TarOptions` structure (at least the code paths that extract archives), we should make sure that we're specifying a value for the field.

thanks for the feedback.  I've changed it to be set when the structure is created and added a patch for doing it from overlay as well

giuseppe:
@nalind is the new version fine for you?

nalind:
I think we'd also want to be setting this in `NaiveDiffDriver.ApplyDiff()`.

giuseppe:
> I think we'd also want to be setting this in `NaiveDiffDriver.ApplyDiff()`.

I've pushed another patch to change it

mtrmac:
How about `pkg/archive/Archiver.UntarPath`?  (That would make https://github.com/containers/image/pull/532 unnecessary.)

giuseppe:
> How about `pkg/archive/Archiver.UntarPath`? (That would make [containers/image#532](https://github.com/containers/image/pull/532) unnecessary.)

yes good idea.  I've tested that it is enough to solve the ostree problem, so we can close containers/image#532 as the new patch I've added is enough.

rhatdan:
LGTM

nalind:
I think that'll work.  LGTM.

PR #238, opened 2018-11-14
DO NOT MERGE: Dump reason for ChangeModify entries

mtrmac:
Debugging for https://github.com/containers/storage/pull/228 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,30 +28,30 @@ env:
         - GO_VERSION="stable"
           DISTRO="ubuntu"
 
-        - GO_VERSION="1.8"
+        - GO_VERSION="1.10"
           DISTRO="ubuntu"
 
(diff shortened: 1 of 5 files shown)


mtrmac:
DO NOT MERGE: Now includes #239 to get closer to the actual problems.

mtrmac:
See more discussion in #228.

**DO NOT JUST MERGE THIS**.  This only shows where the failure manifests; as detailed in #228, AFAICS in at least some important code paths it seems that this relaxed check should not be used at all.  (Maybe the relaxed check should only exists in tests, even?)

**DO NOT JUST MERGE THIS**. I can’t tell that it is safe. I wash my hands of it.

**DO NOT JUST MERGE THIS**. If you ended up merging something similar, only ever do so based on your own knowledge or analysis of how the code works and why.

PR #243, opened 2018-11-26
Refactor ParseIDMap, allow parsing multiple maps at once

isimluk:
I found that other projects, tend to parse multiple maps at once. So, we may want to allow the base library to do so in order to decrease complexity in the upper layers. (For comparison please see: https://github.com/kubernetes-sigs/cri-o/blob/4cd5a7c60349be0678d9f1b0657683324c1a2726/server/server.go#L248 https://github.com/containers/libpod/blob/e8e16fcc789df6a705fbbf5c5597f16a4255d690/pkg/util/utils.go#L172 https://github.com/containers/buildah/blob/c116133dd5d62e3791ff792468c9a11b1bdffafd/util/util.go#L421 )

This is follow-up on previous refactoring in 7b209d36fd98ff0a969, I didn't got it right on first try, sry.

Signed-off-by: Šimon Lukašík <isimluk@fedoraproject.org>

### DIFF
--- a/cmd/containers-storage/create.go
+++ b/cmd/containers-storage/create.go
@@ -56,11 +56,11 @@ func paramIDMapping() (*storage.IDMappingOptions, error) {
 		options.UIDMap = mappings.UIDs()
 		options.GIDMap = mappings.GIDs()
 	}
-	parsedUIDMap, err := idtools.ParseIDMap(paramUIDMap, "uid")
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@nalind @giuseppe @TomSweeneyRedHat PTAL

PR #245, opened 2018-11-27
Should still delete container if layer does not exist.

rhatdan:
We have seen situations in buildah where a container is being
built and user hits ^c, then he ends up in a situation where
he can not delete container, since layer does not exist.


rhatdan:
FIxes https://github.com/containers/buildah/issues/1200

nalind:
LGTM

PR #246, opened 2018-11-29
Rebase vfs code to use Moby definition of copy driver

rhatdan:
This new version of copy will attempt to use reflinks
if the underlying file system supports reflink, which
should optimize the vfs storage driver for rootless
podman and buildah.

### DIFF
--- a/drivers/vfs/copy_linux.go
+++ b/drivers/vfs/copy_linux.go
@@ -1,0 +1,7 @@
+package vfs
+
+import "github.com/containers/storage/drivers/copy"
+
+func dirCopy(srcDir, dstDir string) error {
+	return copy.DirCopy(srcDir, dstDir, copy.Content, false)
+}
--- a/drivers/copy/copy.go
+++ b/drivers/copy/copy.go
@@ -1,0 +1,277 @@
+// +build linux
+
+package copy
+
+/*
+#include <linux/fs.h>
+
+#ifndef FICLONE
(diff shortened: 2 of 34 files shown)


rhatdan:
@nalind @vrothberg @giuseppe @rhvgoyal PTAL

giuseppe:
I've tested it with Podman and XFS with reflinks enabled and it works as expected

LGTM

vrothberg:
The build errors look really odd:
```
[...]
vendor/github.com/ostreedev/ostree-go/pkg/otbuiltin/init.go:59: cannot use r (type C.gboolean) as type C.int in argument to isOk
[...]
```

giuseppe:
there is PR to address it: https://github.com/containers/storage/pull/247

but needs another re-vendoring as it doesn't include the fix for go 1.8

rhatdan:
@giuseppe The strange thing here is that these tests no longer exists in moby code.  So I am not sure what is correct or not here.

nalind:
LGTM

drivers/vfs/copy_linux.go
```diff
@@ -0,0 +1,7 @@
+package vfs
+
+import "github.com/containers/storage/drivers/copy"
+
+func dirCopy(srcDir, dstDir string) error {
+	return copy.DirCopy(srcDir, dstDir, copy.Content, false)
```
nalind: I'm not 100% that this should be `false` rather than `true`, but it's not a blocker.

PR #250, opened 2018-12-04
Add list of supported graph drivers to manpage

mheon:
People looking to modify storage.conf might be confused as to what the valid graph drivers and their names are. List all supported drivers in the manpage to assist them.

Shouldn't be necessary elsewhere, most of the other fields are just paths.


mheon:
@rhatdan @nalind PTAL - I don't know if my list of drivers here is actually what we should be putting in (maybe remove devmapper, maybe add a disclaimer on btrfs/zfs that support might not be built into the library)

mheon:
Updated to add a note about kernel support

rhatdan:
LGTM
@TomSweeneyRedHat PTAL

PR #251, opened 2018-12-04
graphtest: insulate against filesystems with low timestamp granularity

nalind:
When we test for change detection on top of filesystems with low timestamp granularity, we can end up updating a file without forcing a change to its modification timestamp.  Check for that when we attempt to
modify the files, and just keep rewriting it until the timestamp changes.


nalind:
@rhatdan I think this'll fix the problems you're hitting in #246.

giuseppe:
LGTM

rhatdan:
A little on the yuck side, but LGTM

PR #254, opened 2018-12-12
archive: ignore setxattr EPERM in a user namespace

giuseppe:
old kernels, such as 4.9 used by Debian 9 do not support extended
attributes in a user namespace.  In this case, ignore the error and
print a warning.  This should be fine, as we are already doing it for
file systems that do not support xattrs.

Closes: https://github.com/containers/libpod/issues/1941

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -642,11 +642,12 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 	var errors []string
 	for key, value := range hdr.Xattrs {
 		if err := system.Lsetxattr(path, key, []byte(value), 0); err != nil {
-			if err == syscall.ENOTSUP {
+			if err == syscall.ENOTSUP || (err == syscall.EPERM && inUserns) {
 				// We ignore errors here because not all graphdrivers support
 				// xattrs *cough* old versions of AUFS *cough*. However only
 				// ENOTSUP should be emitted in that case, otherwise we still
-				// bail.
+				// bail.  We also ignore EPERM errors if we are running in a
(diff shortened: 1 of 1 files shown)


vrothberg:
LGTM :+1:

mheon:
LGTM, thanks @giuseppe

maxamillion:
Thank you!

pkg/archive/archive.go
```diff
@@ -642,11 +642,12 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 	var errors []string
 	for key, value := range hdr.Xattrs {
 		if err := system.Lsetxattr(path, key, []byte(value), 0); err != nil {
-			if err == syscall.ENOTSUP {
+			if err == syscall.ENOTSUP || (err == syscall.EPERM && inUserns) {
 				// We ignore errors here because not all graphdrivers support
 				// xattrs *cough* old versions of AUFS *cough*. However only
 				// ENOTSUP should be emitted in that case, otherwise we still
-				// bail.
+				// bail.  We also ignore EPERM errors if we are running in a
```
rhatdan: Where are you printing the warning?

giuseppe: they are printed after the loop:

```
	if len(errors) > 0 {
		logrus.WithFields(logrus.Fields{
			"errors": errors,
		}).Warn("ignored xattrs in archive: underlying filesystem doesn't support them")
	}
```

rhatdan: Ok, now I get it.

rhatdan:
LGTM

PR #255, opened 2018-12-13
vendor: use github.com/klauspost/pgzip instead of compress/gzip

giuseppe:
from my tests, I've seen a net improvement of around 30% on the wall
clock time in decompressing layers.

These additional packages will need to be re-vendored:

github.com/klauspost/pgzip v1.2.1
github.com/klauspost/compress v1.4.1
github.com/klauspost/cpuid v1.2.0


rhatdan:
LGTM
@nalind @vrothberg @TomSweeneyRedHat @umohnani8 @mtrmac PTAL

giuseppe:
WIP PR for skopeo here: https://github.com/containers/skopeo/pull/572

rhatdan:
@giuseppe Could you create a PR against CRI-O with this request and the containers/image PR vendored in and see if you can get it to pass the CRI-O Test suite.

giuseppe:
@rhatdan done here: https://github.com/kubernetes-sigs/cri-o/pull/1963

giuseppe:
> The changes LGTM, my only concern is the author of this utility says to "Never use on input smaller than 1MB." [here](https://blog.klauspost.com/go-gzipdeflate-benchmarks/). I couldn't parse out if there were just slowness problems at that level (I'm assuming so), or real problems. Any way to create really small layers and try this against those?

I'd not be too concerned about <1M layers, that are anyway too small to make any significant difference to justify us to use two different implementations.  In any case, I'll try to collect some numbers.

giuseppe:
I've done some tests.  It really depends on how the uncompressed data is used as it seems all the difference is coming from the read ahead functionality of the parallel version.

If I just `io.Copy` the uncompressed data from the gzip reader to `io.Discard`, then the stdlib gzip decompressor is always faster, no matter what is the size of the compressed data.  This is probably because the read ahead cannot read faster than what we consume.

Instead if I add a delay, even small one like 1 millisecond for each Mb of data uncompressed, then the parallel version is significantly faster, ~2x, also here the size of the compressed layer doesn't make any difference.

Layers smaller than 1Mb are <1ms with both versions so I think it makes no sense to worry of them.

TomSweeneyRedHat:
@giuseppe TYVM for running through all the test.  The < 1MB layers results are very good to hear.  The warning in his blog was a little concerning, but I feel comfortable going forward with this based on your tests.

vrothberg:
To add some confidence to that. umoci is using it, so I consider the library to be quite battle-tested given umoci is used to build all openSUSE (and SUSE) containers.

vrothberg:
LGTM (https://github.com/containers/skopeo/pull/572 passes)

TomSweeneyRedHat:
LGTM

rhatdan:
Since the CRI-O Tests is passing and the skopeo, I will merge.

PR #256, opened 2018-12-17
layers: use the package name pgzip instead of gzip

giuseppe:
based on some comments got on the related PR for containers/image.


vrothberg:
LGTM :+1:

PR #258, opened 2019-01-03
handle ENOSPC for overlay driver

baude:
when detecting if the overlay driver can be used a temporary file is
created.  if the system is ENOSPC, that fails but the error message
response was misleading.  this patch makes the error message in that
case more obvious.

Signed-off-by: baude <bbaude@redhat.com>


TomSweeneyRedHat:
LGTM assuming happy tests

rhatdan:
LGTM

PR #261, opened 2019-01-10
PoC/WIP: add an LVM driver

nalind:
Add a driver named "lvm" which calls the "lvm" and "losetup" binaries to implement layer storage using an LVM thin pool.


nalind:
Closing in favor of #285.

PR #262, opened 2019-01-12
Use mount options when mounting the chown layer.

rhatdan:
If we set metacopyup flag in storage.conf, we need to use
this field, when creating the intermediate storage layer.


rhatdan:
@nalind @rhvgoyal PTAL

nalind:
LGTM

PR #263, opened 2019-01-14
Update zfs driver to be closer to moby driver

rhatdan:
There is a bug report reporting issues on zfs driver on reboot.
Wanted to make sure we have close to the same code as the Moby Project.


drivers/zfs/zfs.go
```diff
@@ -360,11 +359,25 @@ func (d *Driver) Remove(id string) error {
...
-func (d *Driver) Get(id string, options graphdriver.MountOpts) (string, error) {
+func (d *Driver) Get(id string, options graphdriver.MountOpts) (_ string, retErr error) {
+
 	mountpoint := d.mountPath(id)
 	if count := d.ctr.Increment(mountpoint); count > 1 {
 		return mountpoint, nil
 	}
+	defer func() {
+		if retErr != nil {
+			if c := d.ctr.Decrement(mountpoint); c <= 0 {
```
vrothberg: This looks like a bug fix, which may fixes the ZFS issue we're facing.

rhatdan:
https://github.com/containers/libpod/issues/2150

vrothberg:
@nalind PTAL (the update seems to fix the linked Podman issue)

drivers/zfs/zfs.go
```diff
@@ -109,9 +111,6 @@ func Init(base string, opt []string, uidMaps, gidMaps []idtools.IDMap) (graphdri
 		return nil, fmt.Errorf("Failed to create '%s': %v", base, err)
 	}
 
-	if err := mount.MakePrivate(base); err != nil {
-		return nil, err
-	}
```
nalind: Are we sure that this is something that we no longer need to do for zfs filesystems for our use cases?

drivers/zfs/zfs.go
```diff
@@ -38,7 +38,7 @@ type Logger struct{}
 
 // Log wraps log message from ZFS driver with a prefix '[zfs]'.
 func (*Logger) Log(cmd []string) {
-	logrus.Debugf("[zfs] %s", strings.Join(cmd, " "))
+	logrus.WithField("storage-driver", "zfs").Debugf("[zfs] %s", strings.Join(cmd, " "))
```
nalind: Non-blocker: remove the `[zfs]` from the log message if we're going to use a field for it.

drivers/zfs/zfs.go
rhatdan: I don't know.

rhatdan: Added it back, since other drivers all do it.  Will remove if the problem comes back to the user.

drivers/zfs/zfs.go
rhatdan: Fixed

nalind:
Switching the error handling logic from attempting to unwind the things that we've done successfully to always unwinding everything (plus removing the mountpoint directory) seems like it might be papering over something, but it's hard to argue with that when it fixes the bug for @vrothberg.  Looks fine otherwise.

drivers/zfs/zfs.go
vrothberg: Here's the Moby commit removing the code: https://github.com/moby/moby/commit/9803272f2db84df7955b16c0d847ad72cdc494d1

rhatdan: Well that is the exact code I added back, and that is the difference between the patch he tested and the final result.

The reason that code was in the original storage drivers is to prevent leaked mount points from propagating out of the storage directory.  @rhvgoyal I believe questions if we still need it.  I say remove it to fix this issue and then see if we see mount leak problems in the future.

PR #265, opened 2019-01-16
Remove mounts private from zfs, it is causing issues on reboot with zfs

rhatdan:
This was originally put in to prevent potential leaks of mount points to
different mount namespaces.  Lets remove it and fix the issue with
zfs, and see if we experience leaked mounts.  oci-umount should also
clean up any leaks, if installed.


TomSweeneyRedHat:
LGTM, but would like a head nod for someone else.

vrothberg:
LGTM, thanks Dan!

PR #267, opened 2019-01-16
Check for the OS when setting btrfs/libdm/ostree tags

nalind:
Before checking for btrfs/libdm/ostree availability, check which OS we're on, and if it isn't Linux, disable them.  This should simplify cross-compiling at least a bit.


rhatdan:
LGTM

PR #268, opened 2019-01-16
overlay: check if metacopy is happening

nalind:
In the overlay driver, check if metacopy is enabled, and report it along with other status information.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -105,6 +105,7 @@ type Driver struct {
 	options       overlayOptions
 	naiveDiff     graphdriver.DiffDriver
 	supportsDType bool
+	usingMetacopy bool
 	locker        *locker.Locker
 	convert       map[string]bool
 }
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@rhvgoyal PTAL

drivers/overlay/overlay.go
```diff
@@ -174,6 +175,18 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
...
+	usingMetacopy, err := doesMetacopy(home, opts.mountOptions)
+	if err == nil {
+		if usingMetacopy {
+			logrus.Debugf("overlay test mount indicated that metacopy is being used")
+		} else {
+			logrus.Debugf("overlay test mount indicated that metacopy is not being used")
+		}
+	} else {
+		logrus.Debugf("overlay test mount did not indicate that metacopy is being used: %v", err)
+		usingMetacopy = false
```
rhvgoyal: This is strange. If you get err returned from doesMetacopy(), then it should be returned back to higher layers?

nalind: Oh, right, we suppress an ENODATA in `pkg/system`.  Yeah, no harm in reporting it, then.  Changing it.

PR #269, opened 2019-01-16
Don't set "override_kernel_check" any more

nalind:
The logic that depended on `override_kernel_check` was changed to test for the feature at runtime, so we don't need to be suggesting to people that they need to set this option.


storage.conf
```diff
@@ -30,7 +30,7 @@ size = ""
 #mount_program = "/usr/bin/fuse-overlayfs"
 
 # OverrideKernelCheck tells the driver to ignore kernel checks based on kernel version
```
rhatdan: Lets just remove this code chunk from the storage.conf.  Since their is no code that does anything with it.

nalind:
Changed to scrub all references to this.  We weren't doing anything when the option was set, other than remembering and reporting that it was set.

PR #270, opened 2019-01-17
Don't fail the mounter test because the kernel added options

nalind:
On SELinux-enabled kernels, the kernel will helpfully add a "seclabel" option to mounted filesystems to indicate that labeling is supported.  Don't let that break our tests.


rhatdan:
LGTM

PR #274, opened 2019-01-18
Move copytar functions from buildah to storage

rhatdan:
We want to use these functions in both podman and buildah
So move them out of buidlah into storage.

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"syscall"
 
 	"github.com/containers/storage/pkg/fileutils"
@@ -23,6 +24,7 @@ import (
 	"github.com/containers/storage/pkg/system"
 	gzip "github.com/klauspost/pgzip"
 	rsystem "github.com/opencontainers/runc/libcontainer/system"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


rhatdan:
@QiWang19 PTAL

rhatdan:

giuseppe:
LGTM

nalind:
Needs tests.

TomSweeneyRedHat:
Overall change LGTM, but will let you and @nalind hash out the appropriate location for the bits.  If landing here, I second the test request.

rhatdan:
@nalind @TomSweeneyRedHat I have added tests.

pkg/archive/archive.go
```diff
@@ -1331,3 +1333,108 @@ const (
 	// HeaderSize is the size in bytes of a tar header
 	HeaderSize = 512
 )
+
+// NewArchiver returns a new Archiver which uses chrootarchive.Untar
```
nalind: Comment is incorrect.

pkg/chrootarchive/archive.go
```diff
@@ -81,3 +79,82 @@ func untarHandler(tarArchive io.Reader, dest string, options *archive.TarOptions
...
+	}
+	return archiver.UntarPath
+}
+
+// TarPath returns a function which creates an archive of a specified
+// location in the container's filesystem, mapping permissions using the
+// container's ID maps
+func TarPath(uidmap []idtools.IDMap, gidmap []idtools.IDMap) func(path string) (io.ReadCloser, error) {
+	return archive.TarPath(uidmap, gidmap)
+}
```
nalind: Having this here could be misleading, since there is no chroot involved when this function is called.

pkg/archive/archive.go
rhatdan: Fixed

pkg/chrootarchive/archive.go
rhatdan: Removed

PR #275, opened 2019-01-21
overlay: do not attempt metacopyup when using a mount program

giuseppe:
Signed-off-by: Giuseppe Scrivano <giuseppe@scrivano.org>

PR #276, opened 2019-01-21
Don't check metacopy if in rootless/mountProgram mode

rhatdan:
Currently podman rootless will not work if you set metacopyup in
the /etc/containers/storage.conf file.


rhatdan:
Closed for https://github.com/containers/storage/pull/275

PR #278, opened 2019-01-31
reader-writer file locks

vrothberg:
ready for review :^)


lockfile.go
```diff
@@ -13,7 +13,11 @@ import (
 // identifier of the last party that made changes to whatever's being protected
 // by the lock.
 type Locker interface {
-	sync.Locker
+	Lock()
+
+	Unlock()
+
+	//	RLock()
```
rhatdan: Why the comment?

lockfile_unix.go
```diff
@@ -37,24 +39,40 @@ type lockfile struct {
...
-		Type:   l.locktype,
+		Type:   l_type,
 		Whence: int16(os.SEEK_SET),
 		Start:  0,
 		Len:    0,
 		Pid:    int32(os.Getpid()),
 	}
-	l.mu.Lock()
-	l.locked = true
+	if !l.ro {
```
rhatdan: So we only lock with write lock?

lockfile_unix.go
```diff
@@ -37,24 +39,40 @@ type lockfile struct {
...
+func (l *lockfile) Lock() {
+	if l.ro {
+		l.RLock()
+	} else {
+		l.lock(unix.F_WRLCK, unix.F_SETLKW)
+	}
+}
+
+// LockRead locks the lock file as a reader
+func (l *lockfile) RLock() {
```
rhatdan: Why not have a function RWLock()?

lockfile_unix.go
```diff
@@ -67,8 +85,10 @@ func (l *lockfile) Unlock() {
 	for unix.FcntlFlock(l.fd, unix.F_SETLKW, &lk) != nil {
 		time.Sleep(10 * time.Millisecond)
 	}
-	l.locked = false
-	l.mu.Unlock()
+	if !l.ro {
+		l.locked = false
```
rhatdan: Why not a function l.Unock()

vrothberg: Just for testing.

lockfile_unix.go
vrothberg: The idea is to have `Lock()` for acquiring a write lock and `RLock()` for the read counter-part. It makes transitioning the code easier as we don't need to change all calls but only the ones where we access reading.

lockfile_unix.go
vrothberg: Also testing but, if we have a read-only lock, then I don't see why we need to have mutex which would serialize reads of the same process.

lockfile.go
vrothberg: Testing. We need that when the PR progresses. The interface needs to change to expose RLock() but not all implementations of the interface yet support it.

rhatdan:
Nice work.

rhatdan:
@vrothberg One thing you might want to try, would be to vendor this into CRI-O in a test build and then fire off the tests against it.  Might give us a better feeling that we will not have issues.

vrothberg:
For the record, I opened the following CI PRs to check for potential regressions:
- CRI-O: https://github.com/kubernetes-sigs/cri-o/pull/2051
- Podman: https://github.com/containers/libpod/pull/2254
- Buildah: https://github.com/containers/buildah/pull/1322

lockfile.go
```diff
@@ -13,7 +12,14 @@ import (
 // identifier of the last party that made changes to whatever's being protected
 // by the lock.
 type Locker interface {
-	sync.Locker
+	// Acquire a writerts lock.
```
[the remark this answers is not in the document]
vrothberg: :-D Thanks!

vrothberg:
Alright. I fixed some issues when using `defer x.Unlock()` on loop variables causing the Unlock() to always be called on the last item of the slice which caused various issues. I further polished reader-writer locks and did some more optimizations on the code (mostly avoiding the fcntl(2) syscall from the same process space), so that a single push is now **30 percent** faster on my notebook. I will update the CI PRs but consider this PR to be the done now (unless CI disagrees).

@rhatdan @nalind @giuseppe Can you have a look at the three commits of this PR?

vrothberg:
Podman and Buildah are green. CRI-O is green but the e2e tests which have a regression from upstream (as all other PRs for CRI-O).

lockfile.go
```diff
@@ -39,44 +45,37 @@ var (
...
 	if lockfiles == nil {
 		lockfiles = make(map[string]Locker)
 	}
 	cleanPath := filepath.Clean(path)
 	if locker, ok := lockfiles[cleanPath]; ok {
-		if locker.IsReadWrite() {
-			return nil, fmt.Errorf("lock %q is a read-write lock", cleanPath)
+		if ro && locker.IsReadWrite() {
+			return nil, errors.Errorf("lock %q is not a read-only lock", cleanPath)
+		} else if !ro && !locker.IsReadWrite() {
```
rhatdan: Nit , no reason for the else here.

store.go
```diff
@@ -2808,7 +2833,8 @@ func (s *store) Image(id string) (*Image, error) {
 	if err != nil {
 		return nil, err
 	}
-	for _, store := range append([]ROImageStore{istore}, istores...) {
+	for _, s := range append([]ROImageStore{istore}, istores...) {
```
rhatdan: Why make this change?

vrothberg: `&s` always point to the same memory as the loop variable is allocated only once, even for non-primitive types. What happens when using `defer s.Unlock()` in the loop is that we call unlock the very same lock (always the last item in the slice) for `len(append([]ROImageStore{istore}, istores...))` times.

Ultimately, I want to remove all `defer lock.Unlock()` occurances in the code as it delays releasing the lock. The "trick" used in this commit is allocate a new pointer on the heap `store := s` and use that in the defer statement.

Note that the issue of using loop variables in go applies to other cases as well. I wrote a blog about that last April (https://vrothberg.github.io/2018/04/02/pointers-in-golang-considered-harmful.html).

vrothberg: Note that this particular problem of unlocking the same lock multiple times predates this PR. I am a bit surprised that it did not cause (major) issues yet as it implies that some lock remained unlocked and others have been unlocked too often.

lockfile.go
vrothberg: Nice catch, thanks! Just updated the PR.

rhatdan:
LGTM
@nalind WDYT?

lockfile_unix.go
```diff
@@ -24,38 +25,84 @@ func getLockFile(path string, ro bool) (Locker, error) {
...
+	locktype   int16
+	locked     bool
+	ro         bool
 }
 
-// Lock locks the lock file
-func (l *lockfile) Lock() {
+// lock locks the lockfile via FCTNL(2) based on the specified type and
+// command.
+func (l *lockfile) lock(l_type int16, l_cmd int) {
```
nalind: `l_cmd` is currently ignored.  Is it supposed to be?

nalind:
Good catch!  We also need the locking loop fix added to `store.ContainerSize()` and `store.LayerSize()`.  LGTM with a couple of pieces added.

lockfile_unix.go
vrothberg: Wow, good catch! That's an artifact of an earlier version but not needed anymore as we're always setting the lock via `unix.F_SETLKW`. Will clean that up, thanks a lot!

vrothberg:
> Good catch! We also need the locking loop fix added to `store.ContainerSize()` and `store.LayerSize()`. LGTM with a couple of pieces added.

Thanks for the review, @nalind! I updated the commits accordingly and repushed.

lockfile_unix.go
```diff
@@ -64,19 +111,40 @@ func (l *lockfile) Unlock() {
 		Len:    0,
 		Pid:    int32(os.Getpid()),
 	}
-	for unix.FcntlFlock(l.fd, unix.F_SETLKW, &lk) != nil {
-		time.Sleep(10 * time.Millisecond)
+	l.stateMutex.Lock()
+	if l.locked == false {
+		// Panic when unlocking and unlocked lock.  That's a vioalation
```
giuseppe: nitpick: "and" and "vioalation"

vrothberg: Thanks! Just fixed it :)

rhatdan:
LGTM

PR #282, opened 2019-02-05
Don't ignore errors reloading layer/image/container lists

nalind:
Check for errors returned by various `Load()` methods.


rhatdan:
LGTM

PR #283, opened 2019-02-05
Teach images to hold multiple manifests

nalind:
Change how we compute digests for BigData items with names that start with "manifest" so that we use the image library's manifest.Digest() function, which knows how to preprocess schema1 manifests to get the right value, instead of just trying to finesse it.

Track the digests of multiple manifest-named items for images.


rhatdan:
LGTM
@mtrmac PTAL

images.go
```diff
@@ -27,12 +29,19 @@ type Image struct {
...
+	// Digests is a list of digest values of the image's manifests, and
+	// possibly a manually-specified value, that we can use to locate the
+	// image.  If Digest is set, its value is also in this list.
+	Digests []digest.Digest `json:"-"`
+
 	// Names is an optional set of user-defined convenience values.  The
 	// image can be referred to by its ID or any of its names.  Names are
-	// unique among images.
+	// unique among images, and are typically the text representation of
+	// tagged or canonical references.
```
mtrmac: (FWIW libpod now has code that fails if parsing elements with `reference.ParseNormalizedNamed` fails.  I don’t have an opinion on whether it should fail like that, just that it would be nice everything agreed on whether this format is required or only recommended.)

images.go
```diff
@@ -167,6 +179,31 @@ func (r *imageStore) datapath(id, key string) string {
 	return filepath.Join(r.datadir(id), makeBigDataBaseName(key))
 }
 
+// bigDataNameIsManifest determines if a big data item with the specified name
+// is considered to be representative of the image, in that its digest can be
+// said to also be the image's digest.  Currently, if its name is, or begins
+// with, "manifest", we say that it is.
+func bigDataNameIsManifest(name string) bool {
+	return name == ImageDigestBigDataKey || strings.HasPrefix(name, ImageDigestBigDataKey)
```
mtrmac: (`HasPrefix` includes the `name == ImageDigestBigDataKey` case. OTOH handling it separately doesn’t hurt at all.)

images.go
```diff
@@ -167,6 +179,31 @@ func (r *imageStore) datapath(id, key string) string {
...
+}
+
+// recomputeDigests takes the list of an image's digests and deduplicates them.
+func recomputeDigests(inputDigests []digest.Digest) (validDigests []digest.Digest, err error) {
```
mtrmac: (`s/recompute/deduplicate/`, then? or `validateAndDeduplicate`, or… I don’t know; it’s just that this is definitely not recomputing the values.

Or maybe move more code from both callers, setting `.Digests` based on `.Digest` and `BigDataDigests`, in here.)

images.go
```diff
@@ -333,12 +380,19 @@ func (r *imageStore) Create(id string, names []string, layer, metadata string, c
...
+	}
+	var inputDigests []digest.Digest
+	if searchableDigest != "" {
+		inputDigests = append(inputDigests, searchableDigest)
+	}
+	digests, err := recomputeDigests(inputDigests)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error validating digests (%v) for new image", names)
 	}
-	names = dedupeNames(names)
```
mtrmac: Is dropping this intentional?

images.go
```diff
@@ -15,9 +17,9 @@ import (
 )
 
 const (
-	// ImageDigestBigDataKey is the name of the big data item whose
-	// contents we consider useful for computing a "digest" of the
-	// image, by which we can locate the image later.
+	// ImageDigestBigDataKey is a prefix of big data item names which we
+	// consider to be manifests, used for computing a "digest" value for
+	// the image as a whole, by which we can locate the image later.
 	ImageDigestBigDataKey = "manifest"
```
mtrmac: Should the name of the variable be updated now that it is a prefix of a key?

Or, rather, if the goal is to store all of the manifests and to allow looking them up by digest, maybe c/storage should prescribe the manifest big data element naming, and provide an `ImageBigDataKeyForManifest(manifest []byte)` (or a `…ForDigest(digest digest.Digest)`?) so that this convention is defined in exactly one place.

images.go
```diff
@@ -637,20 +686,31 @@ func (r *imageStore) SetBigData(id, key string, data []byte) error {
...
+			}
+		}
+		image.Digests, err = recomputeDigests(digestList)
+		if err != nil {
+			return errors.Wrapf(err, "error loading recomputing image naming information for %s", image.ID)
+		}
+		for _, newDigest := range image.Digests {
 			// add the image to the list of images in the digest-based index which
 			// corresponds to the new digest for this item, unless it's already there
 			list := r.bydigest[newDigest]
```
mtrmac: Above, all entries of (old) `image.Digests` have been removed from `r.bydigest`; doesn’t that mean that `len == len(imageSliceWithoutValue)` below will always be true?

images.go
nalind: Doh.  Yeah, no need to keep that.

images.go
nalind: This is me walking back some larger changes that would have added a requirement that values passed to `SetNames()` for images also passed `ParseNormalizedNamed()`.  I remain unsure if that's something we want to be enforcing down here.

images.go
nalind: Whoops, it was not.  Restoring it.

images.go
nalind: In practice it _should_, but uniqueness of manifests among images isn't something we're enforcing at this level.

images.go
nalind: Oh, that makes sense.  Moving some of that duplicated logic into `recomputeDigests()`.

images.go
nalind: I wanted to avoid enforcing a convention so that the image library would be the place where that could be decided, but manifests are a concession to that.  One thing I'd been worried about was avoiding cases where a newer version of the image library that takes advantage of this was mixed with an older version of the storage library that wouldn't index the manifest's digest unless its name was exactly "manifest", and changing the name of the constant that an updated containers/image will use will accomplish that.  I'll leave the old name around for now so that we don't immediately break older versions of containers/image.

rhatdan:
@mtrmac could you take one more look and then we can merge.

images.go
```diff
@@ -189,17 +242,27 @@ func (r *imageStore) Load() error {
...
-				digests[digest] = append(digests[digest], images[n])
+			// Compute the digest list.
+			digestList := make([]digest.Digest, 0, len(image.Names)+1)
+			if image.Digest != "" {
+				digestList = append(digestList, image.Digest)
+			}
+			for itemName, itemDigest := range image.BigDataDigests {
+				if bigDataNameIsManifest(itemName) {
+					digestList = append(digestList, itemDigest)
+				}
```
mtrmac: `digestList` is now unused.

images.go
mtrmac: Ah, right; I realized at some point that IDs are not necessarily config digests but did not notice how it applies to this.

images.go
```diff
@@ -637,20 +699,22 @@ func (r *imageStore) SetBigData(id, key string, data []byte) error {
...
+			if list, ok := r.bydigest[oldDigest]; ok {
+				prunedList := imageSliceWithoutValue(list, image)
+				if len(prunedList) == 0 {
+					delete(r.bydigest, oldDigest)
+				} else {
+					r.bydigest[oldDigest] = prunedList
 				}
 			}
+		}
+		image.Digests, err = recomputeDigests(image.Digest, image.BigDataDigests)
```
mtrmac: (Non-blocking nit: `recomputeDigests` could be a method on `image`, making this `err = image.recomputeDigests()` everywhere.)

images.go
nalind: Aargh, missed that.  Cleared.

rhatdan:
@nalind WDYT about moving recomputeDigests to image.recomputeDigests

nalind:
> @nalind WDYT about moving recomputeDigests to image.recomputeDigests

Sure, done.

nalind:
Also changed `ImageStore.ByDigest()` to return copies of the images in its results slice, since callers aren't supposed to be messing with the copies that we keep.

rhatdan:
LGTM

PR #284, opened 2019-02-07
images: always set Digest if we have digests

nalind:
Make sure that an Image that has at least one digest always has a populated Digest field.


nalind:
@rhatdan PTAL.  This got lost in a last-minute change in https://github.com/containers/storage/pull/283.

rhatdan:
LGTM

rhatdan:
@nalind Merge this once it passes the tests.

nalind:
Merging.

PR #285, opened 2019-02-12
LVM Driver with SELinux context

haircommander:
In testing @nalind's LVM driver, I found there was an SELinux denial because the LVM mount was unlabeled and thus couldn't be exec'd on. Fix this by adding SELinux context to the mount. This allows Podman to use the LVM driver with just a change to /etc/containers/storage.conf


haircommander:
@rhatdan @nalind PTAL, I finally found how to do this!

haircommander:
@nalind @rhatdan FINALLY got CI tests working locally. I don't love my solution, however. We weren't passing because /dev wasn't a volume mounted on the host in the container that ran the tests, so the lvm commands just kept crashing and being confused. 

Is the addition of the volume mount for /dev in CI problematic in the long run?

haircommander:
(okay giving the containers even more capabilities and I'm fairly certain it's too much but I'm just testing so definitely WIP DO NOT MERGE)

rhatdan:
Could you try with just /dev/mapper volume mounted into the container, perhaps this will fix the aufs problems?

rhatdan:
Since no one has worked on this in a couple of years, I see no reason to keep this open.
Closing.

PR #286, opened 2019-02-12
Make use of overlay.override_kernel_check a warning instead of an error

nalind:
When we removed all traces of override_kernel_check in #269, we created a situation where older configuration files would suddenly start causing us to emit an error at startup.  Soften that to a warning, for now at least.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -239,6 +239,8 @@ func parseOptions(options []string) (*overlayOptions, error) {
 		}
 		key = strings.ToLower(key)
 		switch key {
+		case ".override_kernel_check", "overlay.override_kernel_check", "overlay2.override_kernel_check":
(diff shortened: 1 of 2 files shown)


rhatdan:
I thought I opened this PR earlier.
LGTM

vrothberg:
@nalind @rhatdan sorry I am late to the game but doesn't it affect other drivers as well?

nalind:
@vrothberg Yes, but the main motivator here was configurations like CRI-O's, where even its tests would set the driver to "overlay" and then explicitly set the driver-options that it passed in at the API level to "overlay.override_kernel_check=1".  Presumably, people who were using other drivers before #269 was merged had never had the option specified in their configurations, because it would have always produced an error.

rhatdan:
Right previously if you had ovelay.override_kernel_check=1 it would break devicemapper, but would not break overlay.  An updated version of storage started breaking with this field, now we just ignore it for overlay.

PR #289, opened 2019-02-20
Clarify Locker.Locked(), and switch some locks to read-only

nalind:
Clarify that our `Locker.Locked()` checks if the calling process is holding a write lock, since that's what we care about in the places where we call it.  Change a number of places where `Store` takes a lock but doesn't expect to be modifying on-disk data to have it take a read-only lock instead.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -320,7 +320,7 @@ func (r *layerStore) Load() error {
 			}
 		}
 	}
-	if shouldSave && !r.IsReadWrite() {
+	if shouldSave && (!r.IsReadWrite() || !r.Locked()) {
 		return ErrDuplicateLayerNames
 	}
 	mpath := r.mountspath()
@@ -351,7 +351,7 @@ func (r *layerStore) Load() error {
 	// Last step: if we're writable, try to remove anything that a previous
 	// user of this storage area marked for deletion but didn't manage to
 	// actually delete.
-	if r.IsReadWrite() {
+	if r.IsReadWrite() && r.Locked() {
 		for _, layer := range r.layers {
 			if layer.Flags == nil {
 				layer.Flags = make(map[string]interface{})
--- a/layers.go
+++ b/layers.go
@@ -862,8 +862,9 @@ func (r *layerStore) Delete(id string) error {
 		return ErrLayerUnknown
 	}
 	id = layer.ID
-	// This check is needed for idempotency of delete where the layer could have been
-	// already unmounted (since c/storage gives you that API directly)
+	// The layer may already have been explicitly unmounted, but if not, we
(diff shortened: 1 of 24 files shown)


vrothberg:
LGTM :+1: 
Would you open test-PRs for CRI-O and Buildah to make sure there's no unexpected side effect?

rhatdan:
LGTM

rhatdan:
@nalind Is this ready merge.

layers.go
```diff
@@ -718,6 +774,14 @@ func (r *layerStore) Unmount(id string, force bool) (bool, error) {
...
 		return false, errors.Wrapf(ErrStoreIsReadOnly, "not allowed to update mount locations for layers at %q", r.mountspath())
 	}
+	r.mountsLockfile.Lock()
+	defer r.mountsLockfile.Unlock()
+	if modified, err := r.mountsLockfile.Modified(); modified || err != nil {
+		if err = r.loadMounts(); err != nil {
+			return false, err
+		}
+	}
+	defer r.mountsLockfile.Touch()
```
vrothberg: In addition to the other panic()s in the lockfile, we might add another one to `Touch()` if the lock isn't locked and writable.

nalind: Hmm, I like that idea.

nalind:
I didn't clock it, but I did count the number of write locks we took during a couple of `buildah images` runs, and while https://github.com/containers/buildah/pull/1352 helped for the `--all` case, this reduced things by more than that one did.

The additional lock is new since I first opened the PR, so if those additions also look good, I'm okay with merging it.

nalind:
https://github.com/kubernetes-sigs/cri-o/pull/2089 is currently failing because the new "attempted to acquire one lock type while holding another" check is being tripped when multiple cri-o threads are interacting with the locks.  We don't want to trigger an error in that case, though.

vrothberg:
> [kubernetes-sigs/cri-o#2089](https://github.com/kubernetes-sigs/cri-o/pull/2089) is currently failing because the new "attempted to acquire one lock type while holding another" check is being tripped when multiple cri-o threads are interacting with the locks. We don't want to trigger an error in that case, though.

This makes sense to me. I just had a look at `func (l *lockfile) lock(l_type int16)` and spotted another deadlock when we let this pass. I will try to fix this in a separate PR where you can pick the commit from.

vrothberg:
Here's the fix: https://github.com/containers/storage/pull/290

nalind:
Cherry picked #290.

nalind:
I'm concerned that we may hit problems with writer starvation, but we may have to see.

rhatdan:
Passes tests on CRI-O.  Merging.

PR #290, opened 2019-02-26
lockfile: fix a deadlock

vrothberg:

### DIFF
--- a/lockfile_unix.go
+++ b/lockfile_unix.go
@@ -32,7 +32,7 @@ func getLockFile(path string, ro bool) (Locker, error) {
 	}
 	return &lockfile{
 		stateMutex: &sync.Mutex{},
-		writeMutex: &sync.Mutex{},
+		rwMutex:    &sync.RWMutex{},
 		file:       path,
 		fd:         uintptr(fd),
 		lw:         stringid.GenerateRandomID(),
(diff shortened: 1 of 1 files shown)


vrothberg:
This commit can be cherry-picked by @nalind for https://github.com/containers/storage/pull/289. I just realized the deadlock and figured to fix it while the east coast is asleep :)

rhatdan:
LGTM
@nalind PTAL

nalind:
If I'm reading this right, when two threads try to acquire locks, one for reading, the other for writing, they can both succeed.

vrothberg:
> If I'm reading this right, when two threads try to acquire locks, one for reading, the other for writing, they can both succeed.

Ew, you are right. We've been shielded from that case before. I'll add a commit on top. Thanks, @nalind !

vrothberg:
I was also actually wrong in the assumption that `fcntl` would block on in this case...

vrothberg:
@nalind can you have another look at the latest commit? We're using a `sync.RWMutex` now which makes the code much simpler.

nalind:
Yeah, that's in line with what I was thinking.  The tip of the version of #289 that's currently being chewed on by CI is similar enough that I'll see if it turns up anything before cherry-picking this and rebasing the tip on it.

nalind:
LGTM

vrothberg:
Thanks, @nalind! Closing as it has been cherry-picked in https://github.com/containers/storage/pull/289.

PR #291, opened 2019-02-28
lockfile: add some unit tests, don't return ENOSPC from Modified()

nalind:
These aren't exhaustive, but they're something.

### DIFF
--- a/lockfile_test.go
+++ b/lockfile_test.go
@@ -1,0 +1,553 @@
+package storage
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/containers/storage/pkg/reexec"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Warning: this is not an exhaustive set of tests.
+
+func TestMain(m *testing.M) {
+	if reexec.Init() {
+		return
+	}
+	os.Exit(m.Run())
+}
+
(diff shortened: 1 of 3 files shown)


lockfile_unix.go
```diff
@@ -194,7 +194,7 @@ func (l *lockfile) Modified() (bool, error) {
 		return true, err
 	}
 	if n != len(id) {
-		return true, unix.ENOSPC
```
TomSweeneyRedHat: Was this change in for the tests?  If not, could you add a little blurb in the PR description for hysterical purposes?

nalind: I'd forgotten that we were returning an error for cases where the lock file is newly-created, which on retrospect didn't make a lot of sense, since we don't enforce a size for the last-writer identifier.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM assuming happy tests

lockfile_test.go
```diff
@@ -0,0 +1,553 @@
...
+	var wg sync.WaitGroup
+	var counter, highest int64
+	var highestMutex sync.Mutex
+	for i := 0; i < 100000; i++ {
+		wg.Add(1)
+		go func() {
+			l.RLock()
+			tmp := atomic.AddInt64(&counter, 1)
+			assert.True(t, tmp >= 0, "counter should never be less than zero")
+			highestMutex.Lock()
```
vrothberg: I think we could do a `time.Sleep(500 * time.Milisecond)` after the mutex acquisition and reduce the loop range to 10? It would stress the machines a bit less while remaining "quite confident" we're forcing a race. Same could be done for the writer test, and maybe the others as well.

PR #295, opened 2019-03-04
Move Storage features in pkg/util from libpod to containers/storage

rhatdan:
In an effort to remove cross vendoring, trying to fix buildah from importing from
libpod.  I believe these libraries make more sense in containers/storage then in libpod.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -1,0 +1,237 @@
+package storage
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"os/user"
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	"github.com/BurntSushi/toml"
+	"github.com/containers/storage/pkg/idtools"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// ParseIDMapping takes idmappings and subuid and subgid maps and returns a storage mapping
+func ParseIDMapping(UIDMapSlice, GIDMapSlice []string, subUIDMap, subGIDMap string) (*IDMappingOptions, error) {
+	options := IDMappingOptions{
+		HostUIDMapping: true,
+		HostGIDMapping: true,
+	}
+	if subGIDMap == "" && subUIDMap != "" {
+		subGIDMap = subUIDMap
+	}
+	if subUIDMap == "" && subGIDMap != "" {
+		subUIDMap = subGIDMap
+	}
+	if len(GIDMapSlice) == 0 && len(UIDMapSlice) != 0 {
+		GIDMapSlice = UIDMapSlice
+	}
+	if len(UIDMapSlice) == 0 && len(GIDMapSlice) != 0 {
+		UIDMapSlice = GIDMapSlice
+	}
+	if len(UIDMapSlice) == 0 && subUIDMap == "" && os.Getuid() != 0 {
+		UIDMapSlice = []string{fmt.Sprintf("0:%d:1", os.Getuid())}
+	}
+	if len(GIDMapSlice) == 0 && subGIDMap == "" && os.Getuid() != 0 {
+		GIDMapSlice = []string{fmt.Sprintf("0:%d:1", os.Getgid())}
+	}
+
+	if subUIDMap != "" && subGIDMap != "" {
+		mappings, err := idtools.NewIDMappings(subUIDMap, subGIDMap)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create NewIDMappings for uidmap=%s gidmap=%s", subUIDMap, subGIDMap)
+		}
+		options.UIDMap = mappings.UIDs()
+		options.GIDMap = mappings.GIDs()
+	}
+	parsedUIDMap, err := idtools.ParseIDMap(UIDMapSlice, "UID")
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create ParseUIDMap UID=%s", UIDMapSlice)
+	}
+	parsedGIDMap, err := idtools.ParseIDMap(GIDMapSlice, "GID")
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create ParseGIDMap GID=%s", UIDMapSlice)
+	}
+	options.UIDMap = append(options.UIDMap, parsedUIDMap...)
+	options.GIDMap = append(options.GIDMap, parsedGIDMap...)
+	if len(options.UIDMap) > 0 {
+		options.HostUIDMapping = false
+	}
+	if len(options.GIDMap) > 0 {
+		options.HostGIDMapping = false
+	}
+	return &options, nil
+}
+
+// GetRootlessRuntimeDir returns the runtime directory when running as non root
+func GetRootlessRuntimeDir(rootlessUid int) (string, error) {
+	runtimeDir := os.Getenv("XDG_RUNTIME_DIR")
+	if runtimeDir == "" {
+		tmpDir := fmt.Sprintf("/run/user/%d", rootlessUid)
+		st, err := os.Stat(tmpDir)
+		if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Getuid() && st.Mode().Perm() == 0700 {
+			return tmpDir, nil
+		}
+	}
+	tmpDir := fmt.Sprintf("%s/%d", os.TempDir(), rootlessUid)
+	if err := os.MkdirAll(tmpDir, 0700); err != nil {
+		logrus.Errorf("failed to create %s: %v", tmpDir, err)
+	} else {
+		st, err := os.Stat(tmpDir)
+		if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Getuid() && st.Mode().Perm() == 0700 {
+			return tmpDir, nil
+		}
+	}
+	home, err := homeDir()
+	if err != nil {
+		return "", errors.Wrapf(err, "neither XDG_RUNTIME_DIR nor HOME was set non-empty")
+	}
+	resolvedHome, err := filepath.EvalSymlinks(home)
+	if err != nil {
+		return "", errors.Wrapf(err, "cannot resolve %s", home)
+	}
+	return filepath.Join(resolvedHome, "rundir"), nil
+}
+
+// getRootlessDirInfo returns the parent path of where the storage for containers and
+// volumes will be in rootless mode
+func getRootlessDirInfo(rootlessUid int) (string, string, error) {
+	rootlessRuntime, err := GetRootlessRuntimeDir(rootlessUid)
+	if err != nil {
+		return "", "", err
+	}
+
+	dataDir := os.Getenv("XDG_DATA_HOME")
+	if dataDir == "" {
+		home, err := homeDir()
+		if err != nil {
+			return "", "", errors.Wrapf(err, "neither XDG_DATA_HOME nor HOME was set non-empty")
+		}
+		// runc doesn't like symlinks in the rootfs path, and at least
+		// on CoreOS /home is a symlink to /var/home, so resolve any symlink.
+		resolvedHome, err := filepath.EvalSymlinks(home)
+		if err != nil {
+			return "", "", errors.Wrapf(err, "cannot resolve %s", home)
+		}
+		dataDir = filepath.Join(resolvedHome, ".local", "share")
+	}
+	return dataDir, rootlessRuntime, nil
+}
+
+// getRootlessStorageOpts returns the storage opts for containers running as non root
+func getRootlessStorageOpts(rootlessUid int) (StoreOptions, error) {
+	var opts StoreOptions
+
+	dataDir, rootlessRuntime, err := getRootlessDirInfo(rootlessUid)
+	if err != nil {
+		return opts, err
+	}
+	opts.RunRoot = rootlessRuntime
+	opts.GraphRoot = filepath.Join(dataDir, "containers", "storage")
+	if path, err := exec.LookPath("fuse-overlayfs"); err == nil {
+		opts.GraphDriverName = "overlay"
+		opts.GraphDriverOptions = []string{fmt.Sprintf("overlay.mount_program=%s", path)}
+	} else {
+		opts.GraphDriverName = "vfs"
+	}
+	return opts, nil
+}
+
+type tomlOptionsConfig struct {
+	MountProgram string `toml:"mount_program"`
+}
+
+func getTomlStorage(storeOptions *StoreOptions) *tomlConfig {
+	config := new(tomlConfig)
+
+	config.Storage.Driver = storeOptions.GraphDriverName
+	config.Storage.RunRoot = storeOptions.RunRoot
+	config.Storage.GraphRoot = storeOptions.GraphRoot
+	for _, i := range storeOptions.GraphDriverOptions {
+		s := strings.Split(i, "=")
+		if s[0] == "overlay.mount_program" {
+			config.Storage.Options.MountProgram = s[1]
+		}
+	}
+
+	return config
+}
+
+// DefaultStoreOptions returns the default storage ops for containers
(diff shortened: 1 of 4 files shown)


rhatdan:
@giuseppe @nalind PTAL

rhatdan:
@vrothberg PTAL

pkg/rootless/rootless_linux.c
```diff
@@ -0,0 +1,319 @@
...
+      _exit (EXIT_FAILURE);
+    }
+
+  pid = fork ();
+  if (pid < 0)
+    fprintf (stderr, "cannot fork: %s\n", strerror (errno));
+  if (pid)
+    return pid;
+
+  setenv ("_LIBPOD_USERNS_CONFIGURED", "init", 1);
```
vrothberg: Can we rename the identifiers to _CONTAINERS_ ?

mheon:
Why are we moving `pkg/util` over? That's general-purpose utility functions in Podman

nalind:
The `util` package also depends on `github.com/containers/image`, which depends on this library.

rhatdan:
@mheon Util package currently is used in buildah, libpod and eventually skopeo.

rhatdan:
Util package also seems mainly concerned with setting up default containers storage.

rhatdan:
@nalind Can we start moving parts of containers image over here if it is needed by both packages.

I would like to get to the point where we have a stack of tools that looks like this. But the vendoring only goes in one direction.

cri-o
libpod
buildah
image
storage

mheon:
@rhatdan Not really? There's some storage in there, but that's where I tell people to stick general utility functions, and a fair bit of it will be libpod-specific

nalind:
> @nalind Can we start moving parts of containers image over here if it is needed by both packages.

The only dependency in that direction so far is a call to `manifest.Digest`, which we could probably change to a callback or a hook or something.

rhatdan:
@mheon I will remove the parts of utils.go that have nothing to do with storage.

mheon:
@rhatdan Ack, SGTM.

I mainly want to keep the package so I can keep telling @baude to put stuff there when nobody knows where to put it

wking:
> In an effort to remove cross vendoring, trying to fix buildah from importing from libpod.

I don't have a great handle on how libraries are distributed among these projects, so take this with a grain of salt, but I don't see a connection between the rootless stuff and storage, except for holding a package-level `skipStorageSetup` boolean.  Maybe that package should instead live in buildah and libpod could pull it from there (since libpod already vendors buildah to support `podman build`)?

giuseppe:
I agree with @wking.  It might help with vendoring and avoid circular dependencies, but containers/storage doesn't seem the best place for `pkg/rootless`, it is very libpod centric and the libpod repository still seem like the best place for it.

We are only using a subset of the package from Buildah, we could probably avoid the circular dependency embedding the few changes inside Buildah itself and not requiring pkg/rootless.

rhatdan:
I am moving pkg/rootless to buildah and just putting the StorageDefaults handling in this PR.
@nalind Requested that we pass in the rootless fields, into the storage calls to help it differentiate between the system storage versus rootless storage.

rhatdan:
Opened https://github.com/containers/buildah/pull/1400 to test this patch out and mv pkg/rootless to buildah.

rhatdan:
@nalind PTAL

pkg/util/utils.go
```diff
@@ -0,0 +1,233 @@
...
+		options.HostUIDMapping = false
+	}
+	if len(options.GIDMap) > 0 {
+		options.HostGIDMapping = false
+	}
+	return &options, nil
+}
+
+// GetRootlessRuntimeDir returns the runtime directory when running as non root
+func GetRootlessRuntimeDir(rootlessUid int) (string, error) {
```
nalind: Does this need to be exported?

pkg/util/utils.go
```diff
@@ -0,0 +1,233 @@
...
+	if runtimeDir == "" {
+		tmpDir := fmt.Sprintf("/run/user/%d", rootlessUid)
+		os.MkdirAll(tmpDir, 0700)
+		st, err := os.Stat(tmpDir)
+		if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Getuid() && st.Mode().Perm() == 0700 {
+			runtimeDir = tmpDir
+		}
+	}
+	if runtimeDir == "" {
+		tmpDir := fmt.Sprintf("%s/user/%d", os.TempDir(), rootlessUid)
```
nalind: Why the "user" intermediate directory under what will likely be "/tmp"?

pkg/util/utils.go
```diff
@@ -0,0 +1,233 @@
...
+			return "", errors.Wrapf(err, "cannot resolve %s", home)
+		}
+		runtimeDir = filepath.Join(resolvedHome, "rundir")
+	}
+	return runtimeDir, nil
+}
+
+// GetRootlessDirInfo returns the parent path of where the storage for containers and
+// volumes will be in rootless mode
+func GetRootlessDirInfo(rootlessUid int) (string, string, error) {
```
nalind: Does this need to be exported?

pkg/util/utils.go
```diff
@@ -0,0 +1,233 @@
...
+		if err != nil {
+			return "", "", errors.Wrapf(err, "cannot resolve %s", home)
+		}
+		dataDir = filepath.Join(resolvedHome, ".local", "share")
+	}
+	return dataDir, rootlessRuntime, nil
+}
+
+// GetRootlessStorageOpts returns the storage opts for containers running as non root
+func GetRootlessStorageOpts(rootlessUid int) (storage.StoreOptions, error) {
```
nalind: Does this need to be exported?

pkg/util/utils.go
```diff
@@ -0,0 +1,233 @@
...
+		if s[0] == "overlay.mount_program" {
+			config.Storage.Options.MountProgram = s[1]
+		}
+	}
+
+	return config
+}
+
+// GetDefaultStoreOptions returns the default storage ops for containers
+func GetDefaultStoreOptions(rootless bool, rootlessUid int) (storage.StoreOptions, error) {
```
nalind: What do we gain by hiding this in a subpackage?  How should a caller know if it should call this function, `GetRootlessStorageOpts()`, or simply read from `storage.DefaultStoreOptions`?

pkg/util/utils.go
```diff
@@ -0,0 +1,233 @@
...
+			if storageOpts.GraphRoot == "" {
+				storageOpts.GraphRoot = defaultRootlessGraphRoot
+			}
+		}
+	}
+	return storageOpts, nil
+}
+
+// StorageConfigFile returns the path to the storage config file used
+func StorageConfigFile(rootless bool) string {
```
nalind: How should a caller know if this is needed instead of trusting `storage.DefaultConfigFile`?

pkg/util/utils.go
```diff
@@ -0,0 +1,233 @@
...
+	if rootless {
+		return filepath.Join(os.Getenv("HOME"), ".config/containers/storage.conf")
```
nalind: `GetRootlessDirInfo()` returns an error if $HOME is not set.  Should this function be checking for that?

pkg/util/utils.go
rhatdan: It is used all over libpod.

```
/home/dwalsh/libpod/libpod/runtime.go:	rootlessRuntimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/runtime.go:		val, err = util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/runtime.go:		runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/runtime.go:		runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
/home/dwalsh/libpod/libpod/oci.go:	runtimeDir, err := util.GetRootlessRuntimeDir()
```

pkg/util/utils.go
rhatdan: Because it was there in the libpod version.  Removed.

pkg/util/utils.go
rhatdan: Nope, Fixed

pkg/util/utils.go
rhatdan: Nope Fixed.

pkg/util/utils.go
rhatdan: Documentation?

pkg/util/utils.go
rhatdan: I could make DefaultConfigFile not be exposed.

pkg/util/utils.go
rhatdan: Fixed

pkg/util/utils.go
rhatdan: Rewritten again to make these functions primary in storage and to remove links to the constants.
Now storage.DefaultStoreOptions is function.

pkg/util/utils.go
rhatdan: Change DefaultConfigFile-> defaultConfigFile.
Now you need to call StorageFile()

giuseppe:
LGTM

utils.go
```diff
@@ -0,0 +1,231 @@
...
+		UIDMapSlice = []string{fmt.Sprintf("0:%d:1", os.Getuid())}
+	}
+	if len(GIDMapSlice) == 0 && subGIDMap == "" && os.Getuid() != 0 {
+		GIDMapSlice = []string{fmt.Sprintf("0:%d:1", os.Getgid())}
+	}
+
+	if subUIDMap != "" && subGIDMap != "" {
+		mappings, err := idtools.NewIDMappings(subUIDMap, subGIDMap)
+		if err != nil {
+			return nil, err
```
nalind: Wrap this with `errors.Wrapf`.

utils.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+		options.UIDMap = mappings.UIDs()
+		options.GIDMap = mappings.GIDs()
+	}
+	parsedUIDMap, err := idtools.ParseIDMap(UIDMapSlice, "UID")
+	if err != nil {
+		return nil, err
```
nalind: Wrap this with `errors.Wrapf`.

utils.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	parsedGIDMap, err := idtools.ParseIDMap(GIDMapSlice, "GID")
+	if err != nil {
+		return nil, err
```
nalind: Wrap this with `errors.Wrapf`.

utils.go
```diff
@@ -0,0 +1,231 @@
...
+	if _, err = os.Stat(storageConf); err == nil {
+		defaultRootlessRunRoot = storageOpts.RunRoot
+		defaultRootlessGraphRoot = storageOpts.GraphRoot
+		storageOpts = StoreOptions{}
+		ReloadConfigurationFile(storageConf, &storageOpts)
+	}
+
+	if rootless {
+		if os.IsNotExist(err) {
+			os.MkdirAll(filepath.Dir(storageConf), 0755)
```
nalind: Check for an error here.

utils.go
```diff
@@ -0,0 +1,231 @@
...
+			// If the file did not specify a graphroot or runroot,
+			// set sane defaults so we don't try and use root-owned
+			// directories
+			if storageOpts.RunRoot == "" {
+				storageOpts.RunRoot = defaultRootlessRunRoot
+			}
+			if storageOpts.GraphRoot == "" {
+				storageOpts.GraphRoot = defaultRootlessGraphRoot
+			}
+		}
```
nalind: We should do something about non-nil errors that aren't `IsNotExist` errors.

nalind:
Some nits around error handling, but otherwise LGTM.

utils.go
rhatdan: Fixed

utils.go
rhatdan: Fixed

utils.go
rhatdan: Fixed

utils.go
rhatdan: Fixed

utils.go
rhatdan: Added handling for this.

rhatdan:
@giuseppe PTAL I rewrote some of the GetRootlessRuntimeDir functions.

utils.go
```diff
@@ -0,0 +1,237 @@
...
+func GetRootlessRuntimeDir(rootlessUid int) (string, error) {
+	runtimeDir := os.Getenv("XDG_RUNTIME_DIR")
+	if runtimeDir == "" {
+		tmpDir := fmt.Sprintf("/run/user/%d", rootlessUid)
+		st, err := os.Stat(tmpDir)
+		if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Getuid() && st.Mode().Perm() == 0700 {
+			return tmpDir, nil
+		}
+	}
+	tmpDir := fmt.Sprintf("%s/%d", os.TempDir(), rootlessUid)
```
giuseppe: filepath.Join?

rhatdan: That fails to build because rootlessUid is an `int`.

giuseppe:
LGTM

utils.go
```diff
@@ -0,0 +1,237 @@
...
+		mappings, err := idtools.NewIDMappings(subUIDMap, subGIDMap)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create NewIDMappings for uidmap=%s gidmap=%s", subUIDMap, subGIDMap)
+		}
+		options.UIDMap = mappings.UIDs()
+		options.GIDMap = mappings.GIDs()
+	}
+	parsedUIDMap, err := idtools.ParseIDMap(UIDMapSlice, "UID")
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create ParseIDMap UID=%s", UIDMapSlice)
```
TomSweeneyRedHat: to differntiate, could you s/ParseIDMap/ParseUidMap/  here and with ParseGidMap below at line 57?

TomSweeneyRedHat:
LGTM other than two small nits.

Test errors are odd.  All the go 1.11 tests pass, the others fail due to not being able to find md2man?

```
Building/Running tests
go get -u -tags " libdm_no_deferred_remove ostree "  github.com/cpuguy83/go-md2man
package github.com/cpuguy83/go-md2man/v2/md2man: cannot find package "github.com/cpuguy83/go-md2man/v2/md2man" in any of:
	/root/.gimme/versions/go1.10.linux.amd64/src/github.com/cpuguy83/go-md2man/v2/md2man (from $GOROOT)
	/root/go/src/github.com/cpuguy83/go-md2man/v2/md2man (from $GOPATH)
Makefile:85: recipe for target 'install.tools' failed
make: *** [install.tools] Error 1
+sudo rm -rf /tmp/storage_iCIjsD
```

utils.go
rhatdan: Fixed.

PR #296, opened 2019-03-04
Check if layer is actually mounted if mount count > 0

rhatdan:
We have an issue with zfs on reboots getting this data wrong.

You can cause this situation also if someone goes out an umounts
a mount point manually.


rhatdan:
@nalind PTAL

rhatdan:
@nalind Is rejecting this.

PR #299, opened 2019-03-05
Drop our dependency on github.com/containers/image

nalind:
Drop our dependency on the image library's manifest package by requiring that callers pass its Digest() function to us as a callback.  This makes our CLI test/diagnostic tool calculate digests of s1 manifests incorrectly, but that's not something that we were testing.


nalind:
This is going to require `github.com/containers/image/storage` to start passing `manifest.Digest` to us in several places, and CRI-O to do the same in one place.

rhatdan:
LGTM.
Need @vrothberg and @mtrmac To approve.

vrothberg:
LGTM

vrothberg:
(FYI @siretart)

store.go
```diff
@@ -105,9 +111,11 @@ type ROBigDataStore interface {
...
 // various types of file-based lookaside stores that we implement.
 type RWBigDataStore interface {
-	// SetBigData stores a (potentially large) piece of data associated with this
-	// ID.
-	SetBigData(id, key string, data []byte) error
+	// SetBigData stores a (potentially large) piece of data associated
+	// with this ID.  If this is an image store, pass
+	// github.com/containers/image/manifest.Digest as digestManifest to
+	// allow ByDigest to find images by their correct digests.
+	SetBigData(id, key string, data []byte, digestManifest func([]byte) (digest.Digest, error)) error
```
mtrmac: (Is there anything that benefits from the shared `{,RW,RO}BigDataStore` interfaces, so that it is worth it to force container store users to supply an unnecessary `digestManifest`? There may well be a good reason for this, I have no idea.

If nothing else, the `nil` parameters added for container stores below, even if correct and ultimately safe, are a bit ugly.)

images.go
```diff
@@ -595,20 +594,22 @@ func (r *imageStore) BigDataSize(id, key string) (int64, error) {
 		return size, nil
 	}
 	if data, err := r.BigData(id, key); err == nil && data != nil {
-		if r.SetBigData(id, key, data) == nil {
+		if err = r.SetBigData(id, key, data, digestManifest); err == nil {
```
mtrmac: Forcing the caller to supply `digestManifest` here, because of an ultimately very internal implementation decision, is not great.

Couldn’t this just do
```go
image.BigDataSizes[key] = int64(len(data))
r.Save()
```
or something like that (possibly with a helper shared with `SetBigData`?

store.go
```diff
@@ -1604,7 +1618,7 @@ func (s *store) ImageSize(id string) (int64, error) {
 		return -1, errors.Wrapf(err, "error reading list of big data items for image %q", id)
 	}
 	for _, name := range names {
-		n, err := imageStore.BigDataSize(id, name)
+		n, err := imageStore.BigDataSize(id, name, nil)
```
mtrmac: This `nil` is not obviously safe AFAICS.  See the `imageStore.BigDataSize` comment first, though.

store.go
nalind: Hmm, at this point I don't think so.  I know at one point we wanted to refactor interfaces so that Store wasn't just one large interface of its own, but that hasn't really bought us much since then.

The logic for quietly backfilling in a value when there isn't one recorded is there for upgrade cases where the previously-used version didn't track item digests, but that was something we added in #109, so it could also be argued that we don't need to worry about those cases any more.

Dropping both of those would let us excise the need for the callback everywhere except for in `SetImageBigData()`, which might be worth it.

mtrmac: (I’m not at all saying that the backfilling is unnecessary; I have no relevant knowledge and no opinion either way. I’ve only really been commenting on the internal API spreading the impact further than it seems necessary, only based on my limited view of how c/storage is used.)

nalind: Okay, dropped the backfill logic.  Kept the interfaces for grouping, but made the interfaces implemented by image and container stores different.

rhatdan:
LGTM

PR #300, opened 2019-03-07
reexec: add CommandContext()

nalind:
Add `CommandContext()`, which is like `Command()`, but which also takes a context, very much in keeping with `os/exec.CommandContext()`.


TomSweeneyRedHat:
LGTM

rhatdan:
LGTM

PR #303, opened 2019-03-11
storage: make [l]chown errors clearer

giuseppe:
if os.[Lc,C]hown are failing with EINVAL, it might be related to an
UID/GID not mapped in the user namespace we are currently using.

It could be possible to detect this issue by inspecting
/proc/self/uid_map or /proc/self/gid_map, but that won't be possible
when we are pulling a new image and extracting it from a chroot where
/proc is not mounted.

### DIFF
--- a/pkg/idtools/idtools.go
+++ b/pkg/idtools/idtools.go
@@ -7,6 +7,9 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"syscall"
+
+	"github.com/pkg/errors"
 )
 
 // IDMap contains a single entry for user namespace range remapping. An array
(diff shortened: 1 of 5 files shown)


AkihiroSuda:
Any chance to use https://github.com/rootless-containers/proto , so that OCI runtime can emulate file permission using ptrace/seccomp ?

giuseppe:
that would work only with umoci and the rootless-containers fork of PRoot.  Is there any other runtime honoring these xattrs?

I don't think at the moment we want to support syscalls emulation either in Podman and Buildah.

For containers/storage it is probably fine if we support it through an explicit configuration setting, if there is anyone using it.  I am only opposed to silently doing it as a fallback when chown fails.

pkg/idtools/idtools.go
```diff
@@ -277,3 +280,18 @@ func parseSubidFile(path, username string) (ranges, error) {
 	}
 	return rangeList, nil
 }
+
+func checkChownErr(err error, name string, uid, gid int) error {
+	if e, ok := err.(*os.PathError); ok && e.Err == syscall.EINVAL {
+		return errors.Wrapf(err, "there might not be enough IDs available in the namespace (requested %d:%d for %s)", uid, gid, name)
```
rhatdan: How about 
"%d:%d dones have have a valid mapping in the user namespace for %s"

giuseppe: we are not sure that is the problem without checking `/proc/self/uid_map`, that is why I've left it as a possibility in the message.  Even though I don't see other reasons why it could fail with that error code.

TomSweeneyRedHat: I don't know if you want to add something about "verify the values in /proc/self/uid_map" or some such to that.

giuseppe: there is need to be in the correct namespace to verify `/proc/self/uid_map`

rhatdan: I just want something more specific for the user. then "Their might not be" ...

rhatdan:
LGTM

PR #304, opened 2019-03-14
move lockfile API into pkg/lockfile

vrothberg:
We need the lockfile API in containers/image.  Moving the implementation
into a dedicated package seems more idiomatic for external users.


vrothberg:
I'll close it for now. Have to rethink the copy-ops detection.

PR #306, opened 2019-03-19
Fix cross compilation code for other platforms

rhatdan:
When trying to vendor into containers/image we found issues
with cross platform compilation.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -7,10 +7,10 @@ import (
 	"os/user"
 	"path/filepath"
 	"strings"
-	"syscall"
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/system"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


rhatdan:
@vrothberg @giuseppe @umohnani8 @nalind @TomSweeneyRedHat PTAL

utils.go
```diff
@@ -72,17 +72,17 @@ func GetRootlessRuntimeDir(rootlessUid int) (string, error) {
...
 		}
 	}
 	tmpDir := fmt.Sprintf("%s/%d", os.TempDir(), rootlessUid)
 	if err := os.MkdirAll(tmpDir, 0700); err != nil {
 		logrus.Errorf("failed to create %s: %v", tmpDir, err)
 	} else {
-		st, err := os.Stat(tmpDir)
-		if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Getuid() && st.Mode().Perm() == 0700 {
+		st, err := system.Stat(tmpDir)
+		if err == nil && int(st.UID()) == os.Getuid() && st.Mode() == 0700 {
```
nalind: This should compile, which is good, but `st.UID()` is hardwired to always return 0, while `os.Getuid()` is documented to always return -1, so our attempt to create this directory ends up either failing and producing an error, or succeeding and then being ignored.

nalind: My preference is to avoid creating a directory that we're just going to ignore.

rhatdan: Why would stating of the directory return UID==0?

nalind: `pkg/system`'s `Stat()` returns a pointer to a `StatT`, which on Windows systems only ever returns 0 from its `UID()` method.

rhatdan: Do you want to make this OS Specific then?

vrothberg:
Do you know if containers/storage has consumers on Windows? If it's for historical reasons, we may have the chance to ditch the windows builds.

utils.go
giuseppe: we end up using this code only when `os.Geteuid() != 0`, on Windows it returns always `-1`.  Could we just check `os.Geteuid()` value and treat `-1` differently?

nalind:
I don't expect `$XDG_RUNTIME_DIR` to be set on Windows, and at any rate we don't create a directory right here if it is set, so LGTM.

PR #307, opened 2019-03-20
store: add digest locking

vrothberg:
Add a mechanism to create digest-specific copy lock files that are being used to synchronize processes when copying individual layers. It is ultimately aimed at avoiding concurrently downloading the same layer.

At the moment, the lock files will remain in $driver-locks/* and won't be deleted. We could combine unlocking with a removal of the lock file.

Note: this PR is part of a larger set of changes to other parts of github.com/containers/*.

FYI: @rhatdan @nalind


rhatdan:
LGTM

store.go
```diff
@@ -701,6 +713,28 @@ func (s *store) load() error {
...
+	return nil
+}
+
+func (s *store) UnlockDigest(d digest.Digest) error {
+	lockPath := filepath.Join(s.lockRoot, d.Encoded())
+	lockfile, err := GetLockfile(lockPath)
+	if err != nil {
+		return err
+	}
+	// this might be called multiple times
```
mtrmac: ?

The conventional wisdom is that trying to unlock a not-locked lock is indicative of incorrect state tracking in the caller, and risks to unlock a lock owned by _some other process_.

Looking at https://github.com/containers/image/pull/607, this is intentional (and I haven’t read that PR at all yet) — but eventually the reason for this design, and the conditions when it is/isn’t save to use the API that way, should be documented within this project as well.

vrothberg: > The conventional wisdom is that trying to unlock a not-locked lock is indicative of incorrect state tracking in the caller, and risks to unlock a lock owned by some other process.

I think we can avoid that with some extra bookkeeping in the containers/image.

Thanks for the comment!

vrothberg: I removed that check, so we'll properly fail.

mtrmac: This was intended as a request for documentation at this point, but removing this behavior works too :)

store.go
```diff
@@ -701,6 +713,25 @@ func (s *store) load() error {
...
+	if err != nil {
+		return err
+	}
+	lockfile.Lock()
+	return nil
+}
+
+func (s *store) UnlockDigest(d digest.Digest) error {
+	lockPath := filepath.Join(s.lockRoot, d.Encoded())
+	lockfile, err := GetLockfile(lockPath)
```
mtrmac: Nalin pointed out in 
https://github.com/containers/image/pull/607#discussion_r267525606 that every call to `GetLockfile` allocates a FD that is never closed.

mtrmac: … allocating only one FD per path, sorry.

vrothberg: This is now being addressed in the latest commit:
```
Add a method to the store to generate a lock file for a specific digest.
Such a digest-specific lock file is needed to synchronize threads and
processes when copying blobs from a registry to the containers-storage.

Whenever a layer is about to get copied, the lock must be acquired which
indicates to other processes and threads that the layer/blob is already
being copied.

To avoid leaking file descriptors, add a new Close() method to the
Locker.  We are now counting references on locks, which are being
decremented in Close().  When the last reference called Close(),
the lock file's FD is being closed as well.  The reference counter is
incremented for each new invocation on Get{RO,}Lockfile().
```

vrothberg: In short: we're now reference counting the lock files and have a `Locker.Close()` method which will close the file descriptor when the reference counter is 0.

The store now just returns a locker, while the storage image destination must take care of book keeping of the Locker objects.

vrothberg: Actually, it may just be much cleaner to open and close the file on the fly. We're alreay refcounting the with each Lock() and Unlock(). Github turns out to be a good rubbderduck.

vrothberg:
We are now opening and closing the file on-demand. The lock-internal ref counter allows us to know if we're the first Lock() and the last Unlock(). I think that's the most transparent way to avoid leaking file descriptors.

Open task: create the lock directory in the tmp storage.

vrothberg:
@rhatdan @nalind @mtrmac PTAL

lockfile_unix.go
```diff
@@ -14,17 +14,13 @@ import (
...
 
 func getLockFile(path string, ro bool) (Locker, error) {
-	var fd int
-	var err error
-	if ro {
-		fd, err = unix.Open(path, os.O_RDONLY, 0)
-	} else {
-		fd, err = unix.Open(path, os.O_RDWR|os.O_CREATE, unix.S_IRUSR|unix.S_IWUSR)
-	}
+	// Create and open the lock file to catch potetnial permission errors
```
nalind: Typo: potetnial

vrothberg: Thanks! Fixed.

lockfile_unix.go
```diff
@@ -14,17 +14,13 @@ import (
...
 
 func getLockFile(path string, ro bool) (Locker, error) {
-	var fd int
-	var err error
-	if ro {
-		fd, err = unix.Open(path, os.O_RDONLY, 0)
-	} else {
-		fd, err = unix.Open(path, os.O_RDWR|os.O_CREATE, unix.S_IRUSR|unix.S_IWUSR)
-	}
+	// Create and open the lock file to catch potential permission errors
```
rhatdan: Nothing changes as far as handling of R/O Locks?

vrothberg: Nothing changes. We already have an internal counter that is incremented for each (successful) lock and decremented for each unlock. In this change, we make use of that by opening and closing the file when the counter == 0 during lock and unlock.

rhatdan:
LGTM

store.go
```diff
@@ -529,6 +532,7 @@ type store struct {
 	imageStore      ImageStore
 	roImageStores   []ROImageStore
 	containerStore  ContainerStore
+	lockRoot        string
```
mtrmac: (Maybe `digestLockRoot`.)

store.go
```diff
@@ -698,9 +702,20 @@ func (s *store) load() error {
...
+	if err := os.MkdirAll(s.lockRoot, 0700); err != nil {
+		return err
+	}
+
 	return nil
 }
 
+// GetDigestLock returns a digest-specific Locker.
+func (s *store) GetDigestLock(d digest.Digest) (Locker, error) {
+	return GetLockfile(filepath.Join(s.lockRoot, d.Encoded()))
```
mtrmac: Is this intentionally dropping the algorithm? (In principle, this may cause a few temporary collisions, which does not really hurt. OTOH adding the algorithm and having a bit longer file names does not hurt either.)

lockfile_unix.go
```diff
@@ -34,7 +30,7 @@ func getLockFile(path string, ro bool) (Locker, error) {
 		stateMutex: &sync.Mutex{},
 		rwMutex:    &sync.RWMutex{},
 		file:       path,
-		fd:         uintptr(fd),
+		fd:         uintptr(0),
```
mtrmac: `0` is a valid file descriptor value, use `-1` please, if possible. Or, at the very least, leave the field uninitialized = default-initialized; it will still be zero, but at least it will not look like a value anything should rely on.

lockfile_unix.go
```diff
@@ -75,6 +71,14 @@ func (l *lockfile) lock(l_type int16) {
 	}
 	l.stateMutex.Lock()
 	if l.counter == 0 {
+		// If we're the first reference on the lock, we need to open the file again.
```
mtrmac: (BTW, above and below, setting `Pid` does not make sense. This is a return-only field.)

lockfile_unix.go
```diff
@@ -14,17 +14,13 @@ import (
...
-		fd, err = unix.Open(path, os.O_RDWR|os.O_CREATE, unix.S_IRUSR|unix.S_IWUSR)
-	}
+	// Create and open the lock file to catch potential permission errors
+	// etc. early.
+	fd, err := unix.Open(path, os.O_RDWR|os.O_CREATE, unix.S_IRUSR|unix.S_IWUSR)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error opening %q", path)
 	}
-	unix.CloseOnExec(fd)
+	unix.Close(fd)
```
mtrmac: This does work AFAICS, but please clearly document somewhere that `getLockFile` is only ever called once for a single value of `path` due to how `getLockfile` works; otherwise this `Close` would remove all locks held by this process.

(And we really must not have two different methods called `getLockfile` and `getLockFile`. That’s awful. Maybe in a separate PR, but soon.)

lockfile_unix.go
```diff
@@ -14,17 +14,13 @@ import (
...
-	var fd int
-	var err error
-	if ro {
-		fd, err = unix.Open(path, os.O_RDONLY, 0)
-	} else {
-		fd, err = unix.Open(path, os.O_RDWR|os.O_CREATE, unix.S_IRUSR|unix.S_IWUSR)
-	}
+	// Create and open the lock file to catch potential permission errors
+	// etc. early.
+	fd, err := unix.Open(path, os.O_RDWR|os.O_CREATE, unix.S_IRUSR|unix.S_IWUSR)
```
mtrmac: (This now requires write privileges for read-only locks. Is that OK? I don’t know enough about `AdditionalImageStores` to tell.)

store.go
vrothberg: Done.

store.go
vrothberg: I did not consider the potential of collisions, great catch!

lockfile_unix.go
vrothberg: Removed the line. Thanks!

lockfile_unix.go
vrothberg: This sounds plausible but I seem to be unable to find information about that without having to read kernel code.

Let's defer that to a future PR.

lockfile_unix.go
vrothberg: > This does work AFAICS, but please clearly document somewhere that getLockFile is only ever called once for a single value of path due to how getLockfile works; otherwise this Close would remove all locks held by this process.

That's an issue that predates this PR but I can add a comment to `getLockFile`. Note that it's not an issue for the API `GetLockfile()` as the Lockers are held in map.

lockfile_unix.go
vrothberg: That is okay, as we are only working on the FDs.

vrothberg:
@nalind @mtrmac @giuseppe  PTAL

lockfile_unix.go
mtrmac: Really? All of fcntl(2), fcntl(3p), (info libc) say this about `l_pid`.

lockfile_unix.go
mtrmac: I meant: can the read-only accessor be assumed to have the privileges to necessary to open the files read-write?

The privilege check happens at `open`, based on the provided flags; whether or not the returned FD is actually used for such operations later is secondary.

mtrmac:
ACK, though I need to catch up on the code using this.

lockfile_unix.go
mtrmac: Looking at https://github.com/containers/storage/commit/162e676330d4eceddb708e3226e958898d968f43 , the read-only open may be actually necessary.

lockfile_unix.go
mtrmac: > (And we really must not have two different methods called `getLockfile` and `getLockFile`. That’s awful. Maybe in a separate PR, but soon.)

https://github.com/containers/storage/pull/310 .  But the more of the code I read, the less I understand it…

giuseppe:
LGTM

lockfile_unix.go
vrothberg: I must have been temporarily blind...it's all over the place.

TomSweeneyRedHat:
LGTM, but would like a head nod from @mtrmac

vrothberg:
Thanks @TomSweeneyRedHat !

vrothberg:
Please wait with merging: https://github.com/containers/buildah/pull/1445 did not pass yet.

vrothberg:
Yesterday, it looked like we had some deadlock there but I have a hard time reproducing as a) Travis takes up to 4 hours to fire off the jobs, and b) Buildah's CI is heavily tied to Travis making it very hard to run the integration tests locally.

lockfile_unix.go
mtrmac: The way I read that, the store may be created with read-write access (which incidentally creates the lock files), and then somehow made available read-only (is that copied? or NFS-exported, with the possibility of concurrent writers which do still have read-write access?).  The read-only users can use the previously-created lock files to allocate locks [BTW, apparently even an exclusive lock does not require write access to the file descriptor?!].

mtrmac:
> LGTM, but would like a head nod from @mtrmac

Right now this seems to break read-only stores, but I know very little about c/storage and I’m somewhat confused by the locking scheme as well. @rhatdan can hopefully review the read-only store impact.

vrothberg:
> Right now this seems to break read-only stores, but I know very little about c/storage and I’m somewhat confused by the locking scheme as well. @rhatdan can hopefully review the read-only store impact.

@mtrmac, the previous behavior has been restored. I refactored it into a dedicated function: https://github.com/containers/storage/pull/307/files#diff-335585833cd0ec27ad7af7688f7e996eR34

mtrmac:
> @mtrmac, the previous behavior has been restored.

Thanks, ACK.

IIRC we did discuss this, but I can’t find this explicitly mentioned in this conversation: This PR will make it unsafe to use more than one (e.g. vendored) copy of c/storage in a single Go process, because all the `Close`s will unlock all locks held by the same process, i.e. the different copies will be unlocking each others’ locks.

vrothberg:
@mtrmac @rhatdan @nalind, I think this is good to merge now. All remaining bits are done in c/image and the users are cleaned up now so that we can propagate the changes.

Please drop a last LGTM.

PR #308, opened 2019-03-26
Add build tag to exclude vfs driver

sseago:
This commit adds a build flag exclude_graphdriver_vfs to allow for
excluding the vfs driver in the same way that this is currently
supported for devicemapper, aufs, overlay, btrfs, and zfs.

This was the one driver lacking this support which was preventing
me from using containers/image (and deps) in a build with CGO_ENABLED=0

Signed-off-by: Scott Seago <sseago@redhat.com>

### DIFF
--- a/drivers/register/register_vfs.go
+++ b/drivers/register/register_vfs.go
@@ -1,3 +1,5 @@
+// +build !exclude_graphdriver_vfs
+
 package register
 
 import (


rhatdan:
LGTM
@nalind @vrothberg @mtrmac PTAL

vrothberg:
LGTM (but want a head nod from @nalind)

nalind:
I'd much prefer being able to guarantee and depend on `vfs` always being available.  Taking a stab at it in #309.  Does the patch there fix your build?

sseago:
@nalind Looks like the patch in #309 works just as well for me. I'll close this PR and comment there.

PR #309, opened 2019-03-26
drivers/copy: add a non-cgo fallback

nalind:
The vfs driver already has logic that avoids use of the copy package when we're not on Linux, so provide the same fallback in the copy package for non-cgo cases.

### DIFF
--- a/drivers/copy/copy_linux.go
+++ b/drivers/copy/copy_linux.go
@@ -1,4 +1,4 @@
-// +build linux
+// +build cgo
 
 package copy
 
--- a/drivers/copy/copy_unsupported.go
+++ b/drivers/copy/copy_unsupported.go
@@ -1,0 +1,19 @@
+// +build !linux !cgo
+
+package copy
+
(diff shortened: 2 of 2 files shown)


sseago:
@nalind This works for my build. I can build it fine, and the build is working with my current test case. I've closed #308 in favor of this one.

drivers/copy/copy_linux.go
```diff
@@ -1,4 +1,4 @@
-// +build linux
+// +build cgo
```
mtrmac: Looking at the `#include <linux/fs.h>` below, this still should be Linux-only.

nalind: The `linux` build constraint is implicit based on the file's name.

mtrmac: Ah, I thought that the explicit comment overrides the tag implied by the file name. Thanks.

rhatdan:
LGTM

PR #310, opened 2019-03-27
Lock file cleanups, notably getLockfile vs. getLockFile naming

mtrmac:

### DIFF
--- a/lockfile.go
+++ b/lockfile.go
@@ -79,7 +79,7 @@ func getLockfile(path string, ro bool) (Locker, error) {
 		}
 		return locker, nil
 	}
-	locker, err := getLockFile(path, ro) // platform dependent locker
+	locker, err := createLockerForPath(path, ro) // platform-dependent locker
 	if err != nil {
 		return nil, err
 	}
--- a/lockfile.go
+++ b/lockfile.go
@@ -58,8 +58,17 @@ func GetROLockfile(path string) (Locker, error) {
 	return getLockfile(path, true)
 }
 
-// getLockfile is a helper for GetLockfile and GetROLockfile and returns Locker
-// based on the path and read-only property.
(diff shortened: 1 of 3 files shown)


mtrmac:
NOTE: Although the documentation to `GetLockfile`/`GetROLockfile` could be similarly extended with more details, I stopped at the private implementation helper level, because I don’t think I understand the API / use.

- AFAICS, `GetLockfile(singlePath); GetROLockfile(singlePath)` will fail. Is that intentional, a bug that matters, a bug that does not matter in practice?
- The idea of “read-only lock” that is actually a read-write lock, except that locking for write will _only lock it for read_ is very counter-intuitive to me, and confuses me greatly. Does it have to be like that? I would expect there to be a clear type-system-enforced separation between simple mutexes and R/W locks, and for the R/W locks to have clearly-separated “read {un,}lock” and “write {un,}lock” methods; I find it pretty scary that a `.Lock` call can silently behave as a `RLock` call, depending on the receiver object. Yet, apparently, code does depend on this behavior, and I have no idea whether it would be practical to untangle.

So, this stops at being explicit about what the helpers do, and leaves the public API, its documentation, and its use throughout the package unchanged for whoever understands it already.

vrothberg:
Please defer merging this to after https://github.com/containers/storage/pull/307.

rhatdan:
@mtrmac #307 has been merged can you update your PR?

mtrmac:
Rebased.

vrothberg:
The lockfile tests are flaky. I opened https://github.com/containers/storage/pull/314 to reduce the chance of hitting them. @mtrmac, feel free to pick the commit from there (and change at your wish).

vrothberg:
Changes here LGTM. @nalind @rhatdan PTAL.

rhatdan:
LGTM

PR #312, opened 2019-03-29
storage: move the logic to detect rootless into utils.go

giuseppe:
do not expect the caller to know what rootless user is used, but
detect it internally.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -6,6 +6,7 @@ import (
 	"os/exec"
 	"os/user"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/BurntSushi/toml"
@@ -156,6 +157,21 @@ func getTomlStorage(storeOptions *StoreOptions) *tomlConfig {
 	}
 
 	return config
+}
+
+func getRootlessUID() int {
+	uidEnv := os.Getenv("_CONTAINERS_ROOTLESS_UID")
+	if uidEnv != "" {
+		u, _ := strconv.Atoi(uidEnv)
+		return u
+	}
+	return os.Geteuid()
+}
+
+// DefaultStoreOptionsAutoDetectUID returns the default storage ops for containers
+func DefaultStoreOptionsAutoDetectUID() (StoreOptions, error) {
+	uid := getRootlessUID()
+	return DefaultStoreOptions(uid != 0, uid)
 }
 
 // DefaultStoreOptions returns the default storage ops for containers
--- a/cmd/containers-storage/main.go
+++ b/cmd/containers-storage/main.go
@@ -67,7 +67,7 @@ func main() {
 	}
 
 	if options.GraphRoot == "" && options.RunRoot == "" && options.GraphDriverName == "" && len(options.GraphDriverOptions) == 0 {
-		options, _ = storage.DefaultStoreOptions(false, 0)
+		options, _ = storage.DefaultStoreOptionsAutoDetectUID()
 	}
 	args := flags.Args()
 	if len(args) < 1 {


utils.go
```diff
@@ -158,16 +159,27 @@ func getTomlStorage(storeOptions *StoreOptions) *tomlConfig {
...
+	if uidEnv != "" {
+		u, _ := strconv.Atoi(uidEnv)
+		return u
+	}
+	return os.Geteuid()
+}
+
 // DefaultStoreOptions returns the default storage ops for containers
-func DefaultStoreOptions(rootless bool, rootlessUid int) (StoreOptions, error) {
+func DefaultStoreOptions() (StoreOptions, error) {
```
vrothberg: @giuseppe, would you open a PR at libpod for this change? Just yesterday I was cleaning up the API breakage there and this will require changes over a libpod to make it build.

utils.go
```diff
@@ -158,16 +159,27 @@ func getTomlStorage(storeOptions *StoreOptions) *tomlConfig {
...
+	uid := getRootlessUID()
+	rootless := uid != 0
```
vrothberg: Elegant :)

rhatdan:
This one will need @nalind To approve.  Since he requested the params.

giuseppe:
@rhatdan in that case, I can add another function so we have both variants

utils.go
giuseppe: this is going to break containers/image as well :-)

giuseppe:
> This one will need @nalind To approve. Since he requested the params.

I've created a new function, so the old functionality is not lost

giuseppe:
anything more holding this PR?

vrothberg:
It is still breaking the current API. Can't we keep `func DefaultStoreOptions(rootless bool, rootlessUid int) (StoreOptions, error)` and add a new one `func DefaultStoreOptionsAutoDetectUID()`?

We really need to consider the *entire dependency chain*:
```
storage -> image -> buildah -> libpod -> CRI-O
```

I did this dance last week and this week too often for changes that I did not perform (but blocked me) to not feel strong emotions on API breakage.

giuseppe:
we will still need to update containers/image as the current call there is broken.  So since we will anyway need an iteration to update containers/image, I also considered fixing the names and have `DefaultStoreOptions` without any argument, as it used to be before we added the rootless stuff.

rhatdan:
@nalind and I discussed and believe this logic should stay out of containers/storage.

giuseppe:
Was this closed by mistake?  It is only some refactoring inside containers/storage

rhatdan:
Can't you put the code in containers/image then and call the existing DefaultStorage Function?

giuseppe:
I could do that, but still I think it belongs better here than in containers/image

vrothberg:
Just to say. My only complaint was regarding the API break. The more shared code we can put down into c/image and c/storage the better it is IMHO. Especially when considering the case of Debian where all packages and their dependencies are also source packages for compilation. In that case, it is easier to move the shared code down (assuming the API is stable) as the tools can just be recompiled without a version bump.

giuseppe:
> Just to say. My only complaint was regarding the API break

if you have a strong opinion about it then I'll change it back.  

My last attempt trying to convince you :-)  For someone upgrading from a containers/storage older than a month will see `DefaultStoreOptions()` renamed to something like `DefaultStoreOptionsAutoDetect()`, while `DefaultStoreOptions()` now needs two new values, that are quite specific to the implementation.
Instead if we go with the current version of this PR, the only cost would be to drop these two values for projects that have already updated to the latest version.

rhatdan:
It is up to Nalin.  I would prefer the DefaultStoreOptions() call not get changed, but I am fine with the new call, as long as @nalind approves.

utils.go
rhatdan: How about you leave DefaultStoreOptions alone, and add DefaultStoreOptionsFromEnv()

giuseppe:
I renamed the new function so it is not a breaking change for master anymore.

vrothberg:
LGTM
Thanks a lot for bearing with me, @giuseppe :pray:

vrothberg:
> [...] the only cost would be to drop these two values for projects that have already updated to the latest version.

Just to elaborate a bit on that and to explain why I feel so strong about breaking the API. Those costs can be much higher than one might think. We first to update image, then buildah, then skopeo, then podman, then CRI-O. That's quite some work to get PRs in as they need to be in order due to the code dependencies among the projects.

Then there is another issue: anybody working on the projects is blocked when the code is not yet updated. This is way more painful because it affects folks who did not perform the API changes but are blocked.

Sometimes we must break the API and that's okay as we do not (yet) give the promise to have stable and LTSed APIs

rhatdan:
LGTM

PR #314, opened 2019-04-03
lockfile: make tests deterministic

vrothberg:
This is based on https://github.com/containers/storage/pull/310 and adds one commit on top to fix some test flakes.

### DIFF
--- a/lockfile_test.go
+++ b/lockfile_test.go
@@ -323,31 +323,35 @@ func TestLockfileReadConcurrent(t *testing.T) {
 	l, err := getTempLockfile()
 	require.Nil(t, err, "error getting temporary lock file")
 	defer os.Remove(l.name)
-	var wg sync.WaitGroup
-	var counter, highest int64
-	var highestMutex sync.Mutex
-	for i := 0; i < 100000; i++ {
-		wg.Add(1)
+	
+	// the test below is inspired by the stdlib's rwmutex tests
+	numReaders := 1000
+	locked := make(chan bool)
+	unlocked := make(chan bool)
+	done := make(chan bool)
+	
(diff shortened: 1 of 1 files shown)


mtrmac:
Adding a `Sleep` to a test to make it more likely to pass makes me suspicious. It seems to me that either the tests are trying to hit a specific scenario (in which case the scenario should be set up deliberately and precisely, without hoping to be done in any particular time), or they are just stress-testing that the code does what the API promises and does not break invariants, in which case the tests should only be asserting what _must_ happen, not what _could_ happen.

---

I can’t reproduce this, what exactly is failing? The “read counter never have gone above 1 at least once” tests? I don’t quite see why we need hundreds of goroutines for that one—just create two that both lock the lock, report back success, and don’t unlock the lock in the goroutines until told to by the parent.

That’s more tricky in `TestLockfileMixedConcurrent`, where we do want to stress-test the lock; I’m tempted to just remove the check for >1 concurrent reader; the lock is not _required_ to allow any particular number of multiple readers at the same time, after all.

Or maybe engineer an exact situation, with a writer holding the lock, two readers blocked on trying to obtain it, and unlock the writer.

vrothberg:
I think we can do both, some stress testing and another one to check for invariants. The latter is probably best done with channels or sth similar instead of sleeps. I did not have much time for a proper fix (while sleeps seem still more robust than the current state). I'll give it another try.

vrothberg:
@nalind @rhatdan PTAL.

rhatdan:
LGTM
@nalin can you merge this?

TomSweeneyRedHat:
LGTM, but would like a @nalind head nod.

nalind:
Looks reasonable.  Merging.

PR #315, opened 2019-04-05
drivers.DirCopy: fix filemode detection

vrothberg:
We failed to detect char devices as those were not covered by the
`ModeDevice` check.  Fix this by using the entire FileMode mask as
suggested by the standard library.

Also change the "unknown file type" error to include the mode to
ease debugging potential regressions in the future.


vrothberg:
@rhatdan @nalind PTAL

rhatdan:
LGTM

vrothberg:
@giuseppe PTAL

giuseppe:
LGTM

PR #317, opened 2019-04-12
runroot: add check that it is on volatile storage

giuseppe:
Make sure the runroot won't persist after a reboot, if it happens then
we can carry wrong information on the current active mounts.

Closes: https://github.com/containers/libpod/issues/2150

### DIFF
--- a/store.go
+++ b/store.go
@@ -148,6 +148,9 @@ type StoreOptions struct {
 	// for use inside of a user namespace where UID mapping is being used.
 	UIDMap []idtools.IDMap `json:"uidmap,omitempty"`
 	GIDMap []idtools.IDMap `json:"gidmap,omitempty"`
+	// SkipRunRootCheck disables the check for the RunRoot to be on a volatile
+	// storage.
+	SkipRunRootCheck bool
 }
 
 // Store wraps up the various types of file-based stores that we use into a
@@ -596,6 +599,12 @@ func GetStore(options StoreOptions) (Store, error) {
 	}
 	for _, subdir := range []string{"mounts", "tmp", options.GraphDriverName} {
 		if err := os.MkdirAll(filepath.Join(options.GraphRoot, subdir), 0700); err != nil && !os.IsExist(err) {
+			return nil, err
+		}
+	}
+
+	if !options.SkipRunRootCheck {
+		if err := validateRunRoot(options.RunRoot); err != nil {
 			return nil, err
 		}
 	}
@@ -3282,6 +3291,12 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 		storeOptions.GraphDriverName = config.Storage.Driver
 	}
 	if config.Storage.RunRoot != "" {
+		if !storeOptions.SkipRunRootCheck {
+			if err := validateRunRoot(config.Storage.RunRoot); err != nil {
+				fmt.Printf("Failed to set runroot to %s %v\n", config.Storage.RunRoot, err.Error())
+				return
+			}
+		}
 		storeOptions.RunRoot = config.Storage.RunRoot
 	}
 	if config.Storage.GraphRoot != "" {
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
@nalind @vrothberg PTAL

rhatdan:

store.go
```diff
@@ -600,6 +601,14 @@ func GetStore(options StoreOptions) (Store, error) {
...
 	}
 
+	if onTmpfs, err := mount.IsOnVolatileStorage(options.RunRoot); err != nil || !onTmpfs {
+		if err != nil {
+			return nil, errors.Wrapf(err, "cannot check if %s is on tmpfs", options.RunRoot)
+		} else {
+			return nil, errors.Wrapf(ErrTargetNotVolatile, "%s must be on tmpfs", options.RunRoot)
+		}
+	}
+
```
vrothberg: Do we also need the check for `ReloadConfigurationFile(...)` (and extend it to return an error)?

giuseppe: yes good idea.  Pushed a new version

rhatdan:
@giuseppe This breaks tests.  I think you will to mount a tmpfs on local storage to get this to pass.  I wonder if this will break CI/CD Systems when we test also.  Should we have a way to ignore this test.

store.go
```diff
@@ -535,6 +536,17 @@ type store struct {
 	digestLockRoot  string
 }
 
+func validateRunRoot(runRoot string) error {
+	if onTmpfs, err := mount.IsOnVolatileStorage(runRoot); err != nil || !onTmpfs {
+		if err != nil {
+			return errors.Wrapf(err, "cannot check if %s is on tmpfs", runRoot)
+		} else {
```
rhatdan: Nit: No need for the else, just keep the return line.

vrothberg:
CI is not yet happy:
```
go build -compiler gc -tags " libdm_no_deferred_remove ostree "  ./cmd/containers-storage
# github.com/containers/storage/pkg/mount
pkg/mount/mountinfo_linux.go:111:21: undefined: err
pkg/mount/mountinfo_linux.go:112:31: undefined: err
```

rhatdan:
@giuseppe Could you set up a WIP vendor of this patch into libpod and make sure the CI system passes.

giuseppe:
opened here: https://github.com/containers/libpod/pull/2976

vrothberg:
> opened here: containers/libpod#2976

CI is not running on a tmpfs :^)

rhatdan:
This is also going to cause issues with running containers within containers, will now require /run have a tmpfs mounted on /run

giuseppe:
> This is also going to cause issues with running containers within containers, will now require /run have a tmpfs mounted on /run

I've pushed a new version where it is possible to disable this check.  I've also updated the WIP PR for podman to set it correctly when the "container" environment variable is set

pkg/mount/mountinfo_linux.go
```diff
@@ -26,6 +30,8 @@ const (
 	   (10) mount source:  filesystem specific information or "none"
 	   (11) super options:  per super block options*/
 	mountinfoFormat = "%d %d %d:%d %s %s %s %s"
+
+	TMPFS_MAGIC = 0x1021994
```
nalind: Should probably reuse `github.com/containers/storage/drivers.FsMagicTmpfs` and the adjacent `GetFSMagic()` function here.

giuseppe: should I move these functions to `github.com/containers/storage/drivers`?  Otherwise it introduces an import cycle.

rhatdan: Maybe pkg/util?

rhatdan: Never mind that does not exists.

rhatdan: Seems to make sense.

nalind:
LGTM, with one nit: we appear to be checking the location if one is specified in the configuration file, but not if one isn't.  If that's intended, then it's fine.

rhatdan:
This still concerns me for the buildah inside of a podman container case, or inside of a docker or other container runtime.  We need a mechanism to figure out whether we are running inside of a container.  Does Docker always set the `container` environment variable?

giuseppe:
docker doesn't seem to set that.  I am going to close this PR for now, as it probably introduces more issues than it solves

PR #319, opened 2019-04-15
Fix possible integer overflow on 32bit builds

### LINKED ISSUE #316: Possible integer overflow for IDMap parsing on 32bit systems
Hey, I'm not sure if 32bit support is a thing for `containers/storage` so feel free to close if out of scope.

The function `ParseIDMap()` parses the G/UID triple and then cast the `uint32` results to `int`:
https://github.com/containers/storage/blob/ed97cd56f06fb41597db9f76f09b3105e52ed53a/pkg/idtools/parser.go#L47-L51

I tested the parser with a 32bit build and got a wrong result because `int` is not large enough:

```go
func main() {
	m, _ := ParseIDMap([]string{"1:4294967285:0"}, "GID")
	fmt.Printf("%v\n", m)
}
```
```
> export GOARCH=386
> go run parser.go
[{1 -11 0}]
```

I fix would break APIs so I would like to ask you which way we want to go here. :)

saschagrunert:
This commit fixes a possible integer overflow during U/GID map parsing.
Casting a 32bit unsigned integer to the resulting IDMap will now be
guarded by a simple overflow check. This check does apply to all
target architectures.

Closes #316

### DIFF
--- a/pkg/idtools/parser.go
+++ b/pkg/idtools/parser.go
@@ -2,6 +2,8 @@ package idtools
 
 import (
 	"fmt"
+	"math"
+	"math/bits"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@nalind @giuseppe PTAL

rhatdan:
@saschagrunert Looks like lots of failures.  Should these checks only be done on a 32 bit system?

saschagrunert:
> @saschagrunert Looks like lots of failures. Should these checks only be done on a 32 bit system?

Alright, I added an additional bit-size check.

giuseppe:
LGTM

PR #320, opened 2019-04-15
Correctly check permissions on rootless directory

rhatdan:
We have not been checking the permissions on the rootless directory
this causes issues in tools trying to access the login directory
if XDG_RUNTIME_DIR is not set correctly.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -74,7 +74,7 @@ func GetRootlessRuntimeDir(rootlessUid int) (string, error) {
 	if runtimeDir == "" {
 		tmpDir := fmt.Sprintf("/run/user/%d", rootlessUid)
 		st, err := system.Stat(tmpDir)
-		if err == nil && int(st.UID()) == os.Getuid() && st.Mode() == 0700 {
+		if err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000 {
 			return tmpDir, nil
 		}
 	}


rhatdan:
@vrothberg @giuseppe @TomSweeneyRedHat @nalind PTAL

rhatdan:
@QiWang19 PTAL

QiWang19:
LGTM

TomSweeneyRedHat:
LGTM, but would like a head nod form @vrothberg or @giuseppe

utils.go
```diff
@@ -74,7 +74,7 @@ func GetRootlessRuntimeDir(rootlessUid int) (string, error) {
 	if runtimeDir == "" {
 		tmpDir := fmt.Sprintf("/run/user/%d", rootlessUid)
 		st, err := system.Stat(tmpDir)
-		if err == nil && int(st.UID()) == os.Getuid() && st.Mode() == 0700 {
+		if err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 {
```
giuseppe: we want to check that the directory is accessible only by the current user.  Now the `&` mask drops the check on the other users. 
We could relax it to accept `0711`, that is the mode we use when using user namespaces, but I don't think we should allow a directory writeable by other users.

rhatdan: The problem is the st.Mode() != 0700 So this always fails.  I was thinking this would fix this, but it fails.

Just unset XDG_RUNTIME_DIR, and even if /run/usr/UID is setup correctly 0700, Golang reports this as a different numer.  So we fail over to /tmp/UID.

giuseppe:
LGTM

rhatdan:
@QiWang19 Could you verify this still works for you?

QiWang19:
It's working

rhatdan:
Ok I will merge once the tests pass.

TomSweeneyRedHat:
LGTM and happy green test buttons @rhdan

TomSweeneyRedHat:
@rhatdan I mean.

PR #321, opened 2019-04-18
Fix handling of additional stores

rhatdan:
When you build an image with a parent layer in read-only stores
and the new image in read/write stores, the first time you try
to create a container based on the image, it fails, since it
cannot find the image in the same store.

This patch looks not only in the same store, but all of the stores
available.

### DIFF
--- a/store.go
+++ b/store.go
@@ -1038,8 +1038,9 @@ func (s *store) imageTopLayerForMapping(image *Image, ristore ROImageStore, crea
 		return reflect.DeepEqual(layer.UIDMap, options.UIDMap) && reflect.DeepEqual(layer.GIDMap, options.GIDMap)
 	}
 	var layer, parentLayer *Layer
+	allStores := append([]ROLayerStore{rlstore}, lstores...)
 	// Locate the image's top layer and its parent, if it has one.
-	for _, s := range append([]ROLayerStore{rlstore}, lstores...) {
+	for _, s := range allStores {
 		store := s
 		if store != rlstore {
 			store.Lock()
@@ -1056,10 +1057,13 @@ func (s *store) imageTopLayerForMapping(image *Image, ristore ROImageStore, crea
 				// We want the layer's parent, too, if it has one.
 				var cParentLayer *Layer
 				if cLayer.Parent != "" {
-					// Its parent should be around here, somewhere.
-					if cParentLayer, err = store.Get(cLayer.Parent); err != nil {
-						// Nope, couldn't find it.  We're not going to be able
-						// to diff this one properly.
+					// Its parent should be in one of the stores, somewhere.
+					for _, ps := range allStores {
+						if cParentLayer, err = ps.Get(cLayer.Parent); err == nil {
+							break
+						}
+					}
+					if cParentLayer == nil {
 						continue
 					}
 				}
--- a/tests/stores.bats
+++ b/tests/stores.bats
@@ -77,6 +77,35 @@ load helpers
 	# Unmount the third layer.
 	storage --storage-opt ${STORAGE_DRIVER}.imagestore=${TESTDIR}/ro-root unmount $upperlayer
 
+	# Create an image using this third layer.
+	run storage --storage-opt ${STORAGE_DRIVER}.imagestore=${TESTDIR}/ro-root --debug=false create-image $upperlayer
+        [ "$status" -eq 0 ]
+        [ "$output" != "" ]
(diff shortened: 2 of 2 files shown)


rhatdan:

store.go
```diff
@@ -1056,11 +1057,14 @@ func (s *store) imageTopLayerForMapping(image *Image, ristore ROImageStore, crea
...
-					if cParentLayer, err = store.Get(cLayer.Parent); err != nil {
-						// Nope, couldn't find it.  We're not going to be able
-						// to diff this one properly.
-						continue
+					// Its parent should be in one of the stores, somewhere.
+					for _, ps := range allStores {
+						if cParentLayer, err = ps.Get(cLayer.Parent); err != nil {
+							// Nope, couldn't find it.  We're not going to be able
+							// to diff this one properly.
+							continue
```
nalind: This `continue` now skips to the next iteration of the loop started at line 1061, instead of the loop that starts at line 1055.

rhatdan: Fixed, now it sets a found flag when it find the parent layer, otherwise when this loop finishes it checks it loops if not found.

store.go
```diff
@@ -1051,15 +1052,20 @@ func (s *store) imageTopLayerForMapping(image *Image, ristore ROImageStore, crea
...
 				var cParentLayer *Layer
 				if cLayer.Parent != "" {
-					// Its parent should be around here, somewhere.
-					if cParentLayer, err = store.Get(cLayer.Parent); err != nil {
-						// Nope, couldn't find it.  We're not going to be able
-						// to diff this one properly.
+					// Its parent should be in one of the stores, somewhere.
+					for _, ps := range allStores {
+						if cParentLayer, err = ps.Get(cLayer.Parent); err == nil {
+							found = true
```
giuseppe: you could use a label for the loop you want to address:
```
layers: for _, candidate := range append([]string{image.TopLayer}, image.MappedTopLayers...) {
[...]
       continue layers;
[...]
```

so you won't need the additional variable

rhatdan: I don't want to breakout of the inner loop unless found.  Now I can check if cParentLayer != nil and then continue outside the loop.

rhatdan:
@nalind @giuseppe PTAL

giuseppe:
LGTM

nalind:
This needs a test.

nalind:
If I'm reading it right, the change modifies things so that we can ask a lower-level driver instance to create a layer with a parent that isn't in its store, which is not supported by devicemapper, which doesn't  try to support additional stores, or vfs, which does try to support them.

rhatdan:
Well this is not new, it is just fixing what was already happening.  
The way I caused this to happen was I created an additional store /var/lib/shared, by 
```
# buildah rmi --force --all
# mkdir /var/lib/shared
# buildah pull fedora
# mv /var/lib/containers/storage/* /var/lib/shared/
# RUN sed -i -e '/additionalimage.*/a "/var/lib/shared",' /etc/containers/storage.conf
# buildah bud -t test -f Dockerfile .
```
Build a container based on Fedora
Everything works.  BUT now try to run the container
```
# podman run -ti fedora /bin/sh
```

And podman blows up, because it can't find the underlying layer.

podman-mail:
@nalind Tests added.

rhatdan:
@nalind Tests added.
My alter ego just added a message @podman-mail

store.go
```diff
@@ -1056,10 +1057,13 @@ func (s *store) imageTopLayerForMapping(image *Image, ristore ROImageStore, crea
 				// We want the layer's parent, too, if it has one.
 				var cParentLayer *Layer
 				if cLayer.Parent != "" {
-					// Its parent should be around here, somewhere.
-					if cParentLayer, err = store.Get(cLayer.Parent); err != nil {
-						// Nope, couldn't find it.  We're not going to be able
-						// to diff this one properly.
+					// Its parent should be in one of the stores, somewhere.
+					for _, ps := range allStores {
+						if cParentLayer, err = ps.Get(cLayer.Parent); err == nil {
```
vrothberg: I think the condition should be `err == nil && cParentLayer != nil`. Currently, unless there's an error, we only look at the first store.

rhatdan: It will throw an error is the cLayer.Parent does not exists in the store.  This just makes sure that you can find the parent layer if needed.

vrothberg: Got it thanks. I did not realize that `err` is declared outside the scope here.

PR #322, opened 2019-04-23
utils: root in a userns uses global conf file

giuseppe:
If the root user created a new user namespace, it still has to use the
system configuration and storage.

Previously if root created a user namespace when missing
CAP_SYS_ADMIN, we would use the configuration file under
/root/.config/containers instead of /etc/containers.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -182,14 +182,14 @@ func DefaultStoreOptions(rootless bool, rootlessUid int) (StoreOptions, error) {
 		err                      error
 	)
 	storageOpts := defaultStoreOptions
-	if rootless {
+	if rootless && rootlessUid != 0 {
 		storageOpts, err = getRootlessStorageOpts(rootlessUid)
 		if err != nil {
 			return storageOpts, err
 		}
 	}
 
-	storageConf, err := DefaultConfigFile(rootless)
+	storageConf, err := DefaultConfigFile(rootless && rootlessUid != 0)
 	if err != nil {
 		return storageOpts, err
 	}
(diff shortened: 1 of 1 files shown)


rhatdan:
Won't 	storageConf, err := DefaultConfigFile(rootless)
still cause issues?

giuseppe:
> Won't storageConf, err := DefaultConfigFile(rootless)
> still cause issues?

yes good catch.  Fixed now

rhatdan:
LGTM

This fixes attempts to run buildah inside a container with fuse-overlay and no SYS_ADMIN Privs.

utils.go
```diff
@@ -182,14 +182,14 @@ func DefaultStoreOptions(rootless bool, rootlessUid int) (StoreOptions, error) {
...
-	if rootless {
+	if rootless && rootlessUid != 0 {
 		storageOpts, err = getRootlessStorageOpts(rootlessUid)
 		if err != nil {
 			return storageOpts, err
 		}
 	}
 
-	storageConf, err := DefaultConfigFile(rootless)
+	storageConf, err := DefaultConfigFile(rootless && rootlessUid != 0)
```
vrothberg: Do we need this fix for all callers of `DefaultConfigFile()`? If so, I think we need to fix it there.

rhatdan: The only place other then Storage where this is currently called is in libpod.

./libpod/info.go:	configFile, err := storage.DefaultConfigFile(rootless.IsRootless())
./libpod/runtime.go:		storageConfFile, err := storage.DefaultConfigFile(rootless.IsRootless())

giuseppe: good point, but we will need to change `DefaultConfigFile()` to accept another argument as the `rootlessUid` is not known

rhatdan: Can't we just fix the callers to call this correctly.  IE Only call it when you are truly in rootless mode.
I am thinking the problem here is in the callers telling us they are in rootless mode,when they are running as root.

rhatdan: Anyways I think this PR is correct and solves the problem we are seeing.  We need to fix libpod to figure out whether it is REALY in rootless mode, or just running within a User Namespace, Which are really two different things.

PR #326, opened 2019-04-23
Recreate symlink when not found

umohnani8:
If the synlinks under /var/lib/containers/storage/overlay/l are lost
during a sudden reboot, we go through all the layers under the overlay
directory and create symlinks point to their respective diff directories
with the same name as they were created before. We get this name from
the "link" file under each layer dir.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -16,7 +16,7 @@ import (
 	"sync"
 	"syscall"
 
-	"github.com/containers/storage/drivers"
+	graphdriver "github.com/containers/storage/drivers"
 	"github.com/containers/storage/drivers/overlayutils"
 	"github.com/containers/storage/drivers/quota"
 	"github.com/containers/storage/pkg/archive"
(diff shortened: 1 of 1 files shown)


umohnani8:
@nalind @rhatdan @mrunalp PTAL.

rhatdan:
The code looks ok to me.  But I want @nalind to ok it.

mrunalp:
I think we may need to create a branch and cherry-pick this on top of what we have vendored in CRI-O release-1.13. We had to revert a newer version of containers/storage couple of weeks ago and we haven't root-caused the problems with that version.

drivers/overlay/overlay.go
```diff
@@ -677,6 +677,40 @@ func (d *Driver) Remove(id string) error {
...
+// recreateSymlinks goes through the driver's home directory and checks if the diff directory
+// under each layer has a symlink created for it under the linkDir. If the symlink does not
+// exist, it creates them
+func (d *Driver) recreateSymlinks() error {
+	// List all the directories under the home directory
+	dirs, err := ioutil.ReadDir(d.home)
+	if err != nil {
+		return fmt.Errorf("error reading driver home directory %q: %v", d.home, err)
+	}
+	for _, dir := range dirs {
```
nalind: Please screen out non-directories here.  If we turn on quota enforcement, other things show up in this directory, and we don't want to trip over them.

drivers/overlay/overlay.go
```diff
@@ -677,6 +677,40 @@ func (d *Driver) Remove(id string) error {
...
+		// Skip over the linkDir
+		if dir.Name() == linkDir {
+			continue
+		}
+		// Read the "link" file under each layer to get the name of the symlink
+		data, err := ioutil.ReadFile(path.Join(d.dir(dir.Name()), "link"))
+		if err != nil {
+			return fmt.Errorf("error reading name of symlink for %q: %v", dir, err)
+		}
+		linkPath := path.Join(d.home, linkDir, strings.Trim(string(data), "\n"))
```
nalind: No need for the `\n` here.

nalind: Aargh, misread where `\n` was being passed.

drivers/overlay/overlay.go
umohnani8: Fixed

nalind:
LGTM

rhatdan:
@mrunalp Can we just add a patch to the RPM to fix this for CRI-O.  That way we don't have to revendor.

rhatdan:
Will merge once tests pass.

umohnani8:
@nalind @rhatdan @mrunalp should I go ahead and create 1.13 branch then for cri-o?

rhatdan:
@umohnani8 Yes, I guess so.

PR #328, opened 2019-04-25
overlay: fix single-lower test

### LINKED ISSUE #327: overlay storage driver doesn't support Windows Subsystem for Linux's lxfs
see https://github.com/containers/libpod/issues/3000

The storage driver was trying to mount overlay on WSL's lxfs but it was unable to do so:

```
DEBU[0000] overlay test mount with multiple lowers failed mountfrom re-exec error: exit status 1: output: invalid argument
DEBU[0000] overlay test mount with a single lower failed mountfrom re-exec error: exit status 1: output: invalid argument
```

Notice that overlay fs does work under WSL:

```

root@HOST:~# uname -a
Linux HOST 4.4.0-17134-Microsoft #706-Microsoft Mon Apr 01 18:13:00 PST 2019 x86_64 x86_64 x86_64 GNU/Linux
root@HOST:~# mount -t overlay -o lowerdir=/tmp/lower,upperdir=/tmp/upper,workdir=/tmp/workdir none /tmp/overlay
root@HOST:~# mount | grep /tmp/overlay
none on /tmp/overlay type overlay (rw,relatime,lowerdir=/tmp/lower,upperdir=/tmp/upper,workdir=/tmp/workdir)
```

nalind:
Somehow, the multiple-lowers test in the overlay driver succeeded even when we didn't pass in an upper or work directory, but the single-lower test definitely needs us to do that.

This should fix #327.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -320,6 +320,8 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 		mergedDir := filepath.Join(layerDir, "merged")
 		lower1Dir := filepath.Join(layerDir, "lower1")
 		lower2Dir := filepath.Join(layerDir, "lower2")
+		upperDir := filepath.Join(layerDir, "upper")
+		workDir := filepath.Join(layerDir, "work")
 		defer func() {
 			// Permitted to fail, since the various subdirectories
 			// can be empty or not even there, and the home might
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM assuming happy tests.

nalind:
Corrected the commit log message.

rhatdan:
You are allowed to mount multiple lower layers without an upper, but the merge directory will always be read/only.

rhatdan:
LGTM

rhatdan:
Not sure how any of this is passing.

rhatdan:
LGTM

PR #332, opened 2019-04-29
Add MountTemp and RemoveTemp interfaces

rhatdan:
These interfaces can be used to setup a graphdriver mountpoint
of the source directory for use within a container.
The RemoveTemp interface umounts the mountpoint and then removes
all of the modified data in the graphdriver for this source directory.

The primary use case of these interfaces is for container engines that
want to mount a directory from the host system into the container. The
source dirctory then can be modified without actually changing the
directory on the host.

Containers will use these interfaces for sharing packaing cache directories
like /var/cache/dnf, to help speed up container builds.

### DIFF
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -100,6 +100,13 @@ type ProtoDriver interface {
 	Cleanup() error
 	// AdditionalImageStores returns additional image stores supported by the driver
 	AdditionalImageStores() []string
+
+	// MountTemp mounts a source directory from the host using
+	// graphdriver and returns the mountpoint
+	MountTemp(sourcedir, source, mountLabel string) (string, error)
+
+	// RemoveTemp removes temporary mountpoint from host using graphdriver
+	RemoveTemp(mountpoint string) error
 }
 
 // DiffDriver is the interface to use to implement graph diffs
--- a/store.go
+++ b/store.go
@@ -446,6 +446,15 @@ type Store interface {
 	// GID maps (if any are defined) don't contain corresponding IDs.
 	ContainerParentOwners(id string) ([]int, []int, error)
 
+	// MountTemp creates a temporarily mount of a source directory from the
+	// host using graphdriver in the container `id` private storage.
+	// MountTemp then mounts up the storage on the system and returns
+	// the mountpoint
+	MountTemp(id, source string) (string, error)
+
+	// RemoveTemp removes temporary mountpoint and any content written to it
+	RemoveTemp(mountpoint string) error
+
 	// Lookup returns the ID of a layer, image, or container with the specified
 	// name or ID.
 	Lookup(name string) (string, error)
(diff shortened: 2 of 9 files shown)


store.go
```diff
@@ -446,6 +446,13 @@ type Store interface {
 	// GID maps (if any are defined) don't contain corresponding IDs.
 	ContainerParentOwners(id string) ([]int, []int, error)
 
+	// MountTempFromSource mounts a source directory from the host using
+	// graphdriver and returns the mountpoint
```
nalind: What is `id` used for?

store.go
```diff
@@ -1455,6 +1462,42 @@ func (s *store) ImageBigDataSize(id, key string) (int64, error) {
 	return -1, ErrSizeUnknown
 }
 
+func (s *store) MountTempFromSource(id, source string) (string, error) {
+	rcstore, err := s.ContainerStore()
+	if err != nil {
+		return "", err
+	}
+	rcstore.RLock()
+	defer rcstore.Unlock()
```
nalind: Missing a `Modified()` check and `Load()` here.

rhatdan: Added.

store.go
rhatdan: Used to select the location for the mount point.  We want to make sure out mountpoints and upper directory are under containers/storage/overlay-containers/UUID/upper directory.

store.go
```diff
@@ -1455,6 +1462,47 @@ func (s *store) ImageBigDataSize(id, key string) (int64, error) {
...
+		return "", err
+	}
+	middleDir := s.graphDriverName + "-containers"
+	sourcedir := filepath.Join(s.GraphRoot(), middleDir, id, strings.Replace(source, "/", "_", -1))
+
+	return driver.MountTemp(sourcedir, source, container.MountLabel())
+}
+
+func (s *store) RemoveTemp(source string) error {
+	rcstore, err := s.ContainerStore()
```
nalind: This doesn't seem to be used in this function.

drivers/driver.go
```diff
@@ -100,6 +100,13 @@ type ProtoDriver interface {
 	Cleanup() error
 	// AdditionalImageStores returns additional image stores supported by the driver
 	AdditionalImageStores() []string
+
+	// MountTemp mounts a source directory from the host using
+	// graphdriver and returns the mountpoint
+	MountTemp(sourcedir, source, mountLabel string) (string, error)
```
nalind: Which parameter is the directory?  What's the other parameter used for?

store.go
nalind: That should be reflected in the godoc.

But why, though?  There's no indication in here of why tying a mount to a particular container is required.

More broadly, why are these methods of lower level drivers when the only driver that does anything is overlay?  Are we dumping this into the lower level drivers because it's where knowledge of the overlay filesystem is currently concentrated?

What's the plan for cases where multiple processes want to use the same host directory as a temporary directory during overlapping periods of time?

rhatdan: @vrothberg Asked for this to be driver based, because he wanted support for BTRFS.

store.go
rhatdan: I thought I should lock the container store while it is deleted, but if you think it is safe, I can remove this.

store.go
rhatdan: How about?
```
	// MountTemp creates a temporarily mount of a source directory from the
	// host using graphdriver in the container `id` private storage.
	// MountTemp then mounts up the storage on the system and returns
	// the mountpoint
```

drivers/driver.go
rhatdan: Fixed by renaming sourcedir to contentDir.

rhatdan:
After discussing with @nalind We have decided that this really should not be in containers/storge, and really would not work well outside of Overlay.  BTRFS would require all volumes to on BTRFS storage, so I doubt we can get any other driver to support it. 

I am going to move these functions back to buildah/pkg/overlay and use them from there.

PR #333, opened 2019-04-29
Fix up recreateSymlink

umohnani8:
Noticed that the link directory may be lost also with the symlinks,
so this fix creates it if not found and then restores all the links.
Also fix up logic to check wether a dir is found or not instead of
a regular file.


umohnani8:
@rhatdan @nalind @mrunalp PTAL

rhatdan:
LGTM
@nalind PTAL

nalind:
LGTM

PR #334, opened 2019-05-01
layers: use the correct idmapping when creating a layer

giuseppe:
prefer the custom idmapping specified for the layer instead of using the global one configured for the driver.

Closes: https://github.com/containers/libpod/issues/2960

### DIFF
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -40,6 +40,7 @@ var (
 type CreateOpts struct {
 	MountLabel string
 	StorageOpt map[string]string
+	*idtools.IDMappings
 }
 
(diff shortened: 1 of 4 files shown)


edsantiago:
I guess this is where I'm forced to confess that I have no idea how to get this into my podman tree for testing. Adding a commit sha to `vendor.conf` and running `make vendor` failed after 20 minutes (I'm on a slow connection): "... is not a tree". There must be something very simple I'm missing. TIA for any advice.

TomSweeneyRedHat:
@edsantiago sent you some notes in a separate email.  Ping me if they're as clear as mud.

nalind:
Makes sense.

rhatdan:
LGTM

Issue #323, opened 2019-04-23
Documentation for usage of OSTree deduplication missing


Silvanoc:
Thanks to [this comment in a test](https://github.com/containers/storage/blob/2f12336bf40037fd5c9b8aa2c50e362b7858bf83/tests/import-layer-ostree.bats#L40) I've discovered that the OSTree deduplication only applies to read-only layers.

The manpages of `container-storage` don't report any command or option to make use of it, but the option exists for commands [`create-layer`](https://github.com/containers/storage/blob/3d342550da009a8035d663b7aa38d01b4b2ff8bf/cmd/containers-storage/create.go#L221) and [`import-layer`](https://github.com/containers/storage/blob/3d342550da009a8035d663b7aa38d01b4b2ff8bf/cmd/containers-storage/create.go#L241).

On my experiments I couldn't find out how to get OSTree deduplication using `create-layer --readonly`.
With `import-layer --readonly` works as expected and tested by `import-layer-ostree.bats`.

---
Out of the previously mentioned issues I'd say that following tasks would be meaningful:

- [x] Document in the manpage of `containers-storage` that a read-only layers are needed to get OSTree deduplication.
- [ ] Update manpage of `create-layer`. At least the option `--readonly, -r` is missing.
- [x] Add a manpage for `import-layer`.
- [ ] Document how to create a read-only layer by using `create-layer`, or otherwise what's the consequence of `create-layer --readonly`.

Silvanoc:
I might help on the first three tasks, but not on the fourth until someone explains me how it's supposed to be used.

Silvanoc:
The current [last commit](https://github.com/containers/storage/commit/3da0b926c7ba5eca0f7144760e4f9fca8a2100be) from @rhatdan catch my attention, since it's also about non-writable storage.

Having a look at the tests, I'm puzzled since I cannot see in [`tests/store.bats`](https://github.com/containers/storage/blob/3da0b926c7ba5eca0f7144760e4f9fca8a2100be/tests/stores.bats) what makes `ro-root` become a read-only storage... Is it the [`shutdown`](https://github.com/containers/storage/blob/593f2796eb3b93c6312fc0d70295ea796fa3b601/tests/stores.bats#L60)?

In general, does it have anything to do with read-only layers? Or is it something completely different?

rhatdan:
[...]

PR #339, opened 2019-05-03
Document OSTree deduplication and add manpage for import-layer

Silvanoc:
These commits address points 1 and 3 of https://github.com/containers/storage/issues/323:
 * Document in the manpage of containers-storage that a read-only layers are needed to get OSTree deduplication.
 * Add a manpage for import-layer.


rhatdan:
LGTM
@giuseppe @nalind PTAL

giuseppe:
The OSTree part LGTM

PR #341, opened 2019-05-07
CreateContainer: don't worry about mapping layers unless necessary

nalind:
When creating a container, don't worry about whether or not the base image's top layer has the right ID mappings in cases where the base image doesn't have a top layer.


rhatdan:
LGTM

rhatdan:
@giuseppe @vrothberg @TomSweeneyRedHat PTAL

PR #344, opened 2019-05-15
tests: fix regression when using idmapping

giuseppe:
fix a regression introduced with d1554f0dcd6a74eaade2181239f8df7df975f7b1 and 7c6d502e9bd4e485eeb291ed3c8f7b44dcecb34d that broke some tests.

Also, skip some tests that are not valid when using fuse-overlayfs as it supports ID shifting.

With this patch for fuse-overlayfs: https://github.com/containers/fuse-overlayfs/pull/76

I was able to run the tests suite as:

```
# export STORAGE_OPTION=overlay.mount_program=/usr/local/bin/fuse-overlayfs
# export STORAGE_DRIVER=overlay
# cd tests
# ./test_runner.bash
```

### DIFF
--- a/drivers/chown.go
+++ b/drivers/chown.go
@@ -55,6 +55,9 @@ func chownByMapsMain() {
 		if err != nil {
 			return fmt.Errorf("error walking to %q: %v", path, err)
 		}
+		if path == "." {
+			return nil
+		}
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@nalind @vrothberg PTAL

PR #345, opened 2019-05-15
Cirrus: Add initial support for Cirrus-CI

cevich:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -79,9 +79,6 @@ testing_task:
 # Update metadata on VM images referenced by this repository state
 meta_task:
 
-    depends_on:
-        - "gating"
-
     container:
         image: "quay.io/libpod/imgts:latest"  # see contrib/imgts
         cpu: 1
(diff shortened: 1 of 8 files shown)


cevich:
***Note to me:*** This is sharing base-images with the libpod project, so need to add meta task using imgts to set last-used-on labels from this project.  Otherwise the base images might get purged.

cevich:
@rhatdan easy part is done PTAL - VMs boot, run commands, and collect logs :smile:   I think I'd prefer to work on the hard-part (VM setup and running tests) via separate PRs.

Note: I did not include CentOS (or RHEL) among the VMs, as they tended to get in the way for libpod being a fast-moving upstream.  Not sure about here. IMHO, it's better to leave the downstream automation to the downstream systems.  But I'll add one/both here if you think they're needed.

rhatdan:
@nalind @TomSweeneyRedHat @vrothberg PTAL

rhatdan:
Nice, LGTM. Now we just need to plum in the tests.
@nalind @giuseppe FYI

cevich:
> LGTM. It's magical!

Please don't unnecessarily over-inflate my ego, it may burst!   TBH, this was mostly a bunch of copy-paste work, and some clicky-clicky on the google cloud web thingie :grin: 

> Now we just need to plum in the tests.

Oooohhhhh, it's more difficult than that I'm afraid.  The next step is to figure out everything that needs to go into ``setup.sh`` for each distro.  Sounds easy, but it can be mind-numbing repetitive.  Unless we get it perfect on the first try, it means re-running the same set of commands over and over and over again.  

@nalind @giuseppe to maintain our sanity, I'll port my `get_ci_vm.sh` script over here from libpod.  This will  let us spin up VMs to play with at will, outside of the CI environment.  We can use these to develop a working `setup.sh` and (later) debug any problems that arise.

PR #346, opened 2019-05-16
Start implementing testing in cirrus

rhatdan:
Hook up regular tests to run on fedora/ubuntu natively.

### DIFF
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -4,24 +4,28 @@ set -e
 
 source $(dirname $0)/lib.sh
 
+install_ooe
+
 show_env_vars
 
 cd $GOSRC
 
+export RPMBuildRequires="podman autoconf automake golang go-md2man gpgme-devel device-mapper-devel btrfs-progs-devel libassuan-devel libseccomp-devel  glib2-devel  ostree-devel  make bats fuse3-devel fuse3"
+export AptBuildRequires="autoconf automake golang go-md2man libgpgme11-dev libdevmapper-dev libseccomp-dev libglib2.0-dev  libostree-dev  make bats aufs-tools fuse3 libfuse3-dev libbtrfs-dev"
+
 case "$OS_REL_VER" in
-    fedora-30)
+    fedora-*)
         echo "Setting up $OS_RELEASE_ID $OS_RELEASE_VER"  # STUB: Add VM setup instructions here
-        ;;
-    fedora-29)
-        echo "Setting up $OS_RELEASE_ID $OS_RELEASE_VER"  # STUB: Add VM setup instructions here
+	dnf -y update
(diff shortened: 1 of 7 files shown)


cevich:
@rhatdan this will help a lot with figuring out what steps are needed: #348

Otherwise the modify-rerun-CI loop will land you repeatedly waiting for the same package installs over and over and over.  Best figure out the steps manually first, then use CI in a PR to do a final check.

contrib/cirrus/setup.sh
```diff
@@ -8,18 +8,30 @@ show_env_vars
...
     ubuntu-19)
         echo "Setting up $OS_RELEASE_ID $OS_RELEASE_VER"  # STUB: Add VM setup instructions here
+	apt-get -qq update
+	apt-get -qq install realpath $BuildRequires
+        install_fuse_overlay_from_git
         ;;
     ubuntu-18)
         echo "Setting up $OS_RELEASE_ID $OS_RELEASE_VER"  # STUB: Add VM setup instructions here
+	apt-get -qq update
+	apt-get -qq install realpath $BuildRequires
```
cevich: apt-get is horrible to try and use under automation.  Other processes on the system will grab it's locks, and there's no provision to wait on them.  I'll scare up a retry/timeout loop I used in libpod for this...

cevich: Okay, #349 should help with this.

cevich:
@rhatdan I see things running, that's good.  I don't see a 10-minute timeout happening anywhere though, they look like regular test failures for one reason or another.  What am I missing?

cevich:
Oh I see it now:

```
[+0791s] *** Test killed with quit: ran too long (10m0s).
```

That's coming from the test harness or the test itself, will have to dig deeper.

podmanadm:
@cevich I think this is an actual bug in container/storage on f29.  I think Overlay is hanging in a deadlock.  For some reason this passes on F30.  I asked @nalind and @vrothberg To look at this, since I think we might have a deadlock causing it.

Not a cirrus problem.

rhatdan:
podmanadm is my alterego

cevich:
> I think this is an actual bug in container/storage on f29

@rhatdan I'm glad to hear CI is fulfilling it's purpose right out of the gate! :grinning:   That should be manually reproducible using that `hack/get_ci_vm.sh` by Nalin, but I don't think I gave Valentin access, I'll fix that now...

rhatdan:
@cevich @nalind @TomSweeneyRedHat 
This test is only running overlay tests, right now.  Lets get this merged and I will continue to update this with enabling more tests.

nalind:
Alright, looks fine as a step.

TomSweeneyRedHat:
LGTM

cevich:
/lgtm

nice work.

PR #347, opened 2019-05-16
lockfile: add RecursiveLock() API

vrothberg:
Add a RecursiveLock() API to allow for recursive acquisitions of a
writer lock within the same process space.  This is yet another
requirement for the copy-detection mechanism in containers/image where
multiple goroutines can be pulling the same blob.  Having a recursive
lock avoids a complex synchronization mechanism as the commit order is
determinted by the corresponding index in the image.


vrothberg:
@nalind @rhatdan @giuseppe PTAL

I need it [here](https://github.com/containers/image/blob/7531b7b2051bbd344ad65324e5a5a363e43c0d8a/storage/storage_image.go#L505).

TomSweeneyRedHat:
LGTM, but definitely want some other head nods.

lockfile_unix.go
```diff
@@ -161,7 +180,7 @@ func (l *lockfile) Unlock() {
 		// Close the file descriptor on the last unlock.
```
nalind: A `unix.FcntlFlock()` call to clear the lock should be unnecessary if we're going to close the descriptor right after.

vrothberg: That's unrelated to this PR. Let's fix that in another one.

vrothberg:
@nalind @rhatdan are we good to go?

vrothberg:
@nalind @rhatdan ping

rhatdan:
I am fine with this, but we need @nalind review.

nalind:
Looks reasonable, but I'd like a test that ensures that mixing recursive and non-recursive write locks in a single process doesn't unexpectedly break.

vrothberg:
Great idea. I will update the PR and re-ping once the new tests are green.

lockfile_test.go
```diff
@@ -408,6 +498,63 @@ func TestLockfileMixedConcurrent(t *testing.T) {
 	}
 }
 
+func TestLockfileMixedConcurrentRecursiveWriters(t *testing.T) {
```
vrothberg: Here's the new test.

rhatdan:
@nalind Can we merge this so we can create a new version?

nalind:
LGTM

PR #349, opened 2019-05-16
cirrus: Add retry/timeout function for apt-get

cevich:


cevich:
@rhatdan PTAL, this will make working with apt-get **much** less painful.

rhatdan:
LGTM

rhatdan:
This is just for testing, so it will not change functionality.

nalind:
We don't seem to be using much of this yet, so, sure.

PR #350, opened 2019-05-16
Fixes to allow unit test to pass on fedora 30

rhatdan:
Fix unit test to ignore relatime
Update vendor to fix unit tests


rhatdan:

nalind:
This seems to be pulling in an updated `github.com/google/go-cmp`.  Is that intended?

rhatdan:
Yes, Needed this to fix copy tests failing to compile.

pkg/mount/mounter_linux_test.go
```diff
@@ -171,6 +171,9 @@ func validateMount(t *testing.T, mnt string, opts, optional, vfs string) {
 		if mi.Opts != "" {
 			for _, opt := range strings.Split(mi.Opts, ",") {
 				opt = clean(opt)
+				if opt == "relatime" {
+					continue
+				}
 				if !has(wantedOpts, opt) && !has(pOpts, opt) {
```
nalind: It'll be easier to follow this in the future if we follow the pattern established below with `volunteeredVFS`: add a `volunteeredOpts` map that "knows" that sometimes we get the `relatime` when we don't specifically ask for it, and screen it out here using `!has(volunteeredOpts, opt)` rather than a one-off check.

rhatdan: Fixed

giuseppe:
LGTM

PR #351, opened 2019-05-21
Add a ReadOnly field to Layers and Images

nalind:
Add a ReadOnly field to Image and Layer structures, for consumers of the library to use in diagnosing why they can't delete them.


nalind:
@baude Does this API addition look like what you need?

TomSweeneyRedHat:
Changes LGTM.
I don't think there's an issue if you drop this code on a vm/host that has pre-existing images, but thought I'd ask if it's OK as far as you know.

nalind:
@TomSweeneyRedHat that's the intent, at least.  The new field is neither read from, nor written to, on-disk storage, which makes sense because a given location can be used as read-write storage by one process and read-only storage by another.

TomSweeneyRedHat:
@nalind thanks for the sanity check.

rhatdan:
LGTM

PR #353, opened 2019-05-22
Add VFS storage driver testing.

rhatdan:


rhatdan:
@nalind @TomSweeneyRedHat @giuseppe @cevich @vrothberg PTAL
Lets get this merged.  All we have left now is devicemapper tests.

rhatdan:
@giuseppe is this test worth running?
#make STORAGE_DRIVER=overlay FUSE_OVERLAYFS_DISABLE_OVL_WHITEOUT=1 STORAGE_OPTION=overlay.mount_program=/usr/bin/fuse-overlayfs local-test-integration

TomSweeneyRedHat:
LGTM, would like a head nod from @giuseppe

giuseppe:
> #make STORAGE_DRIVER=overlay FUSE_OVERLAYFS_DISABLE_OVL_WHITEOUT=1 STORAGE_OPTION=overlay.mount_program=/usr/bin/fuse-overlayfs local-test-integration

yes I think so, we will test with the `.wh.` whiteouts that are used when running as rootless

LGTM

contrib/cirrus/build.sh
```diff
@@ -12,19 +12,16 @@ echo "local-test-integration Overlay"
 make STORAGE_DRIVER=overlay local-test-integration
 echo "local-test-integration Fuse-overlay"
 make STORAGE_DRIVER=overlay STORAGE_OPTION=overlay.mount_program=/usr/bin/fuse-overlayfs local-test-integration
+echo "local-test-unit"
```
cevich: s/unit/integration/

or, maybe easier to `set -x` on the whole script?  Sometimes that can be too noisy though.
Another option could be to add a "print then do" helper in `lib.sh` similar to https://github.com/containers/storage/blob/master/hack/get_ci_vm.sh#L36 (but without `--background`).

contrib/cirrus/build.sh
```diff
@@ -12,19 +12,16 @@ echo "local-test-integration Overlay"
...
+make STORAGE_DRIVER=vfs local-test-integration
+echo "local-test-unit"
+make local-test-unit
 
 case "$OS_REL_VER" in
     ubuntu-19)
 	echo "local-test-integration Aufs"
 	make STORAGE_DRIVER=aufs local-test-integration
```
cevich: I think this means the Ubuntu runs will take twice as long, and there will be two integration-test outputs one after another.  This became a headache in libpod, where we had the remote integration tests following the regular ones.

Brent and I fixed this, using a Cirrus-CI mechanism that makes it [simple to automatically duplicate a task with a slight modification, like a different env. var. value.](https://github.com/containers/libpod/blob/master/.cirrus.yml#L272-L275)  
  Then at [runtime, both run in paralle but behave differently depending on the env. var value.](https://github.com/containers/libpod/blob/master/contrib/cirrus/integration_test.sh#L41-L45)  Therefor they have separate outputs, and all else is equal.

contrib/cirrus/build.sh
rhatdan: Added showrun.

contrib/cirrus/build.sh
rhatdan: I don't see why this would take twice as long, it is just going to run the aufs tests on Ubuntu platforms.

contrib/cirrus/build.sh
rhatdan: https://github.com/containers/storage/pull/354

contrib/cirrus/build.sh
cevich: It could be that I'm reading the script wrong, but I see ``make STORAGE_DRIVER=overlay ... local-test-integration`` always, then for Ubuntu it also does 'make STORAGE_DRIVER=aufs local-test-integration'.  So I'm assuming `local-test-integration` == `local-test-integration` (time-wise) which could be wrong.

So what can happen is, somebody is debugging an Ubuntu problem, and looks at the output in the cirrus WebUI.  They see integration test output, and assume it's the `STORAGE_DRIVER=overlay` variety.  In reality, Cirrus-CI has truncated the output, and is showing the Ubuntu case with `STORAGE_DRIVER=aufs`.  This can cost the poor debugger lots of wasted time.  It isn't a problem if the tests all clearly identify the storage-driver they use, but poking through results here I didn't see that.

rhatdan: Is the showrun enough to tell you which flag is being used?

cevich: It will, at the top of the output, which (depending on length) could be cut-off by Cirrus-CI.  As long as the per-script output is less than ~1000 lines (not sure of actual limit), `showrun` is fine.  Also separating the different actions (i.e. build | unit | integration) will keep the output shorter and help avoid this.

PR #354, opened 2019-05-23
Add test for fuse-overlay without mknod

rhatdan:

### DIFF
--- a/contrib/cirrus/build.sh
+++ b/contrib/cirrus/build.sh
@@ -6,22 +6,16 @@ source $(dirname $0)/lib.sh
 
 cd $GOSRC
 make install.tools
-echo "Build Binary"
-make local-binary
-echo "local-test-integration Overlay"
-make STORAGE_DRIVER=overlay local-test-integration
-echo "local-test-integration Fuse-overlay"
-make STORAGE_DRIVER=overlay STORAGE_OPTION=overlay.mount_program=/usr/bin/fuse-overlayfs local-test-integration
-echo "local-test-unit"
-make STORAGE_DRIVER=vfs local-test-integration
-echo "local-test-unit"
-make local-test-unit
+showrun make local-binary
+showrun make STORAGE_DRIVER=overlay local-test-integration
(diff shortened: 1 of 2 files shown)


rhatdan:
@giuseppe @nalind @TomSweeneyRedHat @cevich PTAL

giuseppe:
LGTM

contrib/cirrus/build.sh
```diff
@@ -6,22 +6,16 @@ source $(dirname $0)/lib.sh
...
-echo "local-test-unit"
-make STORAGE_DRIVER=vfs local-test-integration
-echo "local-test-unit"
-make local-test-unit
+showrun make local-binary
+showrun make STORAGE_DRIVER=overlay local-test-integration
+showrun make STORAGE_DRIVER=overlay STORAGE_OPTION=overlay.mount_program=/usr/bin/fuse-overlayfs local-test-integration
+showrun make STORAGE_DRIVER=overlay FUSE_OVERLAYFS_DISABLE_OVL_WHITEOUT=1 STORAGE_OPTION=overlay.mount_program=/usr/bin/fuse-overlayfs local-test-integration
+showrun make STORAGE_DRIVER=vfs local-test-integration
+showrun make local-test-unit
```
cevich: Ugg, sorry, I'm so slow to do reviews.  Only comment I'd make is to consider moving some of these things into their dedicated scripts, instead of just calling everything `build` :smile:   The poor, poor, `unit.sh` and `integration.sh` scripts are very empty and lonely :cry: 

'Course the other option is to just kill the other scripts completely, if having the distinction in outputs is unnecessary.

rhatdan: Please open a PR to split them the way you would like.

cevich: Sure happy to.  Though let me know if you think it's unnecessary.  I'm just as happy to kill off those extra scripts if having the distinctions isn't useful.

cevich: #355

rhatdan: Probably not important here. Since I don't think we will be adding many more extensive tests.  I rather spend our time to get an infrastructure like this setup for Buildah.

PR #355, opened 2019-05-29
Separate CI ops for ease of debug/maintenance

cevich:


rhatdan:
LGTM
@TomSweeneyRedHat @nalind @vrothberg @giuseppe PTAL

cevich:
> Probably not important here. Since I don't think we will be adding many more extensive tests. I rather spend our time to get an infrastructure like this setup for Buildah.

okay, I'll consolidate the scripts rather than separate.  Keeping things simple and useful for y'all is my goal.  Then I'll copy what's here over to buildah as a starting point...

cevich:
...okay off to buildah land.

giuseppe:
LGTM

PR #357, opened 2019-06-05
Cherry-pick moby/moby #39292 for CVE-2018-15664 fixes

mheon:
As the title says. We'll probably need a new release of c/storage once this lands.

I think `CopyFileWithTarAndChown` and `CopyWithTarAndChown` (and probably `UntarPathAndChown`) need double-checking to make sure they're also fixed, they're not included in the scope of the cherry pick.

### DIFF
--- a/pkg/chrootarchive/archive.go
+++ b/pkg/chrootarchive/archive.go
@@ -1,7 +1,7 @@
 package chrootarchive
 
 import (
-	"archive/tar"
+	stdtar "archive/tar"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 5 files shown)


mheon:
I think `CopyFileWithTarAndChown` might have symlink issues with the `NewReader()` call, but someone else should double-check me here, I'm not terribly familiar with the code.

nalind:
Looks reasonable.  `tar.NewReader()` sets us up to parse of a tar stream, for demuxing headers and file contents.  `CopyFileWithTarAndChown()` uses it to skip over headers and digest only the contents of a file, and doesn't touch the filesystem.

rhatdan:
I was working on a similar PR Yesterday, but this LGTM

PR #358, opened 2019-06-05
Make chown tolerant of errors when doing single user mapping

kpelzel:
**Linked to this PR:** https://github.com/containers/buildah/pull/1645

### Problem:
When building a container as an unprivileged user and without uidmap tools, buildah will fail when chowning system files. This patch ignores those chown errors if buildah has fallen back to single user mapping mode.

### What this patch changes:
**Buildah side:**
unshare.go checks if newuidmap and newgidmap is installed by running "which newuidmap". If uidmap tools are not installed it sets an environment variable to pass down to the child.

**containers/storage side:**
diff_unix.go passes in the hosts environment variables (including the single user mapping environment variable) to the "storage-applyLayer" command. This allows checkChownErr in idtools.go to check if it needs to ignore the chown errors.

### Notes:
I'm not completely sure if using environment variables is the best way to get this information down to checkChownErr, but it seemed to work best in terms minimal code change. I would love to hear feedback on better solutions.

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -36,14 +36,15 @@ type (
 
 	// TarOptions wraps the tar options.
 	TarOptions struct {
-		IncludeFiles     []string
-		ExcludePatterns  []string
-		Compression      Compression
-		NoLchown         bool
-		UIDMaps          []idtools.IDMap
-		GIDMaps          []idtools.IDMap
-		ChownOpts        *idtools.IDPair
-		IncludeSourceDir bool
+		IncludeFiles      []string
+		ExcludePatterns   []string
+		Compression       Compression
+		NoLchown          bool
+		UIDMaps           []idtools.IDMap
+		GIDMaps           []idtools.IDMap
+		IgnoreChownErrors bool
+		ChownOpts         *idtools.IDPair
+		IncludeSourceDir  bool
 		// WhiteoutFormat is the expected on disk format for whiteout files.
 		// This format will be converted to the standard format on pack
 		// and from the standard format on unpack.
@@ -554,7 +555,7 @@ func (ta *tarAppender) addTarFile(path, name string) error {
 	return nil
 }
 
-func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, Lchown bool, chownOpts *idtools.IDPair, inUserns bool) error {
+func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, Lchown bool, chownOpts *idtools.IDPair, inUserns, ignoreChownErrors bool) error {
 	// hdr.Mode is in linux format, which we can use for sycalls,
 	// but for os.Foo() calls we need the mode converted to os.FileMode,
 	// so use hdrInfo.Mode() (they differ for e.g. setuid bits)
@@ -636,8 +637,13 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 		if chownOpts == nil {
 			chownOpts = &idtools.IDPair{UID: hdr.Uid, GID: hdr.Gid}
 		}
-		if err := idtools.SafeLchown(path, chownOpts.UID, chownOpts.GID); err != nil {
-			return err
+		err := idtools.SafeLchown(path, chownOpts.UID, chownOpts.GID)
+		if err != nil {
+			if ignoreChownErrors {
+				fmt.Fprintf(os.Stderr, "Chown error detected. Ignoring due to ignoreChownErrors flag: %v\n", err)
+			} else {
+				return err
+			}
 		}
 	}
 
@@ -984,7 +990,7 @@ loop:
 			chownOpts = &idtools.IDPair{UID: hdr.Uid, GID: hdr.Gid}
 		}
 
-		if err := createTarFile(path, dest, hdr, trBuf, !options.NoLchown, chownOpts, options.InUserNS); err != nil {
+		if err := createTarFile(path, dest, hdr, trBuf, !options.NoLchown, chownOpts, options.InUserNS, options.IgnoreChownErrors); err != nil {
 			return err
 		}
 
--- a/drivers/aufs/aufs.go
+++ b/drivers/aufs/aufs.go
@@ -35,7 +35,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/containers/storage/drivers"
+	graphdriver "github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/chrootarchive"
 	"github.com/containers/storage/pkg/directory"
@@ -526,13 +526,14 @@ func (a *Driver) DiffGetter(id string) (graphdriver.FileGetCloser, error) {
 	return fileGetNilCloser{storage.NewPathFileGetter(p)}, nil
 }
 
-func (a *Driver) applyDiff(id string, idMappings *idtools.IDMappings, diff io.Reader) error {
+func (a *Driver) applyDiff(id string, idMappings *idtools.IDMappings, diff io.Reader, ignoreChownErrors bool) error {
 	if idMappings == nil {
 		idMappings = &idtools.IDMappings{}
 	}
 	return chrootarchive.UntarUncompressed(diff, path.Join(a.rootPath(), "diff", id), &archive.TarOptions{
-		UIDMaps: idMappings.UIDs(),
-		GIDMaps: idMappings.GIDs(),
+		UIDMaps:           idMappings.UIDs(),
+		GIDMaps:           idMappings.GIDs(),
+		IgnoreChownErrors: ignoreChownErrors,
 	})
 }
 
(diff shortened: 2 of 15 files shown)


rhatdan:
Thanks @kpelzel.
Your PR Needs to be signed

git commit -a -s --amend
git push --force

Should fix this.

rhatdan:
I think it would be better to set this flag in the store config rather then rely on environment variables.
@nalind @giuseppe WDYT?

giuseppe:
> I think it would be better to set this flag in the store config rather then rely on environment variables.
> @nalind @giuseppe WDYT?

yes, I think it should be an option in the store config

kpelzel:
> > I think it would be better to set this flag in the store config rather then rely on environment variables.
> > @nalind @giuseppe WDYT?
> 
> yes, I think it should be an option in the store config

Is there currently anyway to access the store config from the idtools package or does this need to be passed in as a function parameter?

rhatdan:
They are unrelated so you would need to pass the flay in.

kpelzel:
This new commit adds an IgnoreChownErrors boolean to store, layerstore, and tarOptions structs. This seemed like the only way to pass this option down to createTarFile where the chown errors need to be ignored.

giuseppe:
you 'd need to adapt tests as well to use the new APIs

pkg/archive/archive.go
```diff
@@ -636,8 +637,13 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 		if chownOpts == nil {
 			chownOpts = &idtools.IDPair{UID: hdr.Uid, GID: hdr.Gid}
 		}
-		if err := idtools.SafeLchown(path, chownOpts.UID, chownOpts.GID); err != nil {
-			return err
+		err := idtools.SafeLchown(path, chownOpts.UID, chownOpts.GID)
+		if err != nil {
+			if ignoreChownErrors {
+				fmt.Fprintf(os.Stderr, "Chown error detected. Ignoring due to ignoreChownErrors flag: %v\n", err)
```
rhatdan: Should we check for specific errors like EINVAL, to ignore.  There could be other errors like ENOEXIST, or others that we might not want to ignore.

rhatdan:
@nalin PTAL

rhatdan:
@giuseppe WDYT?

kpelzel:
> @nalin PTAL

Did you mean @nalind ?

rhatdan:
@nalin oops sorry, Yes @nalind PTAL

rhatdan:
@kpelzel @giuseppe @nalind I think we should have a Chat on this, and discuss what the goals of doing this will be.

@nalind sees issues with this causing images that can not be used and potentially lots of problems with this strategy.

kpelzel:
> @kpelzel @giuseppe @nalind I think we should have a Chat on this, and discuss what the goals of doing this will be.
> 
> @nalind sees issues with this causing images that can not be used and potentially lots of problems with this strategy.

That sounds like a great idea. If you don't mind I'd like to include some of my other team members in on the chat.

rhatdan:
Sounds like a great idea.  Next week is a big vacation week, could we try something on Monday/Tuesday after the 4th of July.

I will bring a few of our guys and some Product Management.

rhatdan:
OK after talking this week, I believe we came to the conclusion that when we squash the UUIDs we need to end up with a different container image name and UUID then what was pulled somehow.  Bottom line if we squash a pulled fedora we need to end up with something like fedora-squashed.

rhatdan:
@mtrmac WDYT?

mtrmac:
I guess there was some conversation and rationale that is not fully recorded?

My immediate reaction is that the basic case is “if a build call chown and it fails, the container build has failed, full stop” / “if a build process doesn’t rely on chown, down call chown in the first place”.  

Sure, someone somewhere may have a legacy image / build process where chown is called, too difficult to remove, and actually not necessary, and may benefit from an **opt-in** flag that ignores such failures, but 90% of such failures seem likely to be real errors that cause really broken containers to be built, so build should not be automatically “helpfully” silently ignoring chown failures and pretending that broken builds succeeded just fine.

I feel fairly strongly that it should never be the case that a privileged `buildah bud` and unprivileged `buildah bud` both silently succeed but produce different images, in any respect, but especially WRT ownership and permissions.

---

I don’t understand the UUID references, and what “squashing UUIDs” means.

---

Can someone write down a comprehensive a cohesive argument for why this is safe? Also including the exact description of the externally-observable behavior of this option, instead of an internal implementation-level view “ignore chown errors” (what errors? during which processes? which chown? what for?) could be useful.

Or, of course, ignore me and make a decision :)

---

In particular, what I’m worried about (as someone who knows _very little_ about c/storage):

parent:
```
ADD setuid-binary # Owned by default by container-root:container-root
RUN chown some-unprivileged-user /setuid-binary
RUN chmod u+s /setuid-binary
```

child:
```
FROM parent
RUN touch /this-does-nothing # but causes a new image to be built
```

Assuming I understand this correctly, and this change would cause unprivileged extractions of `parent` to ignore the `some-unprivileged-user` ownership if the `chown` failed, doesn’t an unprivileged `buildah bud` of `child` cause `/setuid-binary` to be setuid-`root` instead of setuid-`some-unprivileged-user` (as opposed to a privileged build, which does not change this), because the on-FS ownership differs from the in-layer-metadata ownership after the `RUN` command? If not, why not? Does it depend on the storage driver having native diff support (because it changes how the build process determines whether a FS change belongs to the parent vs. to the `RUN touch …` command)? …

Again, maybe this has all been extensively reviewed and is known to be safe — so I’m asking first instead of investigating all of this.

rhatdan:
Basically HPC (High Performance Computing) workloads want to use podman/buildah for management of containers.  But a lot of the HPC folks come from the government security world which does not allow ANY setuid applications, like newuidmap and newgidmap. 

They have a goal to use podman/buildah to pull an image from a registry and flatten the UIDs down to the single UID of the user doing the pull (Root in the user namespace).  This is what I referred to as Squashing (Of UIDS). 

If the images were thus squashed, podman could launch a container on the image.  This would allow HPC customers to participate in the OCI world.

This PR was to allow callers to ask for the Squashing of UID's not have it by default or govern it via rootless or root. 

Basically they would be looking to have podman/buildah pass in a flag that was specified by the user to squash the pulled image.

This could cause issues with users later creating a new layerd image and pushing the image to a registry, and having it based on an different image.

We are looking for ways that the users could do this as simply as possible.

Idea would be to have
`podman pull --ignoreuid fedora` 
Which would create an image called something like fedora-ignoreuid.

mtrmac:
… looking at https://github.com/containers/buildah/pull/1645 (which, currently, does not set the option added by this PR), this does seem to be opt-in; OK, that less scary.

Still…

> They have a goal to use podman/buildah to pull an image from a registry and flatten the UIDs down to the single UID of the user doing the pull (Root in the user namespace). This is what I referred to as Squashing (Of UIDS).

Have one machine with a privileged/user-namespaced builder do a
```
FROM fedora
RUN chmod -R ug-s /
RUN chown root:root /
```
to create a `hpc-fedora-base`, and build everything on that? It seems to me _much_ simpler to do this one-time step, and have a universally usable base image, than to have every single consumer set an option.

Of course, I have no idea how the build process actually works, and if there were thousands of individual `Dockerfile`s that do `yum install` and may introduce other non-root-owned files, a squash-everything option would be easier — **but that’s also exactly the case when it would be too risky**.

mtrmac:
> ```
> FROM fedora
> RUN chmod -R ug-s /
> RUN chown root:root /
> ```

(I’m not even immediately sure that the above is safe enough, with FS capabilities and what not, but buildah/podman would not be to blame :) )

mtrmac:
> > ```
> > FROM fedora
> > RUN chmod -R ug-s /
> > RUN chown root:root /
> > ```
> 
> 
> (I’m not even immediately sure that the above is safe enough, with FS capabilities and what not, but buildah/podman would not be to blame :) )

Probably better, `dnf remove` the packages owning the non-root-owned files.

mtrmac:
Random drive-by comment N+1:
> Idea would be to have
> `podman pull --ignoreuid fedora`
> Which would create an image called something like fedora-ignoreuid.

Fine, differentiating between ordinary and intentionally-explicitly-corrupt images would be helpful, but this can’t be done only at the level of images: pulling an image reuses existing individual layers using `LayersBy{Compressed,Uncompressed}Digests` (which would still be fine if `layer.Diff` returned the original metadata), _and_ whole layer chains, using `store.Layer`/`store.PutLayer` (with an ID based on the DiffIDs of all constituent layers; so, after a `pod man pull --ignoreuid fedora`, there would be a `fedora-ignoreuid` image with a $id layer chain containing squashed IDs on the FS, and a subsequent `podman pull fedora` without `--ignoreuid` would look for a layer chain with the same `$id`, find it, and reuse it, expecting a FS with non-squashed UIDs.

At that point it seems more practical to have this a per-store option (the kind of option that is recorded in the store, with podman etc. refusing to use the store with the option set to a different value), and force the user to have a separate `~/.local/share/containers` and `~/.local/share/containers-ignoreuid`.

mtrmac:
> At that point it seems more practical to have this a per-store option

(To be explicit: This would make the local image storage consistent, but it still does not make the principle of squashing safe; I continue to be suspicious about the whole idea.)

giuseppe:
I am not sure the image name is the correct way to deal with it, it is easy to tag a different name.

Could we treat the single-ID squashed image as we treat images that have a remapped UIDs/GIDs?  That already has the logic to look at other layers.  In the end this can be considered as a particular form of mapping (although not injective anymore), where every ID is remapped to the same user.

I'd leave the users the choice to enable this mode if they wish to, as long as it is enabled explicitly and not the default.  I don't see any big risk involved with this mechanism since the files in the end will be owned by root.

mtrmac:
> Could we treat the single-ID squashed image as we treat images that have a remapped UIDs/GIDs? That already has the logic to look at other layers. In the end this can be considered as a particular form of mapping (although not injective anymore), where every ID is remapped to the same user.

Intriguing, that could be a good way to implement this without affecting too much else.
 
> I'd leave the users the choice to enable this mode if they wish to, as long as it is enabled explicitly and not the default. I don't see any big risk involved with this mechanism since the files in the end will be owned by root.

It’s not necessarily a risk for the _host_ (although… with everything running as the users’ UID, is there enough separation?), but it may easily completely break the security design of the content _inside_ the container (e.g. the way postfix isolates its components with different groups).

mtrmac:
> > Could we treat the single-ID squashed image as we treat images that have a remapped UIDs/GIDs? That already has the logic to look at other layers. In the end this can be considered as a particular form of mapping (although not injective anymore), where every ID is remapped to the same user.
> 
> Intriguing, that could be a good way to implement this without affecting too much else.

… On second thought, I don’t think this would work; we would have to first create the FS layers with all the different UIDs before adding the top remapping layer, and the point is not to use the different UIDs at all, if I understand this correctly.

PR #361, opened 2019-06-06
Merge the disparate parts of all of the Unix-like lockfiles

nalind:
The lockfile implementation was mostly shared across Unix-like operating systems, except for a couple of methods that differed.  The system package already provided abstractions for most of them, so we add a
missing piece and use it to finish merging them.


rhatdan:
LGTM
@TomSweeneyRedHat @giuseppe @vrothberg PTAL

PR #362, opened 2019-06-07
Add function to parse out mount options from graphdriver

rhatdan:
We have a bug in podman that uses the defaultGraphDriver options
for returning the MountOptions rather then the driver overrides
from the user.

This PR adds a new interface GetMountOptions which parses the callers
graphdriveroptions and return the mountoptions


mheon:
Change LGTM on my end

nalind:
Changes look fine.

PR #363, opened 2019-06-07
compression: add support for the zstd algorithm

giuseppe:
zstd is a compression algorithm that has a very fast decoder, while
providing also good compression ratios.  The fast decoder makes it
suitable for container images, as decompressing the tarballs is a very
expensive operation.  This is a first step at supporting zstd as we We
don't yet generate zstd layers.

In my testing, copying the Fedora image from a local dir: repository,
the wall clock time passed from ~8s with gzip compression to ~4.5s
with zstd.

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -98,6 +98,8 @@ const (
 	Gzip
 	// Xz is xz compression algorithm.
 	Xz
+	// Zstd is zstd compression algorithm.
+	Zstd
 )
 
 const (
@@ -141,6 +143,7 @@ func DetectCompression(source []byte) Compression {
 		Bzip2: {0x42, 0x5A, 0x68},
 		Gzip:  {0x1F, 0x8B, 0x08},
 		Xz:    {0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00},
+		Zstd:  {0x28, 0xb5, 0x2f, 0xfd},
 	} {
 		if len(source) < len(m) {
 			logrus.Debug("Len too short")
@@ -200,6 +203,8 @@ func DecompressStream(archive io.Reader) (io.ReadCloser, error) {
 			<-chdone
 			return readBufWrapper.Close()
 		}), nil
+	case Zstd:
+		return zstdReader(buf)
 	default:
 		return nil, fmt.Errorf("Unsupported compression format %s", (&compression).Extension())
 	}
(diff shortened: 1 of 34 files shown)


giuseppe:
PR for c/image: https://github.com/containers/image/pull/639

pkg/archive/archive.go
```diff
@@ -217,6 +223,10 @@ func CompressStream(dest io.Writer, compression Compression) (io.WriteCloser, er
 		gzWriter := gzip.NewWriter(dest)
 		writeBufWrapper := p.NewWriteCloserWrapper(buf, gzWriter)
 		return writeBufWrapper, nil
+	case Zstd:
+		zStdWriter := gzip.NewWriter(dest)
+		writeBufWrapper := p.NewWriteCloserWrapper(buf, zStdWriter)
+		return writeBufWrapper, nil
```
nalind: Shouldn't this be returning a `*zstd.Writer`?

nalind:
Modifying `pkg/archive/archive.go` requires a second commit that updates `archive_ffjson.go`.
The new dependency also needs to be added to our top-level `vendor.conf`.

pkg/archive/archive.go
giuseppe: ouch yes sure, will fix it

giuseppe:
> Modifying `pkg/archive/archive.go` requires a second commit that updates `archive_ffjson.go`.
> The new dependency also needs to be added to our top-level `vendor.conf`.

I've added the dependency, the generated `archive_ffjson.go` doesn't change

giuseppe:
ready for review

pkg/archive/archive_cgo.go
```diff
@@ -0,0 +1,18 @@
...
+package archive
+
+import (
+	"bufio"
+	"io"
+
+	"github.com/DataDog/zstd"
+)
+
+func zstdReader(buf *bufio.Reader) (io.ReadCloser, error) {
```
nalind: Why does this expect a `*bufio.Reader` instead of an `io.Reader`?

pkg/archive/archive_nocgo.go
```diff
@@ -0,0 +1,17 @@
...
+
+package archive
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+)
+
+func zstdReader(buf *bufio.Reader) (io.ReadCloser, error) {
```
nalind: Why does this expect a `*bufio.Reader` instead of an `io.Reader`?

pkg/archive/archive_cgo.go
giuseppe: I had left that as with unbuffered input I see very bad compression results.

But I guess it is better to be more generic, I've changed it to `io.Reader`.

nalind:
LGTM

rhatdan:
LGTM

PR #366, opened 2019-06-12
Preserve image name on matching images

saschagrunert:
If two image names (including their tags) match, then we usually remove the name completely. This means in the end that we have no chance to tell the user what the original image name was. We now strip the tag away and add a `<none>`, which could serve as generic indicator that no tag is available for that image.

I tested it with the latest CRI-O master and this applied patch, we now have the chance to preserve the image name as follows:
```
> sudo crictl images
IMAGE                 TAG                 IMAGE ID            SIZE
localhost:5000/test   <none>              9fbc3a97f1aec       2.07kB
localhost:5000/test   latest              776f147227fde       3.76kB
```

Before it was:
```
> sudo crictl images
IMAGE                 TAG                 IMAGE ID            SIZE
<none>                <none>              9fbc3a97f1aec       2.07kB
localhost:5000/test   latest              776f147227fde       3.76kB
```


rhatdan:
Could you add a test for this?
@mtrmac @nalind PTAL

rhatdan:
@giuseppe PTAL

saschagrunert:
> Could you add a test for this?

Yes sure, I will add unit and integration tests if we decide to implement it in the proposed way.

vrothberg:
I am interested in the motivation behind the change. Can you elaborate a
bit on that?

saschagrunert:
> I am interested in the motivation behind the change. Can you elaborate a bit on that?

Yes sure: A user has mentioned that containerd seems to preserve that kind of information and is therefore able to provide more details about available images on a node (see https://github.com/cri-o/cri-o/pull/2455#discussion_r291808887). So my main intention was to preserve the image name when a tag moved away to another image layer.

Beside this, there is a bug in crictl which does not allow to show the image name when pulling via digest, but this is another story and addressed in https://github.com/kubernetes-sigs/cri-tools/pull/471.

In the end the whole story is a bit mixed up, I guess that's because of 'historical reasons'. Kubernetes expects (currently) always some sort of image name for the repo tag and digest, whereas I discovered the following possible combinations in the wild:

- `<none>:<none>`/`<none>@<none>`: No image information available beside the ID (originally from the moby project and now addressed in https://github.com/cri-o/cri-o/pull/2455)
- `name:<none>` No image tag available but the name (addressed within this PR)
- `<none>:<none>`/`image@sha256:...` Use image name from digest if no tag available (addressed in https://github.com/kubernetes-sigs/cri-tools/pull/471)
- `image:tag` Tag and image name available, works as before

store.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/containers/storage/pkg/parsers"
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/containers/storage/pkg/stringutils"
+	"github.com/docker/distribution/reference"
```
mtrmac: We use `c/image/docker/reference` throughout (IIRC for somewhat historical reasons, the upstream package went through a painful API break at some point).

store.go
```diff
@@ -3187,9 +3188,15 @@ func stringSliceWithoutValue(slice []string, value string) []string {
 	modified := make([]string, 0, len(slice))
 	for _, v := range slice {
 		if v == value {
-			continue
+			ref, err := reference.ParseNamed(v)
+			if err != nil {
+				continue
+			}
+			name := fmt.Sprintf("%s:<none>", reference.TrimNamed(ref).String())
```
mtrmac: `<none>` is not a valid tag (as is already visible from the fact that this needs to use a `Sprintf` instead of `reference.WithTag`).

While it’s not, IIRC, documented explicitly, c/storage names must _all_ be parseable using `ParseNormalizedNamed` (tools like podman just fail when they see a non-compliant name).

store.go
```diff
@@ -3187,9 +3188,15 @@ func stringSliceWithoutValue(slice []string, value string) []string {
 	modified := make([]string, 0, len(slice))
```
mtrmac: Doing a non-trivial edit like this in a function innocuously named `stringSliceWithoutValue` is horribly confusing and almost certain to trip some future caller up. If this needs to exist at all, rename it to be clear about what the function does.

saschagrunert:
I see, I will check if this topic can be addressed in a different way. Thank you for the input.

PR #369, opened 2019-06-19
overlay: cache the results of feature tests

### LINKED ISSUE #368: Improve overlay startup performance.
On every run of tools using container/storage overlay driver, we are running supportsOverlay function which does a whole bunch of processes in mounting of file systems.  Once this is done once we should not have to do it again until the next time the system boots.
       
Basically have driver create

/run/containers/storage/overlay/supportsDtype

/run/containers/storage/overlay/usingMetacopy

And then just do a gettattr on these files, if they exists, no reason to do the check.

Save each podman/buildah/cri-o/skopeo command from having to do them, after the first one in a boot.

nalind:
Instead of passing the driver-specific directory and assorted fields from a `Config` struct to lower-level drivers when we initialize them, pass them the directory and the `Config` struct.  Add the `runroot` to the `Config` struct.  Have the overlay driver use cache files under the runroot to track whether or not overlay is supported, whether or not metacopy is on (for a specific set of mount options), and whether or not we can use native diff (again, for a specific set of mount options).  This attempts to fix #368.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -97,6 +97,7 @@ type overlayOptions struct {
 type Driver struct {
 	name          string
 	home          string
+	runhome       string
 	uidMaps       []idtools.IDMap
 	gidMaps       []idtools.IDMap
 	ctr           *graphdriver.RefCounter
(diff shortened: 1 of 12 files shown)


drivers/overlay/overlay.go
```diff
@@ -157,32 +158,71 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
...
+		} else {
+			supportsDType, err = supportsOverlay(home, fsMagic, rootUID, rootGID)
+			if err != nil {
+				os.Remove(filepath.Join(home, linkDir))
+				os.Remove(home)
+				patherr, ok := err.(*os.PathError)
+				if ok && patherr.Err == syscall.ENOSPC {
+					return nil, err
+				}
+				if err2 := cachedFeatureRecord(runhome, feature, false, fmt.Sprintf("kernel does not support overlay fs: %v", err)); err2 != nil {
```
rhatdan: This should be set to a variable so that every time you run command it outputs same error.

fmt.Sprintf("kernel does not support overlay fs: %v", err))

rhatdan:
Great job, I really like this patch.  One small NIT.
@giuseppe @vrothberg @TomSweeneyRedHat @baude PTAL

baude:
initial feedback is this really helped ... it seems to take c/s out of the performance profiles all together.  Excellent!

drivers/overlay/overlay.go
nalind: This error should be fixed now.  The error reporting when the driver determines that the kernel or underlying filesystem won't work is messier, but I'd like to tackle that later.

rhatdan:
LGTM

PR #372, opened 2019-06-23
utils: fix check for missing conf file

giuseppe:
the previous error condition was causing an early failure when the
configuration file existed.  It prevented some additional checks and
to correctly set the RunRoot and GraphRoot to their default values
when they are not overriden in the configuration file.

We have not noticed it earlier as Podman would still use the
configuration from the DB in this case, so the issue is visible with
Podman only when the RunRoot is not specified in the configuration
file and in the libpod DB as well.

Closes: https://github.com/containers/libpod/issues/3274


rhatdan:
LGTM

rhatdan:

PR #375, opened 2019-06-25
quota: add build tag to disable it

giuseppe:
change backported from Moby.

### DIFF
--- a/drivers/quota/projectquota.go
+++ b/drivers/quota/projectquota.go
@@ -1,4 +1,4 @@
-// +build linux
+// +build linux,!exclude_disk_quota
 
 //
 // projectquota.go - implements XFS project quota controls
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

giuseppe:
the CI failure doesn't seem related to the PR

giuseppe:
@nalind PTAL

nalind:
Looks reasonable.  If we're not going to export the newly-added error for direct comparison (it would have to be moved to a place where it's unconditionally declared), then it might as well be allocated at the sites where it's being returned, so that a backtrace ends at the right place.  It's okay to merge in its current form, though.

giuseppe:
> Looks reasonable. If we're not going to export the newly-added error for direct comparison (it would have to be moved to a place where it's unconditionally declared), then it might as well be allocated at the sites where it's being returned, so that a backtrace ends at the right place. It's okay to merge in its current form, though.

thanks, I've updated the PR and now the error is allocated where it is returned

vrothberg:
Restarted the flaked job. Good to merge once we're green :-)

PR #376, opened 2019-06-29
Add support for UID Matching in /etc/subuid and /etc/subgid

rhatdan:
Administrators can add records to /etc/subuid and /etc/subgid based
on username or UID.  Currently we only support lookup based on username.

This patch will lookup based on UID, if it is available in the /etc/passwd
file on the host.

### DIFF
--- a/pkg/idtools/idtools.go
+++ b/pkg/idtools/idtools.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"os/user"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


rhatdan:
This should help fix https://github.com/containers/buildah/issues/1699

rhatdan:
@nalind @TomSweeneyRedHat @vrothberg @giuseppe PTAL

rhatdan:
@lukeyeager PTAL

giuseppe:
Do newuidmap and newgid support lookup by UID?  Last time I have checked, it didn't work

TomSweeneyRedHat:
bot, retest this please

TomSweeneyRedHat:
LGTM assuming happy tests.

rhatdan:
@giuseppe added this patch to buildah and modified /etc/subuid to use my UID instead of my username and it worked fine.

giuseppe:
> @giuseppe added this patch to buildah and modified /etc/subuid to use my UID instead of my username and it worked fine.

That is great.  We should do the same thing in libpod

giuseppe:
LGTM

PR #377, opened 2019-06-29
Remove unfinished VFS layer if its creation fails.

QazerLab:
Current implementation of VFS driver performs the copy of the parent layer during the creation of new (child) layer, but does not clean the destination if the copy process fails.

From the consuming application perspective, this make storage leaks possible, e.g. an attempt to create new container with `podman`, if failed in the middle of the copy due to disk overflow, leaves unfinished container layer stray.

To avoid obliging the application (or the end user) to take care of such possible leaks, cleanup is added.

The exact scenario of reproducing the leak with current implementation is simple (and in my case the leak was fast, as my disk is almost always close to full):

1. Fill almost all disk (e.g. with `dd`).
2. Try to start new container using the image whose size larger than the remaining disk space. Observe the failure.
3. Check, that disk is now completely full.
4. Check, that you can not simply reclaim the space used by the unfinished layer with `podman`:
* `podman ps` does not know about unfinished container, so you can not delete it with `podman rm`;
* `podman system prune -a` cleans everything... But the leaked layer.
5. Free some space, retry. If freed space is not enough, it again will leak.

### DIFF
--- a/drivers/vfs/driver.go
+++ b/drivers/vfs/driver.go
@@ -118,7 +118,7 @@ func (d *Driver) Create(id, parent string, opts *graphdriver.CreateOpts) error {
 	return d.create(id, parent, opts, true)
 }
 
-func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, ro bool) error {
+func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, ro bool) (retErr error) {
 	if opts != nil && len(opts.StorageOpt) != 0 {
 		return fmt.Errorf("--storage-opt is not supported for vfs")
 	}
@@ -133,6 +133,13 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, ro bool
 	if err := idtools.MkdirAllAndChown(filepath.Dir(dir), 0700, rootIDs); err != nil {
 		return err
 	}
+
+	defer func() {
+		if retErr != nil {
+			os.RemoveAll(dir)
+		}
+	}()
+
 	if parent != "" {
 		st, err := system.Stat(d.dir(parent))
 		if err != nil {


rhatdan:
LGTM.
My only concern is that could the directory have preexisted and had content in it?

Should we do similar for Overlay driver?  Do we know that we created a new Upper directory and ran out of space?

QazerLab:
@rhatdan,

In my understanding, if the directory for specific ID existed before a call to `create()` with this ID, it could not be for the writable layer (as their IDs are generated randomly on container creation and checked for collisions with existing IDs). On the other hand, if this directory was for RO layer, it either already contained the complete layer (and thus should not be attempted to be `create()`d again), or it contained an incomplete layer, which probably is a bug.

As to overlay, it seems like it already has [similar cleanup](https://github.com/containers/storage/blob/8eed0c36d1e3dcc684ebbcbc703c3a1f092a71aa/drivers/overlay/overlay.go#L594) and it works well:

```
-bash-5.0# podman run -it alpine
Error: error creating container storage: error creating read-write layer with ID "0ff6ef8817f570a84bac288e6fe319354d8a0319e805885557e7ce505015c8ff": mkdir /var/lib/containers/storage/overlay/0ff6ef8817f570a84bac288e6fe319354d8a0319e805885557e7ce505015c8ff/diff: no space left on device
-bash-5.0# cd: /var/lib/containers/storage/overlay/0ff6ef8817f570a84bac288e6fe319354d8a0319e805885557e7ce505015c8ff: No such file or directory
```

QazerLab:
@rhatdan, could you please re-trigger the failed `testing image:ubuntu` task? Seems like there was a transient setup issue.

nalind:
LGTM

rhatdan:
@cevich Any idea what is going on here.  We do not seem to be getting Ubuntu to work on container/storage.

cevich:
Looking.

cevich:
@rhatdan I see something similar happened recently on master as well: https://cirrus-ci.com/task/6215525896552448

I discovered this the other day in the Ubuntu 18 image we use in libpod land.  Canonical (or google) in all their wisdom, set up out-of-the-box daily package updates and other [possibly problematic periodic activities we generally don't want in VM images ](https://github.com/containers/libpod/pull/3548) for testing.

Here specifically, I'd guess it's something like `apt-daily-upgrade` or `apt-daily` systemd services that are getting in the way :confounded:  Exit code 124 comes from `timeout`, and since it's already retried several times, I'd speculate there may be some kind of cloud-networking challenges compounding the situation.

Let me get the fix working in libpod for Ubuntu 18, then I'll open something similar for c/storage and c/buildah.  Thanks for letting me know.

rhatdan:
Ok I will manually merge, since I am sure this is fine.

PR #378, opened 2019-07-01
WIP - pkg/flock package

vrothberg:
Introduce the pkg/flock package, a flock(2)-based implementation in
golang that can be used within and across process spaces. The main
motivation for this package is to benefit from the TryLock-semantics
offered by flock(2), which is not possible with c/storage.Locker file
lock since they are based on fcntl(2).

### DIFF
--- a/pkg/flock/flock.go
+++ b/pkg/flock/flock.go
@@ -1,0 +1,63 @@
+// flock exposes an API around flock(2) but allows for synchronizing within and
+// across process spaces, making it suitable for currently running threads and
+// processes alike. To support long-running daemons, the locks' file descriptors
+// are closed when the lock is being released via Unlock() or RUnlock().
+//
+// Note that the flock package supports UNIX-only but compiles on Windows for
+// compatibility reasons.
+
+package flock
+
+import (
+	"sync"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sync/semaphore"
+)
+
+var (
+	errInternal       = errors.New("internal error: flock")
+	errNotImplemented = errors.New("internal error: flock: not implemented")
+)
+
+// Flock implements a flock(2)-based file lock. It can be used to synchronize
+// within the same process space and across process spaces.
+type Flock interface {
+	// Lock locks the exclusive Flock for writing.
+	Lock() error
+	// TryLock locks the exclusive Flock without blocking. The first return
+	// value indicates if the Flock has been acquired.
+	TryLock() (bool, error)
+	// RLock locks the shared Flock for reading.  Note that RLock is busy
+	// looping for acquiring the Flock.
+	RLock() error
+	// Unlock unlocks the Flock. If it's a shared Flock, unlock must be called
+	// by the number of RLock() calls for the underlying lock file to be
+	// unlocked.
+	Unlock() error
+	// Locked indicates if the Flock is locked.
+	Locked() bool
+}
+
+// flockType indicates the type of the Flock.
+type flockType int
+
+const (
+	// flockUnlocked indicates an unlock flock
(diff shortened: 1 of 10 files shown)


pkg/flock/flock_windows.go
```diff
@@ -0,0 +1,27 @@
+// build unix
```
giuseppe: should this be dropped?

pkg/flock/flock_unix.go
```diff
@@ -0,0 +1,176 @@
...
+	return flock{
+		flockType: flockUnlocked,
+		fd:        -1,
+		path:      path,
+		mutex:     new(sync.Mutex),
+		sem:       semaphore.NewWeighted(1),
+	}, nil
+}
+
+// open opens the flock's path and sets the file desriptor. State must be
```
giuseppe: typo in `desriptor`

pkg/flock/flock_unix.go
```diff
@@ -0,0 +1,176 @@
...
+// synchronized by the caller.
+func (f *flock) open() error {
+	if f.fd != -1 || f.flockType != flockUnlocked {
+		return fmt.Errorf("%v: trying to open improper Flock %q %d %d", errInternal, f.path, f.fd, f.flockType)
```
giuseppe: please use `errors.Wrapf`

vrothberg: Initially I had that but the problem is that Wrapf mixes the order. I want to have a "internal error ..." but using Wrapf would move that to the end.

pkg/flock/flock_windows.go
vrothberg: Oh yes, I should. Thanks!

vrothberg:
Marking it as WIP until there's a user of this API, presumably the copy-blob-detection in c/image.

pkg/flock/flock_unix.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/pkg/errors"
+	"golang.org/x/sync/semaphore"
+	"golang.org/x/sys/unix"
+)
+
+// New returns a Flock for path.
+func New(path string) (flock, error) {
+	// Make sure we can operate on the path.
+	fd, err := unix.Open(path, os.O_RDWR|os.O_CREATE, unix.S_IRUSR|unix.S_IWUSR)
```
rhatdan: Should we use an unix.Access check for this rather then opening and closing?  Would that be faster? And then only create the file if it does not exists.

rhatdan:
@nalind WDYT?

pkg/flock/flock.go
```diff
@@ -0,0 +1,63 @@
+// flock exposes an API around flock(2) but allows for synchronizing within and
+// across process spaces, making it suitable for currently running threads and
```
vrothberg: s/currently/concurrently/

rhatdan:
@vrothberg Makes sense, since I envisioned @nalind Blocking this if it was not used in containers/storage.

pkg/flock/flock_unix.go
nalind: Perhaps `errors.WithStack()`?

rhatdan:
@vrothberg Lets talk about this in the next meeting.

vrothberg:
Ping. @rhatdan, let's check if we still/really need this PR.

rhatdan:
@vrothberg Needs a rebase.  New conversations with @saschagrunert might get us a user for this, yet.

vrothberg:
@rhatdan, I prefer to wait until we're certain there will be a user.

rhatdan:
@vrothberg Since this has been around for 2+ years, can we close it?

PR #380, opened 2019-07-01
lockfile: move into pkg/lockfile

nalind:
This PR moves the lockfile implementation into a subpackage so that we can use it in lower-level drivers without creating import loops.

It updates us to ensure that we only try to take read-locks on read-only stores, and adds a panic when we try to take a write lock on a read-only item.  It also removes an unnecessary unlock before closing a file, which already releases the lock.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -579,6 +579,7 @@ func (r *containerStore) RecursiveLock() {
 func (r *containerStore) RLock() {
 	r.lockfile.RLock()
 }
+
 func (r *containerStore) Unlock() {
 	r.lockfile.Unlock()
 }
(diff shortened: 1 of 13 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
bot, retest this please

TomSweeneyRedHat:
Changes LGTM, ubuntu tests are failing to what looks to be a network flake.

nalind:
Looking for another review of the latest addition, which tweaks `contrib/cirrus/setup.sh`.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

nalind:
I don't think we've hooked up homu to this repository, so it's a manual merge process.

TomSweeneyRedHat:
Oops, and it looks like I don't have the juice.  Me thinks we're waiting on @rhatdan or @vrothberg

PR #382, opened 2019-07-08
Switch to go modules

saschagrunert:
Remove the old vendor approach in replacement of go module support. Outdated dependencies have been updated as well. Except docker, where the API changed.


rhatdan:
@nalind PTAL

nalind:
This looks like it pulls in an older `runc` than we were using previously.  Not that the version we had been vendoring was recent, either, but still.  Otherwise LGTM.

rhatdan:
Lets vendor in the latest runc available.

saschagrunert:
> Lets vendor in the latest runc available.

Yes, I vendored in runc 1.0.0-rc8. This was mainly caused because go is not able to "detect" it as latest version correctly.

rhatdan:
LGTM

PR #389, opened 2019-07-16
Keep track of the UIDs and GIDs used in applied layers

nalind:
Add a field to the `Layer` structure that lets us make note of the set of UIDs and GIDs which own files in the layer, populated by scanning the diff that we used to populate the layer, if there was one.  This could eventually be used to decide how many UIDs and GIDs we need to map into a container in order to start a container based on a given image.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -1,15 +1,18 @@
 package storage
 
 import (
+	"archive/tar"
 	"bytes"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
@vrothberg @giuseppe @TomSweeneyRedHat PTAL

giuseppe:
LGTM

PR #390, opened 2019-07-17
Remove merged directory when it is unmounted

rhatdan:
This should cause the kernel to clear any leaked mount points into other
containers.

Also create the merged directory every time you attempt to create it.


rhatdan:
I believe @rhvgoyal suggests this as a fix for https://github.com/containers/libpod/issues/3540

rhvgoyal:
This sounds like the right thing to do to help with unintentional leaking of container rootfs mount point.

PR #392, opened 2019-07-22
zstd: use github.com/klauspost/compress directly

vrothberg:
The DataDog/zstd package added more than 2MB to the binary size.
Reduce that by using the github.com/klauspost/compress directly
from 11MB to 8.7MB.


vrothberg:
Don't merge. I think we don't need cgo anymore.

vrothberg:
It squeezes the same amount from Podman :+1: 

@baude @mheon FYI

vrothberg:
@giuseppe @nalind PTAL

giuseppe:
LGTM

rhatdan:
LGTM

vrothberg:
@rhatdan, can we cut a new release to revendor it across the stack? Codename: slimfast.

PR #394, opened 2019-07-23
Fix cross-compilation

nalind:
The zstd package no longer requires cgo, but we forgot to remove the guards in the archive package that let us build without it.

Should resolve the problem we're seeing at https://github.com/containers/buildah/pull/1739#issuecomment-514300697.


nalind:
@vrothberg PTAL

PR #395, opened 2019-07-24
do not merge - Fix cross-compilation

vrothberg:
The zstd package no longer requires cgo, but we forgot to remove the
guards in the archive package that let us build without it.

Signed-off-by: Nalin Dahyabhai <nalin@redhat.com>


vrothberg:
```[+0068s] gcc: error: unrecognized command line option '-marm'; did you mean '-mabm'?```

@nalind, compiling arm requires setting the CC to the cross compiler. Can we split the `cgo` fix and the cross compiling changes?

PR #398, opened 2019-07-25
overlay: fix small piece of repeated work

haircommander:
we compute d.dir(id) twice, but store the value the first time. use the cached value instead.

Signed-off-by: Peter Hunt <pehunt@redhat.com>


TomSweeneyRedHat:
LGTM
Nice spot and pickup @haircommander .  I love fixes like this.

PR #399, opened 2019-07-26
loadMounts(): reset counts before merging just-loaded data

nalind:
Before applying mount counts that we've just loaded, reset all of the counts.

If a layer that we thought was mounted was unmounted by another process, there won't be a record of it in the mounts list any more, so we wouldn't reset the mount count on our record for that layer.

This may be related to https://github.com/cri-o/cri-o/issues/2584.


rhatdan:
LGTM
@vrothberg @mtrmac @TomSweeneyRedHat @mrunalp PTAL

mrunalp:
:+1:

mtrmac:
LGTM (though, in https://github.com/cri-o/cri-o/issues/2584 , is there another process?)

mrunalp:
Miloslav, not sure what you mean by another process? We saw crio looping even when kubelet was stopped once it hit the issue.

mrunalp:
If you meant another process accessing storage and possibly holding locks that is unlikely.

layers.go
```diff
@@ -387,6 +387,10 @@ func (r *layerStore) loadMounts() error {
 	}
 	layerMounts := []layerMountPoint{}
 	if err = json.Unmarshal(data, &layerMounts); len(data) == 0 || err == nil {
+		for _, layer := range r.layers {
```
vrothberg: Can we add a comment here explaining why it's safe/correct to reset the counter?

mtrmac: … why it is necessary. Yes, please.

mtrmac:
> Miloslav, not sure what you mean by another process? We saw crio looping even when kubelet was stopped once it hit the issue.

AFAICS this could affect the code path in https://github.com/cri-o/cri-o/issues/2584 only if, while CRI-O is unmounting layers, _some other_ process unmounted the layer concurrently to CRI-O. (But I still can’t see that it could use an infinite loop in that case, although I haven’t read the code nearly attentively enough—I haven’t noticed the problem fixed by this PR.)

mrunalp:
As an aside, I think that we should move to stable branch for CRI-O 1.14 like we did for 1.13. That way we don't regress by unrelated changes and only pull in what we need for CRI-O.

mrunalp:
We are at `v1.12.10` in CRI-O `release-1.14` right now.

layers.go
nalind: Added.

nalind:
> LGTM (though, in [cri-o/cri-o#2584](https://github.com/cri-o/cri-o/issues/2584) , is there another process?)

I didn't think so.  I waffled on emphasizing the "may" in the initial comment for exactly that reason.

Rebased and squashed.

rhatdan:
Will merge once this passes tests and cut a new release.

PR #401, opened 2019-07-31
[CRI-O 1.14] loadMounts(): reset counts before merging just-loaded data

haircommander:
Before applying mount counts that we've just loaded, reset all of the
counts.

If a layer that we thought was mounted was unmounted by another process,
there won't be a record of it in the mounts list any more, so we
wouldn't reset the mount count on our record for that layer.

Signed-off-by: Nalin Dahyabhai <nalin@redhat.com>

pick of: https://github.com/containers/storage/pull/399
per: https://github.com/containers/storage/pull/399#issuecomment-516153182

PR #403, opened 2019-08-01
[cri-o-release-1.14] loadMounts(): reset counts before merging just-loaded data

nalind:
Before applying mount counts that we've just loaded, reset all of the counts.

If a layer that we thought was mounted was unmounted by another process, there won't be a record of it in the mounts list any more, so we wouldn't reset the mount count on our record for that layer.

Cherry-picked from #399.


nalind:
Closing in favor of #401

PR #404, opened 2019-08-02
Don't chown unless you have to

rhatdan:
When running with readonly additional stores, containers/storage is trying to
create directories that already exists with the correct UIDs. This causes
buildah/podman to throw an error when committing an image.

This fix will allow buildah/podman to work with readonly additional image stores.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -839,6 +839,16 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	if _, err := os.Stat(dir); err != nil {
 		return "", err
 	}
+	readWrite := true
+	// fuse-overlayfs doesn't support working without an upperdir.
+	if d.options.mountProgram == "" {
+		for _, o := range options.Options {
+			if o == "ro" {
+				readWrite = false
+				break
+			}
+		}
+	}
 
 	lowers, err := ioutil.ReadFile(path.Join(dir, lowerFile))
 	if err != nil && !os.IsNotExist(err) {
@@ -910,20 +920,31 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 
 	// If the lowers list is still empty, use an empty lower so that we can still force an
 	// SELinux context for the mount.
+
+	// if we are doing a readOnly mount, and there is only one lower
+	// We should just return the lower directory, no reason to mount.
+	if !readWrite {
+		if len(absLowers) == 0 {
+			return path.Join(dir, "empty"), nil
+		}
+		if len(absLowers) == 1 {
+			return absLowers[0], nil
+		}
+	}
 	if len(absLowers) == 0 {
 		absLowers = append(absLowers, path.Join(dir, "empty"))
 		relLowers = append(relLowers, path.Join(id, "empty"))
 	}
-
 	// user namespace requires this to move a directory from lower to upper.
 	rootUID, rootGID, err := idtools.GetRootUIDGID(d.uidMaps, d.gidMaps)
 	if err != nil {
 		return "", err
 	}
-
 	diffDir := path.Join(dir, "diff")
-	if err := idtools.MkdirAs(diffDir, 0755, rootUID, rootGID); err != nil && !os.IsExist(err) {
-		return "", err
+	if readWrite {
+		if err := idtools.MkdirAllAs(diffDir, 0755, rootUID, rootGID); err != nil && !os.IsExist(err) {
+			return "", err
+		}
 	}
 
 	mergedDir := path.Join(dir, "merged")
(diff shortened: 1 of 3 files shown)


rhatdan:

nalind:
What's the problem we're trying to avoid with the new logic here?

drivers/overlay/overlay.go
```diff
@@ -995,9 +995,13 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
...
-	// chown "workdir/work" to the remapped root UID/GID. Overlay fs inside a
-	if err := os.Chown(path.Join(workDir, "work"), rootUID, rootGID); err != nil {
-		return "", err
+	workPath := path.Join(workDir, "work")
+	stat, err := system.Stat(workPath)
+	if err != nil || stat.UID() != uint32(rootUID) || stat.GID() != uint32(rootGID) {
+		// chown "workdir/work" to the remapped root UID/GID. Overlay fs inside a
+		if err := os.Chown(workPath, rootUID, rootGID); err != nil {
+			return "", err
+		}
```
nalind: This looks like a good place to replace `os.Chown()` with a call to the updated `idtools.SafeChown()`.

rhatdan:
I setup some additional stores and then I mount them read/only.
Do:
```
# rm -rf /var/lib/containers/storage
# buildah pull fedora
# mv /var/lib/containers/storage /var/lib/shared
# mount -o ro --bind /var/lib/shared /var/lib/shared
```
Now modify mounts.conf to use additionalstores as /var/lib/shared
```
# ctr=$(buildah from fedora)
# buildah commit $ctr fedora1
ERRO[0000] error unmounting /var/lib/shared/overlay/18bc230da79c23576d7c3a31a3725ae5f133899adb0f54a98654e0ec880183e0/merged: invalid argument 
error committing container "fedora-working-container" to "fedora1": error copying layers and metadata for container "635323ed1884d98c933866525f71838f3b1d26a980336eaef5d96015632bd4f7": Error initializing source containers-storage:fedora-working-container: error extracting layer "3e9dbe1d66c63d50e17635899dc4cf598d934e86705bed51d5c558a84ac5617c": error creating overlay mount to /var/lib/shared/overlay/18bc230da79c23576d7c3a31a3725ae5f133899adb0f54a98654e0ec880183e0/merged: read-only file system
```
With this change read/only shares will just work as long as the content exists on the upper layer with the right UID, even if the UIDS are mounted read/only.
```

**Comment by nalind** [comment]
That looks like an error from a `Mkdir` for a "merged" directory, while this patch tries to skip a `Chown` on the "work" directory.  Are they the same problem?

**Comment by rhatdan** [comment]
@nalind Once I fixed the safechown calls the other workdir blew up.

When you commit the image, the driver is using the naiveDiff Driver, which mounts the parent directory which is from the ReadOnly store and this hits the two mkdiras calls even though they are not needed.  Since the file system is readonly, the calls fails, and buildah is not able to create the image.  If we check before attempting to create that the directories exists, and they have the correct ownership, then the chown call does not happen and does not fail, buildah can commit.

File: drivers/overlay/overlay.go (hunk 1 above)
Comment (giuseppe) [reply drivers/overlay/overlay.go]: I think it would be faster if you try the `SafeChown` in any case then check whether it was needed only on error:
```
if err := SafeChown(...); err != nil {
    stat, err := system.Stat(workPath)
    ... raise the error only if uid!=stat.uid || gid!=stat.gid, otherwise ignore it
}
```

Comment (rhatdan) [reply drivers/overlay/overlay.go]: Why, I would figure it would be much more often the case where the file already exists with root, root and the chown takes more time.

**Comment by rhatdan** [comment]
Ok I found more problems with metacopyup=true and naivediff.

File: drivers/overlay/overlay.go (hunk 1 above)
Comment (rhatdan) [reply drivers/overlay/overlay.go]: @nalind Changes to use SafeChown for overlay.

**Comment by rhatdan** [comment]
@rhvgoyal FYI

File: drivers/overlay/overlay.go (hunk 1 above)
Comment (giuseppe) [reply drivers/overlay/overlay.go]: > Why, I would figure it would be much more often the case where the file already exists with root, root and the chown takes more time.

ok then in that case it is fine.  I assumed we were chown'ing most of the time

File: drivers/overlay/overlay.go [hunk 2]
```diff
@@ -839,6 +839,13 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	if _, err := os.Stat(dir); err != nil {
 		return "", err
 	}
+	readWrite := true
```
giuseppe: works if I do:

```
diff --git a/drivers/overlay/overlay.go b/drivers/overlay/overlay.go
index 754d63ad0..404102376 100644
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -840,10 +840,14 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
                return "", err
        }
        readWrite := true
-       for _, o := range options.Options {
-               if o == "ro" {
-                       readWrite = false
-                       break
+
+       // fuse-overlayfs doesn't support working without an upperdir.
+       if d.options.mountProgram == "" {
+               for _, o := range options.Options {
+                       if o == "ro" {
+                               readWrite = false
+                               break
+                       }
                }
        }
 ```

File: drivers/overlay/overlay.go [hunk 3]
```diff
@@ -839,6 +839,16 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
...
+	// fuse-overlayfs doesn't support working without an upperdir.
```
nalind: Not a blocker: we should add a note so that we don't forget to come back and remove this workaround later.

drivers/overlay/overlay.go
```diff
@@ -910,20 +920,31 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
...
 	rootUID, rootGID, err := idtools.GetRootUIDGID(d.uidMaps, d.gidMaps)
 	if err != nil {
 		return "", err
 	}
-
 	diffDir := path.Join(dir, "diff")
-	if err := idtools.MkdirAs(diffDir, 0755, rootUID, rootGID); err != nil && !os.IsExist(err) {
-		return "", err
+	if readWrite {
+		if err := idtools.MkdirAllAs(diffDir, 0755, rootUID, rootGID); err != nil && !os.IsExist(err) {
```
nalind: Non-blocker: I think we only care about the Chown() here, since the "diff" directory is created when we create the layer.

nalind:
A couple of nits that can be fixed later, otherwise LGTM.

PR #405, opened 2019-08-03
We should not be touching internal work dir

rhatdan:
Vivek Goyal says that we should not be touching the internal WorkDir/work directory.
This is internal to the kernel and should not be modified.  I believe this was done
to make sure that the "/" of the container is writable by the root user of the user
namespace that is running the container.  Changing the ownership of the diffDir
should give us the same behavior and is safe to do.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -840,7 +840,6 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		return "", err
 	}
 
-	diffDir := path.Join(dir, "diff")
 	lowers, err := ioutil.ReadFile(path.Join(dir, lowerFile))
 	if err != nil && !os.IsNotExist(err) {
 		return "", err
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind @rhvgoyal @giuseppe @vrothberg @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
Changes LGTM.
Weird go version error in one of the test suites though....
Do we hold Buildah 1.10 and Podman 1.5 for this?  (FYI @mheon)

mheon:
All our tests in Podman land are still red, so we have time!

rhatdan:
We have been running with this for a while, so I don't see this as a huge deal.  
I kicked the retest.

giuseppe:
LGTM

rhatdan:
@cevich We can no longer get ubuntu images to pass, they are always locking up on conflict of dpkg.

rhatdan:
Done fighting with ubuntu image.  Forcing merge.

PR #406, opened 2019-08-03
pkg/tarlog: use io.Pipe()

vrothberg:
Using `os.Pipe()` has a side effect when writing tar files surfacing in
differing digests (see [1] for reference).  Instead use `io.Pipe()` with
a workaround to avoid writes after the reader has been closed - which is
supported by `os.Pipe()` and causes the tar package to error otherwise.

[1] https://github.com/containers/libpod/pull/3705#issuecomment-517954910


pkg/tarlog/tarlogger.go
```diff
@@ -1,47 +1,65 @@
...
+	t.stateMutex.Unlock()
 	return t.writer.Write(b)
 }
 
 func (t *tarLogger) Close() error {
-	err := t.writer.Close()
-	t.wg.Wait()
-	return err
+	t.writer.Close()
+	t.closeMutex.Lock()
```
TomSweeneyRedHat: OK, maybe late Saturday brain, but should this be Unlock?

vrothberg: That's intended. I changed the waiting group to a mutex as it was used to always sync two routines. The lock makes sure to wait for the other routine.

vrothberg:
We are having some serious issues with `ffjson`:
```
754
STDERR:
755
ffjson-inception316066608/ffjson-inception303213507.go:9:2: cannot find package "github.com/pquerna/ffjson/inception" in any of:
756
	/root/go/src/github.com/containers/storage/vendor/github.com/pquerna/ffjson/inception (vendor tree)
757
	/root/.gimme/versions/go1.12.linux.amd64/src/github.com/pquerna/ffjson/inception (from $GOROOT)
758
	/root/go/src/github.com/pquerna/ffjson/inception (from $GOPATH)
759
layers_ffjson_expose.go:9:2: cannot find package "github.com/pquerna/ffjson/shared" in any of:
760
	/root/go/src/github.com/containers/storage/vendor/github.com/pquerna/ffjson/shared (vendor tree)
761
	/root/.gimme/versions/go1.12.linux.amd64/src/github.com/pquerna/ffjson/shared (from $GOROOT)
762
	/root/go/src/github.com/pquerna/ffjson/shared (from $GOPATH)
763
764
```

I see those errors in Travis builds for a longer time and can even reproduce in master.

vrothberg:
I added another commit to trick go modules into vendoring the missing deps.

vrothberg:
libpod CI is passing now : https://github.com/containers/libpod/pull/3705

vrothberg:
Ubuntu is failing to install, so I we could/should(?) ignore it.

rhatdan:
Yes I think we have to ignore this until @cevich fixes it.  Need @nalind ok and then we can merge.

nalind:
At some point we should root-cause this and add a test for it, but as the libpod folks need this to unbreak their tests, and this doesn't break any of ours, this'll do for now.  LGTM.

cevich:
> his until @cevich fixes it

Oh geeze this again?  okay, It's gone past my PITA-o-meter...https://github.com/containers/storage/pull/408

PR #408, opened 2019-08-05
Cirrus: Use libpod cache-images

cevich:
Previously automation on this project used the low-level VM images
mostly as-is from google.  However, due to many items fixed by
https://github.com/containers/libpod/pull/3632 it makes more sense
to use the cache-images instead.

Non-deterministic behavior is a severe detriment to testing
reliability.  By using the libpod cache-images, this projects inherits
the libpod PR-based image-build workflow, a slew of pre-installed
packages, along with disabled default background services.

This should result in both faster and more reliable testing.


cevich:
Hmmm, dang, was hoping this would be an easy switchero :frowning_face:

My guess: We've installed lots of packages into these images, likely one/more is getting in the way of items CI in this project was setup to build from source.

Any other clues?  (I'm not proficient in reading the test failures here)

vrothberg:
> rm: cannot remove '/tmp/tmp.qrl.vpbzc16rlcnw/ro-root/overlay/d25b8f55353206b375e84d2452bdeb89b13354d5a1e5c97ae06ca52f773fdb28/merged': Device or resource busy

@nalind @rhatdan ?

> [+0003s]     UBUNTU_CACHE_IMAGE_NAME=ubuntu-18-libpod-5479475851034624
[+0003s] Unknown/Unsupported distro. ubuntu and/or version 18 for setup.sh

This one looks like we need to rename or adjust the script?

nalind:
> > rm: cannot remove '/tmp/tmp.qrl.vpbzc16rlcnw/ro-root/overlay/d25b8f55353206b375e84d2452bdeb89b13354d5a1e5c97ae06ca52f773fdb28/merged': Device or resource busy
> 
> @nalind @rhatdan ?

The previous line's attempt to mount the middle layer produced an error result, but the cleanup attempt is failing because another layer (likely the one mounted at line 23) is still mounted.

cevich:
> This one looks like we need to rename or adjust the script?

Yes, we were using Ubuntu 19, I don't have that as a cache-image (yet) for libpod, so dropped to 18.

cevich:
> The previous line's attempt to mount the middle layer produced an error result, but the cleanup attempt is failing because another layer (likely the one mounted at line 23) is still mounted.

So my fear is the problem is caused at a higher-level, since the new images have MANY more packages pre-installed.  It's easily possible there are multiple copies of "things" on the system, and being used incorrectly different contexts.

Hmmm.  I think we need more data.

cevich:
I think we probably need to focus on "setup" instead of the actual test failure.

> multiple copies of "things" on the system

Looking...I wonder about the fuse/overlay stuff.  This is what goes into a Fedora image:
https://github.com/containers/libpod/blob/master/contrib/cirrus/packer/fedora_setup.sh

and this is an Ubuntu image:
https://github.com/containers/libpod/blob/master/contrib/cirrus/packer/ubuntu_setup.sh

cevich:
note to me:  Looks like only ubuntu 18 needs some `$DEBS_REQUIRED` adjustments (or make 19 available).  The Fedora's packaging looks "okay" to me, still not sure why they're dying during test :confused:

cevich:
rebased, removed fuse3 and btrfs from ubuntu.

cevich:
rebased, added libfuse2

cevich:
@nalind It looks to me like the Fedora's are in better shape than Ubuntu 18.  I'm not finding fuse 3 available (standard) there, and it seems this is a requirement for fuse-overlay?  Since we're borrowing cache-images from libpod's workflow, there is no Ubuntu 19 available yet.  Just to be clear, if that's a hard-requirement, implementing it there will slow down this PR.  There are a few third-party repos available though, so I may go check those for fuse3 first before giving up.

On the Fedora's front, any hints at what could be going wrong with test?

I was thinking it might be helpful if I had the setup script splat out a list of the currently installed packages and versions.  If that would be helpful, what packages would you like to see included in the list?

P.S. My unspoken goal with this PR is also positional - so I can bring in a secondary partition for additional use in test :smile:

cevich:
Assuming we need Ubuntu 19 here, opened https://github.com/containers/libpod/pull/3754

cevich:
waiting on some bleeding-edge VM images to build, then will update/rebase and try again here...

cevich:
Package updates of note (from `fedora-cloud-base-30-1-2-1556821664`):
* glibc
* glibc-common
* coreutils
* dbus
* systemd
* kernel-core

rhatdan:
@cevich What's happening with this?

cevich:
@rhatdan blocked by: #419

cevich:
(rebased, added missing 'automake' package)

cevich:
Note to me: update with images from https://github.com/containers/libpod/pull/4028

cevich:
rebased/force-pushed w/ latest libpod VM images

cevich:
(rebased to include #439)

cevich:
Hehe, okay then, well dang, that was easy.

rhatdan:
LGTM
@nalind PTAL

nalind:
LGTM

PR #409, opened 2019-08-05
Nonspecific storage driver for ignore_chown_errors option

kpelzel:
This pull request allows a user to not have to specify the storage driver for the ignore_chown_errors option.

Signed-off-by: Kevin Pelzel <kevinpelzel22@gmail.com>

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -322,7 +322,7 @@ func parseOptions(options []string) (*overlayOptions, error) {
 				return nil, fmt.Errorf("overlay: ostree_repo specified but support for ostree is missing")
 			}
 			o.ostreeRepo = val
-		case "overlay2.ignore_chown_errors", "overlay.ignore_chown_errors":
+		case ".ignore_chown_errors", "overlay2.ignore_chown_errors", "overlay.ignore_chown_errors":
 			logrus.Debugf("overlay: ignore_chown_errors=%s", val)
 			o.ignoreChownErrors, err = strconv.ParseBool(val)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


drivers/overlay/overlay.go
```diff
@@ -322,7 +322,7 @@ func parseOptions(options []string) (*overlayOptions, error) {
 				return nil, fmt.Errorf("overlay: ostree_repo specified but support for ostree is missing")
 			}
 			o.ostreeRepo = val
-		case "overlay2.ignore_chown_errors", "overlay.ignore_chown_errors":
+		case ".ignore_chown_errors", "overlay2.ignore_chown_errors", "overlay.ignore_chown_errors":
```
rhatdan: This is fine, but containers/storage/store.go was appending the driver.  Did you see an error with this?

kpelzel: I didn't realize this till after the changes got merged in, but Ideally we'd like to be able to pass in this option as a command line flag in buildah without specifying the driver. Currently I can do this:
`buildah --storage-opt "vfs.ignore_chown_errors=true"`
and this:
`buildah --storage-opt "overlay2.ignore_chown_errors=true"`
but I'd like to be able to run something like this:
`buildah --storage-opt "ignore_chown_errors=true"`

I got the impression that containers/storage/store.go just parsed the storage.conf file and not the command line parameters. I could have missed something though, is there someway of doing this currently? I also didn't know for sure if you wanted to keep --storage-opt to be driver specific options.

rhatdan: No you are probably correct.

rhatdan:
LGTM

vrothberg:
LGTM

PR #410, opened 2019-08-05
Disable cgo-requiring bits when cgo is not enabled

nalind:
Adjust build tags in drivers and pkg so that builds with CGO_ENABLED=0 won't fail outright.  This ends up disabling btrfs (which needs kernel headers), ostree (which uses libostree), overlayfs (which uses C headers to define fs_disk_quota_t), and devicemapper (which uses libdevmapper and loopback) by default.


rhatdan:
LGTM, although I would love to get overlay to work without cgo.

rhatdan:
@vrothberg @TomSweeneyRedHat @giuseppe PTAL

nalind:
This should help with https://github.com/containers/image/pull/675#issuecomment-518281596.

PR #412, opened 2019-08-06
Don't report IsNotExists errors, when unmounting

rhatdan:


rhatdan:
@TomSweeneyRedHat That test VM is always unhappy.  I have asked @cevich to see if we can fix it. Basically the vm starts a apt update when it boots, and it takes so long that the tests suite times out.

PR #415, opened 2019-08-09
Cirrus: Update VM images

cevich:
These were produced as an intended byproduct of:
https://github.com/containers/libpod/pull/3754

Where 'systemd_banish.sh' was run during base-image production.

This is a short-term fix, as there is much setup duplication
here which can (hopefully) be avoided with:
https://github.com/containers/storage/pull/408

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -19,9 +19,9 @@ env:
     ####
     # GCE project where images live
     IMAGE_PROJECT: "libpod-218412"
-    FEDORA_CACHE_IMAGE_NAME: "fedora-cloud-base-30-1-2-1556821664"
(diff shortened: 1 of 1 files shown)


cevich:
Now...this is curious.  Looks like the same/similar failures as in #408

So I would hazard to guess, the problems are due to new packages/changes in the base-level image (which was fully updated).   So...new kernel, different default kernel options/capibilities...that kind of thing.

cevich:
(I need help with a fix here and in #408...in case that wasn't clear)

rhatdan:
@cevich Does this need to be updated now?

cevich:
@rhatdan yes, but I'm going to do in #408 - has same change and (is/was) reproducing the same problem.

PR #416, opened 2019-08-09
Ignore ro mount options in btrfs and windows drivers

rhatdan:
Since now we always set the "ro" mount option, we need to ignore
these options on drivers that do not support them.

### DIFF
--- a/drivers/btrfs/btrfs.go
+++ b/drivers/btrfs/btrfs.go
@@ -645,7 +645,15 @@ func (d *Driver) Get(id string, options graphdriver.MountOpts) (string, error) {
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


rhatdan:
Fixes https://github.com/containers/libpod/issues/3773

rhatdan:
@marcov Could you try this out?

rhatdan:
@vrothberg @giuseppe @nalind @marcov @mheon PTAL

mheon:
Code LGTM

TomSweeneyRedHat:
LGTM but would like a head nod from @nalind and/or a confirmation from @marcov

vrothberg:
I tested this PR successfully in a VM.

PR #422, opened 2019-08-28
Add names-history support

saschagrunert:
This commit adds a new `NamesHistory` field to the `images.json`, which
is basically a deduped list of names the image had in the past. The
first entry of the list is the latest history entry.

The main use case for this feature is to tell the end-user which
names/tags an image had in the past if it does not contain any `names`
any more.

Detailed use case:
1. Pulling `image:v1` into the local registry: `names: [ "image:v1" ]`
2. Pushing a new image as `image:v1` into the remote registry
3. Pulling `image:v1` again into the local registry:
    - first image: `names: [ "image:v1" ]`
    - previous v1 image: `names: [], names-history: [ "image:v1" ]`
4. An consumer of the storage API can now process the image name and
   still display `image` as REPOSITORY, like:
   Before:
   ```
   > podman images
   REPOSITORY      TAG      IMAGE ID       CREATED          SIZE
   image           v1       25b62d1b654a   13 seconds ago   2.07 kB
   <none>          <none>   b134eff7b955   17 seconds ago   2.07 kB
   ```
   After:
   ```
   > podman images
   REPOSITORY      TAG      IMAGE ID       CREATED          SIZE
   image           v1       25b62d1b654a   13 seconds ago   2.07 kB
   image           <none>   b134eff7b955   17 seconds ago   2.07 kB
   ```
5. Since the `NamesHistory` is a slice we would be able to tell the
   end-user which names an image ID had before.

The change should be backwards compatible with previous versions of
containers/storage.

/cc @mtrmac

### DIFF
--- a/images.go
+++ b/images.go
@@ -47,6 +47,11 @@ type Image struct {
 	// or canonical references.
 	Names []string `json:"names,omitempty"`
 
+	// NamesHistory is an optional set of Names the image had in the past. The
+	// contained names are free from any duplicates, whereas the newest entry
+	// is the first one.
+	NamesHistory []string `json:"names-history,omitempty"`
+
 	// TopLayer is the ID of the topmost layer of the image itself, if the
 	// image contains one or more layers.  Multiple images can refer to the
 	// same top layer.
@@ -155,6 +160,7 @@ func copyImage(i *Image) *Image {
 		Digest:          i.Digest,
 		Digests:         copyDigestSlice(i.Digests),
 		Names:           copyStringSlice(i.Names),
+		NamesHistory:    copyStringSlice(i.NamesHistory),
 		TopLayer:        i.TopLayer,
 		MappedTopLayers: copyStringSlice(i.MappedTopLayers),
 		Metadata:        i.Metadata,
@@ -481,6 +487,10 @@ func (r *imageStore) removeName(image *Image, name string) {
 	image.Names = stringSliceWithoutValue(image.Names, name)
 }
 
+func (i *Image) addNameToHistory(name string) {
+	i.NamesHistory = dedupeNames(append([]string{name}, i.NamesHistory...))
+}
+
 func (r *imageStore) SetNames(id string, names []string) error {
 	if !r.IsReadWrite() {
 		return errors.Wrapf(ErrStoreIsReadOnly, "not allowed to change image name assignments at %q", r.imagespath())
@@ -495,6 +505,7 @@ func (r *imageStore) SetNames(id string, names []string) error {
 				r.removeName(otherImage, name)
 			}
 			r.byname[name] = image
+			image.addNameToHistory(name)
 		}
 		image.Names = names
 		return r.Save()
--- a/images_ffjson.go
+++ b/images_ffjson.go
@@ -59,6 +59,22 @@ func (j *Image) MarshalJSONBuf(buf fflib.EncodingBuffer) error {
 		}
 		buf.WriteByte(',')
 	}
+	if len(j.NamesHistory) != 0 {
+		buf.WriteString(`"names-history":`)
+		if j.NamesHistory != nil {
+			buf.WriteString(`[`)
+			for i, v := range j.NamesHistory {
+				if i != 0 {
+					buf.WriteString(`,`)
+				}
+				fflib.WriteJsonString(buf, string(v))
+			}
+			buf.WriteString(`]`)
+		} else {
+			buf.WriteString(`null`)
+		}
+		buf.WriteByte(',')
+	}
 	if len(j.TopLayer) != 0 {
 		buf.WriteString(`"layer":`)
 		fflib.WriteJsonString(buf, string(j.TopLayer))
@@ -171,6 +187,8 @@ const (
 
 	ffjtImageNames
 
+	ffjtImageNamesHistory
+
 	ffjtImageTopLayer
 
 	ffjtImageMappedTopLayers
@@ -193,6 +211,8 @@ var ffjKeyImageID = []byte("id")
 var ffjKeyImageDigest = []byte("digest")
 
 var ffjKeyImageNames = []byte("names")
+
+var ffjKeyImageNamesHistory = []byte("names-history")
 
 var ffjKeyImageTopLayer = []byte("layer")
 
@@ -348,6 +368,11 @@ mainparse:
 						currentKey = ffjtImageNames
 						state = fflib.FFParse_want_colon
 						goto mainparse
+
+					} else if bytes.Equal(ffjKeyImageNamesHistory, kn) {
+						currentKey = ffjtImageNamesHistory
+						state = fflib.FFParse_want_colon
+						goto mainparse
 					}
 
 				}
@@ -400,6 +425,12 @@ mainparse:
 					goto mainparse
 				}
 
+				if fflib.EqualFoldRight(ffjKeyImageNamesHistory, kn) {
+					currentKey = ffjtImageNamesHistory
+					state = fflib.FFParse_want_colon
+					goto mainparse
+				}
+
 				if fflib.EqualFoldRight(ffjKeyImageNames, kn) {
 					currentKey = ffjtImageNames
 					state = fflib.FFParse_want_colon
@@ -443,6 +474,9 @@ mainparse:
 
 				case ffjtImageNames:
 					goto handle_Names
+
+				case ffjtImageNamesHistory:
+					goto handle_NamesHistory
 
 				case ffjtImageTopLayer:
 					goto handle_TopLayer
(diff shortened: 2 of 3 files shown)


saschagrunert:
@rhatdan @vrothberg @nalind @mtrmac WDYT? It is mainly another approach to improve the information density about images to end users without breaking past versions.

images.go
```diff
@@ -492,6 +504,7 @@ func (r *imageStore) SetNames(id string, names []string) error {
 		}
 		for _, name := range names {
 			if otherImage, ok := r.byname[name]; ok {
+				r.addNameToHistory(otherImage, name)
```
vrothberg: I suggest calling `addNameToHistory(...)` from `removeName(...)`. This way, we're sure to not miss an execution path.

saschagrunert:
> What happens if I tag the image?

So from my local experiments, if I do:

1. Create a base image without names:
```
> ./bin/podman build -f Dockerfile-empty -t img:v1
```
2. Now the `images.json` contains the name correctly but no history:
```json
[
  {
    "id": "a13e0f4e4ce6ce32820739c1b892a2c24f694dfc676216a9e5f19d0596da6968",
    "digest": "sha256:108f68845a70455485fb8677c3db83f596dc05ec257ddb4626254ccbd9a0c686",
    "names": [
      "localhost/img:v1"
    ],
    "layer": "5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef",
    ...
  }
]
```
3. Tag multiple other names:
```
> ./bin/podman tag img:v1 img:v2 img:v3
```
4. We still have it as names, because they are all distinct:
```json
[
  {
    "id": "a13e0f4e4ce6ce32820739c1b892a2c24f694dfc676216a9e5f19d0596da6968",
    "digest": "sha256:108f68845a70455485fb8677c3db83f596dc05ec257ddb4626254ccbd9a0c686",
    "names": [
      "localhost/img:v1",
      "localhost/img:v2",
      "localhost/img:v3"
    ],
    "layer": "5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef",
    ...
  }
]
```
```
> ./podman images
REPOSITORY      TAG   IMAGE ID       CREATED              SIZE
localhost/img   v1    a13e0f4e4ce6   About a minute ago   2.07 kB
localhost/img   v2    a13e0f4e4ce6   About a minute ago   2.07 kB
localhost/img   v3    a13e0f4e4ce6   About a minute ago   2.07 kB
```
4. Now we steal the img:v1 tag by creating a new empty image:
```
> ./bin/podman build -f Dockerfile-empty -t img:v1
```
```json
[
  {
    "id": "a13e0f4e4ce6ce32820739c1b892a2c24f694dfc676216a9e5f19d0596da6968",
    "digest": "sha256:108f68845a70455485fb8677c3db83f596dc05ec257ddb4626254ccbd9a0c686",
    "names": [
      "localhost/img:v2",
      "localhost/img:v3"
    ],
    "names-history": [
      "localhost/img:v1"
    ],
    ...
  },
  {
    "id": "12d9c90a46f52a4d6abaf2605fd0bf9a89fe8763e3f13993fb894b739b6c1111",
    "digest": "sha256:df4879cb75d87edc08639e0c7252cb70945cd304f6eceba43a2349c93f7777c3",
    "names": [
      "localhost/img:v1"
    ],
    "layer": "5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef",
    ...
  }
]
```
5. Take another tag from the first image:
```
> ./bin/podman tag img:v1 img:v2
```
```json
[
  {
    "id": "a13e0f4e4ce6ce32820739c1b892a2c24f694dfc676216a9e5f19d0596da6968",
    "digest": "sha256:108f68845a70455485fb8677c3db83f596dc05ec257ddb4626254ccbd9a0c686",
    "names": [
      "localhost/img:v3"
    ],
    "names-history": [
      "localhost/img:v2",
      "localhost/img:v1"
    ],
    ...
  },
  {
    "id": "12d9c90a46f52a4d6abaf2605fd0bf9a89fe8763e3f13993fb894b739b6c1111",
    "digest": "sha256:df4879cb75d87edc08639e0c7252cb70945cd304f6eceba43a2349c93f7777c3",
    "names": [
      "localhost/img:v1",
      "localhost/img:v2"
    ],
    "layer": "5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef",
    ...
  }
]
```

> If I can tag it, can I still push it? Can I remove it via tag name?

I think we could add dedicated command line options to buildah/podman to interact with _old images_, for example something like `podman rmi --cleanup`.

> Why is this a list, how could I have multiple former image names?

See my example above, I also thought about adding something like a timestamp to indicate when it has been retagged, but this would clash with the current de-duplication approach.

images.go
saschagrunert: Yes, I changed it and bound the new method to the `*Image` type, which should be more clear. :+1:

mtrmac:
Can we first explicitly clarify whether the goal of the PR is intended to mirror existing `docker images`, or whether it is a new thing being designed?

For the former, I suppose we would get more similar behavior by implementing `RepoDigests` the way Docker does (notably `docker tag` does not allow creating a “name:<none>” entry locally I think.

For the latter, it just seems weird to me to see a $name entry in _any_ form in the output after an explicit rm/untag of $name . (That’s an UI concern not _directly_ impacting the design of the c/storage data, but still.)

mtrmac:
> For the former, I suppose we would get more similar behavior by implementing `RepoDigests` the way Docker does (notably `docker tag` does not allow creating a “name:” entry locally I think.

… and we would _still_ need to implement `RepoDigests` correctly even after this PR and using the history data in `docker images` output would mask _some_ of the impacts of not recording `RepoDigests`.

saschagrunert:
> Can we first explicitly clarify whether the goal of the PR is intended to mirror existing `docker images`, or whether it is a new thing being designed?

I think the major use case is: Displaying the image name to the user if an image has no `Names` any more due to re-tag, [point 4](https://github.com/containers/storage/pull/422#issue-311735338)

In CRI-O, we could add these values to the ListImages RPC, whereas in podman we could either add an additional colum to `podman images` or stripping of the tag and showing it as `REPOSITORY` value.

images.go
```diff
@@ -47,6 +47,11 @@ type Image struct {
 	// or canonical references.
 	Names []string `json:"names,omitempty"`
 
+	// NamesHistory is an optional set of Names the image had in the past. The
+	// contained names are free from any duplicates, whereas the newest entry
+	// is the first one.
+	NamesHistory []string `json:"names-history,omitempty"`
```
mtrmac: Currently (and especially with the changes what I expect to happen for `RepoDigests`), `Names` can contain both name:tag and name@digest values. So, `NamesHistory` would currently contain both as well. Is that OK?

(Ideally, I’d like be to insist on strongly typed values like `reference.NamedTagged`, encoding the semantics in the type system, but with `Names []string` that doesn’t seem reasonable right now.)

mtrmac:
>  I also thought about adding something like a timestamp to indicate when it has been retagged, but this would clash with the current de-duplication approach.

Well, the deduplication is really not all that much code; if we need to make it a bit more complex to get the data structure we want, we should just do that.

On the one hand, if nothing needs a timestamp now, it seems rather pointless to add them now.

On the other hand, if we do end up adding timestamps in the future, it can’t be done within the current `NamesHistory` array and we would have to either break the ABI by changing the field, or maintain two independent fields with redundant values forever. So, getting this correct right now is worth a bit of thought before committing to the API.

(One possible way to defer that decision: make the field private or documented-to-be-private, and add a public getter method; then we can enrich the recorded data without breaking the API.)

--- 

Up to you, @nalind and others, I really understand neither the UI ambitions nor the API design of c/storage sufficiently.

images.go
saschagrunert: > Currently (and especially with the changes what I expect to happen for `RepoDigests`), `Names` can contain both name:tag and name@digest values. So, `NamesHistory` would currently contain both as well. Is that OK?

I think so.

> (Ideally, I’d like be to insist on strongly typed values like `reference.NamedTagged`, encoding the semantics in the type system, but with `Names []string` that doesn’t seem reasonable right now.)

Yeah, we could change both `Names` and `NamesHistory` to be forcibly valid references, but I wanted to add the change as non-breaking as possible.

mtrmac: ACK.

rhatdan:
@saschagrunert You still interested in this?

saschagrunert:
Yes, I would stick to the current implementation approach and rebased on top of the latest master branch.

@mtrmac Do you see any blocker for this getting merged (if the CI passes)?
@nalind May I ask you to please take a look at this one?

images.go
```diff
@@ -478,9 +484,14 @@ func (r *imageStore) SetMetadata(id, metadata string) error {
 }
 
 func (r *imageStore) removeName(image *Image, name string) {
+	image.addNameToHistory(name)
```
mtrmac: `removeName` is, AFAICS, only called on a _name conflict_ (e.g. if `SetNames` adds another name to an image, and that name has been used by a different image); it is not called on _untag_ (`SetNames` removing an existing name).

That’s good enough as far as the original motivation goes:
> I think the major use case is: Displaying the image name to the user if an image has no Names any more due to re-tag, point 4

but not consistent with how `NamesHistory` is documented:
> `NamesHistory is an optional set of Names the image had in the past. `

saschagrunert: Alright, I updated the documentation as suggested.

mtrmac: I was somewhat expecting an implementation change instead of documenting the behavior, but, sure, this is consistent. Up to @nalind.

saschagrunert: Yeah I can also change the implementation to match the previous documentation, this was not clear for me in the first place.

saschagrunert: Changed the implementation and added tests on top.

rhatdan:
LGTM

saschagrunert:
Hey, may I ask regarding a new release that I can bring this feature into CRI-O and libpod?

vrothberg:
The last release was just 7 days ago but I think Dan wants throw out some of the recent changes soon. @rhatdan, is anything blocking us from cutting another release?

rhatdan:
I will cut a release.

rhatdan:
https://github.com/containers/storage/pull/468

PR #425, opened 2019-09-10
layerStore.Load(): avoid double-locking the mounts list for Save

nalind:
If we need to re-save the layers list when we've loaded it, to either solve a duplicate name issue or to clean up a partially-constructed layer, don't make the mistake of attempting to take another lock on the mounts list.


umohnani8:
LGTM
@rhatdan @mrunalp PTAL

mrunalp:
LGTM

layers.go
```diff
@@ -416,6 +416,16 @@ func (r *layerStore) loadMounts() error {
 }
 
 func (r *layerStore) Save() error {
+	r.mountsLockfile.Lock()
+	defer r.mountsLockfile.Unlock()
+	defer r.mountsLockfile.Touch()
+	if err := r.saveLayers(); err != nil {
```
vrothberg: We could move the call to `saveLayers()` before acquiring the mountsLockfile (i.e., to line 419). It would  further shorten the critical section.

TomSweeneyRedHat:
LGTM, but one of the tests is having an unrelated lock issue.  I'll try restarting.

rhatdan:
LGTM, and would like to see separate patch with @vrothberg optimization.

umohnani8:
@nalind the ubuntu test is failing :/

vrothberg:
Same for https://github.com/containers/storage/pull/426. I vote for force merging.

umohnani8:
Force merging since the test is failing on other PRs as well.

PR #426, opened 2019-09-11
wrap ID or digest to ErrImageUnkown errors

vrothberg:
Wrap the ID or the digest to ErrImageUnknown errors to avoid ambiguity
which image is unknown.  Consumers of the storage library may have
multiple subsequent calls to the storage API where it can be unclear
which image is unknown.  Wrapping the ID and digest attempts to avoid
this ambiguity.

Related-to:    github.com/containers/libpod/issues/2979

@giuseppe @nalind @rhatdan PTAL

### DIFF
--- a/images.go
+++ b/images.go
@@ -372,7 +372,7 @@ func (r *imageStore) ClearFlag(id string, flag string) error {
 	}
 	image, ok := r.lookup(id)
 	if !ok {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

nalind:
LGTM.

PR #427, opened 2019-09-11
[cri-o-release-1.14] layerStore.Load(): avoid double-locking the mounts list for Save

nalind:
If we need to re-save the layers list when we've loaded it, to either solve a duplicate name issue or to clean up a partially-constructed layer, don't make the mistake of attempting to take another lock on the mounts list.

Cherry picks #425 for the cri-o-release-1.14 branch.


rhatdan:
LGTM
@umohnani8 @vrothberg PTAL Not sure why Ubuntu is breaking all the time, this is usually where I whine at @cevich

umohnani8:
LGTM

PR #434, opened 2019-09-24
ImageSize: don't get tripped up by images with no layers

nalind:
Images don't have to have layers, so they don't have to have top layers, and we shouldn't return an error when attempting to determine the size of such an image.


rhatdan:
LGTM, Ubuntu is a known failure.  @vrothberg @giuseppe @TomSweeneyRedHat PTAL

nalind:
Added a test.

vrothberg:
> Added a test.

Thanks, Nalin :1st_place_medal:

PR #436, opened 2019-09-25
ImageBigData: distinguish between no-such-image and no-such-item

nalind:
When reporting an error from store.ImageBigData(), distinguish between cases where we can't find the specified image, and where we found the image, but it didn't have a matching requested item.


store.go
```diff
@@ -1515,6 +1516,12 @@ func (s *store) ImageBigData(id, key string) ([]byte, error) {
 		if err == nil {
 			return data, nil
 		}
+		if store.Exists(id) {
+			foundImage = true
+		}
+	}
+	if foundImage {
```
TomSweeneyRedHat: should this be `if !foundImage {`  ?

nalind: That wouldn't match the error that we return in the block.

TomSweeneyRedHat: ugh, coulda sworn that said 'error locating image...'  Darn dyslexia!  Thx, all good.

TomSweeneyRedHat:
LGTM assuming happy tests.

rhatdan:
LGTM

PR #438, opened 2019-10-03
new interface for MountImage added

kunalkushwaha:
Interface added to mount image rootfs without any container.
image can be mounted in ReadOnly or ReadWrite mode.

Signed-off-by: Kunal Kushwaha <kunal.kushwaha@gmail.com>

### DIFF
--- a/store.go
+++ b/store.go
@@ -254,6 +254,10 @@ type Store interface {
 	// Wipe removes all known layers, images, and containers.
 	Wipe() error
 
+	// MountImage mounts an image to temp folder and returns the mount point.
+	// Currently only supported option in "ro" i.e. Read Only.
+	MountImage(id string, mountOptions []string) (string, error)
+
 	// Mount attempts to mount a layer, image, or container for access, and
 	// returns the pathname if it succeeds.
 	// Note if the mountLabel == "", the default label for the container
@@ -2463,6 +2467,40 @@ func (s *store) Status() ([][2]string, error) {
 
 func (s *store) Version() ([][2]string, error) {
 	return [][2]string{}, nil
+}
+
+func (s *store) MountImage(id string, mountOptions []string) (string, error) {
+	// Append ReadOnly option to mountOptions
+	img, err := s.Image(id)
+	if err != nil {
+		return "", err
+	}
+
+	mountOptions = append(mountOptions, "ro")
+	if err := validateMountOptions(mountOptions); err != nil {
+		return "", err
+	}
+
+	rlstore, err := s.LayerStore()
+	if err != nil {
+		return "", err
+	}
+
+	rlstore.Lock()
+	defer rlstore.Unlock()
+	if modified, err := rlstore.Modified(); modified || err != nil {
+		if err = rlstore.Load(); err != nil {
+			return "", err
+		}
+	}
+
+	if rlstore.Exists(img.TopLayer) {
+		options := drivers.MountOpts{
+			Options: mountOptions,
+		}
+		return rlstore.Mount(img.TopLayer, options)
+	}
+	return "", ErrLayerUnknown
 }
 
 func (s *store) Mount(id, mountLabel string) (string, error) {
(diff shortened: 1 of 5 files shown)


kunalkushwaha:
This PR is for https://github.com/containers/libpod/issues/1433.

@rhatdan Please see if this looks okay.
- Is options need to be checked here or functions that invoke this Interface should take care of that?

rhatdan:
You need to update cmd/containers-storage to support this mount.  Then we  need tests to test this.

Not sure if we would want to support mounting the image read/write?  

@nalind WDYT?

store.go
```diff
@@ -253,6 +253,10 @@ type Store interface {
 	// Wipe removes all known layers, images, and containers.
 	Wipe() error
 
+	// MountImage mounts an image to temp folder and returns the mount point.
+	// Options  supported are ro, rw.
```
vrothberg: Double whitespace between "Options" and "supported".

rhatdan:
I can see mounting an image Read/Only mounting it read/write seems like a bad idea unless their is a upper layer. I don't think we would want the Image Storage modified at all.

@vrothberg @nalind WDYT?

kunalkushwaha:
@rhatdan I got your point. 

Earlier I thought, storage lib should provide the simple API, so consumer of API can choose the mode. 

But with without additional layer, mounting image should not be allowed as it will change the image. For all other cases `Mount()` should be used. I will make changes for `ReadOnly`.

kunalkushwaha:
I am little confused towards naming the API.
`MountImage()` by name do not suggest it will return `ReadOnly` mounted image.
- So currently it takes `mountOptions` but is ignored and set to `ReadOnly` in function.
- In future, if this function need to add support of ReadWrite Mode, no need to change the Interface Name.

- We can remote `mountOptions` by changing the interfaceName as `MountImageRO()` or `MountROImage()`

@rhatdan @vrothberg @nalind PTAL

store.go
```diff
@@ -2473,6 +2478,32 @@ func (s *store) Version() ([][2]string, error) {
 	return [][2]string{}, nil
 }
 
+func (s *store) MountImage(id string, mountOptions []string) (string, error) {
+	// Currently only ReadOnly option is set
+	// and value passed to mountOptions is ignored.
+	mountOptions = []string{"ro"}
```
rhatdan: This should be appended on, we still want to pass in the mount options.  We should support NODEV, NOEXEC, NOSUID ...

rhatdan:
LGTM

rhatdan:
We need tests on this.  I want to make sure that I can mount an image and then create a container based on that image, umount the container.  Mount the image multiple times, umount it multiple times and keep track of how many times it is mounted.

kunalkushwaha:
Test cases failing.. mostly while creating new layer. This works fine if run by command.

`populate` function fails.

rhatdan:
@kunalkushwaha Could you try to rebase and push.  TO see if this fixes the issue.

kunalkushwaha:
@rhatdan This PR is stuck with this test failing issue. PTAL

vrothberg:
Restarted the tests to check if the errors are flakes. Thanks for the pin!

rhatdan:
@kunalkushwaha tests seem to be failing.

vrothberg:
Friendly ping. Any updates?

kunalkushwaha:
Sorry, was not able to work on this . I will start on this now and update.

vrothberg:
Awesome, thanks a lot, @kunalkushwaha !

rhatdan:
@kunalkushwaha needs a rebase.

mrunalp:
@kunalkushwaha Could you rebase this?

rhatdan:
@ashley-cui PTAL

rhatdan:
@kunalkushwaha We need this PR to move forward, are you still working on it?  If not we will take it over.

kunalkushwaha:
@rhatdan : @sujil02 asked me a few days back regarding this and had explained to him. 
I thought he is already working on this.

The major blocker on this for merging was test case failure. (https://github.com/containers/storage/pull/438#issuecomment-550970311)

I may not be able to pick it up very soon. So if anyone can pick it up, it will be nice.

Issue #419, opened 2019-08-13
CI Test Failure under updated kernel 5.x


cevich:
Steps to reproduce (using same VM images as master):

```
$ hack/get_ci_vm.sh fedora-cloud-base-30-1-2-1556821664
...cut...
```

Everything works fine, including [the package install/update.](https://paste.fedoraproject.org/paste/uEMBW0XVfAbR2RD4GNQyfg)

```
[root@cevich-fedora-cloud-base-30-1-2-1556821664 storage]# contrib/cirrus/build_and_test.sh
```

[The tests pass as normal](https://paste.fedoraproject.org/paste/ZtRhNHrBQWsnOfNyLPLZyQ), then I execute `reboot`.  After waiting a minute, reconnect to the VM and execute tests a second time:

```
[root@cevich-fedora-cloud-base-30-1-2-1556821664 storage]# reboot
Connection to 35.226.8.25 closed by remote host.
Connection to 35.226.8.25 closed.
ERROR: (gcloud.compute.ssh) [/bin/ssh] exited with return code [255].

Offering to Delete cevich-fedora-cloud-base-30-1-2-1556821664 (Might take a minute or two)

Note: It's safe to answer N, then re-run script again later.
+  sudo podman run -it --rm -e AS_ID=4179 -e AS_USER=cevich --security-opt label=disable -v /tmp/get_ci_vm.sh_tmpdir_NsUYzF:/home/cevich -v /home/cevich/.config/gcloud:/home/cevich/.config/gcloud -v /home/cevich/.config/gcloud/ssh:/home/cevich/.ssh -v /home/cevich/devel/storage:/home/cevich/devel/storage quay.io/cevich/gcloud_centos:latest --configuration=storage --project=storage-240716 compute instances delete --zone us-central1-b --delete-disks=all cevich-fedora-cloud-base-30-1-2-1556821664
The following instances will be deleted. Any attached disks configured
 to be auto-deleted will be deleted unless they are attached to any 
other instances or the `--keep-disks` flag is given and specifies them
 for keeping. Deleting a disk is irreversible and any data on the disk
 will be lost.
 - [cevich-fedora-cloud-base-30-1-2-1556821664] in [us-central1-b]

Do you want to continue (Y/n)?  n

ERROR: (gcloud.compute.instances.delete) Deletion aborted by user.
[cevich@cevich storage]$ hack/get_ci_vm.sh fedora-cloud-base-30-1-2-1556821664
...cut...
[...]

PR #439, opened 2019-10-05
storage: drop ostree deduplication

giuseppe:
it was an attempt to use OSTree to deduplicate files, at the time we already had a dependency on OSTree for system containers in containers/image.  Since the feature never really took off, let's just drop it.

Closes: https://github.com/containers/storage/issues/419
 

rhatdan:
I am fine with this, although it is API Breakage and we will need to bounce the major version.
LGTM

nalind:
LGTM.  If we're doing a bump, we ended up never using the `ParentOwner` methods, so we might want to drop them as well, as a separate PR.

rhatdan:
SGTM

jordemort:
Would it be unwelcome if I attempted to revive this later in the year, with the intent of using it and maintaining it? I'm working on https://github.com/hello-seam/podracer right now and it mostly wouldn't need to exist if there was an OSTree storage driver.

rhatdan:
We would consider it.

PR #442, opened 2019-10-14
Release v1.13.5

vrothberg:
@nalind @TomSweeneyRedHat @rhatdan PTAL

I want push out the fix to return the lock-opening errors that @cevich and @mheon found.


TomSweeneyRedHat:
LGTM assuming happy tests

PR #443, opened 2019-10-15
Add pure file matches indicator

saschagrunert:
The `pure` indicator for a match now returns `true` if it's the only
one. This additional information will be used by consumers to decide
more fine-granular if a match can be excluded completely or not.

Needed by https://github.com/containers/buildah/pull/1914

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -821,11 +821,12 @@ func TarWithOptions(srcPath string, options *TarOptions) (io.ReadCloser, error) 
 				// is asking for that file no matter what - which is true
 				// for some files, like .dockerignore and Dockerfile (sometimes)
 				if include != relFilePath {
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM, but would like a @nalind head nod.

nalind:
Basically what @vrothberg said.  The logic looks right, but the tests should exercise it.

saschagrunert:
I added the test cases and improved the documentation.

saschagrunert:
> Code LGTM although I wonder whether we could change the return value to a tuple of `(counter int, err error)` with `counter` being the number of matches. This combines the two booleans into int + comparison.

Sure thing, I changed it to return the amount of matches. Tests have been adapted as well.

rhatdan:
LGTM

PR #445, opened 2019-10-16
Add excluded information to matches

saschagrunert:
This is a follow up of https://github.com/containers/storage/pull/443, because the information of how many matches we have is not sufficient enough to fix the issue in https://github.com/containers/buildah/pull/1914. :innocent: 

We now add the information if an match has been excluded or not, which is
needed for more fine-granular filepath matching.

### DIFF
--- a/pkg/fileutils/fileutils.go
+++ b/pkg/fileutils/fileutils.go
@@ -58,9 +58,10 @@ func NewPatternMatcher(patterns []string) (*PatternMatcher, error) {
 }
 
 // Matches verifies the provided filepath against all patterns.
-// It returns the amount of `matches` or an error in case of any failure.
+// It returns the amount of `matches` and `excludes` for the patterns on
+// success, otherwise an error.
 // It is not safe to be called concurrently.
-func (pm *PatternMatcher) Matches(file string) (matches uint, err error) {
+func (pm *PatternMatcher) Matches(file string) (matches, excludes uint, err error) {
 	file = filepath.FromSlash(file)
 	parentPath := filepath.Dir(file)
 	parentPathDirs := strings.Split(parentPath, string(os.PathSeparator))
@@ -74,7 +75,7 @@ func (pm *PatternMatcher) Matches(file string) (matches uint, err error) {
 
 		match, err := pattern.match(file)
 		if err != nil {
-			return 0, err
+			return 0, 0, err
 		}
 
 		if !match && parentPath != "." {
@@ -85,9 +86,8 @@ func (pm *PatternMatcher) Matches(file string) (matches uint, err error) {
 		}
 
 		if match {
-			// Subtract negative matches if possible
-			if matches > 0 && negative {
-				matches--
+			if negative {
+				excludes++
 			} else {
 				matches++
 			}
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@nalind @TomSweeneyRedHat @vrothberg @giuseppe PTAL

TomSweeneyRedHat:
LGTM but I'd like a head nod from @vrothberg or @nalind

vrothberg:
Is the problem that the amount of matches is not sufficient, as it could still be positive even if one pattern was excluded?

I have troubles wrapping my head around this specific API, which makes me believe we can simplify it a bit more. What I find most confusing is that we decrement matches for each exclusion and additionally set the boolean. This won't help callers who want to know the exact number of matches.

Maybe we can use two ints instead, one for _matches_ another one for _exclusions_? This way, I would find the API more approachable and callers can interpret the data for their specific needs.

saschagrunert:
> Is the problem that the amount of matches is not sufficient, as it could still be positive even if one pattern was excluded?

Yes! :grin: 

> Maybe we can use two ints instead, one for _matches_ another one for _exclusions_? This way, I would find the API more approachable and callers can interpret the data for their specific needs.

Yeah okay this adds even more information so I changed the implementation to now add the amount of exclusions as integer value.

saschagrunert:
I decided to an additional API function for convenience usage, wheras the buildah fix will use the modified method.

pkg/fileutils/fileutils.go
```diff
@@ -98,7 +98,22 @@ func (pm *PatternMatcher) Matches(file string) (matches uint, err error) {
 		logrus.Debugf("Skipping excluded path: %s", file)
 	}
 
-	return matches, nil
+	return matches, excludes, nil
+}
+
+// IsMatch verifies the provided filepath against all patterns and returns true
+// if it overall matches. It returns an error on failure and is not safe to be
```
vrothberg: "overall" is a bit loose. Can we describe exactly what it means? Some user might expect exclusions to yield an unmatch.

pkg/fileutils/fileutils.go
```diff
@@ -98,7 +98,22 @@ func (pm *PatternMatcher) Matches(file string) (matches uint, err error) {
...
+// called concurrently.
+func (pm *PatternMatcher) IsMatch(file string) (matched bool, err error) {
+	matches, excludes, err := pm.Matches(file)
+	if err != nil {
+		return false, err
+	}
+
+	if matches >= excludes && matches-excludes > 0 {
```
vrothberg: The two conditions seem to contradict a bit (one allows equality while the other doesn't) and are redundant. If we expect more matches than exclusions, `matches-excludes > 0` is sufficient.
```golang
    ...
    return matches-excludes > 0, nil
}
```

pkg/fileutils/fileutils.go
saschagrunert: Okay, I will extend the explanations.

pkg/fileutils/fileutils.go
saschagrunert: I see you point and will remove the `=`, but it will overflow if `excludes > matches` because we're dealing with unsigned integers here.

vrothberg: Eeeeek, good catch. Didn't think about the underflow :+1:

saschagrunert:
Thanks again for reviewing! :pray: I added the new function to the existing test cases to verify that it works correctly.

saschagrunert:
Thanks, can someone please re-trigger the tests for me? I think I don't have enough power.

vrothberg:
Done :heavy_check_mark:

rhatdan:
LGTM

PR #448, opened 2019-10-25
Add make vendor-in-container

rhatdan:
In order to always run vendor with the latest supported version of golang,
add the ability to run the vendor command within a podman container.

After running the command include vendor changes.


rhatdan:
@vrothberg PTAL

rhatdan:

nalind:
LGTM, but the commit log calls it "validate-in-container", not "vendor-in-container", so please correct that.

rhatdan:
Oops fixed.

rhatdan:
Previously we were running with golang 1.12, now we are running with golang 1.13, so these changes are caused by the new golang compiler.

We want everyone using 1.13, buf if you are running on an older OS, you can use 
make vendor-in-container to use it if your golang < 1.13

TomSweeneyRedHat:
@rhatdan, so you did `make` in your sandbox and golang vendored these in then you did  git add/commit/push, etc?

rhatdan:
Correct.

Issue #450, opened 2019-10-29
opts.go: build fails against docker 19.03


elboulangero:
```
# github.com/containers/storage/opts
src/github.com/containers/storage/opts/opts.go:299:15: undefined: filters.ToParam
src/github.com/containers/storage/opts/opts.go:309:18: undefined: filters.ParseFlag
```

These functions were removed in Docker `19.03`: https://github.com/moby/moby/commit/c334a87

It seems `ToParam` can be replaced by`ToJSON` as is, but the move from `ParseFlag` to `Args.Add`  seems to be a bit more involving. I'm reaching the limit of my Go knowledge already here...

vrothberg:
Hi @elboulangero, thanks for the report. We'll look into the issue and try to resolve it. Out of curiosity, how did you detect the issue?

elboulangero:
I package Docker for Debian, we have the version `19.03.4` in the experimental suite: https://packages.debian.org/experimental/docker.io

 Now we're at the stage where we spot broken reverse build dependencies (ie. packages that depend on docker, used to build against docker `18.09`, and fail to build against docker `19.03`), and try to fix it. That's the story :)

Thank for the quick answer (although I wonder if it's an automated answer :p )

vrothberg:
Great, thanks for reaching out!

No automated answer but the right caffeine level :) The dependency is small enough to just fork the package. I'll prepare a PR.

PR #451, opened 2019-10-29
Move opts to internal/opts and fork docker filters

vrothberg:
The dependency on docker's filter package prevents us from using newer
versions of docker/moby where the package's API has changed.  The filter
package is small enough to justify maintaining a small fork of it. Also,
since the opts package is only used for the `containers-storage` binary
and is not part of the actual storage library, move it to the ./internal
directory to make it clearer that this is an internal API.

Fixes: #450


vrothberg:
@rhatdan @nalind @TomSweeneyRedHat @giuseppe PTAL

internal/opts/parse.go
```diff
@@ -1,18 +1,18 @@
...
 	"encoding/json"
 	"errors"
 	"regexp"
 	"strings"
-
-	"github.com/docker/docker/api/types/versions"
 )
 
-// Args stores a mapping of keys to a set of multiple values.
+// args stores a mapping of keys to a set of multiple values.
```
rhatdan: Nit: Why change case?

rhatdan:
LGTM except for one Nit.

internal/opts/parse.go
vrothberg: Good catch. I was playing with the idea of making those types private but had to revert it at some point. I missed this comment.

vrothberg: Fixed now.

TomSweeneyRedHat:
LGTM but would like a more storage savy head nod.

giuseppe:
LGTM.  Do we need to bump the major version?

vrothberg:
> LGTM. Do we need to bump the major version?

It depends if we consider this to be a bug fix or not. I think it's a bug that we exposed an API that it's used by a binary for testing and I personally don't consider this package to be part of the storage library. From the user perspective, I don't know of any tool that depends on this package.

rhatdan:
I would say this is a bug and should not need a bump.

PR #452, opened 2019-10-29
Don't check overlay if using fuse-overlay

rhatdan:
We want to be able to use fuse-overlay on top of overlay file system.

If code is configured to use fuse-overlay then do not do overlay check.


rhatdan:
This is not the correct fix.

Issue #447, opened 2019-10-22
overlay: mount counter checks for overlay, even if we're using fuse-overlayfs


nalind:
When we're using `fuse-overlayfs`, we skip the is-supported test so that we can run on top of an overlay filesystem, but we may also skip mount attempts because when we ask the mount counter if a container's rootfs is mounted in `get()`, it returns true because that location is already on an overlay filesystem.

This is affecting people who try to run the `quay.io/buildah/stable:latest` image.

rhatdan:
Do you have a proposed fix?

nalind:
We can skip using the `ctr` field if `mountProgram` is set, or we can teach it to check for a different filesystem magic value for those cases, or @giuseppe may have another option to suggest.

giuseppe:
I think checking for a different filesystem would be better.  I think it should be fine to assume it is a FUSE filesystem when `mountProgram` is used.

rhatdan:
@giuseppe Could you open a PR for this?  I think this is one thing that will break podman in podman.

nalind:
I'm okay with that, so long as it won't set up a problem for rootless-in-rootless, if we want to support that.  The magic value for FUSE is 0x65735546, FWIW.

giuseppe:
PR opened here: https://github.com/containers/storage/pull/453

I've tested it manually, and verified that it is possible to use a storage on top of overlay for the storage directory.

> I'm okay with that, so long as it won't set up a problem for rootless-in-rootless

yes it will be a problem if we will want rootless-in-rootless, but for now it should not matter since fuse-overlayfs doesn't support running on top of fuse-overlayfs.

It could also be a problem if the storage directory is stored on top of another FUSE file system

PR #453, opened 2019-10-30
overlay: check for FUSE when using mountProgram

giuseppe:
if a mountProgram is specified, check that the file system is already
mounted using the FUSE magic number instead of overlay.  It enables
using fuse-overlayfs on top of overlay.

Closes: https://github.com/containers/storage/issues/447


rhatdan:
LGTM
@nalind PTAL

rhatdan:
Fixes: https://github.com/containers/storage/issues/447 I believe.

nalind:
LGTM

PR #459, opened 2019-11-04
overlay: use fusermount to unmount FUSE file systems

giuseppe:
when a mountProgram is used, attempt to unmount the file system using
fusermount or fusermount3.

If the FUSE writeback cache is used (fuse-overlayfs uses it by default
since version 0.5), some changes might not be flushed from the kernel
to the FUSE file system.

Using umount we could lose these changes as the file system is
terminated before these changes are visible to the FUSE file system.

Going through fusermount|fusermount3 instead makes sure FUSE flushes
all the pending requests before the file system is unmounted.

If the fusermount attempts fail, make sure the file system is flushed
before being unmounted.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1021,8 +1021,39 @@ func (d *Driver) Put(id string) error {
 	if _, err := ioutil.ReadFile(path.Join(dir, lowerFile)); err != nil && !os.IsNotExist(err) {
 		return err
 	}
-	if err := unix.Unmount(mountpoint, unix.MNT_DETACH); err != nil && !os.IsNotExist(err) {
-		logrus.Debugf("Failed to unmount %s overlay: %s - %v", id, mountpoint, err)
+
+	unmounted := false
+
+	if d.options.mountProgram != "" {
+		// Attempt to unmount the FUSE mount using either fusermount or fusermount3.
+		// If they fail, fallback to unix.Unmount
+		for _, v := range []string{"fusermount3", "fusermount"} {
+			err := exec.Command(v, "-u", mountpoint).Run()
+			if err != nil && !os.IsNotExist(err) {
+				logrus.Debugf("Error unmounting %s with %s - %v", mountpoint, v, err)
+			}
+			if err == nil {
+				unmounted = true
+				break
+			}
+		}
+		// If fusermount|fusermount3 failed to unmount the FUSE file system, make sure all
+		// pending changes are propagated to the file system
+		if !unmounted {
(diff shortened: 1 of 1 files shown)


giuseppe:
@AkihiroSuda PTAL

drivers/overlay/overlay.go
```diff
@@ -1021,8 +1021,37 @@ func (d *Driver) Put(id string) error {
...
 	}
-	if err := unix.Unmount(mountpoint, unix.MNT_DETACH); err != nil && !os.IsNotExist(err) {
-		logrus.Debugf("Failed to unmount %s overlay: %s - %v", id, mountpoint, err)
+
+	unmounted := false
+
+	if d.options.mountProgram != "" {
+		// Attempt to unmount the FUSE mount using either fusermount or fusermount3.
+		// If they fail, fallback to unix.Unmount
+		for _, v := range []string{"fusermount", "fusermount3"} {
```
AkihiroSuda: `fusermount3` should be prioritized over `fusermount`?

giuseppe: I didn't think of that.  That is a good idea, I've switched them

drivers/overlay/overlay.go
```diff
@@ -1021,8 +1021,39 @@ func (d *Driver) Put(id string) error {
...
+
+	if d.options.mountProgram != "" {
+		// Attempt to unmount the FUSE mount using either fusermount or fusermount3.
+		// If they fail, fallback to unix.Unmount
+		for _, v := range []string{"fusermount3", "fusermount"} {
+			err := exec.Command(v, "-u", mountpoint).Run()
+			if err != nil && !os.IsNotExist(err) {
+				logrus.Debugf("Error unmounting %s with %s - %v", mountpoint, v, err)
+			}
+			if err == nil {
```
rhatdan: Seems like you are doing a lot of work if os.IsNotExists(err) is true, Should we set unmounted=trued under those circumstances as well, it is just going to cause everything below to fail multiple times?

giuseppe: the `os.IsNotExists(err)` should not be treated differently than the case `err != nil` as the mount is still not freed.

When we could not unmount using `fusermount|fusermount3` then we fallback to syncfs+unmount(2) that frees the mount and the `syncfs` makes sure any pending request is flushed to the file system

rhatdan: Is the err IsNotExists referring to the Mount point does not exists, won't it still not exists?  So the rest of the calls will fail as well.

rhatdan:
LGTM

PR #465, opened 2019-11-19
Bump to v1.14.0

rhatdan:


rhatdan:
I did not know if counting started at 0 or 1, I can change it back.  I would think this is a minor update since a new interface was added?

vrothberg:
> I did not know if counting started at 0 or 1, I can change it back. I would think this is a minor update since a new interface was added?

Your're right. The new interface requires the bump to 14. Counting starts at 0 though.

PR #467, opened 2019-11-22
Move runroot containers storage into containers directory

rhatdan:
Currently we are throwing lots of files/directories into the
top level of /run/user/UID/  And this makes it difficult to cleanup.
Switching to using /run/user/UID/containers, makes it easier and cleaner.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -70,6 +70,18 @@ func ParseIDMapping(UIDMapSlice, GIDMapSlice []string, subUIDMap, subGIDMap stri
 
 // GetRootlessRuntimeDir returns the runtime directory when running as non root
 func GetRootlessRuntimeDir(rootlessUid int) (string, error) {
+	path, err := getRootlessRuntimeDir(rootlessUid)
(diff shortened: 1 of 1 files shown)


rhatdan:
@mheon @vrothberg @giuseppe @nalind @TomSweeneyRedHat PTAL

This might cause issues with the pause container on rootless.

vrothberg:
containers/libpod#4562 is green :heavy_check_mark:

rhatdan:
@nalind @giuseppe @TomSweeneyRedHat @mheon PTAL

@vrothberg I was testing this with the `podman system reset` PR https://github.com/containers/libpod/pull/4558.  Basically this PR is needed to allow successfully cleaning up storage.

TomSweeneyRedHat:
LGTM
Any doc changes necessary in other places?

mheon:
I don't think this fixes existing containers, but it's a massive improvement over what we have now.

LGTM here.

PR #469, opened 2019-11-26
remove pkg/system/filesys*

vrothberg:
This API caused build regressions as golang.org/x dependencies
introduced breaking changes.  We don't have any callers in our
projects and I consider it a bug that those APIs were there,
which might prevent us from causing a major bump.


vrothberg:
@rhatdan @nalind PTAL. It's meant to fix https://github.com/containers/image/pull/758#issuecomment-558654968

rhatdan:
Tests are not happy with this PR.

vrothberg:
Now missing non-linux cleanups and we should be ready

rhatdan:
Will this cause a Major Version bump?

vrothberg:
> Will this cause a Major Version bump?

On paper, yes. But there are no users outside of c/storage of this specific API in the containers umbrella. At some point, I want to analyze those packages a bit more and move internal APIs to ./internal and remove unused ones entirely. Hence, I consider this specific change a bug fix - although breaking unused APIs.

rhatdan:
LGTM
@nalind @giuseppe @TomSweeneyRedHat @saschagrunert PTAL

zmedico:
I suspect changes from this PR triggered the failure in https://github.com/containers/libpod/pull/4598/checks?check_run_id=325880381.

rhatdan:
@vrothberg PTAL

vrothberg:
Doesn't make much sense to me. This PR should fix exactly those errors.

vrothberg:
This error actually comes from `vendor/github.com/docker/docker/pkg/system/filesys_windows.go:112:24`. This means we either need to downgrade x/sys OR find a way to prune github.com/docker/docker/pkg/system/ from our dependencies.

PR #470, opened 2019-11-27
pkg/archive: limit user xattr value size to math.MaxUint16 bytes

zmedico:
This should avoid the following panic in Lgetxattr:

    panic: runtime error: makeslice: len out of range

Fixes: 0da4bc60b3ba ("pkg/archive: strictly handle errors reading xattrs")
See: https://travis-ci.org/containers/buildah/jobs/617652963
Signed-off-by: Zac Medico <zmedico@gmail.com>

### DIFF
--- a/pkg/system/xattrs_linux.go
+++ b/pkg/system/xattrs_linux.go
@@ -2,12 +2,16 @@ package system
 
 import (
 	"bytes"
+	"math"
 	"syscall"
 
 	"golang.org/x/sys/unix"
 )
 
 const (
+	// Value is larger than the maximum size allowed
+	E2BIG syscall.Errno = unix.E2BIG
+
 	// Operation not supported
 	EOPNOTSUPP syscall.Errno = unix.EOPNOTSUPP
 )
@@ -22,6 +26,9 @@ func Lgetxattr(path string, attr string) ([]byte, error) {
 		return nil, nil
 	}
 	if errno == unix.ERANGE {
+		if sz > math.MaxUint16 {
+			return nil, unix.E2BIG
+		}
 		dest = make([]byte, sz)
 		sz, errno = unix.Lgetxattr(path, attr, dest)
 	}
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

rhatdan:
Still breaking libpod.
https://github.com/containers/libpod/pull/4598/checks?check_run_id=325880381

zmedico:
> Still breaking libpod.
> https://github.com/containers/libpod/pull/4598/checks?check_run_id=325880381

It doesn't seem like 091491bbf78481696a8970150d9326cae11671bc could have triggered
```[+0020s] vendor/github.com/docker/docker/pkg/system/filesys_windows.go:112:24: cannot use uintptr(unsafe.Pointer(&sd[0])) (type uintptr) as type *"golang.org/x/sys/windows".SECURITY_DESCRIPTOR in assignment```
Looking at changes in https://github.com/containers/storage/compare/v1.15.0...v1.15.1, https://github.com/containers/storage/pull/469 seems more likely to blame.

**Comment by rhatdan** [comment]
I think this is still causing issues in Buildah this time.

https://travis-ci.org/containers/buildah/jobs/620107658?utm_medium=notification&utm_source=github_status

File: pkg/system/xattrs_linux.go [hunk 1]
```diff
@@ -22,6 +26,9 @@ func Lgetxattr(path string, attr string) ([]byte, error) {
 		return nil, nil
 	}
 	if errno == unix.ERANGE {
+		if sz > math.MaxUint16 {
+			return nil, unix.E2BIG
```
rhatdan: Shouldn't this be

return unix.E2BIG, nil

rhatdan: Never mind stupid question

pkg/system/xattrs_linux.go
```diff
@@ -22,6 +26,9 @@ func Lgetxattr(path string, attr string) ([]byte, error) {
...
+		}
 		dest = make([]byte, sz)
```
saschagrunert: The issue is that `sz` can still be `-1` here so the above check does not fix.

zmedico: https://github.com/containers/storage/pull/483 looks good. Thanks!

PR #478, opened 2019-12-02
revert golang.org/x/sys

vrothberg:
Revert golang.org/x/sys to a known-to-work version.  Recent breaking
changes in this package introduce build regressions.  Thoss have already
been fixed in containers/storage but other (transitive) dependencies of
consumers of c/storage are not, and hence break builds.

See https://github.com/containers/libpod/pull/4598#issuecomment-560304335
as an example.


vrothberg:
@rhatdan PTAL

rhatdan:
LGTM

PR #480, opened 2019-12-02
Overlayfs has dropped support for skip_mount_home options

rhatdan:
Remove all references to the option in man pages in default configuration.


rhatdan:
@nalind @mheon @vrothberg @TomSweeneyRedHat @giuseppe @saschagrunert PTAL

giuseppe:
LGTM

rhatdan:
Why would we document a NOOP.  I think we should add the ignore back to overlay though.

vrothberg:
> Why would we document a NOOP. I think we should add the ignore back to overlay though.

The intention is to communicate it to the user. I wonder if it shouldn't be conditional on the version

TomSweeneyRedHat:
LGTM but lint is kicking up its heals.

PR #483, opened 2019-12-04
Fix possible runtime panic in Lgetxattr

saschagrunert:
Implementation references:
- Moby: https://github.com/moby/moby/blob/master/pkg/system/xattrs_linux.go
  If [lgetxattr(2)](https://linux.die.net/man/2/lgetxattr) always returns `-1` on any failure, then we cause a runtime panic in L15:  `dest = make([]byte, sz)`
- runc: https://github.com/opencontainers/runc/blob/master/libcontainer/system/xattrs_linux.go
  Looks more correct to me.

Relates to: https://github.com/containers/buildah/pull/1998, https://github.com/containers/storage/pull/470, https://github.com/moby/moby/pull/40283

### DIFF
--- a/pkg/system/xattrs_linux.go
+++ b/pkg/system/xattrs_linux.go
@@ -2,7 +2,6 @@ package system
 
 import (
 	"bytes"
-	"math"
 	"syscall"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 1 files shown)


saschagrunert:
Ready for review, PTAL

rhatdan:
LGTM

TomSweeneyRedHat:
Took me a while to convince myself, but a neat little fix @saschagrunert .  LGTM

saschagrunert:
Actually I believe that this is a bug in moby, I'll address that later on too there.

PR #484, opened 2019-12-04
chrootarchive: invokeUnpack: close pipe on all return paths

vrothberg:
Make sure that the created pipe is closed on all return paths.

### DIFF
--- a/pkg/chrootarchive/archive_unix.go
+++ b/pkg/chrootarchive/archive_unix.go
@@ -96,11 +96,13 @@ func invokeUnpack(decompressedArchive io.Reader, dest string, options *archive.T
 	cmd.Stderr = output
 
 	if err := cmd.Start(); err != nil {
+		w.Close()
 		return fmt.Errorf("Untar error on re-exec cmd: %v", err)
 	}
 
 	//write the options to the pipe for the untar exec to read
 	if err := json.NewEncoder(w).Encode(options); err != nil {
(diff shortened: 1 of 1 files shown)


vrothberg:
@nalind @rhatdan PTAL Cc @baude 

Note, that's not fixing the osp issue.

pkg/chrootarchive/archive_unix.go
```diff
@@ -96,11 +96,13 @@ func invokeUnpack(decompressedArchive io.Reader, dest string, options *archive.T
 	cmd.Stderr = output
 
 	if err := cmd.Start(); err != nil {
+		w.Close()
```
rhatdan: Why not defer?

vrothberg: We must close the stream before `cmd.Wait()` in line 110. A defer would cause two calls to `Close()` in the non-error case.

rhatdan:
LGTM
@nalind @saschagrunert @giuseppe @TomSweeneyRedHat @baude PTAL

TomSweeneyRedHat:
LGTM

giuseppe:
LGTM, is this the root cause for the race issue we are seeing?

vrothberg:
> LGTM, is this the root cause for the race issue we are seeing?

I don't think so. The error happens after. But I noticed this specific issue while reading the code.

PR #486, opened 2019-12-05
Cleanup handling of default driver name, when storage.conf has drivername=""

rhatdan:
If the user specifies drivername="" in the storage.conf
then we default to the first driver name based on priority list, usually
overla.

The problem with this, is now the storage.conf file has options that are
driver specific.  The code does not catch this so ends up looking for a driver
name of "", and finds no matching options.

This code will set the drivername that the storage layer would choose earlies,
so that storage will get the correct options out of the config file.

### DIFF
--- a/store.go
+++ b/store.go
@@ -3278,7 +3278,7 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 	data, err := ioutil.ReadFile(configFile)
 	if err != nil {
 		if !os.IsNotExist(err) {
-			fmt.Printf("Failed to read %s %v\n", configFile, err.Error())
+			fmt.Fprintf(os.Stderr, "Failed to read %s %v\n", configFile, err.Error())
 			return
 		}
 	}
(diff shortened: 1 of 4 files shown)


store.go
```diff
@@ -3354,6 +3354,20 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
...
+	if storeOptions.GraphDriverName == "" {
+		options := drivers.Options{
+			Root:    storeOptions.GraphRoot,
+			RunRoot: storeOptions.RunRoot,
+			UIDMaps: storeOptions.UIDMap,
+			GIDMaps: storeOptions.GIDMap,
+		}
+		name, err := drivers.GetDefaultDriverName(options)
+		if err != nil {
+			fmt.Print(err)
```
nalind: Should the error be returned here, too?

rhatdan: The errors above for parsing the UID Maps were just printing errors, so I was matching, but I am fine with returning an error.  Problem is ReloadConfigurationFile currently does not return errors, so this would need to be an API Breaking change.

rhatdan:
This does not seem to be possible, without a lot of horrible hacking.  So I am going to give up and tell users to specify the driver in the storage name.

PR #487, opened 2019-12-06
tarlogger: drop state mutex

giuseppe:
detect when the read pipe got closed by using the Write error code.
    
It solves a race, where the pipe could be closed between the check and
the actual write.

### DIFF
--- a/pkg/tarlog/tarlogger.go
+++ b/pkg/tarlog/tarlogger.go
@@ -11,7 +11,6 @@ import (
 type tarLogger struct {
 	writer     *io.PipeWriter
 	closeMutex *sync.Mutex
-	stateMutex *sync.Mutex
 	closed     bool
 }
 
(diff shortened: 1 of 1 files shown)


giuseppe:
dropped the WIP, pushed a simpler implementation.

vrothberg:
Amazing work, @giuseppe! I believe we have isolated the sources of the two errors; this and the proxy.

vrothberg:
@nalind @rhatdan PTAL

rhatdan:
LGTM
I would like @nalind to give his ok, and then we need to cut a new release.

vrothberg:
@rhatdan let's wait with a new release until after the tests. I'll set some new ones up with this fix. I ran some without the proxy over night on my machine and didn't see any error anymore. We're getting close.

vrothberg:
All tests were successful :heavy_check_mark:

nalind:
LGTM

mrunalp:
Do we need this for CRI-O 1.16 (for 4.3.0) ?

vrothberg:
> Do we need this for CRI-O 1.16 (for 4.3.0) ?

I'd say, yes. I will identify all impacted releases tomorrow and prepare storage releases for CRI-O and Podman.

PR #488, opened 2019-12-09
Lazy initialize the layer store

saschagrunert:
The layer store gets memoized in any case so we can skip the initial
setup on store load.


saschagrunert:
This seems a good addition to avoid initial locking on storage load.

PTAL

rhatdan:
LGTM
@vrothberg @nalind PTAL

nalind:
Sounds reasonable.  LGTM.

PR #489, opened 2019-12-09
overlay: allow storing images with more than 127 layers

vrothberg:
Note that such images can now be stored but they cannot be
pushed yet.  Supporting both use cases would be ideal for
source-container images that are known to have a large number
of layers, likely exceeding the current limit of 127 layers.

This limit is inherited from Docker to proactively restrict the
amount and size of arguments passed when mounting.  All (lower)
layers must be specified at mount-time but we only have one page
to pass them.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -676,9 +676,6 @@ func (d *Driver) getLower(parent string) (string, error) {
 		parentLowers := strings.Split(string(parentLower), ":")
 		lowers = append(lowers, parentLowers...)
 	}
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

giuseppe:
I think we could also bump `maxDepth` to 500 to reflect the hard limit in the kernel

in any case: LGTM

nalind:
LGTM

saschagrunert:
Do you think you could add an integration test for that as well?

rhatdan:
We can do this in a separate PR.

PR #491, opened 2019-12-10
1.13 update

vrothberg:
Update the 1.13 branch update with all fixes, and add the latest storage fix on-top, so we can release a 1.13.6 needed for Podman and CRI-O.


vrothberg:
@mrunalp @nalind @rhatdan PTAL

nalind:
`release-1.13` is the maintenance branch for CRI-O 1.13.  Is updating the version that it uses a goal here?

vrothberg:
I got confused by the naming. Will open a dedicated branch again.

PR #493, opened 2019-12-12
Remove locking from layer store creation

saschagrunert:
The `layerstore.Load()` on `new[RO]LayerStore` required a held lock on
store initialization. Since the consumers of the layer storage already
call `Load()`, it should not be necessary to lock on initialization of
the layer store.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -239,6 +239,10 @@ type LayerStore interface {
 	// ApplyDiff reads a tarstream which was created by a previous call to Diff and
 	// applies its changes to a specified layer.
 	ApplyDiff(to string, diff io.Reader) (int64, error)
+
+	// LoadLocked wraps Load in a locked state. This means it loads the store
+	// and cleans-up invalid layers if needed.
+	LoadLocked() error
 }
 
 type layerStore struct {
@@ -346,6 +350,7 @@ func (r *layerStore) Load() error {
 	r.byname = names
 	r.bycompressedsum = compressedsums
 	r.byuncompressedsum = uncompressedsums
+
 	// Load and merge information about which layers are mounted, and where.
 	if r.IsReadWrite() {
 		r.mountsLockfile.RLock()
@@ -353,22 +358,23 @@ func (r *layerStore) Load() error {
 		if err = r.loadMounts(); err != nil {
 			return err
 		}
-	}
-	// Last step: if we're writable, try to remove anything that a previous
-	// user of this storage area marked for deletion but didn't manage to
-	// actually delete.
-	if r.IsReadWrite() && r.Locked() {
-		for _, layer := range r.layers {
-			if layer.Flags == nil {
-				layer.Flags = make(map[string]interface{})
-			}
-			if cleanup, ok := layer.Flags[incompleteFlag]; ok {
-				if b, ok := cleanup.(bool); ok && b {
-					err = r.deleteInternal(layer.ID)
-					if err != nil {
-						break
+
+		// Last step: as we’re writable, try to remove anything that a previous
+		// user of this storage area marked for deletion but didn't manage to
+		// actually delete.
+		if r.Locked() {
+			for _, layer := range r.layers {
+				if layer.Flags == nil {
+					layer.Flags = make(map[string]interface{})
+				}
+				if cleanup, ok := layer.Flags[incompleteFlag]; ok {
+					if b, ok := cleanup.(bool); ok && b {
+						err = r.deleteInternal(layer.ID)
+						if err != nil {
+							break
+						}
+						shouldSave = true
 					}
-					shouldSave = true
 				}
 			}
 		}
@@ -376,7 +382,14 @@ func (r *layerStore) Load() error {
 			return r.saveLayers()
 		}
 	}
+
 	return err
+}
+
+func (r *layerStore) LoadLocked() error {
+	r.lockfile.Lock()
+	defer r.lockfile.Unlock()
+	return r.Load()
 }
 
 func (r *layerStore) loadMounts() error {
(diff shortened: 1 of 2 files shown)


saschagrunert:
Let's see if this works...

saschagrunert:
Ready for review, PTAL

layers.go
```diff
@@ -378,6 +378,48 @@ func (r *layerStore) Load() error {
...
+		}
+		err = nil
+	}
+	r.layers = layers
+	r.idindex = truncindex.NewTruncIndex(idlist)
+	r.byid = ids
+	r.byname = names
+	r.bycompressedsum = compressedsums
+	r.byuncompressedsum = uncompressedsums
+	return err
```
saschagrunert: I would appreciate a closer look to this function @nalind. I basically tried to isolate the read-only operations from `Load()`.

vrothberg:
Thanks a lot, @saschagrunert!

Would you mind opening test PRs for Skopeo, Buildah, Podman and CRI-O vendoring this branch? This will help us gain confidence in the change and make sure we didn't miss any potential side effect.

saschagrunert:
> Thanks a lot, @saschagrunert!
> 
> Would you mind opening test PRs for Skopeo, Buildah, Podman and CRI-O vendoring this branch? This will help us gain confidence in the change and make sure we didn't miss any potential side effect.

Yep sure, see the linked PRs. Let's see what the tests say.

layers.go
giuseppe: would it make sense to refactor the two functions into one?  There are not many differences between `Load` and `LoadReadOnly`

rhatdan:
@saschagrunert Looks like something went wrong with your rebase push?

saschagrunert:
Yes, thanks for the hint. I'll have to check if the podman test failures in https://github.com/containers/libpod/pull/4696 are caused by this PR. :thinking:

saschagrunert:
Looks like we have something which finally works. I will check it out with libpod

saschagrunert:
Ready for review PTAL @vrothberg @rhatdan @nalind

TomSweeneyRedHat:
@saschagrunert could you rebase the other project just to double check those too?

saschagrunert:
> @saschagrunert could you rebase the other project just to double check those too?

Yes, I updated those as well:
- https://github.com/cri-o/cri-o/pull/3047
- https://github.com/containers/buildah/pull/2021
- https://github.com/containers/skopeo/pull/788

layers.go
```diff
@@ -346,39 +351,48 @@ func (r *layerStore) Load() error {
...
-			if cleanup, ok := layer.Flags[incompleteFlag]; ok {
-				if b, ok := cleanup.(bool); ok && b {
-					err = r.deleteInternal(layer.ID)
-					if err != nil {
-						break
+
+		// Last step: if we're writable, try to remove anything that a previous
+		// user of this storage area marked for deletion but didn't manage to
+		// actually delete.
+		if r.Locked() {
```
TomSweeneyRedHat: comment doesn't fit any more.  Should the read/write lock really be removed here?

saschagrunert: Hm, no I don't think so because we still check for `r.IsReadWrite()` in Line 363. :)

TomSweeneyRedHat: Ah, I see, I missed the removal of '}' on old line 356.  Can you help my future confusion and change the comment from "if we're writable" -> "as we're writeable"?

saschagrunert: Sure, I changed it as suggested :+1:

layers.go
```diff
@@ -318,9 +321,11 @@ func (r *layerStore) Load() error {
 			ids[layer.ID] = layers[n]
 			idlist = append(idlist, layer.ID)
 			for _, name := range layer.Names {
-				if conflict, ok := names[name]; ok {
-					r.removeName(conflict, name)
-					shouldSave = true
+				if r.Locked() {
```
vrothberg: Please add a comment why this only applies when the store is locked.

layers.go
```diff
@@ -337,7 +342,7 @@ func (r *layerStore) Load() error {
 		}
 		err = nil
 	}
-	if shouldSave && (!r.IsReadWrite() || !r.Locked()) {
+	if r.Locked() && shouldSave && (!r.IsReadWrite() || !r.Locked()) {
```
vrothberg: A comment would be good

layers.go
```diff
@@ -239,6 +239,9 @@ type LayerStore interface {
 	// ApplyDiff reads a tarstream which was created by a previous call to Diff and
 	// applies its changes to a specified layer.
 	ApplyDiff(to string, diff io.Reader) (int64, error)
+
+	// Cleanup re-loads the store and cleans invalid layers
+	Cleanup() error
```
vrothberg: I don't like to name it `Cleanup`. It's effectively a wrapper around `Load` but locked. `LoadLocked` feels more accurate to me. The comment should reflect that it's a wrapper around Load.

saschagrunert: Sure, I fixed it and adapted as suggested.

layers.go
saschagrunert: Probably not needed, I will give a removal of this part another try.

saschagrunert: Looks like it is working, I'm now vendoring it into the related PRs.

rhatdan:
Worked on Buildah.

store.go
```diff
@@ -2788,6 +2788,9 @@ func (s *store) Layers() ([]Layer, error) {
 	if err != nil {
 		return nil, err
 	}
+	if err := lstore.LoadLocked(); err != nil {
+		return nil, err
+	}
```
nalind: This unconditionally reloads the store contents using a write lock here, in addition to where it's conditionally done under a read lock on line 2805.  We'll probably want to drop `lstore` from the initial value of the slice that's constructed on line 2800 to avoid that extra bit of work.

nalind:
I think there's a bit of work we can drop from `Layers()`, otherwise looks fine.

store.go
saschagrunert: Alright I put another commit on top of my changes to see if the new logic breaks anything.

saschagrunert:
Seems to work, squashing commits.

rhatdan:
LGTM

PR #495, opened 2019-12-18
Give devices more time/retries to remove mounts

saschagrunert:
In terms of overlay, we see a lot of flakes in CRI-O CI because there is
no mount available any more, but the directory to be removed is still busy.

To avoid such failure cases we now give the system more time/retries.

---

I saw during debugging no wrong behavior, it was just about a timing
issue during unmount because I think that the disk pressure is pretty
high when running the CRI-O integration tests in parallel.


rhatdan:
@nalind @vrothberg @giuseppe PTAL

giuseppe:
LGTM

PR #496, opened 2019-12-18
Check if the layer is actually mounted.

rhatdan:
There are cases where the storage database gets out of whack with
whether or not the storage is actually mounted.  We need to check
before returning the mount point.

1 A user could go in and umount the storage.
2 If the storage was mounted in a different mount namespace and then
the mount namespace goes away the counter will never get decremented
even though the mount point was removed.
3. If storage runtime is on non tmpfs storage a system reboot could be
done that will not clear the mount count.

This patch will fix the problem with the layer not being mounted, but
we still have a problem in that we can't figure out when to umount the
image. Not sure that is a solveable problem.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -18,6 +18,7 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/ioutils"
+	"github.com/containers/storage/pkg/mount"
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/containers/storage/pkg/system"
 	"github.com/containers/storage/pkg/tarlog"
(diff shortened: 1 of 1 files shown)


rhatdan:
@saschagrunert @giuseppe @vrothberg @nalind PTAL

layers.go
```diff
@@ -776,8 +777,17 @@ func (r *layerStore) Mount(id string, options drivers.MountOpts) (string, error)
...
-		return layer.MountPoint, r.saveMounts()
+		mounted, err := mount.Mounted(layer.MountPoint)
+		if err != nil {
+			return "", err
+		}
+		// If the container is not mounted then we have a condition
+		// where the kernel umounted the mount point. This means
+		// that the mount count never got decremented.
+		if mounted {
+			layer.MountCount++
```
giuseppe: shouldn't we still increase the counter when we mount it again?

rhatdan: I don't think so, since we know that it was unmounted once, that we did not take into account.  At this point the mount count is useless, since we don't know if their are any processes still out their that have mounted the image.  In this case we are likely to leak the mount or at least force the user to have to unmount it more then is necessary to get the mount point to unmount.

The mount count was always an estimation of how many processes needed to mount the storage.

rhatdan: And the mountcount is incremented below. when it falls through to mount the layer

PR #497, opened 2019-12-19
store: keep graph lock during Mount

giuseppe:
This solves a race condition where a mountpoint is created without the
home mount being present.

The cause is that another process could be calling the graph driver
cleanup as part of store.Shutdown() causing the unmount of the
driver home directory.

The unmount could happen between the time the rlstore is retrieved and
the actual mount, causing the driver mount to be done without a home
mount below it.

A third process then would re-create again the home mount, shadowing
the previous mount.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1757845

### DIFF
--- a/store.go
+++ b/store.go
@@ -2479,6 +2479,10 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
+	s.graphLock.Lock()
+	defer s.graphLock.Unlock()
+
 	rlstore.Lock()
 	defer rlstore.Unlock()
 	if modified, err := rlstore.Modified(); modified || err != nil {
@@ -2486,6 +2490,18 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
 			return "", err
 		}
 	}
+
+	/* We need to make sure the home mount is present when the Mount is done.  */
+	if s.graphLock.TouchedSince(s.lastLoaded) {
+		s.graphDriver = nil
+		s.layerStore = nil
+		s.graphDriver, err = s.getGraphDriver()
+		if err != nil {
+			return "", err
+		}
+		s.lastLoaded = time.Now()
+	}
+
 	if rlstore.Exists(id) {
 		options := drivers.MountOpts{
 			MountLabel: mountLabel,


rhatdan:
@nalind PTAL

rhatdan:
This looks good, but we need @vrothberg @nalind and @saschagrunert To take a look.

giuseppe:
dropped the WIP, it is ready for review

store.go
```diff
@@ -2479,13 +2479,29 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
...
 	rlstore.Lock()
 	defer rlstore.Unlock()
 	if modified, err := rlstore.Modified(); modified || err != nil {
 		if err = rlstore.Load(); err != nil {
 			return "", err
 		}
 	}
+
+	/* We need to make sure the home mount is present when the Mount is done.  */
+	if s.graphLock.TouchedSince(s.lastLoaded) {
```
saschagrunert: We probably want to refactor this into an own method because it’s used in multiple places in this file. :)

giuseppe: I had refactored the code in a single function in one of the versions I've pushed yesterday but I reverted the change as the new block is different than other similar cases as it also re-initialize the graphDriver

giuseppe:
> Now everytime we mount a layer, we are locking down the entire storage. While the lock is only necessary the first time, if we leak.

That was my assumption too, but in reality we are already locking/unlocking `s.graphLock` every time we take a store, so the extra cost here should not be noticeable.

> I’m not sure if it’s highly affecting performance since I guess that `Mount()` is not used excessively by API consumers, right? We could vendor it into podman and check if we see any drawbacks in the CI.

I am afraid we are not running enough tests in parallel to detect such issues.

Given the amount of locking that is going on in containers/storage, is there any advantage for Podman to have a separate locking mechanism (via `/dev/shm`)?  If it has clear performance benefits, should we investigate its usage in `containers/storage` as well?

store.go
```diff
@@ -2479,13 +2479,29 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
+	s.graphLock.Lock()
```
vrothberg: Can we move this down to line 2493 and immediately unlock as soon as we don't need? Deferring is comfortable but we could try to keep the critical sections as short as possible - if possible :)

vrothberg:
> > I’m not sure if it’s highly affecting performance since I guess that `Mount()` is not used excessively by API consumers, right? We could vendor it into podman and check if we see any drawbacks in the CI.
> 
> I am afraid we are not running enough tests in parallel to detect such issues.
> 
> Given the amount of locking that is going on in containers/storage, is there any advantage for Podman to have a separate locking mechanism (via `/dev/shm`)? If it has clear performance benefits, should we investigate its usage in `containers/storage` as well?

The only advantage I'd see is that shm locks can be fair while the fcntl file locks are not, such that a process can monopolize a lock. But this specific locks doesn't seem like a high-frequency candidate.

store.go
giuseppe: that could cause a deadlock.  The locks taken on the stores are always taken after the graphLock

giuseppe:
> The only advantage I'd see is that shm locks can be fair while the fcntl file locks are not, such that a process can monopolize a lock. But this specific locks doesn't seem like a high-frequency candidate.

I'd rather see the shm lock go away if possible, as it is rather difficult to use it correctly from nested containers as it needs the same IPC namespace

rhatdan:
Ok I will merge this, since it is critical we move it forward.  @giuseppe @vrothberg Lets continue discussion on improving the patch in a different PR>

PR #508, opened 2020-01-12
drivers: btrfs: Allow unprivileged user to delete subvolumes

t-msn:
rootless podman does not work correctly with btrfs backend since unprivileged subvolume deletion fails: https://github.com/containers/libpod/issues/3963 https://github.com/containers/libpod/issues/4764

This commit fallbacks to rm/rmdir syscalls if subvolume deletion ioctls fails
so that unprivileged user can delete their owned subvolumes.

### DIFF
--- a/drivers/btrfs/btrfs.go
+++ b/drivers/btrfs/btrfs.go
@@ -627,7 +627,12 @@ func (d *Driver) Remove(id string) error {
 	d.updateQuotaStatus()
 
 	if err := subvolDelete(d.subvolumesDir(), id, d.quotaEnabled); err != nil {
-		return err
+		if d.quotaEnabled {
+			return err
+		}
+		// If quota is not enabled, fallback to rmdir syscall to delete subvolumes.
+		// This would allow unprivileged user to delete their owned subvolumes
+		// in kernel >= 4.18 without user_subvol_rm_alowed mount option.
 	}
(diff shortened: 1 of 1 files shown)


rhatdan:

drivers/btrfs/btrfs.go
```diff
@@ -627,7 +627,12 @@ func (d *Driver) Remove(id string) error {
 	d.updateQuotaStatus()
 
 	if err := subvolDelete(d.subvolumesDir(), id, d.quotaEnabled); err != nil {
-		return err
+		if d.quotaEnabled {
+			return err
+		}
+		// If quota is not enabled, fallback to rmdir syscall to delete subvolumes.
+		// This would allow unprivileged user to delete their owned subvolumes
+		// in kernel >= 4.18 without user_subvol_rm_alowed mount option.
```
TomSweeneyRedHat: Is this going to be OK if the kernel is less than 4.18?

rhatdan: I would figure that container/storage with btrfs does not work well on this old of kernel anyways.

t-msn: prior to 4.18, system.EnsureRmoveall() would return error since subvolume cannot be removed (though subvolume's contents are cleaned up). So the difference is from where error retuens and I don't think this would cause any problem.

PR #511, opened 2020-01-14
Remove read-write check for mounts load

saschagrunert:
Loading the mounts should not require write store and may also succeed
on read-only stores.


saschagrunert:
Ah at least now I know what not to do. Closing this now

PR #513, opened 2020-01-17
Handle mount options for overlay2 driver.

rhatdan:
Overlay and overlay2 drivers are the same, but we
were not procesessing mount options correctly for overlay2.


rhatdan:
@mheon @vrothberg @nalind @TomSweeneyRedHat PTAL

rhatdan:
@giuseppe PTAL

mheon:
LGTM

TomSweeneyRedHat:
I think, but I am not 100% sure that all the options in that case statement work for Overlay2 also.  If that's the case, then LGTM.

PR #518, opened 2020-01-24
[WIP] Add `rootless_storage_path` directive to storage.conf

rhatdan:
This allows rootless admins to setup alternative
paths to content in the homedir.

Rootless users on NFS homedirs will not be allowed to run
podman, if an admin wants to setup alternative directory say
in /var/tmp on local storage, they could configure the storage.conf
file and then all users would automatically get storage in /var/tmp.

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -34,6 +34,20 @@ The `storage` table supports the following options:
   container storage graph dir (default: "/var/lib/containers/storage")
   Default directory to store all writable content created by container storage programs.
 
+**rootless_storage_path**="$HOME/.local/share/containers/storage"
+  Storage path for rootless users. By default the graphroot for rootless users
+is set to `$XDG_DATA_HOME/containers/storage`, if XDG_DATA_HOME is set.
+Otherwise the `$HOME/.local/share/containers/storage` is used.  This field can
+be used if administrators need to change the storage location for all users.
+
+    The rootless storage path supports three substations:
+    * `$HOME` => Replaced by the users home directory.
+    * `$UID`  => Replaced by the users UID
+    * `$USER` => Replaced by the users name
+
+  A common use case for this field is `NFS home directories`, which do not work
+with rootless container storage.
+
 **runroot**=""
   container storage run dir (default: "/var/run/containers/storage")
   Default directory to store all temporary writable content created by container storage programs.
--- a/utils.go
+++ b/utils.go
@@ -228,7 +228,21 @@ func DefaultStoreOptions(rootless bool, rootlessUid int) (StoreOptions, error) {
 			}
 			if storageOpts.GraphRoot == "" {
 				storageOpts.GraphRoot = defaultRootlessGraphRoot
+			} else if storageOpts.RootlessStoragePath != "" {
+				home, err := homeDir()
+				if err != nil {
+					return storageOpts, err
+				}
+				rootlessStoragePath := strings.Replace(storageOpts.RootlessStoragePath, "$HOME", home, -1)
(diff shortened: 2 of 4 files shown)


rhatdan:
I wrote this about a week ago, but have not been able to test it.  If anyone wants to grab it and see if it works, It would be appreciated.

The goal is to allow for NFS homedirs support, by specifying an alternate directory for container storage.

docs/containers-storage.conf.5.md
```diff
@@ -34,6 +34,20 @@ The `storage` table supports the following options:
   container storage graph dir (default: "/var/lib/containers/storage")
   Default directory to store all writable content created by container storage programs.
 
+**rootless_storage_path**="$HOME/.local/share/containers/storage"
```
giuseppe: should it be specific to rootless or can we name it `user_storage_path` so that root containers also look it up if `GraphRoot` is not specified?

rhatdan: I would think the graphroot would just get default if not specified, rather then looking for this path.

docs/containers-storage.conf.5.md
```diff
@@ -34,6 +34,20 @@ The `storage` table supports the following options:
...
+  Storage path for rootless users. By default the graphroot for rootless users
+is set to `$XDG_DATA_HOME/containers/storage`, if XDG_DATA_HOME is set.
+Otherwise the `$HOME/.local/share/containers/storage` is used.  This field can
+be used if administrators need to change the storage location for all users.
+
+    The rootless storage path supports three substitions:
```
TomSweeneyRedHat: substitions -> substitutions

docs/containers-storage.conf.5.md
```diff
@@ -34,6 +34,20 @@ The `storage` table supports the following options:
...
+    * `$HOME` => Replaced by the users home directory.
+    * `$UID`  => Replaced by the users UID
+    * `$USER` => Replace by the users name
```
TomSweeneyRedHat: Replace -> Replaced

docs/containers-storage.conf.5.md
```diff
@@ -34,6 +34,20 @@ The `storage` table supports the following options:
...
+
+  A common use case for this field is `NFS home directories`, which is not
+supported for container storage on rootless users.
```
TomSweeneyRedHat: "on rootless" -> "for rootless", or perhaps end with "container storage in a rootles environment."

utils.go
```diff
@@ -256,7 +270,7 @@ func homeDir() (string, error) {
 	if home == "" {
 		usr, err := user.Current()
 		if err != nil {
-			return "", errors.Wrapf(err, "neither XDG_RUNTIME_DIR nor HOME was set non-empty")
+			return "", errors.Wrapf(err, "unable to retrieve the current user information")
```
TomSweeneyRedHat: nice!

docs/containers-storage.conf.5.md
rhatdan: fixed

docs/containers-storage.conf.5.md
rhatdan: Fixed

docs/containers-storage.conf.5.md
rhatdan: I wrote
```  A common use case for this field is `NFS home directories`, which do not work
with rootless container storage.
```

rhatdan:
@QiWang19 could you take this over and make sure it works. Perhaps add tests.

QiWang19:
> @QiWang19 could you take this over and make sure it works. Perhaps add tests.

ok. I'll take a look

rhatdan:
https://github.com/containers/storage/pull/529 replaces this PR.

PR #519, opened 2020-01-26
overlay: drop fuse-overlayfs exception

giuseppe:
now fuse-overlayfs supports running without an upper dir so it doesn't
need to be treated specially.


rhatdan:
@TomSweeneyRedHat @nalind @vrothberg @mheon PTAL

TomSweeneyRedHat:
Nice!  LGTM

PR #521, opened 2020-01-28
Update golangci-lint to v1.23.1 and remove from vendoring

saschagrunert:
It is an upstream recommendation to not vendor golangci-lint and use the
pre-built binary instead. This means we now remove it from the vendored
sources and update the config as well.


.golangci.yml
```diff
@@ -8,18 +8,25 @@ linters:
     - bodyclose
     - depguard
     - gofmt
+    - goprintffuncname
     - interfacer
+    - rowserrcheck
     - typecheck
     # - deadcode
```
vrothberg: Can you change it to a blacklist instead of a whitelist? This way, we can keep track easier of which checks we don't yet enable while making updates easier. Could be a separate PR if you prefer.

saschagrunert: Sure, I added the comments to the `disable` section in the file.

PR #525, opened 2020-01-31
Fix ffjson compilation

saschagrunert:
The ffjson generated sources were broken since quite some time, which is
now fixed again.


rhatdan:

PR #529, opened 2020-02-12
Add `rootless_storage_path` directive to storage.conf

QiWang19:
This allows rootless admins to setup alternative
paths to content in the homedir.

Rootless users on NFS homedirs will not be allowed to run
podman, if an admin wants to setup alternative directory say
in /var/tmp on local storage, they could configure the storage.conf
file and then all users would automatically get storage in /var/tmp.

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -34,6 +34,19 @@ The `storage` table supports the following options:
   container storage graph dir (default: "/var/lib/containers/storage")
   Default directory to store all writable content created by container storage programs.
 
+**rootless_storage_path**="$HOME/.local/share/containers/storage"
+  Storage path for rootless users. By default the graphroot for rootless users
+is set to `$XDG_DATA_HOME/containers/storage`, if XDG_DATA_HOME is set.
+Otherwise `$HOME/.local/share/containers/storage` is used.  This field can
+be used if administrators need to change the storage location for all users.
+
+    The rootless storage path supports three substations:
+    * `$HOME` => Replaced by the users home directory.
+    * `$UID`  => Replaced by the users UID
+    * `$USER` => Replaced by the users name
+
+  A common use case for this field is to provide a local storage directory when user home directories are NFS-mounted (podman does not support container storage over NFS).
+
 **runroot**=""
   container storage run dir (default: "/var/run/containers/storage")
   Default directory to store all temporary writable content created by container storage programs.
--- a/storage.conf
+++ b/storage.conf
@@ -13,6 +13,10 @@ runroot = "/var/run/containers/storage"
 # Primary Read/Write location of container storage
 graphroot = "/var/lib/containers/storage"
 
+# Storage path for rootless users
+#
+# rootless_storage_path = "$HOME/.local/share/containers/storage"
+
 [storage.options]
 # Storage options to be passed to underlying storage drivers
 
@@ -115,7 +119,7 @@ mountopt = "nodev"
 # size = ""
 
 # use_deferred_removal marks devicemapper block device for deferred removal.
-# If the thinpool is in use when the driver attempts to remove it, the driver 
+# If the thinpool is in use when the driver attempts to remove it, the driver
 # tells the kernel to remove it as soon as possible. Note this does not free
 # up the disk space, use deferred deletion to fully remove the thinpool.
 # use_deferred_removal = "True"
--- a/store.go
+++ b/store.go
@@ -139,6 +139,9 @@ type StoreOptions struct {
 	// GraphRoot is the filesystem path under which we will store the
 	// contents of layers, images, and containers.
 	GraphRoot string `json:"root,omitempty"`
+	// RootlessStoragePath is the storage path for rootless users
+	// default $HOME/.local/share/containers/storage
+	RootlessStoragePath string `toml:"rootless_storage_path"`
 	// GraphDriverName is the underlying storage driver that we'll be
 	// using.  It only needs to be specified the first time a Store is
 	// initialized for a given RunRoot and GraphRoot.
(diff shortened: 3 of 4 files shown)


rhatdan:
@QiWang19 Is this supposed to replace https://github.com/containers/storage/pull/518?

We need to have a PR for podman that proves that this works.  Or at least experiment with it to make sure it works correctly.

rhatdan:
@QiWang19 this needs a rebase.

rhatdan:

docs/containers-storage.conf.5.md
```diff
@@ -34,6 +34,20 @@ The `storage` table supports the following options:
   container storage graph dir (default: "/var/lib/containers/storage")
   Default directory to store all writable content created by container storage programs.
 
+**rootless_storage_path**="$HOME/.local/share/containers/storage"
+  Storage path for rootless users. By default the graphroot for rootless users
+is set to `$XDG_DATA_HOME/containers/storage`, if XDG_DATA_HOME is set.
+Otherwise the `$HOME/.local/share/containers/storage` is used.  This field can
```
TomSweeneyRedHat: "Otherwise the" -> "Otherwise"

docs/containers-storage.conf.5.md
```diff
@@ -34,6 +34,20 @@ The `storage` table supports the following options:
...
+be used if administrators need to change the storage location for all users.
+
+    The rootless storage path supports three substations:
+    * `$HOME` => Replaced by the users home directory.
+    * `$UID`  => Replaced by the users UID
+    * `$USER` => Replaced by the users name
+
+  A common use case for this field is `NFS home directories`, which do not work
+with rootless container storage.
```
TomSweeneyRedHat: "do not work" -> "does not work"
or do we want to soften it a bit with something like "which is not currently supported with rootless container storage"?

docs/containers-storage.conf.5.md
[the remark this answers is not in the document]
QiWang19: Fixed. The path is not used, the default is set to "".

storage.conf
```diff
@@ -13,6 +13,10 @@ runroot = "/var/run/containers/storage"
 # Primary Read/Write location of container storage
 graphroot = "/var/lib/containers/storage"
 
+# Storage path for rootless users
+#
+# rootless_storage_path = "$HOME/.config/containers/storage"
```
TomSweeneyRedHat: Maybe I've over code-reviewed today, but the man page seems to say this should be `$HOME/.local/share/containers/storage`?

QiWang19: Are they talking about no need to specify rootless_storage_path = "$HOME/.config/containers/storage" in storage.conf?  I need ask @rhatdan

rhatdan: Yes Tom is correct this should be something like

 rootless_storage_path = "/var/tmp/$USER/containers/storage"

QiWang19: /var/tmp? not $HOME/.config/containers/storage? also change thoses man pages to /var/tmp/$USER/containers/storage?

rhatdan: Let's talk today.  If we are showing the default then it should be
$HOME/.local/share/containers/storage

store.go
```diff
@@ -139,6 +139,9 @@ type StoreOptions struct {
 	// GraphRoot is the filesystem path under which we will store the
 	// contents of layers, images, and containers.
 	GraphRoot string `json:"root,omitempty"`
+	// RooltessStoragePath is the storage path for rootless users
```
edsantiago: typo

utils.go
```diff
@@ -226,6 +229,15 @@ func DefaultStoreOptions(rootless bool, rootlessUID int) (StoreOptions, error) {
 			}
 			if storageOpts.GraphRoot == "" {
 				storageOpts.GraphRoot = defaultRootlessGraphRoot
+			} else if storageOpts.RootlessStoragePath != "" {
+				rootlessStoragePath := strings.Replace(storageOpts.RootlessStoragePath, "$HOME", homedir.Get(), -1)
```
edsantiago: Oh, ouch. This sort of string replacement isn't as simple as that: for instance, what if someone writes `$HOMEDIR` or `$USERNAME`? What you probably want here is something closer to:
```
    look for dollar sign followed by any number of letters
    if those letters are HOME, UID, or USER, replace as needed
    otherwise, throw an error
```

docs/containers-storage.conf.5.md
```diff
@@ -34,6 +34,19 @@ The `storage` table supports the following options:
   container storage graph dir (default: "/var/lib/containers/storage")
   Default directory to store all writable content created by container storage programs.
 
+**rootless_storage_path**="$HOME/.local/share/containers/storage"
```
edsantiago: What is the RHS of the `=` supposed to be? In all other options it's simply an empty string `""`. Why is this one different? Especially why is it misleadingly different? The `=` here implies that it's a default, but right below it says otherwise.

docs/containers-storage.conf.5.md
```diff
@@ -34,6 +34,19 @@ The `storage` table supports the following options:
...
+is set to `$XDG_DATA_HOME/containers/storage`, if XDG_DATA_HOME is set.
+Otherwise `$HOME/.local/share/containers/storage` is used.  This field can
+be used if administrators need to change the storage location for all users.
+
+    The rootless storage path supports three substations:
+    * `$HOME` => Replaced by the users home directory.
+    * `$UID`  => Replaced by the users UID
+    * `$USER` => Replaced by the users name
+
+  A common use case for this field is `NFS home directories`, which is not currently supported with rootless container storage.
```
edsantiago: This is confusing. Perhaps something more like "A common use case for this field is to provide a local storage directory when user home directories are NFS-mounted (podman does not support container storage over NFS)."

docs/containers-storage.conf.5.md
QiWang19: the value of rootless_storage_path will not be used unless the user overwrites it in storage.conf. I think it can be described as "". But I'n not sure. I need ask @rhatdan

edsantiago:
LGTM. Thank you for addressing my concerns.

PR #530, opened 2020-02-13
reexec: drop Pdeathsig

giuseppe:
it can lead to a race condition where the process is killed even if
the parent process is still running.

It can happen as the parent in the prctl(PR_SET_PDEATHSIG) case is
considered the *thread* that created the process.  In the case of the
Go runtime the thread could be freed by the runtime at any time
causing the re-execed process to receive the SIGTERM.

### DIFF
--- a/pkg/reexec/command_linux.go
+++ b/pkg/reexec/command_linux.go
@@ -5,9 +5,6 @@ package reexec
 import (
 	"context"
 	"os/exec"
-	"syscall"
-
-	"golang.org/x/sys/unix"
 )
 
 // Self returns the path to the current process's binary.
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe Do we loose any functionality from removing Pdeathsig?

giuseppe:
> @giuseppe Do we loose any functionality from removing Pdeathsig?

yes the re-exec'ed process won't automatically get a SIGTERM when the parent process exits, but I think that is a better behaviour than getting spurious signals

Before we merge, let's wait for @nalind, I'd like his review on this one

nalind:
Nice catch!  LGTM.

TomSweeneyRedHat:
LGTM
this needs one more approved review and I thought we would have had this by now with the reviews already.

PR #531, opened 2020-02-17
add pkg/homedir

vrothberg:
pkg/homedir is a fork from docker/docker and part of un-dockering the  
direct dependencies of the containers projects.  There is nothing wrong
at all with the Docker packages but there is an increasing risk of     
non-deterministic behavior as Docker does not provide a go module and  
it's release related only to the binary but not to the go libraries.   
Enforcing control over that will prevent us from regressions in the    
future.                                                                
                                                                       
c/storage seems like the best place for this package since it's at the 
very end of the containers/ dependency chain.                          
                                                                       
Note that all forked files have a copyright note to correctly attribute
the original authors.                                                  
                                                                       
### DIFF
--- a/pkg/homedir/homedir_linux.go
+++ b/pkg/homedir/homedir_linux.go
@@ -1,23 +1,96 @@
-// +build linux
-
 package homedir
 
-import (
-	"os"
+// Copyright 2013-2018 Docker, Inc.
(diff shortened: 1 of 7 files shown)


vrothberg:
@nalind @rhatdan @giuseppe PTAL

PR #540, opened 2020-02-27
ioutils: use Fdatasync instead of Sync

giuseppe:
since we don't care of other inode attributes, but only those required
to retrieve the data, use the slightly faster fdatasync.

### DIFF
--- a/pkg/ioutils/fswriters_linux.go
+++ b/pkg/ioutils/fswriters_linux.go
@@ -1,0 +1,11 @@
+package ioutils
+
+import (
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+func fdatasync(f *os.File) error {
+	return unix.Fdatasync(int(f.Fd()))
+}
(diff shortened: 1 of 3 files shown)


rhatdan:
@giuseppe You need to deal with cross compile.

TomSweeneyRedHat:
LGTM

rhatdan:
I'm fine with this, but would like to hear @nalind opinion?

pkg/ioutils/fswriters_linux.go
```diff
@@ -0,0 +1,11 @@
+package ioutils
+
+import (
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+func fdatasync(f *os.File) error {
+	return unix.Fdatasync(int(f.Fd()))
```
nalind: fdatasync(3p) says this is in POSIX, but in golang.org/x/sys/unix, only Linux, Solaris, and AIX have a corresponding function.  I don't think we target Solaris or AIX, so this should be fine.

nalind:
LGTM

PR #541, opened 2020-02-27
store: make DeleteContainer parallel

giuseppe:
the operations have no dependency on each other so they can be
performed in parallel.

### DIFF
--- a/store.go
+++ b/store.go
@@ -2320,24 +2320,53 @@ func (s *store) DeleteContainer(id string) error {
 
 	if rcstore.Exists(id) {
 		if container, err := rcstore.Get(id); err == nil {
+			errChan := make(chan error)
+			var wg sync.WaitGroup
+
 			if rlstore.Exists(container.LayerID) {
-				if err = rlstore.Delete(container.LayerID); err != nil {
-					return err
-				}
+				wg.Add(1)
+				go func() {
+					errChan <- rlstore.Delete(container.LayerID)
+					wg.Done()
+				}()
 			}
-			if err = rcstore.Delete(id); err != nil {
-				return err
-			}
+			wg.Add(1)
+			go func() {
+				errChan <- rcstore.Delete(id)
+				wg.Done()
+			}()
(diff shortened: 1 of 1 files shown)


rhatdan:
@saschagrunert @nalind @vrothberg PTAL

giuseppe:
marked as WIP for now, need to think more of it.

The cost I want to avoid is the double sync that each store does on the atomic write

store.go
```diff
@@ -2320,24 +2320,53 @@ func (s *store) DeleteContainer(id string) error {
 
 	if rcstore.Exists(id) {
 		if container, err := rcstore.Get(id); err == nil {
+			var (
+				wg        sync.WaitGroup
+				rlError   error
+				rcError   error
+				rmGcError error
+				rmRcError error
+			)
```
vrothberg: Can we replace the variables above with one error channel? I think that would be a bit more idiomatic and ultimately less lines of code. We can wrap the errors into one before returning.

giuseppe: yes sure, fixed in the new version.

rhatdan:
@saschagrunert PTAL

nalind:
It gets harder to clean up things if we fail to delete the layer for any reason.  I don't know how often that comes up, but it leaves me a bit uneasy.

I'm not entirely sure about doing the waitgroup `Wait` in a goroutine and using a channel instead of just using variables to hold the nil-or-error return values we want to examine, but that's not a blocker.

saschagrunert:
> I'm not entirely sure about doing the waitgroup `Wait` in a goroutine and using a channel instead of just using variables to hold the nil-or-error return values we want to examine, but that's not a blocker.

I thought the same and concluded for myself that a wait group is appropriate here. :innocent:

rhatdan:
@nalind Did you just give a luke warm LGTM?

nalind:
The inability to retry a failed cleanup still makes me uneasy, but I'm not going to block on that.

PR #542, opened 2020-02-28
fix else case for rootless storage path and path match

QiWang19:
move the assignment of rootless_storage_path out of `else` block to make the config work even the graphroot is empty.
fix the path match error

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -230,17 +230,11 @@ func DefaultStoreOptions(rootless bool, rootlessUID int) (StoreOptions, error) {
 			}
 			if storageOpts.GraphRoot == "" {
 				storageOpts.GraphRoot = defaultRootlessGraphRoot
-			} else if storageOpts.RootlessStoragePath != "" {
-				splitPaths := strings.SplitAfter(storageOpts.RootlessStoragePath, "$")
-				validEnv := regexp.MustCompile(`^(HOME|USER|UID)[^a-zA-Z]`).MatchString
-				if len(splitPaths) > 1 {
-					for _, path := range splitPaths {
-						if !validEnv(path) {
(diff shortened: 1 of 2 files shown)


utils.go
```diff
@@ -230,11 +230,12 @@ func DefaultStoreOptions(rootless bool, rootlessUID int) (StoreOptions, error) {
 			}
 			if storageOpts.GraphRoot == "" {
 				storageOpts.GraphRoot = defaultRootlessGraphRoot
-			} else if storageOpts.RootlessStoragePath != "" {
+			}
+			if storageOpts.RootlessStoragePath != "" {
 				splitPaths := strings.SplitAfter(storageOpts.RootlessStoragePath, "$")
-				validEnv := regexp.MustCompile(`^(HOME|USER|UID)[^a-zA-Z]`).MatchString
+				validEnv := regexp.MustCompile(`^(HOME|USER|UID)([^a-zA-Z]|$)`).MatchString
```
vrothberg: Why do we need to change the regex?

@QiWang19, could you break the code in this if branch into a dedicated function and add unit tests? We have a long tradition in regressing with parsing code.

utils_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	validPaths := []string{
+		"$HOME",
+		"$HOME/",
+		"/test/path",
+		"/test/$HOME",
+		"/test/$HOME/path",
+		"/test/$HOME/$USER/$UID",
+		"/test/$HOME/$USER/$UID/path",
+		"$HOME/$USER/$UID/path",
```
rhatdan: What about these valid paths?
/tmp/$HOME$USER$UID
/tmp/$HOME-$USER-$UID

PR #546, opened 2020-03-02
lockfile: drop call to unix.Fsync

giuseppe:
do not flush the modified metadata to disk each time the store is
modified.

It is not needed for other processes stat'ing the lock file to see the
updated mtime.

introduced with commit 9b0633295b55d9b52d909a4fb8e3b6e84265c2cf

### DIFF
--- a/pkg/lockfile/lockfile_unix.go
+++ b/pkg/lockfile/lockfile_unix.go
@@ -206,10 +206,6 @@ func (l *lockfile) Touch() error {
 	if n != len(id) {
 		return unix.ENOSPC
 	}
-	err = unix.Fsync(int(l.fd))
-	if err != nil {
-		return err
-	}
 	return nil
 }
 


giuseppe:
@nalind it is almost 4 years old patch now, but do you remember why the Fsync was needed?

`TouchedSince` should still work without the `fsync` as it sees the updated mtime.

It seems to be very expensive operation:

without the patch:
```
$ \time bin/podman  run --rm -ti alpine true
0.13user 0.11system 0:02.55elapsed 9%CPU (0avgtext+0avgdata 59320maxresident)k
16inputs+936outputs (0major+9130minor)pagefaults 0swaps
```

with the patch:
```
$ \time bin/podman  run --rm -ti alpine true
0.19user 0.12system 0:01.78elapsed 17%CPU (0avgtext+0avgdata 38492maxresident)k
16inputs+896outputs (0major+4194minor)pagefaults 0swaps

```

TomSweeneyRedHat:
I'll go with whatever @nalind says, but I'm leery about not waiting for the updated lock information to get written out to storage if we remove that fsync.

nalind:
> @nalind it is almost 4 years old patch now, but do you remember why the Fsync was needed?
> 
> `TouchedSince` should still work without the `fsync` as it sees the updated mtime.

What's ensuring that the file contents and mtime are updated after we remove this call?  I'm referencing the notes in write(2) and the comparison between the behaviors of `fsync` and `fdatasync` in fsync(2).

giuseppe:
> What's ensuring that the file contents and mtime are updated after we remove this call? I'm referencing the notes in write(2) and the comparison between the behaviors of `fsync` and `fdatasync` in fsync(2).

my understanding of it is that `fsync/fdatasync` force the flush to the permanent storage, so the data can be retrieved even in case of a system crash.  And we are not covering this use case since the crash can happen between the write and the fsync itself, the correct way of doing it would be like in `pkg/ioutils/NewAtomicFileWriter` that is write+fsync+rename.  In memory updates to the inode should be immediately visible to the other processes.

@rhvgoyal could you please confirm it?  Or do we need fsync here?

rhvgoyal:
I am not sure what's the full context of conversation. But I agree that fsync is required only if we want to ensure that changes are persistent on disk after crash. It should not be required for mtime update. That will be in VFS as inode property and should be visible to other users.

So will need little more context of what operation we are doing and what kind of crash guarantee we need to be able to tell what fsync operation is required.

In general, Jeff Moyer in this LWN article gives following summary if we want to make sure a temp file is persistent after a rename.

https://lwn.net/Articles/457667/

1. create a new temp file (on the same file system!)
2. write data to the temp file
3. fsync() the temp file
4. rename the temp file to the appropriate name
5. fsync() the containing directory

rhatdan:
After talking to Vivek, he agrees with @giuseppe that the Fsync is not needed and we should drop it.

PR #550, opened 2020-03-04
go mod tidy

kolyshkin:
Looks like go.mod and go.sum contain a lot of cruft
accumulated over time, like indirect dependencies
or packages that were removed later.

This patch is a result of running `go mod tidy`
(from go-1.13.6) in root and `tests/tools` directories.


rhatdan:

TomSweeneyRedHat:
It looks OK to me, but would like @vrothberg to merge after his review.

PR #551, opened 2020-03-07
pkg/mount: optimize/cleanup

kolyshkin:
This tries to bring `pkg/mount` package to a better shape, mostly reshuffling things here and there, removing redundant and excessive calls to parse `/proc/self/mountinfo`, etc.

Please see individual commits for detailed descriptions.

:warning: **This needs a very thorough review**

PS In fact, this might fix some bugs, too, since reading `/proc/self/mountinfo` from the kernel proved to be unreliable/racy. In case there are multiple mounts and unmounts happening while reading /proc/self/mountinfo, the kernel can only give you half the mounts (if you're curious, see https://github.com/kolyshkin/procfs-test for more details and repro).

PPS Once this lands, I want to optimize the mountinfo parser which is still used for at least `Mount()`. Probably makes sense to also drop Solaris support.

### DIFF
--- a/pkg/mount/sharedsubtree_linux.go
+++ b/pkg/mount/sharedsubtree_linux.go
@@ -1,5 +1,3 @@
-// +build linux
-
 package mount
 
 // MakeShared ensures a mounted filesystem has the SHARED mount option enabled.
--- a/pkg/mount/sharedsubtree_linux.go
+++ b/pkg/mount/sharedsubtree_linux.go
@@ -3,62 +3,62 @@ package mount
 // MakeShared ensures a mounted filesystem has the SHARED mount option enabled.
 // See the supported options in flags.go for further reference.
 func MakeShared(mountPoint string) error {
-	return ensureMountedAs(mountPoint, "shared")
+	return ensureMountedAs(mountPoint, SHARED)
 }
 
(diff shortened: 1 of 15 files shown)


rhatdan:
Removing ForceUmount will cause a API change.

rhatdan:
@vrothberg @nalind @TomSweeneyRedHat PTAL

pkg/mount/sharedsubtree_linux.go
```diff
@@ -1,69 +1,64 @@
-// +build linux
-
 package mount
 
 // MakeShared ensures a mounted filesystem has the SHARED mount option enabled.
 // See the supported options in flags.go for further reference.
 func MakeShared(mountPoint string) error {
-	return ensureMountedAs(mountPoint, "shared")
+	return ensureMountedAs(mountPoint, SHARED)
```
TomSweeneyRedHat: love these changes!

TomSweeneyRedHat:
LGTM
Really nice clean up overall.  I would NOT remove ForceUnmount, but would mark it with a deprecated comment or some such.
Definitely want a head nod from @vrothberg and/or @nalind

kolyshkin:
Agree wrt `ForceUnmount()` removal; I now made it deprecated instead.

PR #554, opened 2020-03-09
pkg/system, pkg/mount: assorted improvements

kolyshkin:
* improved error reporting for mount.Mount()/Unmount() and system.[L]Stat()
* dropped Solaris and FreeBSD support for pkg/mount
* optimizations for `system.EnsureRemoveAll()` and `mount.RecursiveUnmount()` (same as https://github.com/moby/moby/pull/40637)


pkg/system/rm.go
```diff
@@ -63,12 +63,8 @@ func EnsureRemoveAll(dir string) error {
...
 		}
 
-		if mounted, _ := mount.Mounted(pe.Path); mounted {
-			if e := mount.Unmount(pe.Path); e != nil {
-				if mounted, _ := mount.Mounted(pe.Path); mounted {
-					return errors.Wrapf(e, "error while removing %s", dir)
-				}
-			}
+		if e := mount.Unmount(pe.Path); e != nil {
+			return errors.Wrapf(e, "error while removing %s", dir)
```
rhatdan: In the existing code, the Unmount command would not be called, and we would not get an error if the code was not Mounted,  Why not check if the error is EINVAL, and then check if it is Mounted, and not return an error.
Then you are fast in the default path, and continue the functionality of not returning EINVAL errors.

pkg/mount/mount.go
```diff
@@ -116,9 +116,7 @@ func RecursiveUnmount(target string) error {
 			continue
 		}
 		if err := Unmount(m.Mountpoint); err != nil && i == len(mounts)-1 {
-			if mounted, err := Mounted(m.Mountpoint); err != nil || mounted {
-				return err
-			}
+			return err
```
rhatdan: We do not want to bubble errors when some other tool unmounted the file system.

pkg/system/rm.go
kolyshkin: `mount.Unmount()` already ignores `EINVAL`

pkg/mount/mount.go
kolyshkin: same here, `Mount()` ignores EINVAL

kolyshkin:
Do we want to support darwin here? Currently, CI is failing on darwin as I removed some support for "unsupported" platforms.

rhatdan:
Yes, when Container/Storage gets vendored into podman, it currently gets installed for MAC/WIndows platforms. Eventually this should be cleaned up, since we don't actually support C/S on a MAC, but for now podman-remote requires this to compile correctly.

rhatdan:
LGTM

PR #555, opened 2020-03-09
pkg/mount: mountinfo parser speed up

kolyshkin:
The mountinfo parser implemented via `fmt.Sscanf()` is slower than the one
using `strings.Split()` and `strconv.Atoi()`. This rewrite helps to speed it
up to a factor of 8x, here is a result from go bench:

> BenchmarkParsingScanf-4            300          22294112 ns/op
> BenchmarkParsingSplit-4           3000           2780703 ns/op

I also tried other approaches, such as using `fmt.Sscanf()` for the first
three (integer) fields and `strings.Split()` for the rest, but it slows
things down considerably:

> BenchmarkParsingMixed-4           1000           8827058 ns/op

Note the old code uses `fmt.Sscanf`, when a linear search for '-' field,
when a split for the last 3 fields. The new code relies on one single
split.

I have also added more comments to aid in future development.

Finally, the test data is fixed to now have white space before the first field.

### DIFF
--- a/pkg/mount/mountinfo_linux.go
+++ b/pkg/mount/mountinfo_linux.go
@@ -5,25 +5,10 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"strings"
-)
 
-const (
-	/* 36 35 98:0 /mnt1 /mnt2 rw,noatime master:1 - ext3 /dev/root rw,errors=continue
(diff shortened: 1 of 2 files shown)


giuseppe:
great, thanks for looking at it.  I also found out the parsing of mountinfo to be one of the slowest operations we have.

LGTM

PR #556, opened 2020-03-11
Make linters happy

rhatdan:
Check lots of potential error situations.

### DIFF
--- a/pkg/mount/mount_unix_test.go
+++ b/pkg/mount/mount_unix_test.go
@@ -42,14 +42,22 @@ func TestMounted(t *testing.T) {
 		targetPath = path.Join(targetDir, "file.txt")
 	)
 
-	os.Mkdir(sourceDir, 0777)
-	os.Mkdir(targetDir, 0777)
+	if err := os.Mkdir(sourceDir, 0777); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := os.Mkdir(targetDir, 0777); err != nil {
+		t.Fatal(err)
+	}
 
 	f, err := os.Create(sourcePath)
 	if err != nil {
 		t.Fatal(err)
 	}
-	f.WriteString("hello")
+	if _, err = f.WriteString("hello"); err != nil {
+		t.Fatal(err)
+	}
+
 	f.Close()
 
 	f, err = os.Create(targetPath)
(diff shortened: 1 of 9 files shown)


rhatdan:
@vrothberg @giuseppe @nalind @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
This looks like it's ready to be merged, but GitHub is "blocking" it for approvers?

pkg/mount/mount_unix_test.go
```diff
@@ -122,7 +137,7 @@ func TestMountReadonly(t *testing.T) {
 		}
 	}()
 
-	f, err = os.OpenFile(targetPath, os.O_RDWR, 0777)
+	_, err = os.OpenFile(targetPath, os.O_RDWR, 0777)
```
kolyshkin: That way you may leak an opened file descriptor, which is totally OK in this test, but I guess some other linter is going to complain.

So maybe something like
```go
if f, err := os.OpenFile(...); err == nil {
    f.Close()
￼		t.Fatal("Should not be able to open a ro file as rw")
￼	}
```

pkg/reexec/reexec_test.go
```diff
@@ -51,7 +51,7 @@ func TestCommand(t *testing.T) {
 func TestCommandContext(t *testing.T) {
 	stdout := &bytes.Buffer{}
 
-	ctx, _ := context.WithDeadline(context.TODO(), time.Now().Add(5*time.Second))
+	ctx, finish := context.WithDeadline(context.TODO(), time.Now().Add(5*time.Second))
```
kolyshkin: Add `defer finish()` here, otherwise it's not called in case of an error and some linter is going to complain

pkg/system/rm.go
```diff
@@ -29,7 +29,9 @@ func EnsureRemoveAll(dir string) error {
 	maxRetry := 100
 
 	// Attempt to unmount anything beneath this dir first
-	mount.RecursiveUnmount(dir)
+	if err := mount.RecursiveUnmount(dir); err != nil {
```
kolyshkin: The comment says "attempt" so I guess we'd rather ignore an error here and still try to remove what we can (and then return an error from `RemoveAll()` if there is any).

rhatdan: Fixed to just print debug messages on failures.

pkg/reexec/reexec_test.go
rhatdan: Fixed

pkg/mount/mount_unix_test.go
rhatdan: Fixed

pkg/system/rm.go
```diff
@@ -29,7 +30,9 @@ func EnsureRemoveAll(dir string) error {
...
+		logrus.Debugf("RecusiveUnmount on %s failed: %v", dir, err)
```
kolyshkin: Since commit 8b67934cbcc1df the unmount error is verbose enough per se, producing something like

`umount <dir>, flags 0x01: device busy`

so you don't have to print `dir` anymore. IOW

```suggestion
		logrus.Debugf("EnsureRemoveAll warning: %v", err)
```

PR #561, opened 2020-03-13
Make mount.Mounted() capable of handling files

saschagrunert:
The function `fileutils.ReadSymlinkedDirectory` makes `mount.Mounted()`
fail if the mount is a file. We now only check that if the mount is a
directory to allow handling files in `mount.Mounted()`.

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -1,6 +1,7 @@
 package mount
 
 import (
+	"os"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/mount/mount.go
```diff
@@ -56,10 +57,17 @@ func Mounted(mountpoint string) (bool, error) {
...
 	if err != nil {
 		return false, err
 	}
+	if info.IsDir() {
+		mountpoint, err = fileutils.ReadSymlinkedDirectory(mountpoint)
+		if err != nil {
+			return false, err
+		}
+	}
+
```
TomSweeneyRedHat: If it is NOT a dir, should we `return false, nil` in an else statement?

saschagrunert: No, then we just should look if the file is in the mounts :)

TomSweeneyRedHat: @saschagrunert I'll trust your judgement on this, but I thought all mountpoints had to be directories in Linux?  New learning for me?

saschagrunert: Yeah for pinned linux namespaces in CRI-O we use one file mount per namespace. The issue was that the previous implementation would error out for the pinned namespaces, which cause  stale mounts.

rhatdan:
@nalind @giuseppe @vrothberg @TomSweeneyRedHat @mrunalp PTAL

PR #562, opened 2020-03-13
config: honor XDG_CONFIG_HOME

giuseppe:
when the env variable is set, use its value for locating the user
configuration file.

### DIFF
--- a/store.go
+++ b/store.go
@@ -3308,6 +3308,9 @@ const defaultConfigFile = "/etc/containers/storage.conf"
 // DefaultConfigFile returns the path to the storage config file used
 func DefaultConfigFile(rootless bool) (string, error) {
 	if rootless {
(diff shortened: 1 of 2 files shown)


store.go
```diff
@@ -3308,6 +3308,9 @@ const defaultConfigFile = "/etc/containers/storage.conf"
 // DefaultConfigFile returns the path to the storage config file used
 func DefaultConfigFile(rootless bool) (string, error) {
 	if rootless {
+		if configHome := os.Getenv("XDG_CONFIG_HOME"); configHome != "" {
```
rhatdan: Should this be documented in the containers.storage.conf man page?

TomSweeneyRedHat: LGTM
I'd be OK with a man page addition, but I am OK without if it causes angst in others.

giuseppe: yes, I've pushed a new version with a change for the man page

PR #566, opened 2020-03-17
mount.Mounted: normalize mountpoints

vrothberg:
Commit 90513cbb96a7 introduced a regression as specified mountpoints
must be normalized (e.g., "" -> "/") before we can deal with them.
Instead of working around the fileutils package and special casing
files, add a new `fileutils.ReadSymlinkedPath` which can work with
directories and files.

### DIFF
--- a/pkg/fileutils/fileutils.go
+++ b/pkg/fileutils/fileutils.go
@@ -1,7 +1,6 @@
 package fileutils
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -10,6 +9,7 @@ import (
 	"strings"
 	"text/scanner"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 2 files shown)


vrothberg:
@saschagrunert @giuseppe @rhatdan PTAL

pkg/fileutils/fileutils.go
```diff
@@ -358,6 +358,25 @@ func ReadSymlinkedDirectory(path string) (string, error) {
 	return realPath, nil
 }
 
+// ReadSymlinkedDirectory returns the target directory of a symlink.
+// The target of the symbolic link can be a file and a directory.
+func ReadSymlinkedPath(path string) (string, error) {
```
saschagrunert:
```suggestion
func ReadSymlinkedPath(path string) (realpath string, err error) {
```

pkg/fileutils/fileutils.go
```diff
@@ -358,6 +358,25 @@ func ReadSymlinkedDirectory(path string) (string, error) {
...
+	var realPath string
+	var err error
+	if realPath, err = filepath.Abs(path); err != nil {
+		return "", fmt.Errorf("unable to get absolute path for %s: %s", path, err)
```
saschagrunert:
```suggestion
		return "", errors.Wrapf(err, "unable to get absolute path for %s", path)
```

vrothberg:
Updated. After fighting VSCode which is not behaving lately :-D

mtrmac:
If this was an important regression (was it?), should there be a test?

vrothberg:
> If this was an important regression (was it?), should there be a test?

It was an important one as it broke large parts of Podman. Tests would be ideal, absolutely!

PR #568, opened 2020-03-17
Use temp instead of run as fallback directory for rootless mode

andrew-aladev:
[`XDG_*` directories](https://github.com/moby/moby/issues/40708) will be a big question. But please do not use `/run/user` as fallback directory, use something like `/tmp/storage-user` instead.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -2,6 +2,7 @@ package storage
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"os/user"
@@ -82,31 +83,92 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
 	return path, nil
 }
 
-func getRootlessRuntimeDir(rootlessUID int) (string, error) {
-	runtimeDir, err := homedir.GetRuntimeDir()
+type rootlessRuntimeDirEnvironment interface {
+	getProcCommandFile() string
+	getRunUserDir() string
+	getTmpPerUserDir() string
+
+	homeDirGetRuntimeDir() (string, error)
+	systemLstat(string) (*system.StatT, error)
+	homedirGet() string
+}
+
+type rootlessRuntimeDirEnvironmentImplementation struct {
+	procCommandFile string
+	runUserDir      string
+	tmpPerUserDir   string
+}
+
+func (env rootlessRuntimeDirEnvironmentImplementation) getProcCommandFile() string {
+	return env.procCommandFile
+}
+func (env rootlessRuntimeDirEnvironmentImplementation) getRunUserDir() string {
+	return env.runUserDir
+}
+func (env rootlessRuntimeDirEnvironmentImplementation) getTmpPerUserDir() string {
+	return env.tmpPerUserDir
+}
+func (rootlessRuntimeDirEnvironmentImplementation) homeDirGetRuntimeDir() (string, error) {
+	return homedir.GetRuntimeDir()
+}
+func (rootlessRuntimeDirEnvironmentImplementation) systemLstat(path string) (*system.StatT, error) {
+	return system.Lstat(path)
+}
+func (rootlessRuntimeDirEnvironmentImplementation) homedirGet() string {
+	return homedir.Get()
+}
+
+func isRootlessRuntimeDirOwner(dir string, env rootlessRuntimeDirEnvironment) bool {
+	st, err := env.systemLstat(dir)
+	return err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000
+}
+
+// getRootlessRuntimeDirIsolated is an internal implementation detail of getRootlessRuntimeDir to allow testing.
+// Everyone but the tests this is intended for should only call getRootlessRuntimeDir, never this function.
+func getRootlessRuntimeDirIsolated(env rootlessRuntimeDirEnvironment) (string, error) {
+	runtimeDir, err := env.homeDirGetRuntimeDir()
 	if err == nil {
 		return runtimeDir, nil
 	}
-	tmpDir := fmt.Sprintf("/run/user/%d", rootlessUID)
-	st, err := system.Stat(tmpDir)
-	if err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000 {
-		return tmpDir, nil
+
+	initCommand, err := ioutil.ReadFile(env.getProcCommandFile())
+	if err != nil || string(initCommand) == "systemd" {
+		runUserDir := env.getRunUserDir()
+		if isRootlessRuntimeDirOwner(runUserDir, env) {
+			return runUserDir, nil
+		}
 	}
-	tmpDir = fmt.Sprintf("%s/%d", os.TempDir(), rootlessUID)
-	if err := os.MkdirAll(tmpDir, 0700); err != nil {
-		logrus.Errorf("failed to create %s: %v", tmpDir, err)
-	} else {
-		return tmpDir, nil
+
+	tmpPerUserDir := env.getTmpPerUserDir()
+	if _, err := env.systemLstat(tmpPerUserDir); os.IsNotExist(err) {
+		if err := os.Mkdir(tmpPerUserDir, 0700); err != nil {
+			logrus.Errorf("failed to create temp directory for user: %v", err)
+		} else {
+			return tmpPerUserDir, nil
+		}
+	} else if isRootlessRuntimeDirOwner(tmpPerUserDir, env) {
+		return tmpPerUserDir, nil
 	}
-	home := homedir.Get()
-	if home == "" {
-		return "", errors.Wrapf(err, "neither XDG_RUNTIME_DIR nor HOME was set non-empty")
+
+	homeDir := env.homedirGet()
+	if homeDir == "" {
+		return "", errors.New("neither XDG_RUNTIME_DIR not temp dir nor HOME was set non-empty")
 	}
-	resolvedHome, err := filepath.EvalSymlinks(home)
+	resolvedHomeDir, err := filepath.EvalSymlinks(homeDir)
 	if err != nil {
-		return "", errors.Wrapf(err, "cannot resolve %s", home)
+		return "", errors.Wrapf(err, "cannot resolve %s", homeDir)
 	}
-	return filepath.Join(resolvedHome, "rundir"), nil
+	return filepath.Join(resolvedHomeDir, "rundir"), nil
+}
+
+func getRootlessRuntimeDir(rootlessUID int) (string, error) {
+	return getRootlessRuntimeDirIsolated(
+		rootlessRuntimeDirEnvironmentImplementation{
+			"/proc/1/comm",
+			fmt.Sprintf("/run/user/%d", rootlessUID),
+			fmt.Sprintf("%s/containers-user-%d", os.TempDir(), rootlessUID),
+		},
+	)
 }
 
 // getRootlessDirInfo returns the parent path of where the storage for containers and
--- a/utils_test.go
+++ b/utils_test.go
@@ -1,8 +1,14 @@
 package storage
 
 import (
+	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"testing"
 
+	"github.com/containers/storage/pkg/homedir"
+	"github.com/containers/storage/pkg/system"
 	"gotest.tools/assert"
 )
 
@@ -41,4 +47,251 @@ func TestValidStoragePathFormat(t *testing.T) {
 		err := validRootlessStoragePathFormat(path)
 		assert.NilError(t, err)
 	}
+}
+
+type homeRuntimeData struct {
+	dir   string
+	error error
+}
+
+type rootlessRuntimeDirEnvironmentTest struct {
+	homeRuntime     homeRuntimeData
+	procCommandFile string
+	runUserDir      string
+	tmpPerUserDir   string
+	homeDir         string
+	result          string
+}
+
+func (env rootlessRuntimeDirEnvironmentTest) getProcCommandFile() string {
+	return env.procCommandFile
+}
+func (env rootlessRuntimeDirEnvironmentTest) getRunUserDir() string {
+	return env.runUserDir
+}
+func (env rootlessRuntimeDirEnvironmentTest) getTmpPerUserDir() string {
+	return env.tmpPerUserDir
+}
+func (env rootlessRuntimeDirEnvironmentTest) homeDirGetRuntimeDir() (string, error) {
+	return env.homeRuntime.dir, env.homeRuntime.error
+}
+func (env rootlessRuntimeDirEnvironmentTest) systemLstat(path string) (*system.StatT, error) {
+	return system.Lstat(path)
+}
+func (env rootlessRuntimeDirEnvironmentTest) homedirGet() string {
+	return env.homeDir
+}
+
+func TestRootlessRuntimeDir(t *testing.T) {
+	testDir, err := ioutil.TempDir("", "rootless-runtime-dir-test")
+	assert.NilError(t, err)
+	defer os.Remove(testDir)
+
+	homeRuntimeDir := filepath.Join(testDir, "home-rundir")
+	err = os.Mkdir(homeRuntimeDir, 0700)
+	assert.NilError(t, err)
+
+	homeRuntimeDisabled := homeRuntimeData{error: errors.New("homedirGetRuntimeDir is disabled")}
+
+	systemdCommandFile := filepath.Join(testDir, "systemd-command")
+	err = ioutil.WriteFile(systemdCommandFile, []byte("systemd"), 0644)
+	assert.NilError(t, err)
+
+	initCommandFile := filepath.Join(testDir, "init-command")
+	err = ioutil.WriteFile(initCommandFile, []byte("init"), 0644)
+	assert.NilError(t, err)
+
+	dirForOwner := filepath.Join(testDir, "dir-for-owner")
+	err = os.Mkdir(dirForOwner, 0700)
+	assert.NilError(t, err)
+
+	dirForAll := filepath.Join(testDir, "dir-for-all")
+	err = os.Mkdir(dirForAll, 0777)
+	assert.NilError(t, err)
+
+	dirToBeCreated := filepath.Join(testDir, "dir-to-be-created")
+
+	envs := []rootlessRuntimeDirEnvironmentTest{
+		{
+			homeRuntime: homeRuntimeData{dir: homeRuntimeDir},
+			result:      homeRuntimeDir,
+		},
+
+		// Reading proc command file fails
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: "",
+			runUserDir:      dirForOwner,
+			result:          dirForOwner,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: "",
+			runUserDir:      "", // Accessing run user dir fails
+			tmpPerUserDir:   dirForOwner,
+			result:          dirForOwner,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: "",
+			runUserDir:      dirForAll,
+			tmpPerUserDir:   dirForOwner,
+			result:          dirForOwner,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: "",
+			runUserDir:      "", // Accessing run user dir fails
+			tmpPerUserDir:   dirToBeCreated,
+			result:          dirToBeCreated,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: "",
+			runUserDir:      "", // Accessing run user dir fails
+			tmpPerUserDir:   "", // Accessing tmp per user dir fails
+			homeDir:         dirForOwner,
+			result:          filepath.Join(dirForOwner, "rundir"),
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: "",
+			runUserDir:      "", // Accessing run user dir fails
+			tmpPerUserDir:   dirForAll,
+			homeDir:         dirForOwner,
+			result:          filepath.Join(dirForOwner, "rundir"),
+		},
+
+		// systemd
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: systemdCommandFile,
+			runUserDir:      dirForOwner,
+			result:          dirForOwner,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: systemdCommandFile,
+			runUserDir:      "", // Accessing run user dir fails
+			tmpPerUserDir:   dirForOwner,
+			result:          dirForOwner,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: systemdCommandFile,
+			runUserDir:      dirForAll,
+			tmpPerUserDir:   dirForOwner,
+			result:          dirForOwner,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: systemdCommandFile,
+			runUserDir:      "", // Accessing run user dir fails
+			tmpPerUserDir:   dirToBeCreated,
+			result:          dirToBeCreated,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: systemdCommandFile,
+			runUserDir:      "", // Accessing run user dir fails
+			tmpPerUserDir:   "", // Accessing tmp per user dir fails
+			homeDir:         dirForOwner,
+			result:          filepath.Join(dirForOwner, "rundir"),
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: systemdCommandFile,
+			runUserDir:      "", // Accessing run user dir fails
+			tmpPerUserDir:   dirForAll,
+			homeDir:         dirForOwner,
+			result:          filepath.Join(dirForOwner, "rundir"),
+		},
+
+		// init
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: initCommandFile,
+			tmpPerUserDir:   dirForOwner,
+			result:          dirForOwner,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: initCommandFile,
+			tmpPerUserDir:   dirToBeCreated,
+			result:          dirToBeCreated,
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: initCommandFile,
+			tmpPerUserDir:   "", // Accessing tmp per user dir fails
+			homeDir:         dirForOwner,
+			result:          filepath.Join(dirForOwner, "rundir"),
+		},
+		{
+			homeRuntime:     homeRuntimeDisabled,
+			procCommandFile: initCommandFile,
+			tmpPerUserDir:   dirForAll,
+			homeDir:         dirForOwner,
+			result:          filepath.Join(dirForOwner, "rundir"),
+		},
+	}
+
+	for _, env := range envs {
+		os.Remove(dirToBeCreated)
+
+		resultDir, err := getRootlessRuntimeDirIsolated(env)
+		assert.NilError(t, err)
+		assert.Assert(t, resultDir == env.result)
+	}
+}
+
+type rootlessRuntimeDirEnvironmentRace struct {
+	procCommandFile string
+	tmpPerUserDir   string
+}
+
+func (env rootlessRuntimeDirEnvironmentRace) getProcCommandFile() string {
+	return env.procCommandFile
+}
+func (rootlessRuntimeDirEnvironmentRace) getRunUserDir() string {
+	return ""
+}
+func (env rootlessRuntimeDirEnvironmentRace) getTmpPerUserDir() string {
+	return env.tmpPerUserDir
+}
+func (rootlessRuntimeDirEnvironmentRace) homeDirGetRuntimeDir() (string, error) {
+	return "", errors.New("homedirGetRuntimeDir is disabled")
+}
+func (env rootlessRuntimeDirEnvironmentRace) systemLstat(path string) (*system.StatT, error) {
+	if path == env.tmpPerUserDir {
+		st, err := system.Lstat(path)
+		// We can simulate that race directory was created immediately after system.Lstat call.
+		if err := os.Mkdir(path, 0700); err != nil {
+			return nil, err
+		}
+		return st, err
+	}
+	return system.Lstat(path)
+}
+func (rootlessRuntimeDirEnvironmentRace) homedirGet() string {
+	return homedir.Get()
+}
+
+func TestRootlessRuntimeDirRace(t *testing.T) {
+	raceDir, err := ioutil.TempDir("", "rootless-runtime-dir-race-test")
+	assert.NilError(t, err)
+	defer os.Remove(raceDir)
+
+	procCommandFile := filepath.Join(raceDir, "command")
+	err = ioutil.WriteFile(procCommandFile, []byte("init"), 0644)
+	assert.NilError(t, err)
+
+	tmpPerUserDir := filepath.Join(raceDir, "tmp")
+
+	resultDir, err := getRootlessRuntimeDirIsolated(rootlessRuntimeDirEnvironmentRace{
+		procCommandFile,
+		tmpPerUserDir,
+	})
+	assert.NilError(t, err)
+	assert.Assert(t, resultDir != tmpPerUserDir, "Rootless runtime dir shouldn't follow race dir.")
 }


utils.go
```diff
@@ -87,12 +87,11 @@ func getRootlessRuntimeDir(rootlessUID int) (string, error) {
 	if err == nil {
 		return runtimeDir, nil
 	}
-	tmpDir := fmt.Sprintf("/run/user/%d", rootlessUID)
+	tmpDir := fmt.Sprintf("%s/storage-user-%d", os.TempDir(), rootlessUID)
```
AkihiroSuda: What will happen if your UID is 1000, and another user (UID=2000) has already created `/tmp/storage-user-1000`?
Can 2000 trick to obtain privileges for 1000?

vrothberg:
What's the issue with using `/run/user`?

andrew-aladev:
@vrothberg `/run/user` is not stable across user sessions by design. It is not suitable as fallback path.

utils.go
rhatdan: Podman should verify that the path has the correct ownership when it is done.  If it is not owned by the current user then it should throw an error

rhatdan:
@giuseppe WDYT?

utils.go
AkihiroSuda: TOCTOU?

rhatdan: If I check if a file/directory is owned by me and has 0700, then no other non privileged user can change it.  If it is a privileged process screwing around with me, I have far greater problems.

giuseppe:
> @giuseppe WDYT?

I think we should use `/run`.  When the user session terminates, also the containers are terminated not only `/run/user/X` so using `/tmp` doesn't solve the issue.  If you'd like the session to not be terminated, you could enable the linger mode for the user.  Would that work for you?

andrew-aladev:
@giuseppe `/tmp` solves the issue. If you are using `tmux` or `screen` than your process won't be terminated.

You can't enable linger mode if you have no systemd. The path mentioned above is **fallback** path and it should work across user sessions in **all possible cases**. Giuseppe, did you catch?

giuseppe:
> @giuseppe `/tmp` solves the issue. If you are using `tmux` or `screen` than your process won't be terminated.

If the `/run/user/X` directory cannot be used we already fallback to `tmpDir = fmt.Sprintf("%s/%d", os.TempDir(), rootlessUID)`. 

So does `/run/user/X` exist but you don't want to use it?

andrew-aladev:
> If the `/run/user/X` directory cannot be used we already fallback to

We shouldn't even try to use this directory, because its **behaviour is undefined**. This directory is not suitable to be fallback.

giuseppe:
We want this directory to be tied to the user session.  There is no point that it survives a logout if all the user processes are anyway terminated.

Also we had issues in the past where `XDG_RUNTIME_DIR=` was not correctly set for some podman invocations and didn't use the files under `/run/user/X`.  So it makes sense to default to the most common case.

andrew-aladev:
> We want this directory to be tied to the user session. There is no point that it survives a logout if all the user processes are anyway terminated.

Users want this directory to survive across user sessions. Have you ever used `screen` or `tmux`?

> Also we had issues in the past where `XDG_RUNTIME_DIR=` was not correctly set for some podman invocations and didn't use the files under `/run/user/X`. So it makes sense to default to the most common case.

It will happen with podman anyway because `if err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000` can fail and `/tmp` will be used anyway. This is podman issue.

rhatdan:
So you are saying on your systems it works correctly?  IE It fails on XDG_RUNTIME_DIR and /run/usr and eventually hits /tmp.

But you want to change the default fallback for all systemd based systems.  Where systemd will kill all processes left in the user session when the user logs out, unless the admin enables linger mode.   I believe that your change would not solve the no linger mode issue with tmux and screen, so why make the change?

andrew-aladev:
> So you are saying on your systems it works correctly? IE It fails on XDG_RUNTIME_DIR and /run/usr and eventually hits /tmp.

No, everything is broken on my systems without systemd. I have to set `XDG_RUNTIME_DIR=/tmp/something` to fix it and this _solution_ is not acceptable.

- Do you support running applications with **systemd without lingering**? Yes.
- Do you support running applications with **systemd with lingering**? Yes.

So `/tmp/something` can be a fallback, because its behaviour is the same as systemd with lingering on any system with/without systemd.

> But you want to change the default fallback for all systemd based systems.

Yes, it works perfect on any system with/without systemd. It fixed `tmux` and `screen` issues.

giuseppe:
what is the equivalent of lingering mode on your system?  What is the way to prevent `/run/user/X` to be deleted?

andrew-aladev:
> what is the equivalent of lingering mode on your system? What is the way to prevent `/run/user/X` to be deleted?

Bingo, we are definitely on the right way.

There is no lingering mode equivalent on system without systemd. There is no way to prevent `/run/user/X` to be deleted.

Lingering mode is just a dirty hack implemented by systemd only. `/run/user/X` should always be re-mounted on user login. Nothing can stop initialization system to re-mount it. It don't care about sticky bits or something else. `/run/user/X` was designed for user session only.

Lingering mode was introduced by Lennart Poettering in systemd v183. He didn't care about using a new name for directory with new behaviour. He just used the same directory `/run/user/X` and broke  compatibility in any application.

I know that you are working somewhere in Red Hat. Maybe you are sitting near Lennart, say hello to him. People remember how he broke standards.

rhatdan:
@andrew-aladev Let's keep this civil.  @giuseppe what would break if we moved from using /run/usr/X as the fall back to /tmp/user/X?

Most use cases we will just use XDG_RUNTIME_DIR.  So the issue would be if XDG_RUNTIME_DIR is not turned on for a systemd based system.  

There might be risk if /tmp is not mount on a tmpfs, since the storage drivers rely on information in this directory being destroyed on reboot.

There is potential risks of hostile users on the system trying to attack the /tmp directory, but hopefully we are handling this correctly.

Any other issues that I am not thinking of?

@nalind WDYT?

rhatdan:
@mtrmac Any thoughts?

utils.go
mtrmac: It’s still a DOS at the very least.

utils.go
```diff
@@ -87,12 +87,11 @@ func getRootlessRuntimeDir(rootlessUID int) (string, error) {
...
 	st, err := system.Stat(tmpDir)
```
mtrmac: `Stat` follows symlinks, which makes the ID check below not safe with predictable file names.

giuseppe:
> @giuseppe what would break if we moved from using /run/usr/X as the fall back to /tmp/user/X?

I am concerned about not causing issues when `XDG_RUNTIME_DIR=` is unset for any reason (e.g. using `os/exec` from a Golang program with the wrong Env block).

With systemd all user processes are terminated on logout if lingering mode is not enabled, so we want the runtime files to go away as well and `/run/user/X` is the ideal place to store them.

I don't see any problem in setting `XDG_RUNTIME_DIR=` to something else if the user wishes so

andrew-aladev:
> Could you describe your OS and what exactly is doing what and why it is failing, please?

Any OS with consolekit/elogind without systemd.

> /run/user/X is, AFAIK a systemd invention, maintained by pam_systemd. So, on a non-systemd system, I would expect the per-user directories not to exist at all, and the code to automatically fall back to other alternatives, starting with /tmp already.

`/run/user/X` is maintained by many applications. Its behaviour is well defined in the [freedesktop standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). There is **no lingering mode** in that standard, lingering mode is a compatibility breaker.

Systemd may use other directory like `/run/user-linger/X` for lingering mode and it won't provide issues. But it looks like nobody cares.

> You seem to have some kind of imperfect emulation of /run/user — my initial impression is that that’s something to be fixed in that emulation.

This _emulation_ is working according to standard and not related to systemd.

> The lifetime of the directory MUST be bound to the user being logged in. It MUST be created when the user first logs in and if the user fully logs out the directory MUST be removed. If the user logs in more than once he should get pointed to the same directory, and it is mandatory that the directory continues to exist from his first login to his last logout on the system, and not removed in between. Files in the directory MUST not survive reboot or a full logout/login cycle.

I have posted same text many times.

> With systemd all user processes are terminated on logout if lingering mode is not enabled, so we want the runtime files to go away as well and `/run/user/X` is the ideal place to store them.

We are talking about the same situation but can't understand each other. This is a way how systemd is broken. System without systemd follows standard and have no ability to make directory survive. But systemd has a special ability to make directory survive.

Than we are talking about **fallback** directory. This is a bit **another** talk. This directory should allow users to use application in any environment by default. Today `buildah` is broken with `screen` and `tmux` without systemd lingering mode, so `/run/user/X` is not suitable as fallback path.

utils.go
andrew-aladev: I see no difference between `/run` and `/tmp` prefix. This patch adds no additional problems.

mtrmac:
> > Could you describe your OS and what exactly is doing what and why it is failing, please?
> 
> Any OS with consolekit/elogind without systemd.

We are not fixing “any OS” in this PR. One OS at a time.

> `/run/user/X` is maintained by many applications. Its behaviour is well defined in the [freedesktop standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). There is **no lingering mode** in that standard, lingering mode is a compatibility breaker.
> 
> > The lifetime of the directory MUST be bound to the user being logged in. It MUST be created when the user first logs in and if the user fully logs out the directory MUST be removed. If the user logs in more than once he should get pointed to the same directory, and it is mandatory that the directory continues to exist from his first login to his last logout on the system, and not removed in between. Files in the directory MUST not survive reboot or a full logout/login cycle.

The way I read that, `XDG_RUNTIME_DIR` is supposed to survive `screen`/`tmux` sessions. (And this is consistent with my view that having to explicitly enable “linger” in systemd is a workaround for their bug / design defect.)

> I have posted same text many times.

No idea where.

> Than we are talking about **fallback** directory. This is a bit **another** talk. This directory should allow users to use application in any environment by default.

There is no directory that is guaranteed to be removed on reboot on pre-systemd FHS (or on non-init lightweight containers / “Linux from scratch” setups), so “any environment” is not realistic.

utils.go
mtrmac: The big difference is that unprivileged users are not allowed to create new directories in `/run/user`.

(If you don’t understand the function and the minute details of the various directories, consider whether it is reasonable to be strongly demanding specific solutions to be chosen.)

andrew-aladev:
> We are not fixing “any OS” in this PR. One OS at a time.

There are only 3 systems: without systemd, with systemd + without lingering mode, systemd + with lingering mode.

> The way I read that, `XDG_RUNTIME_DIR` is supposed to survive `screen`/`tmux` sessions. (And this is consistent with my view that having to explicitly enable “linger” in systemd is a workaround for their bug / design defect.)

`XDG_RUNTIME_DIR` won't and shouldn't survive:

> Files in the directory MUST not survive reboot **or a full logout/login cycle**.

> No idea where.

This was an explanation for @giuseppe, sorry.

> There is no directory that is guaranteed to be removed on reboot on pre-systemd FHS (or on non-init lightweight containers / “Linux from scratch” setups), so “any environment” is not realistic.

I can't understand why `/tmp` can't work as this folder. If `/tmp` is not `tmpfs` than this case should be just ignored.

mtrmac:
> > We are not fixing “any OS” in this PR. One OS at a time.
> 
> There are only 3 systems: without systemd, with systemd + without lingering mode, systemd + with lingering mode.

I’m afraid I just lost interest, then, other than making sure this PR does not introduce security vulnerabilities.

> > The way I read that, `XDG_RUNTIME_DIR` is supposed to survive `screen`/`tmux` sessions. (And this is consistent with my view that having to explicitly enable “linger” in systemd is a workaround for their bug / design defect.)
> 
> `XDG_RUNTIME_DIR` won't and shouldn't survive:
> 
> > Files in the directory MUST not survive reboot **or a full logout/login cycle**.

This is the relevant bit:
>  continues to exist from his first login to his last logout on the system,

I can’t see why that wouldn’t include `screen` sessions. Sure, systemd gets it wrong. That’s bad for systemd users, they have to enable linger mode.

andrew-aladev:
> I can’t see why that wouldn’t include `screen` sessions. Sure, systemd gets it wrong. That’s bad for systemd users, they have to enable linger mode.

This directory continue existing after user logout but in unmounted state. When user coming - systemd mounts new tpmfs again. Content will be lost immediately after logout. Same for non 
 systemd systems. `screen` and `tmux` will become broken for both systems without systemd and systemd without lingering mode, i've tested.

rhatdan:
Would a check for whether or not this is a systemd system be a compromize for this?

andrew-aladev:
@rhatdan yes, it will be good too.

rhatdan:
How does one determine this?  And would this work with hidepid?

vrothberg:
I'd look at `/proc/1/comm`. We might want to put that into an `init()` to only do it once.

giuseppe:
> `screen` and `tmux` will become broken for both systems without systemd and systemd without lingering mode, i've tested.

if lingering mode is not enabled, screen and tmux themselves will be terminated, so there is no gain to let these files survive on `/tmp`

andrew-aladev:
> if lingering mode is not enabled, screen and tmux themselves will be terminated, so there is no gain to let these files survive on `/tmp`

No, `screen` and `tmux` will never be stopped. I am very surprised to read such assumptions.

rhatdan:
So checking if /proc/1/command == "systemd" would work.  But not in hidepid envionments?

rhatdan:
@andrew-aladev I believe systemd will kill all processes in the session including tmux and screen.

giuseppe:
> No, `screen` and `tmux` will never be stopped. I am very surprised to read such assumptions.

they are terminated with systemd and no lingering mode.

andrew-aladev:
systemd runs `screen` and `tmux` as service with `RemainAfterExit=yes`.

giuseppe:
> systemd runs `screen` and `tmux` as service with `RemainAfterExit=yes`.

I don't see how `RemainAfterExit` prevents processes to be terminated.

In any case, I've just tried on Fedora Rawhide and tmux doesn't survive the logout

andrew-aladev:
The only reason of using `screen` or `tmux` is that they survives logout. Some maintainers provides service with `RemainAfterExit`, gentoo disables `-Ddefault-kill-user-processes=false` completely.

rhatdan:
In the risk of adding to your hate of systemd, I think tmux and screen will work as long as the login session is not started via systemd.
IE No pam_systemd.

rhatdan:
Lets just do the check for /proc/1/command == "systemd" for now.
@giuseppe Is there a way to check if you are in hidepid?

giuseppe:
> @giuseppe Is there a way to check if you are in hidepid?

we could check the mount flags for `/proc`, but if we are going this path we need to make sure we use `/run/user/X` when `/proc/1/command` can't be read.

rhatdan:
@andrew-aladev Are you still working on this?  I think we can add it with the check for systemd?

andrew-aladev:
@rhatdan, done, please review.

utils.go
rhatdan: @mtrmac If we switch the os.Stat to os.Lstat and verify that the path is a directory, it should be save, correct?

rhatdan:
@giuseppe PTAL

utils.go
```diff
@@ -82,17 +83,39 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+func getInitCommand() (string, error) {
+	file, err := os.Open("/proc/1/comm")
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+	buf, err := ioutil.ReadAll(file)
+	if err != nil {
+		return "", err
+	}
```
giuseppe: could you use directly `ioutil.ReadFile(...)` here?

utils.go
```diff
@@ -82,17 +83,39 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+
 func getRootlessRuntimeDir(rootlessUID int) (string, error) {
 	runtimeDir, err := homedir.GetRuntimeDir()
 	if err == nil {
 		return runtimeDir, nil
 	}
-	tmpDir := fmt.Sprintf("/run/user/%d", rootlessUID)
+	initCommand, err := getInitCommand()
+	var tmpDir string
+	if err == nil && initCommand == "systemd" {
```
giuseppe: let's invert the logic here and default to systemd: `if err != nil || initCommand == "systemd"`

utils.go
```diff
@@ -82,17 +83,39 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+		// Systemd maintains lingering mode for /run/user/X.
+		// It is up to user whether this directory will survive logout.
+		tmpDir = fmt.Sprintf("/run/user/%d", rootlessUID)
+	} else {
+		// Other systems follows standard https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html.
+		// /run/user/X can't survive logout, so it is possible to use temp directory only.
+		tmpDir = fmt.Sprintf("%s/storage-user-%d", os.TempDir(), rootlessUID)
+	}
```
giuseppe: we already have a fallback to TempDir so please drop the else branch completely.

If you'd like to change the directory name, please change the existing `tmpDir = fmt.Sprintf("%s/%d", os.TempDir(), rootlessUID)` to `tmpDir = fmt.Sprintf("%s/containers-user-%d", os.TempDir(), rootlessUID)` (let's use containers instead of storage).

utils.go
```diff
@@ -82,17 +83,39 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
 	st, err := system.Stat(tmpDir)
 	if err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000 {
 		return tmpDir, nil
 	}
-	tmpDir = fmt.Sprintf("%s/%d", os.TempDir(), rootlessUID)
 	if err := os.MkdirAll(tmpDir, 0700); err != nil {
```
mtrmac: This is insecure: an attacker can create a directory with this predictable name, `MkdirAll` will succeed, and the attacker can then attack the use of the runtime directory (e.g. by placing symlinks inside it, or replacing the directory itself.

(Using `os.Mkdir` _should_ be safe even in presence of other files or symlinks, as long as errors are not ignored. But please have a test for that. (Really, this function would benefit a 100% code coverage testing against various attacks — or even instrumentation to explicitly pause it in various points, like above this line, to let the “attacking” test always win a race.)

(If it were up to me, I would delete most of the fallback code in here and just expect `/run/user` and/or $HOME to work. OTOH I haven’t looked into what functionality that would break, so I’m clearly not qualified to make such a recommendation.)

utils.go
mtrmac: It’s certainly _more_ safe. But if we are considering systems without `/run/user`, the `/run/user` path could be an attacker-created symlink. Ugh.

utils.go
andrew-aladev: done

utils.go
andrew-aladev: done

utils.go
andrew-aladev: done

utils.go
```diff
@@ -88,11 +89,16 @@ func getRootlessRuntimeDir(rootlessUID int) (string, error) {
 		return runtimeDir, nil
 	}
 	tmpDir := fmt.Sprintf("/run/user/%d", rootlessUID)
+	initCommand, err := ioutil.ReadFile("/proc/1/comm")
+	if err != nil || string(initCommand) != "systemd" {
+		// Non systemd systems follows standard https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html.
+		// /run/user/X can't survive logout, so it is possible to use temp directory only.
+		tmpDir = fmt.Sprintf("%s/containers-user-%d", os.TempDir(), rootlessUID)
+	}
```
giuseppe: I had in mind something like:

```
@@ -87,12 +88,15 @@ func getRootlessRuntimeDir(rootlessUID int) (string, error) {
        if err == nil {
                return runtimeDir, nil
        }
-       tmpDir := fmt.Sprintf("/run/user/%d", rootlessUID)
-       st, err := system.Stat(tmpDir)
-       if err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000 {
-               return tmpDir, nil
+       initCommand, err := ioutil.ReadFile("/proc/1/comm")
+       if err != nil || string(initCommand) != "systemd" {
+               tmpDir := fmt.Sprintf("/run/user/%d", rootlessUID)
+               st, err := system.Stat(tmpDir)
+               if err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000 {
+                       return tmpDir, nil
+               }
        }
-       tmpDir = fmt.Sprintf("%s/%d", os.TempDir(), rootlessUID)
+       tmpDir := fmt.Sprintf("%s/containers-user-%d", os.TempDir(), rootlessUID)
```

Does it work for you?  If so, please amend my change in your patch

utils.go
mtrmac: Actually, `/run` is root-writable-only (if it exists at all), so this should be safe then. But yes, this **absolutely must** be `Lstat` ⚠️

utils.go
mtrmac: @giuseppe Did you mean `== "systemd"`?

rhatdan: Ye this logic is backwards 
if err == nil && string(initCommand) == "systemd" {

mtrmac: I think Giuseppe wanted `if err != nil || … == "systemd"`, to fall back to the Systemd logic.

giuseppe: @mtrmac yes correct, I want that so we use systemd either on errors or if it is explicitly set.

I'll fix the patch

andrew-aladev: updated, please review.

utils.go
```diff
@@ -87,12 +88,15 @@ func getRootlessRuntimeDir(rootlessUID int) (string, error) {
 	if err == nil {
 		return runtimeDir, nil
 	}
-	tmpDir := fmt.Sprintf("/run/user/%d", rootlessUID)
-	st, err := system.Stat(tmpDir)
-	if err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000 {
-		return tmpDir, nil
+	initCommand, err := ioutil.ReadFile("/proc/1/comm")
+	if err == nil && string(initCommand) == "systemd" {
```
giuseppe, line 92: this must be: `err != nil || string(initCommand) == "systemd"`

utils.go
```diff
@@ -87,12 +88,15 @@ func getRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+		tmpDir := fmt.Sprintf("/run/user/%d", rootlessUID)
+		st, err := system.Stat(tmpDir)
```
giuseppe, line 94: please use `Lstat` instead of `Stat`

utils.go
andrew-aladev: If there is no error and systemd than using `/run/user/X`. I can't understand why `err != nil || _ == "systemd"`. `err != nil` looks like system is unknown, I don't know why we have to use `/run/user/X` in unknown system.

giuseppe: `err != nil` means there was an error while trying to detect it, we still want to default to systemd in such case

andrew-aladev: Yes, that's the question. Why we should try to use systemd path if we can't detect system?

rhatdan: I agree with @giuseppe, we believe /run is more secure then /tmp. So verifying that the init system is NOT systemd is the only way we should use /tmp.  If we can not verify it is not systemd we assume it is, which is probably a safe assumption.

andrew-aladev: ok, updated

utils.go
andrew-aladev: updated

mtrmac:
Wait, what about https://github.com/containers/storage/pull/568#discussion_r400156458 ?

andrew-aladev:
@mtrmac updated, please review

utils.go
```diff
@@ -87,13 +88,16 @@ func getRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+		tmpDir := fmt.Sprintf("/run/user/%d", rootlessUID)
+		st, err := system.Lstat(tmpDir)
+		if err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000 {
+			return tmpDir, nil
+		}
 	}
-	tmpDir = fmt.Sprintf("%s/%d", os.TempDir(), rootlessUID)
-	if err := os.MkdirAll(tmpDir, 0700); err != nil {
+	tmpDir := fmt.Sprintf("%s/containers-user-%d", os.TempDir(), rootlessUID)
+	if err := os.Mkdir(tmpDir, 0700); err != nil && !os.IsExist(err) {
```
mtrmac, line 100: Still https://github.com/containers/storage/pull/568#discussion_r400156458 . This is no better than `MkdirAll`.

I’m increasingly convinced this does require 100% code coverage and attack scenario simulation — except that the previous version is also broken.

rhatdan: How about if we check if the directory exists, and proper permissions, we use it.
If not then we create a random tmporary directory ioutil.TempDir() https://golang.org/pkg/io/ioutil/#TempDir
Then do a os.Rename(newTmpDir,tmpDir)
Does that solve the race condition.

rhatdan: @mtrmac WYT?

mtrmac: That would probably resolve the race condition (write tests!) but I don’t think it would work, in `S_ISVTX` `/tmp` the user is not allowed to remove another users’ files.

Back to https://github.com/containers/storage/pull/568#discussion_r394967980 .

(To be clear, it would clearly be more safe than the current `MkdirAll` or the current PR. Better DOS than a privilege escalation.)

rhatdan: RIght it would fail, which would tell the user he is under attack.
If I saw this happen, I would see permission denied on /tmp/container-user-3267, and then I would check what user created this file with my UID, and I would have a good idea something suspicions is going on.

mtrmac: If that is OK, the code could just as well raise an alarm without trying to create the temporary file and `Rename`.

(As for whether such a behavior in `/tmp`, exposing the non-Systemd user to a trivial DOS, is desirable: as established elsewhere in the PR, I’m not spending time on making it work, only on not making it insecure.)

rhatdan: I believe this would be secure and would satisfy the need.  These issues on creating content in /tmp in well known locations have always existed.  We just need to make sure we don't make a mistake in a potentially risky environment.

andrew-aladev: @mtrmac You mean that other user created `/tmp/folder` with 777 permissions and watching what content will appear in this folder. I can't understand why you are talking about this issue here. You can grep all storage project and find 10-20 `!os.IsExist(err)`. Please create new issue and we can discuss here what to do with `!os.IsExist(err)`.

andrew-aladev: @mtrmac, BTW I see no difference between `/tmp` and `/run/user`. Another user can fool software and redirect system calls from `/proc/1/comm` to fake file and create symlink from `/tmp` to `/run/user`. I am sure that solution for `os.IsExist(err)` shouldn't depend on any parameter like initialization system detection.

andrew-aladev: @mtrmac We can check whether target folder was created by previous `Mkdir` call. It is possible to verify permissions after `os.IsExist`. We can create project scope utility and use it instead of `Mkdir` everywhere. What do you think about it?

rhatdan:
@andrew-aladev Did you have an issue with what I suggested. Doing the rename should fix the issue altogether and then we can merge this PR>

andrew-aladev:
@mtrmac, @rhatdan done, please review

rhatdan:
Failing lint.

utils.go
```diff
@@ -82,25 +83,39 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
-		logrus.Errorf("failed to create %s: %v", tmpDir, err)
-	} else {
+	tmpDir := fmt.Sprintf("%s/containers-user-%d", os.TempDir(), rootlessUID)
+	if _, err := os.Stat(tmpDir); os.IsNotExist(err) {
+		newTmpDir, err := ioutil.TempDir(os.TempDir(), "containers-user-*")
+		if err != nil {
+			logrus.Errorf("failed to create temp directory: %v", err)
+		}
+		if err := os.Rename(newTmpDir, tmpDir); err != nil {
+			logrus.Errorf("failed to rename %s to %s: %v", newTmpDir, tmpDir, err)
```
mtrmac, line 110: I don’t see what the `TempDir` + `Rename` does that straight `Mkdir` doesn’t.

utils_test.go
```diff
@@ -42,3 +50,202 @@ func TestValidStoragePathFormat(t *testing.T) {
...
 }
+
+func TestRootlessRuntimeDir(t *testing.T) {
+	tmpDir, err := ioutil.TempDir(os.TempDir(), "containers-user-*")
+	defer os.RemoveAll(tmpDir)
+	assert.NilError(t, err)
+
+	// testing homedir.GetRuntimeDir binding.
+
+	monkey.Patch(homedir.GetRuntimeDir, func() (string, error) {
```
mtrmac, line 61: I guess this all can work, but the implementation of monkey patching is fairly scary (`mprotect` to RWX, is that even allowed with SELinux, and will it continue to be?) — and mostly not necessary.

In this case, `os.Setenv("XDG_RUNTIME_DIR")`

utils_test.go
```diff
@@ -42,3 +50,202 @@ func TestValidStoragePathFormat(t *testing.T) {
...
+	defer monkey.Unpatch(ioutil.ReadFile)
+
+	var lstatGuard *monkey.PatchGuard
+	lstatGuard = monkey.Patch(system.Lstat, func(path string) (*system.StatT, error) {
+		lstatGuard.Unpatch()
+		defer lstatGuard.Restore()
+		if path == systemdRuntimeDir {
+			err = os.Chmod(tmpDir, 0700)
+			assert.NilError(t, err)
+			err = os.Chown(tmpDir, os.Getuid(), os.Getgid())
```
mtrmac, line 98: This probably does nothing, and if it does something, it won’t work as non-root users; that seems rather problematic for unit tests at least, for any kind of edit+compile+test cycle with an IDE.

@nalind what’s a good way to integrate such tests in this project? Most of the CI runs in dedicated VMs…

utils_test.go
```diff
@@ -42,3 +50,202 @@ func TestValidStoragePathFormat(t *testing.T) {
...
+
+	rootlessRuntimeDir, err = getRootlessRuntimeDir(rootlessUID)
+	assert.NilError(t, err)
+	assert.Assert(t, rootlessRuntimeDir != targetTmpDir, "Rootless runtime dir shouldn't follow invalid target temp dir.")
+
+	// testing target home dir.
+
+	targetHomeDir := tmpDir + "/rundir"
+
+	monkey.Patch(homedir.Get, func() string {
```
mtrmac, line 236: In this case, os.Setenv("HOME")

utils_test.go
```diff
@@ -42,3 +50,202 @@ func TestValidStoragePathFormat(t *testing.T) {
...
+		defer readFileGuard.Restore()
+		if path == "/proc/1/comm" {
+			return []byte("systemd"), nil
+		}
+		return ioutil.ReadFile(path)
+	})
+	defer monkey.Unpatch(ioutil.ReadFile)
+
+	var lstatGuard *monkey.PatchGuard
+	lstatGuard = monkey.Patch(system.Lstat, func(path string) (*system.StatT, error) {
```
mtrmac, line 92: For the `ReadFile`/`Lstat` patching, instead `getRootlessRuntimeDir` can be split into a “production” caller with the real paths, and `getRootlessRuntimeDirImplementation` where the relevant paths are parameters, and the tests can point those parameters to test data.

utils_test.go
```diff
@@ -42,3 +50,202 @@ func TestValidStoragePathFormat(t *testing.T) {
 		assert.NilError(t, err)
 	}
 }
+
+func TestRootlessRuntimeDir(t *testing.T) {
```
mtrmac, line 54: (Non-blocking? This feels long and laborious and a bit hard to follow; it could probably be shorter and clearer as a Go-style table-driven test.)

mtrmac, line 54: From a quick skim, I don’t see this hitting the “`tmpDir` does not exist, trying to create it, and racing against an attacker” case. That would require even more parametrization of the implementation under test (and indirectly making the test more different from the actual system behavior, so I’m not sure how much we need that. What do others think?

utils.go
mtrmac: > @mtrmac You mean that other user created `/tmp/folder` with 777 permissions and watching what content will appear in this folder. I can't understand why you are talking about this issue here. You can grep all storage project and find 10-20 `!os.IsExist(err)`. 

I expect _most_ of them not to happen in world-writable paths (but, full disclosure, I didn’t check).

Still, sure, at least the original `MkdirAll` here was insecure, and I _somewhat_ agree that however that code was introduced, it should be fixed the same way. But then this PR wants to _change_ the behavior away from `/run/user` towards that path on more systems, and that requires that path to, at least, not be known unsafe.

mtrmac: > @mtrmac, BTW I see no difference between `/tmp` and `/run/user`. Another user can fool software and redirect system calls from `/proc/1/comm` to fake file and create symlink from `/tmp` to `/run/user`.

How exactly? Root user attacking other users, VM creators attacking contents of VMs, users attacking themselves, and similar things are not in the threat model I care about, at least.

mtrmac:
> @andrew-aladev Did you have an issue with what I suggested. Doing the rename should fix the issue altogether and then we can merge this PR>

Linking back to https://github.com/containers/storage/pull/568#discussion_r406488248 because it seems there isn’t quite consensus on this yet. I may well be missing something.

vendor/bou.ke/monkey/LICENSE.md
```diff
@@ -0,0 +1,5 @@
+Copyright Bouke van der Bijl
+
+I do not give anyone permissions to use this tool for any purpose. Don't use it.
+
+I’m not interested in changing this license. Please don’t ask.
```
mtrmac, line 5: That’s not good.

utils_test.go
nalind: The license for the module also looks... incompatible.

utils.go
```diff
@@ -82,25 +83,39 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+		if isRootlessRuntimeDirOwner(tmpDir) {
+			return tmpDir, nil
+		}
 	}
-	tmpDir = fmt.Sprintf("%s/%d", os.TempDir(), rootlessUID)
-	if err := os.MkdirAll(tmpDir, 0700); err != nil {
-		logrus.Errorf("failed to create %s: %v", tmpDir, err)
-	} else {
+	tmpDir := fmt.Sprintf("%s/containers-user-%d", os.TempDir(), rootlessUID)
+	if _, err := os.Stat(tmpDir); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
andrew-aladev: We need to know whether directory exist and using `Stat`. What is the problem with it?

utils.go
andrew-aladev: I don't know too. Please write test in go that clearly defines what problem we need to handle. Because without it we can work on this pull request forever.

utils_test.go
andrew-aladev: Yes, but I can't find any other tool for monkey patching for go. It is not possible to mock system calls (as I know) without monkey patching.

utils.go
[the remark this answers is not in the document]
mtrmac: Well, that it doesn’t _do so_. This is checking whether the target of a symlink exists.

Sure, the code ends up doing the right thing because the following code on both branches rejects symlinks, but one shouldn’t need to do such analysis to verify security-critical code is correct when there is a readily available alternative — discussed in this PR several times already — that locally does the right thing without extra reasoning.

utils.go
mtrmac: “I don’t know why the code is here” is a potential security risk for my review.

utils_test.go
andrew-aladev: `tmpDir` was created by current user, so this user can change 2 group and others permissions in any possible way. It shouldn't provide problems.

Why we have to have `tmpDir`? Because `system.StatT` type has all fields private and we can't mock it without using any of 2 public functions. The easiest way is to have `tmpDir` and change its permissions.

utils_test.go
andrew-aladev: I can't agree, this won't be cross platform. We shouldn't try to mock using any internal business logic of dependant software.

utils_test.go
mtrmac: > `tmpDir` was created by current user, so this user can change 2 group and others permissions in any possible way. It shouldn't provide problems.

I’m not worried about the human running this unit test attacking the operation of the unit test. What would be the point? And such an attacker could just as well change the permissions _after_ this line executes anyway.

> Why we have to have `tmpDir`? Because `system.StatT` type has all fields private and we can't mock it without using any of 2 public functions. The easiest way is to have `tmpDir` and change its permissions.

Sure, that’s fair for the `Chmod`, but if the `Chown` is not really intended to _change_ permissions, the intent would be more clearly expressed by just calling `Lstat` and verifying the ownership has the expected values.

nalind: Ideally we'd be running the unit tests for filesystem-specific logic along with the integration tests that are geared for it.  In that context, we'd probably have to extend the `local-test-unit` target to also compile tests which can/should be run without privileges with `go test -c`, and then invoke the result using a `-test.run` flag that skipped tests which are expected to fail when they're run without root privileges.

utils_test.go
mtrmac: Fine, then parametrize this path by splitting the implementation into two, as discussed for reading `/proc`

utils.go
andrew-aladev: I've read previous comments about security problem and implemented solution that everyone agree about. Now another security problem appeared (as I see) so I am requesting people to **describe security problem in actual code**.

utils_test.go
andrew-aladev: > calling Lstat and verifying the ownership has the expected values

Sure, we will fix it. `Chown` call is redundant here.

andrew-aladev: > which can/should be run without privileges

This test should work without priviledges, sure, we can remove `Chown`.

utils_test.go
andrew-aladev: Sure, that will be better solution.

vendor/bou.ke/monkey/LICENSE.md
andrew-aladev: This is the limitation of `go`: there are no other tools for monkey patching (as I know) and it is forbidden by language itself. I don't know why you are using `go`.

utils_test.go
andrew-aladev: I am not very experienced in `go`. What table tests do you mean?

utils_test.go
andrew-aladev: Yes, we can split this function and remove several monkey patches.

utils.go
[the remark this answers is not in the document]
andrew-aladev: > This is checking whether the target of a symlink exists.

Yes, what is the security problem with symlink?

utils_test.go
andrew-aladev: > racing against an attacker

If you will describe in in a strict way that it will be implemented.

andrew-aladev:
@rhatdan Can you please help with understanding what `racing against an attacker` @mtrmac means? I can't understand it, so I can't implement it.

utils.go
andrew-aladev: I've read it about 10 times, I see no position anywhere which can be kept or changed. You points to some attack, but nobody can declear actual attack.

utils_test.go
mtrmac: https://github.com/golang/go/wiki/TableDrivenTests : have shared code that performs the test, with the different situations described as just data.

utils.go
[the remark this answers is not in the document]
mtrmac: > Sure, the code ends up doing the right thing… but…

utils_test.go
andrew-aladev: Aaa, no. This test can't be implemented in that way: it is branched and next branch depends on previous. We can make it smaller and more readable but can't make it table like.

mtrmac: Consider how
```go
if err := os.Rename(newTmpDir, tmpDir); err != nil {
```
can fail if `tmpDir` has been created in the meantime.

But, again, I’m not at all sure adding a test for this one is worth it. It is certainly _doable_ but the cost (adding an extra pair of channels for communicating with the test?) may be making the code too difficult to read; the security-critical bit is ultimately the `isRootlessRuntimeDirOwner` check after the `Rename`.

utils.go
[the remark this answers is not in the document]
andrew-aladev: Please describe steps one-by-one to reproduce attack. I will convert this steps into actual code. Because now I can't understand what you want to say. I see no problem with symlinks.

utils_test.go
mtrmac: Do the individual cases really _strongly_ depend on the sequence they happen?
A
```go
[]struct {
   runtimeDirPresent bool
   proc1CommReadable bool
   proc1Comm string
   runUserUIDDirPresent bool
   runUserUIDDirPermissions int
   tmpUIDDirPresent bool
   tmpUIDDirPermissions int
   expectedPath string // "" on expected error
}
```
or something like that, seems possible. Or maybe I misunderstand what “it is branched” refers to.

utils.go
mtrmac: At this point I’m not pointing at a specific attack; I think that a plain `os.Mkdir` (not `All`!) with a strict error check would work fine.

If Dan is asking for a `TmpDir`+`Rename`, I don’t know why that is necessary, which indicates I may well be making a mistake, and missing another attack I should be thinking about.

utils_test.go
andrew-aladev: It is possible but we will receive branched table: first item will have only one `runtimeDirPresent` filled, last item will have all filled. I can implement it if you like, no problem.

andrew-aladev: Yes, that's sounds reasonable. We need to use `tmpDir` only if renamed successfully. It will be fixed.

utils.go
andrew-aladev: I see the only possible attack - user attacks himself. We can't do anything with such user - he can attack himself in any folder with sticky bit or without it.

utils_test.go
mtrmac: As mentioned at the start, this is non-blocking for me. Looking at all the members above, it’s not even obvious that it would be shorter … but it should be easier to follow the logic.

(BTW filling all of the fields with valid possibilities along with `runtimeDirPresent` and testing that the `runtimeDir` is the one chosen would also be a reasonable test — and much easier to add in the table-driven implementation. But then, it’s not a security-relevant test so I’m not asking for that to be added.)

utils.go
mtrmac: Well, either this approach is necessary, in which case I will need to rethink things, or it is not, and the code should be simpler.

Let’s wait for Dan.

rhatdan: Yes mkdir should do the right thing.  IE Create the directory with the correct permissions, and fail if an attacker has created the directory since you checked if it exists.  The rename is overkill, sorry for suggesting it.

rhatdan:
If the monkey patching is just for testing, then I say drop it.

utils.go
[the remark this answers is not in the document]
andrew-aladev: Dan provided good solution to use Mkdir, so resolved.

utils_test.go
andrew-aladev: I will try to write parameters version of function to test it without monkey patching.

vendor/bou.ke/monkey/LICENSE.md
andrew-aladev: I will try to write parameters version of function to test it without monkey patching.

utils.go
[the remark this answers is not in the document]
mtrmac: This is unrelated to the `Rename` conversation.

andrew-aladev:
@rhatdan, @mtrmac Tests about order of directories and trojan directory implemented, please review.

rhatdan:
Tests are failing
@mtrmac @giuseppe PTAL

LGTM

giuseppe:
I've restarted the tests.  LGTM

giuseppe:
@andrew-aladev could you rebase and push again?  Let's see if it helps the tests

andrew-aladev:
@giuseppe Ubuntu tests are failing:

```
E: The repository 'http://us-central1.gce.archive.ubuntu.com/ubuntu disco Release' no longer has a Release file.
```

Something is wrong with ubuntu test images.

andrew-aladev:
@giuseppe, now tests are working, thank you.

giuseppe:
@mtrmac PTAL

utils_test.go
```diff
@@ -42,3 +49,203 @@ func TestValidStoragePathFormat(t *testing.T) {
...
+	homeRunDir := homedir.Get() + "/rundir"
+
+	datas := []RootlessRuntimeDirSyscallsTest{
+		{
+			HomeRuntime: HomeRuntime{Dir: homeRunDir},
+			Result:      homeRunDir,
+		},
+
+		// systemd
+		{
```
mtrmac, line 126: Non-blocking: Nothing in here exercises the “temporary directory was missing and successfully created” case I think. (But that would require moving the tests from real `os.TempDir()`, see the race test comment.)

rhatdan:
@andrew-aladev Could you implement @mtrmac changes and then we can get this merge.

andrew-aladev:
Hello, I will implement it soon, thank you.

utils.go
```diff
@@ -82,33 +83,95 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
 	return path, nil
 }
 
-func getRootlessRuntimeDir(rootlessUID int) (string, error) {
-	runtimeDir, err := homedir.GetRuntimeDir()
+// RootlessRuntimeDirSyscalls can be used to test getRootlessRuntimeDirIsolated.
+type RootlessRuntimeDirSyscalls interface {
```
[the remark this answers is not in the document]
andrew-aladev: updated

utils.go
```diff
@@ -82,33 +83,95 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+	systemLstat(string) (*system.StatT, error)
+	osGetuid() int
+	osTempDir() string
+	osStat(string) (os.FileInfo, error)
+	osMkdir(string, os.FileMode) error
+	filePathEvalSymlinks(string) (string, error)
+}
+
+// RootlessRuntimeDirSyscallsImplementation is the default RootlessRuntimeDirSyscalls implementation.
+type RootlessRuntimeDirSyscallsImplementation struct{}
```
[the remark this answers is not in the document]
andrew-aladev: updated

utils_test.go
```diff
@@ -42,3 +49,203 @@ func TestValidStoragePathFormat(t *testing.T) {
...
+type CommandFile struct {
+	Data  string
+	Error error
+}
+type Stat struct {
+	Path        string
+	Replacement string
+}
+
+type RootlessRuntimeDirSyscallsTest struct {
```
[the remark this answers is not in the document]
andrew-aladev: updated

utils.go
```diff
@@ -82,33 +83,95 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+
+func (RootlessRuntimeDirSyscallsImplementation) homeDirGetRuntimeDir() (string, error) {
+	return homedir.GetRuntimeDir()
+}
+func (RootlessRuntimeDirSyscallsImplementation) homeDirGet() string {
```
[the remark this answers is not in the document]
andrew-aladev: removed

utils.go
```diff
@@ -82,33 +83,95 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+	return homedir.Get()
+}
+func (RootlessRuntimeDirSyscallsImplementation) ioUtilReadFile(path string) ([]byte, error) {
+	return ioutil.ReadFile(path)
+}
+func (RootlessRuntimeDirSyscallsImplementation) systemLstat(path string) (*system.StatT, error) {
+	return system.Lstat(path)
+}
+func (RootlessRuntimeDirSyscallsImplementation) osGetuid() int {
```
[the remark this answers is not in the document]
andrew-aladev: removed

utils.go
```diff
@@ -82,33 +83,95 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+	return os.Getuid()
+}
+func (RootlessRuntimeDirSyscallsImplementation) osTempDir() string {
```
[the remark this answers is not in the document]
andrew-aladev: removed

utils.go
```diff
@@ -82,33 +83,95 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+	return os.TempDir()
+}
+func (RootlessRuntimeDirSyscallsImplementation) osStat(path string) (os.FileInfo, error) {
+	return os.Stat(path)
+}
+func (RootlessRuntimeDirSyscallsImplementation) osMkdir(path string, mode os.FileMode) error {
```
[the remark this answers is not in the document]
andrew-aladev: removed

utils.go
```diff
@@ -82,33 +83,95 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+	return os.Mkdir(path, mode)
+}
+func (RootlessRuntimeDirSyscallsImplementation) filePathEvalSymlinks(path string) (string, error) {
```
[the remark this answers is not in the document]
andrew-aladev: removed

utils_test.go
```diff
@@ -42,3 +49,203 @@ func TestValidStoragePathFormat(t *testing.T) {
...
+	RootlessRuntimeDirSyscallsImplementation
```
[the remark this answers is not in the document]
andrew-aladev: removed

utils_test.go
```diff
@@ -42,3 +49,203 @@ func TestValidStoragePathFormat(t *testing.T) {
...
+	}
+
+	for _, data := range datas {
+		resultDir, err := getRootlessRuntimeDirIsolated(rootlessUID, data)
+		assert.NilError(t, err)
+		assert.Assert(t, resultDir == data.Result)
+	}
+}
+
+type RootlessRuntimeDirSyscallsTrojan struct {
```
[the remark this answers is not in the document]
andrew-aladev: updated

utils.go
```diff
@@ -82,33 +83,95 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+	return filepath.EvalSymlinks(path)
+}
+
+func isRootlessRuntimeDirOwner(dir string, syscalls RootlessRuntimeDirSyscalls) bool {
+	st, err := syscalls.systemLstat(dir)
+	return err == nil && int(st.UID()) == syscalls.osGetuid() && st.Mode()&0700 == 0700 && st.Mode()&0066 == 0000
+}
+
+func getRootlessRuntimeDirIsolated(rootlessUID int, syscalls RootlessRuntimeDirSyscalls) (string, error) {
```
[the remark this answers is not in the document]
andrew-aladev: updated

utils.go
```diff
@@ -82,33 +83,95 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
+			return tmpDir, nil
+		}
 	}
-	tmpDir = fmt.Sprintf("%s/%d", os.TempDir(), rootlessUID)
-	if err := os.MkdirAll(tmpDir, 0700); err != nil {
-		logrus.Errorf("failed to create %s: %v", tmpDir, err)
-	} else {
+
+	tmpDir := fmt.Sprintf("%s/containers-user-%d", syscalls.osTempDir(), rootlessUID)
+	if _, err := syscalls.osStat(tmpDir); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
andrew-aladev: updated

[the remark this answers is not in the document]
mtrmac: How was this fixed? I still see `syscalls.osStat` == `os.Stat(path)`

utils.go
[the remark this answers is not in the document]
mtrmac: I see no change.

mtrmac:
@andrew-aladev I can’t see any of the changes you said you have made. Is that a work in progress, or have you maybe forgotten to push your finished work?

andrew-aladev:
Several questions are not yet updated, thinking about them =)

utils_test.go
andrew-aladev: updated

utils_test.go
```diff
@@ -42,3 +49,203 @@ func TestValidStoragePathFormat(t *testing.T) {
...
+			HomeRuntime: HomeRuntime{Dir: homeRunDir},
+			Result:      homeRunDir,
+		},
+
+		// systemd
+		{
+			HomeRuntime: HomeRuntime{Error: errors.New("homedirGetRuntimeDir is disabled")},
+			CommandFile: CommandFile{Data: "systemd"},
+			Stats: []Stat{
+				{Path: runUserDir, Replacement: dirForOwnerOnly},
```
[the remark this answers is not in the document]
andrew-aladev: updated but with additional methods for testing

utils_test.go
```diff
@@ -42,3 +49,203 @@ func TestValidStoragePathFormat(t *testing.T) {
...
+		return fakeST, fakeErr
+	}
+	return os.Stat(path)
+}
+
+func TestRootlessRuntimeDirTrojan(t *testing.T) {
+	rootlessUID := math.MaxUint16
+
+	trojanDir := fmt.Sprintf("%s/containers-user-%d", os.TempDir(), rootlessUID)
+	defer os.Remove(trojanDir)
```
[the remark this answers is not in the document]
andrew-aladev: updated

andrew-aladev:
@mtrmac seems like all questions was resolved, please review.

rhatdan:
@mtrmac PTAL  Lets finally get this in.

utils.go
```diff
@@ -82,31 +83,92 @@ func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
...
-	return filepath.Join(resolvedHome, "rundir"), nil
+	return filepath.Join(resolvedHomeDir, "rundir"), nil
+}
+
+func getRootlessRuntimeDir(rootlessUID int) (string, error) {
+	return getRootlessRuntimeDirIsolated(
+		rootlessRuntimeDirEnvironmentImplementation{
+			"/proc/1/comm",
+			fmt.Sprintf("/run/user/%d", rootlessUID),
+			fmt.Sprintf("%s/containers-user-%d", os.TempDir(), rootlessUID),
```
mtrmac, line 169: Note to self: the old path was `"%s/%d"`. I don’t think it matters, this one is at least a bit more attributable.

rhatdan:
@andrew-aladev Thanks for working on this.

PR #571, opened 2020-03-23
userns: add support for auto

giuseppe:
automatically pick an empty range and create an user namespace for the container.
    
For root containers, it is necessary to specify an entry in the /etc/subuid and /etc/subgid files to use for picking the range of available IDs.  This is necessary to avoid collisions with IDs used for rootless containers.  This setting is ignored for rootless containers, since it is not possible to use arbitrary IDs, and the initial set is always picked by the IDs assigned to the rootless user.
    
When using auto userns, a container will use a range of IDs that is not used by any other container user namespace, also those that are not using auto userns, this is checked at creation time.

A successive container that doesn't use auto userns feature can still collide with IDs used by an auto userns container.


storage.conf
```diff
@@ -46,6 +46,14 @@ additionalimagestores = [
 # remap-user = "storage"
 # remap-group = "storage"
 
+# Root-auto-userns-user is a user name which can be used to look up one or more UID/GID
+# ranges in the /etc/subuid and /etc/subgid file.  These ranges will be partioned
+# to containers configured to create automatically an user namespace.  Containers
+# configured to automatically create an user namespace can still overlap with containers
+# having an explicit mapping set.
+# This setting is ignored when running as rootless.
+# root-auto-userns-user = "storage"
```
rhatdan: I would prefer default should be "containers",

userns.go
```diff
@@ -0,0 +1,220 @@
...
+				id = strconv.Itoa(os.Geteuid())
+			}
+			userID, err := user.LookupId(id)
+			if err == nil {
+				username = userID.Username
+			}
+		}
+	} else {
+		if username == "" {
+			username = "storage"
```
rhatdan: Containers, and should be a constant.

userns.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	var usedUIDs, usedGIDs []idtools.IDMap
+	for _, c := range containers {
+		usedUIDs = append(usedUIDs, c.UIDMap...)
+		usedGIDs = append(usedGIDs, c.GIDMap...)
+	}
+
+	if size <= 0 {
+		size = 4096
```
rhatdan: Should this be defined in containers/storage?

rhatdan:
@vrothberg @saschagrunert @nalind @TomSweeneyRedHat PTAL

rhatdan:
I think we need a better description of what we are after here.

The basic idea is to allow tools like Podman and CRI-O to automatically grab a unique UID/GID range for a user namespace large enough for the container, based on the UIDs/GIDs defined in the image.  We need to go through all existing containers and make sure of no overlap.

Also we will preallocate a range of uids in the /etc/subuid and /etc/subgid file when tools like CRI-O, Buildah, Podman get installed.

Something like:
```
cat /etc/subuid
3267:100000:8000
containers:2147483646:2147483648
```

Then the container engines can grab ranges from 2147483646 on, and useradd and systemd can continue to use the bottom range of UIDs.

userns.go
giuseppe: should we allow to override it (via storage.conf) as part of this PR or can it be done later?

giuseppe:
> I think we should also check the highest uid in the /etc/passwd defined in the container as well, to make sure that is not greater then the range of UIDs.

There are other heuristic we should add (like check `/etc/sub*id`) but we need to do it at pull time when we first create the layer.

giuseppe:
libpod implementation: https://github.com/containers/libpod/pull/5478

rhatdan:
I am fine with doing it later.

I think we need more tests.  How do we check what happens when we run out of ranges?

userns.go
giuseppe: it is the name we are already using for `RemapUser`.

I am fine with changing it but I think we should be consistent.  Should we change the default for `RemapUser` too?

giuseppe:
> I think we need more tests. How do we check what happens when we run out of ranges?

I'll add more tests

userns.go
giuseppe: I've added  a patch that allows setting the minimum and maximum value for the user namespace.

The maximum size is necessary as a malicious image could consume the entire space.

docs/containers-storage.conf.5.md
```diff
@@ -76,6 +76,15 @@ The `storage.options` table supports the following options:
...
      remap-group = "storage"
 
+**root-auto-userns-user**=""
+  Root-auto-userns-user is a user name which can be used to look up one or more UID/GID ranges in the /etc/subuid and /etc/subgid file.  These ranges will be partioned to containers configured to create automatically an user namespace.  Containers configured to automatically create an user namespace can still overlap with containers having an explicit mapping set.  This setting is ignored when running as rootless.
+
+**auto-userns-min-size**=4096
+  Auto-userns-min-size is the minimum size for an user namespace created automatically.
+
+**auto-userns-max-size**=65536
+  Auto-userns-max-size is the maximum size for an user namespace created automatically.
```
vrothberg: Here and above s/an user/a user/

vrothberg: And some of the code comments.

storage.conf
vrothberg: :+1: on "containers"

store.go
```diff
@@ -1151,6 +1179,34 @@ func (s *store) CreateContainer(id string, names []string, image, layer, metadat
...
+			if s < 0 {
+				return nil, errors.Errorf("invalid size specified for the user namespace %q", val)
+			}
+			size = uint32(s)
+		}
+		options.UIDMap, options.GIDMap, err = s.getAutoUserNS(size, image)
+		if err != nil {
+			return nil, err
+		}
+	}
```
vrothberg: Could we turn this code into a anonymous function/lambda? This would reduce the critical section and hence the time we keep the lock (and ultimately speed it up a bit more).

vrothberg:
```golang
uidMap, gidMap, err := func () ... {
   lock()
   defer unlock()
}

File: store.go [hunk 6]
```diff
@@ -3305,6 +3361,9 @@ func copyStringInterfaceMap(m map[string]interface{}) map[string]interface{} {
 // defaultConfigFile path to the system wide storage.conf file
 const defaultConfigFile = "/etc/containers/storage.conf"
 
+const AutoUserNsMinSize = 1024
```
vrothberg: The docs state `4096`. Which value is the right one?

userns.go
vrothberg: SGTM

userns.go
```diff
@@ -0,0 +1,235 @@
...
+	"os/user"
+	"strconv"
+
+	"github.com/containers/common/pkg/unshare"
+	"github.com/containers/storage/pkg/idtools"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+func getAutoUserNsMappings(username string) ([]idtools.IDMap, []idtools.IDMap, error) {
```
vrothberg: A comment would be great.

userns.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		logrus.Errorf("cannot find mappings for user %q: %v", username, err)
+	} else {
+		uids = mappings.UIDs()
+		gids = mappings.GIDs()
+	}
+	return uids, gids, nil
+}
+
+func (s *store) getAvailableMappings() ([]idtools.IDMap, []idtools.IDMap, error) {
```
vrothberg: A comment would be great.

userns.go
```diff
@@ -0,0 +1,235 @@
...
+		return nil, nil, errors.Wrapf(err, "cannot read mappings")
+	}
+
+	var usedUIDs, usedGIDs []idtools.IDMap
+	for _, c := range containers {
+		usedUIDs = append(usedUIDs, c.UIDMap...)
+		usedGIDs = append(usedGIDs, c.GIDMap...)
+	}
+
+	if size <= 0 {
```
vrothberg: A (verbose) comment would be great.

userns.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	if s.autoNsMaxSize > 0 && size > s.autoNsMaxSize {
+		size = s.autoNsMaxSize
+	}
+
+	return findAvailableRange(size, size, availableUIDs, availableGIDs, usedUIDs, usedGIDs)
+}
+
+func findAvailableRange(sizeUID, sizeGID uint32, availableUIDs, availableGIDs, usedUIDs, usedGIDs []idtools.IDMap) ([]idtools.IDMap, []idtools.IDMap, error) {
```
vrothberg: A comment would be great.

userns.go
```diff
@@ -0,0 +1,235 @@
...
+
+	GIDMap, err := findAvailableIDRange(sizeGID, availableGIDs, usedGIDs)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return UIDMap, GIDMap, nil
+}
+
+func findAvailableIDRange(size uint32, availableIDs, usedIDs []idtools.IDMap) ([]idtools.IDMap, error) {
```
vrothberg: A comment would be great.

userns.go
```diff
@@ -0,0 +1,235 @@
...
+	// ContainerID will be adjusted later.
+	for _, i := range availableIDs {
+		n := idtools.IDMap{
+			ContainerID: 0,
+			HostID:      i.HostID,
+			Size:        i.Size,
+		}
+		avail = append(avail, n)
+	}
+	for _, u := range usedIDs {
```
vrothberg: A (verbose) comment would be great.

userns.go
```diff
@@ -0,0 +1,235 @@
...
+			avail[i].Size = int(remaining)
+			return avail[:i+1], nil
+		}
+		remaining -= uint32(avail[i].Size)
+	}
+
+	return nil, errors.New("could not find enough available IDs")
+}
+
+func subtractIDs(avail idtools.IDMap, used idtools.IDMap) []idtools.IDMap {
```
vrothberg: A comment would be great.

store.go
giuseppe: we need to serialize the entire function.  We can release the lock only after the container is created in the store to prevent another process to pick the same range.

giuseppe:
thanks for the review.  I've pushed another version that addresses the points above.

I've added a patch to change the default user from `storage` to `containers`

store.go
giuseppe: updated the docs to the correct value

giuseppe:
marked WIP until Podman tests pass

vrothberg:
@giuseppe, mind squashing the commits into one?

giuseppe:
> @giuseppe, mind squashing the commits into one?

isn't clearer to keep the changes separate?

The first patch adds the basic support for auto-userns, the second one extends auto-userns with a new option (there are likely more patches to be added in the future for other options) and the the third one changes an existing setting.  The current third patch should probably be moved first before the other two patches

giuseppe:
> @giuseppe, mind squashing the commits into one?

I've squashed the two patches for auto-userns and left the one that changes `storage` -> `containers`

giuseppe:
@rhatdan ready for review

rhatdan:
LGTM

PR #572, opened 2020-03-23
store: memoize ReloadConfigurationFile

giuseppe:
avoid reloading the same configuration multiple times if the file was
not modified.

### DIFF
--- a/store.go
+++ b/store.go
@@ -3417,12 +3417,44 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 	}
 }
 
+var prevReloadConfig = struct {
+	storeOptions *StoreOptions
+	mod          time.Time
+	mutex        sync.Mutex
+	configFile   string
+}{}
+
+func reloadConfigurationFileIfNeeded(configFile string, storeOptions *StoreOptions) {
+	prevReloadConfig.mutex.Lock()
+	defer prevReloadConfig.mutex.Unlock()
+
+	fi, err := os.Stat(configFile)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			fmt.Printf("Failed to read %s %v\n", configFile, err.Error())
+		}
+		return
+	}
+
+	mtime := fi.ModTime()
+	if prevReloadConfig.storeOptions != nil && prevReloadConfig.mod == mtime && prevReloadConfig.configFile == configFile {
+		*storeOptions = *prevReloadConfig.storeOptions
+		return
+	}
+
+	ReloadConfigurationFile(configFile, storeOptions)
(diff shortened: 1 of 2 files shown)


store.go
```diff
@@ -3417,12 +3417,44 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
...
+	fi, err := os.Stat(configFile)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			fmt.Printf("Failed to read %s %v\n", configFile, err.Error())
+		}
+		return
+	}
+
+	mtime := fi.ModTime()
+	if prevStoreOptions != nil && prevMod == mtime {
```
rhatdan: Shouldn't you test if prevConfigFIle == configFile?

giuseppe: good catch.  Yes I need to do that.  Pushed a fixed version

rhatdan:
@nalind @saschagrunert @vrothberg @TomSweeneyRedHat @mrunalp PTAL

store.go
```diff
@@ -3417,12 +3417,44 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 	}
 }
 
+var (
+	prevStoreOptions *StoreOptions
+	prevMod          time.Time
+	prevMutex        sync.Mutex
+	prevConfigFile   string
+)
```
saschagrunert: It’s a bit but I’d prefer to have just one single global mutable variable as struct.

store.go
```diff
@@ -3417,12 +3417,44 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
...
+
+func maybeReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
```
TomSweeneyRedHat: nit "reloadConfigurationFileIfNeeded" perhaps?

TomSweeneyRedHat:
Couple nits for consideration.  Could you touch up the PR title too?  "memoize" needs to be removed/changed.

rhatdan:
@TomSweeneyRedHat memoize is a real word and means exactly what @giuseppe is trying to do.
I had to look up the word as well, first time I saw it.
https://en.wikipedia.org/wiki/Memoization

TomSweeneyRedHat:
Well dang, a new word for me.  Nicely done @giuseppe!

store.go
rhatdan: Ok @giuseppe Change to a struct and then we can merge.

store.go
giuseppe: I've used the same naming we have for other functions, but I am fine with changing it

PR #580, opened 2020-03-31
Fix CGo builds on non-Linux

mtrmac:
Go refuses to compile a subpackage with `CGO_ENABLED=1` if there is a `*.c` file but no `import "C"`.  OTOH if we did have an `import "C"`, the Linux-only code would fail to compile.

So, satisfy the Go compiler by using `import "C"` but `#ifdef`-ing out all of the code.


mtrmac:
BTW the package currently allows compilation of non-CGo Linux builds (but that fails tests). This PR does nothing to change that.

Issue #575, opened 2020-03-26
panic when trying to update last-writer without write-lock


haircommander:
we somehow encountered this in CRI-O:
```
Mar 26 16:46:25 ip-172-18-1-199.ec2.internal crio[26020]: time="2020-03-26 16:46:25.447845722Z" level=info msg="About to del CNI network lo (type=loopback)"
Mar 26 16:46:25 ip-172-18-1-199.ec2.internal crio[26020]: panic: attempted to update last-writer in lockfile without the write lock
Mar 26 16:46:25 ip-172-18-1-199.ec2.internal crio[26020]: goroutine 139197 [running]:
Mar 26 16:46:25 ip-172-18-1-199.ec2.internal crio[26020]: panic(0x1b3d9c0, 0x2166ba0)
Mar 26 16:46:25 ip-172-18-1-199.ec2.internal crio[26020]: 	/usr/local/go/src/runtime/panic.go:1060 +0x420 fp=0xc000825d50 sp=0xc000825ca8 pc=0x43c860
Mar 26 16:46:25 ip-172-18-1-199.ec2.internal crio[26020]: github.com/containers/storage/pkg/lockfile.(*lockfile).Touch(0xc000350eb0, 0xc000fc6720, 0x0)
Mar 26 16:46:25 ip-172-18-1-199.ec2.internal crio[26020]: 	/go/src/github.com/cri-o/cri-o/vendor/github.com/containers/storage/pkg/lockfile/lockfile_unix.go:193 +0x235 fp=0xc000825dd0 sp=0xc000825d50 pc=0x7a2ae5
```

[ref](https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3482/test_pull_request_crio_e2e_crun_fedora/2409/artifacts/journals/crio.service)

What's going on here, why is this a panic, and is there anything callers of c/storage can do to prevent this?

mrunalp:
@nalind @TomSweeneyRedHat @rhatdan ptal.

rhatdan:
@vrothberg PTAL

vrothberg:
> What's going on here, why is this a panic, and is there anything callers of c/storage can do to prevent this?

We _must_ panic when detecting violations of the lock protocol to prevent data corruption. `(*lockfile).Touch()` has either been called without owning the lock OR at the time of being a reader. But we need to own the lock as a writer to update its metadata.

@haircommander, is it happening in master?

[...]

PR #581, opened 2020-03-31
DeleteContainer(): wait for all goroutines

nalind:
DeleteContainer() does most of its heavy lifting in a number of goroutines.  Previously, if any of them returned an error result, that result was returned to the caller immediately, and DeleteContainer()
would release any locks that it had acquired as part of that.  This could all happen while other goroutines were still running, and some of those goroutines depended on having those locks held until they finished their work, and would panic() when they detected that the locks had been released out from under them.

Fixes https://github.com/containers/storage/issues/575.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/Microsoft/hcsshim v0.8.7
 	github.com/containers/common v0.6.1
 	github.com/docker/go-units v0.4.0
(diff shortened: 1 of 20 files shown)


umohnani8:
LGTM. I tested it out and no longer see the failure :D thank you @nalind!

umohnani8:
@rhatdan @vrothberg PTAL

PR #583, opened 2020-04-01
Run bats jobs in parallel

saschagrunert:
We now run the bats jobs in parallel to have a faster test output.


rhatdan:
Hopefully this does not lead to flakes.

rhatdan:

giuseppe:
LGTM

PR #585, opened 2020-04-02
Add docs about when locks can panic

saschagrunert:
This partially fixes a CRI-O panic on container restore if the sandbox
could not restored successfully and CRI-O tries to remove the sandbox
afterwards. To not have the need to recover the panic in CRI-O (or
possible other consumers of this library) we should avoid panic'ing here
and stick to error handling.

Reproduce the panic with the latest CRI-O master:

```json
{
  "metadata": {
    "name": "nginx-sandbox",
    "namespace": "default",
    "attempt": 1,
    "uid": "hdishd83djaidwnduwk28bcsb"
  },
  "log_directory": "/tmp",
  "linux": {}
}
```

```
# Start CRI-O
> sudo ./bin/crio
> sudo crictl runp sandbox.json
# Stop CRI-O, start CRI-O again
panic: attempted to update last-writer in lockfile without the write lock

goroutine 86 [running]:
github.com/containers/storage/pkg/lockfile.(*lockfile).Touch(0xc0007149b0, 0xc0006e4000, 0x0)
        /home/sascha/go/src/github.com/cri-o/cri-o/vendor/github.com/containers/storage/pkg/lockfile/lockfile_unix.go:193 +0x235
github.com/containers/storage.(*containerStore).Touch(0xc00014af60, 0x3e, 0xc0005b4030)
        /home/sascha/go/src/github.com/cri-o/cri-o/vendor/github.com/containers/storage/containers.go:588 +0x33
github.com/containers/storage.(*containerStore).Save(0xc00014af60, 0x0, 0x0)
        /home/sascha/go/src/github.com/cri-o/cri-o/vendor/github.com/containers/storage/containers.go:228 +0x202
github.com/containers/storage.(*containerStore).Delete(0xc00014af60, 0xc00042e500, 0x39, 0x6d20746c75616665, 0x72662073746e756f)
        /home/sascha/go/src/github.com/cri-o/cri-o/vendor/github.com/containers/storage/containers.go:405 +0x26c
github.com/containers/storage.(*store).DeleteContainer.func2(0xc0000aaa80, 0x2220400, 0xc00014af60, 0xc00042e500, 0x39, 0xc00045c4e0)
        /home/sascha/go/src/github.com/cri-o/cri-o/vendor/github.com/containers/storage/store.go:2335 +0x49
created by github.com/containers/storage.(*store).DeleteContainer
        /home/sascha/go/src/github.com/cri-o/cri-o/vendor/github.com/containers/storage/store.go:2334 +0x540
```

cc @rhafer

### DIFF
--- a/pkg/lockfile/lockfile.go
+++ b/pkg/lockfile/lockfile.go
@@ -13,6 +13,9 @@ import (
 // by the lock.
 type Locker interface {
 	// Acquire a writer lock.
+	// The default unix implementation panics if:
+	// - opening the lockfile failed
+	// - tried to lock a read-only lock-file
 	Lock()
 
 	// Acquire a writer lock recursively, allowing for recursive acquisitions
@@ -20,6 +23,9 @@ type Locker interface {
 	RecursiveLock()
 
 	// Unlock the lock.
+	// The default unix implementation panics if:
+	// - unlocking an unlocked lock
+	// - if the lock counter is corrupted
 	Unlock()
 
 	// Acquire a reader lock.


saschagrunert:
> I mentioned it before.
> 
> The panics are intentional. If we have a code path that is touching a lock without having acquired it as a writer, then it's a clear violation of the lock protocol. Such panics are common in go (e.g., [unlocking an unlocked mutex](https://play.golang.org/p/5wCarFJyrbD)) and `pkg/lockfile` panics in various other conditions as well.
> 
> It really looks like a bug in c/storage that needs to be fixed. Having the reproducer is great!

At least there should be some documentation which APIs of c/storage are panicing, that we have a chance to recover in CRI-O.

There are multiple issues, where as some of them are part of CRI-O (working on fixes). I'm not sure if there is a bug in c/storage, but I guess we will trace down the path.

vrothberg:
> I'm not sure if there is a bug in c/storage, but I guess we will trace down the path.

Thanks, @saschagrunert! Do you have a full trace leading to the panic?

> At least there should be some documentation which APIs of c/storage are panicing, that we have a chance to recover in CRI-O.

Sounds like a good idea :+1: The docs of the `Locker` interface are not explicit about the panics.

saschagrunert:
> > I'm not sure if there is a bug in c/storage, but I guess we will trace down the path.
> 
> Thanks, @saschagrunert! Do you have a full trace leading to the panic?
> 

Unfortunately, my example in the first post contains already the "full" trace. The issue starts here:

https://github.com/cri-o/cri-o/blob/0ed20e629f1c783b698859297ba1316c16878f53/server/server.go#L185

The error is not equal to nil because the unmarshalling of the annotations failed there:

https://github.com/cri-o/cri-o/blob/0ed20e629f1c783b698859297ba1316c16878f53/internal/lib/container_server.go#L157

So I think the crash starts to happen at this line:

https://github.com/cri-o/cri-o/blob/0ed20e629f1c783b698859297ba1316c16878f53/server/server.go#L205

(Sorry for the messy links)

> > At least there should be some documentation which APIs of c/storage are panicing, that we have a chance to recover in CRI-O.
> 
> Sounds like a good idea  The docs of the `Locker` interface are not explicit about the panics.

Alright I'll change that PR to update the docs.

pkg/lockfile/lockfile.go
```diff
@@ -13,13 +13,20 @@ import (
 // by the lock.
 type Locker interface {
 	// Acquire a writer lock.
+	// The default unix implementation panics if:
+	// - opening the lockfile failed
+	// - tried to lock a read-only lock-file
 	Lock()
```
saschagrunert, line 19: Theoretically we would have to add those notes to every method of `Store` as well but I think this is somehow overkill. :thinking:

PR #586, opened 2020-04-03
Downgrade logrus to v1.4.2

saschagrunert:
The upgrade of logrus forces to update the dependency in go module
consumers as well.

The new locking in logrus 1.5.0 introduces regressions in conjunction to
deadlocks when using custom hooks, which is the case in CRI-O for
example.

Let's downgrade logrus until the [referencing issue is
fixed](https://github.com/sirupsen/logrus/issues/1122).


umohnani8:
LGTM

TomSweeneyRedHat:
LGTM and happy green tests buttons.
Should a CRIO issue be spun up to remind us of the upgrade and what it's waiting on?

saschagrunert:
> LGTM and happy green tests buttons.
> Should a CRIO issue be spun up to remind us of the upgrade and what it's waiting on?

Yeah why not. :)

PR #587, opened 2020-04-03
Use containers/common unshare package and remove internal

saschagrunert:
This fixes build issues where projects vendor containers/storage and
containers/common because the unshare packages clash when trying to
build them via cgo.

PS: Thats the reason why I cannot test https://github.com/containers/storage/pull/473 in https://github.com/containers/libpod/pull/5706 and https://github.com/cri-o/cri-o/pull/3521 right now.

See: https://circleci.com/gh/openSUSE/cri-o/42914?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link


rhatdan:
containers/common v1.0.0 was reverted, and I have no idea where it is coming from.  Something is strange on github, that this is showing up.

rhatdan:
If we update buildah/storage/common at the same time, I think it will get through this issue.

rhatdan:
Hopefully https://github.com/containers/libpod/pull/5700
fixes the problem

saschagrunert:
Okay thank you for the heads up

PR #596, opened 2020-04-15
internalDelete needs the r.byname id deletion

rphillips:
r.lookup references r.byid, r.byname, and r.idindex to lookup a layer. The resulting Delete() function call misses the `delete(r.byname, name)` causing crio to say a layer is in use, when it actually is not.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -992,6 +992,9 @@ func (r *layerStore) deleteInternal(id string) error {
 	if err == nil {
 		os.Remove(r.tspath(id))
 		delete(r.byid, id)
+		for _, name := range layer.Names {
+			delete(r.byname, name)
+		}
 		r.idindex.Delete(id)
 		mountLabel := layer.MountLabel
 		if layer.MountPoint != "" {


rphillips:
cc @nalind

tedyu:
I wonder if the deletion from byname is done by (around line 395):
```
func (r *containerStore) Delete(id string) error {
```

rhatdan:
Looks like this is causing breakage?

rphillips:
The Kubelet is getting into a state where if the first sandbox creation fails (context deadline exceeded), then a subsequent namespace creation (with the same name) fails with a name in use error. This patch looks like the likely culprit. Is there a unit test for layers?

nalind:
> I wonder if the deletion from byname is done by (around line 395):
> 
> ```
> func (r *containerStore) Delete(id string) error {
> ```

You're right that this handles the `Names` in the storage library's `Container` structures, and we have similar logic for its `Image` structures, but we didn't have it for the `Layer` structures.

I didn't think that CRI-O was even using names for `Layer` structures, but either way this patch LGTM.

giuseppe:
I've restarted the failing tests.

The patch LGTM

rhatdan:
We need to get the tests to pass in order to merge.

rhatdan:
There are issues in Ubuntu images, that are unrelated to this PR.

PR #597, opened 2020-04-15
[crio-release-1.17] DeleteContainer(): wait for all goroutines

haircommander:
DeleteContainer() does most of its heavy lifting in a number of
goroutines.  Previously, if any of them returned an error result, that
result was returned to the caller immediately, and DeleteContainer()
would release any locks that it had acquired as part of that.  This
could all happen while other goroutines were still running, and some of
those goroutines depended on having those locks held until they finished
their work, and would panic() when they detected that the locks had been
released out from under them.

Signed-off-by: Nalin Dahyabhai <nalin@redhat.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ require (
 	github.com/Microsoft/go-winio v0.4.15-0.20190919025122-fc70bd9a86b5
 	github.com/Microsoft/hcsshim v0.8.7
 	github.com/docker/go-units v0.4.0
+	github.com/hashicorp/go-multierror v1.0.0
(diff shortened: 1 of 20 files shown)


haircommander:
`[+0316s] E: The repository 'http://us-central1.gce.archive.ubuntu.com/ubuntu disco-updates Release' no longer has a Release file.`
failures unrelated to this PR

haircommander:
PTAL @nalind @rhatdan @vrothberg @giuseppe 
we need this for cri-o 1.17 and 1.16

mrunalp:
LGTM

PR #598, opened 2020-04-16
overlay: recreate workdir on UpdateLayerIDMap

giuseppe:
make sure the work dir is also recreated with the diff dir.

When using index=on, overlay records a reference to the upperdir inode
into work/index, and raises an error if on a successive mount there is
a mismatch between what is stored in the work/index xattr and the
upper dir inode.

Closes: https://github.com/containers/libpod/issues/5816


nalind:
LGTM

rhvgoyal:
So this is in the context of setting up a new overlay mount. That is old upper directory is gone and a fresh upper directory is being setup. If yes, then it makes sense to start with a clean work/ directory as well.

PR #600, opened 2020-04-17
storage:provided an interface to specify the configuration file path

zklei:
Add an interface to sets the default storage configuration to the specified path

In some scenarios, the default configuration path of the storage needs to be changed.
Therefore, provided an interface to specify the configuration file path.

Signed-off-by: leizhongkai <leizhongkai@huawei.com>

### DIFF
--- a/store.go
+++ b/store.go
@@ -3397,7 +3397,7 @@ func copyStringInterfaceMap(m map[string]interface{}) map[string]interface{} {
 }
 
 // defaultConfigFile path to the system wide storage.conf file
-const defaultConfigFile = "/etc/containers/storage.conf"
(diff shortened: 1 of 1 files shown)


rhatdan:
I am fine with this.  For some reason the ubuntu images are screwed up right now.

TomSweeneyRedHat:
LGTM

zklei:
Thank you everyone~  
@rhatdan @TomSweeneyRedHat I wonder why four testing errors occur. According to the detailed logs, it seems that the error is caused by the CI environment problem?

rhatdan:
Merging since ths PR is not causing the problem.

rhatdan:
Ubuntu apt update seems to be screwed up right now.

PR #603, opened 2020-04-21
Cirrus: Update Fedora and Ubuntu VM images

cevich:


.cirrus.yml
```diff
@@ -19,9 +19,9 @@ env:
     ####
     # GCE project where images live
     IMAGE_PROJECT: "libpod-218412"
-    _BUILT_IMAGE_SUFFIX: "libpod-5874660151656448"
-    FEDORA_CACHE_IMAGE_NAME: "fedora-31-${_BUILT_IMAGE_SUFFIX}"
-    PRIOR_FEDORA_CACHE_IMAGE_NAME: "fedora-30-${_BUILT_IMAGE_SUFFIX}"
+    _BUILT_IMAGE_SUFFIX: "libpod-6301182083727360"
+    FEDORA_CACHE_IMAGE_NAME: "fedora-32-${_BUILT_IMAGE_SUFFIX}"
+    PRIOR_FEDORA_CACHE_IMAGE_NAME: "fedora-31-${_BUILT_IMAGE_SUFFIX}"
```
TomSweeneyRedHat, line 24: Just curious, where's the libpod number coming from?

.cirrus.yml
```diff
@@ -19,9 +19,9 @@ env:
...
     UBUNTU_CACHE_IMAGE_NAME: "ubuntu-19-${_BUILT_IMAGE_SUFFIX}"
```
saschagrunert, line 25: Do we have to change this as well?

.cirrus.yml
cevich: It's a valid question.  It comes from https://github.com/containers/libpod/pull/5737 Cirrus-CI task `test_build_cache_images` which is only activated when `CI:IMG` is present in the PR description.  The number directly correspond to the webpage in the WebUI, so:

`https://cirrus-ci.com/task/6301182083727360`...which doesn't exist for some reason(?)  You'll have to trust me that it did exist at some point before it got pruned :grin: 

In short, for normal purposes, you can just copy the value from libpod's .cirrus-ci.yml.  This is a special-case, since we really only needed the new Ubuntu 19.10 image.

.cirrus.yml
cevich: Nope, it only uses the major version number, so 19.04 -> 19.10 still shows as 19.  Thanks for checking.

PR #608, opened 2020-04-28
lost mountopts when read json from file

shsjshentao:
The type of container flag is map[string]interface{}.
json unmarshal from container.json will convert []string to []interface{}, thus MountOpts will be lost.
We should assert it in case lost MountOpts data

Signed-off-by: blade <blade.shen@ucloud.cn>

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -148,10 +148,20 @@ func (c *Container) ProcessLabel() string {
 }
 
 func (c *Container) MountOpts() []string {
-	if mountOpts, ok := c.Flags["MountOpts"].([]string); ok {
+	switch c.Flags["MountOpts"].(type) {
+	case []string:
+		return c.Flags["MountOpts"].([]string)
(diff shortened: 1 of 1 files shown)


rhatdan:
@shsjshentao You need to sign your PRs

git commit -a --amend -s
git push --force

shsjshentao:
@rhatdan done, is it ok?

rhatdan:
@giuseppe @nalind @vrothberg @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
Tests aren't happy.  Speaking of which, could a test be added for this?

nalind:
LGTM.  This is consistent with the docs for `encoding/json.Unmarshal()`.

rhatdan:
Ubuntu tests have been failing for a while, unrelated to this PR.

PR #609, opened 2020-04-28
only chown when mounting read write on zfs

### LINKED ISSUE #602: zfs driver fails at chown
Dear developers,
I get following error message when building from docker files, for example, https://github.com/diegogslomp/samba-ad-dc

```
Error: error committing container for step {Env:[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin] Command:run Args:[yum update -y] Flags:[] Attrs:map[] Message:RUN yum update -y Original:RUN yum update -y}: error copying layers and metadata for container "40baf6bf2a7f60d455279f160223354decae52d17e4a966cb680506bf9968ec6": Error initializing source containers-storage:centos-working-container: error extracting layer "9a6dda6e1eb9e097fbe3374c9aa20abab3365585845121bd28f30b8af5ef17a8": error modifying zfs mountpoint (/var/lib/containers/storage/zfs/graph/aff9833d464191900b4a3b6e68aad912f930de639b411124e433aeb3757001cd) directory ownership: chown /var/lib/containers/storage/zfs/graph/aff9833d464191900b4a3b6e68aad912f930de639b411124e433aeb3757001cd: read-only file system
```

The command that fails is in https://github.com/containers/storage/blob/847e58e7c9560a8e3ab105a3927da902b32bdd59/drivers/zfs/zfs.go#L411

Same issue on on CentOS 7 as well as CentOS 8

```
$ yum info containers-common.x86_64 
Name       : containers-common
Architektur : x86_64
Epoche             :2
Version    : 0.2.0
Ausgabe    : 2.1.el7
Größe : 44 k
Quelle      : installed
Aus Quelle  : devel_kubic_libcontainers_stable
```

my /etc/containers/storage.conf looks like this:
```
[storage]
driver = "zfs"
runroot = "/var/run/containers/storage"

[storage.options]
additionalimagestores = [
]

[storage.options.zfs]
fsname="testpool/containers"
mountopt="nodev"
```

I have not set up rootless containers and subuid/subgid, just running as root.

asteven:
fixes #602

Signed-off-by: Steven Armstrong <steven.armstrong@id.ethz.ch>

### DIFF
--- a/drivers/zfs/zfs.go
+++ b/drivers/zfs/zfs.go
@@ -384,9 +384,21 @@ func (d *Driver) Get(id string, options graphdriver.MountOpts) (_ string, retErr
 		}
 	}()
 
+	// In the case of a read-only mount we first mount read-write so we can set the
+	// correct permissions on the mount point and remount read-only afterwards.
+	remountReadOnly := false
 	mountOptions := d.options.mountOptions
 	if len(options.Options) > 0 {
-		mountOptions = strings.Join(options.Options, ",")
+		var newOptions []string
+		for _, option := range options.Options {
+			if option == "ro" {
+				// Filter out read-only mount option but remember for later remounting.
+				remountReadOnly = true
+			} else {
+				newOptions = append(newOptions, option)
+			}
+		}
+		mountOptions = strings.Join(newOptions, ",")
 	}
 
 	filesystem := d.zfsPath(id)
@@ -409,7 +421,14 @@ func (d *Driver) Get(id string, options graphdriver.MountOpts) (_ string, retErr
 	// this could be our first mount after creation of the filesystem, and the root dir may still have root
 	// permissions instead of the remapped root uid:gid (if user namespaces are enabled):
 	if err := os.Chown(mountpoint, rootUID, rootGID); err != nil {
-		return "", fmt.Errorf("error modifying zfs mountpoint (%s) directory ownership: %v", mountpoint, err)
+		return "", errors.Wrapf(err, "modifying zfs mountpoint (%s) ownership", mountpoint)
+	}
+
+	if remountReadOnly {
+		opts = label.FormatMountLabel("remount,ro", options.MountLabel)
+		if err := mount.Mount(filesystem, mountpoint, "zfs", opts); err != nil {
+			return "", errors.Wrap(err, "error remounting zfs mount read-only")
+		}
 	}
 
 	return mountpoint, nil


drivers/zfs/zfs.go
```diff
@@ -406,10 +406,20 @@ func (d *Driver) Get(id string, options graphdriver.MountOpts) (_ string, retErr
...
+			readWrite = false
+			break
+		}
+	}
+
+	if readWrite {
+		// this could be our first mount after creation of the filesystem, and the root dir may still have root
+		// permissions instead of the remapped root uid:gid (if user namespaces are enabled):
+		if err := os.Chown(mountpoint, rootUID, rootGID); err != nil {
+			return "", fmt.Errorf("error modifying zfs mountpoint (%s) directory ownership: %v", mountpoint, err)
```
saschagrunert, line 421:
```suggestion
			return "", errors.Wrapf(err, "modifying zfs mountpoint (%s) ownership", mountpoint)
```

drivers/zfs/zfs.go
```diff
@@ -406,10 +406,20 @@ func (d *Driver) Get(id string, options graphdriver.MountOpts) (_ string, retErr
...
+		if option == "ro" {
+			readWrite = false
+			break
+		}
+	}
+
+	if readWrite {
+		// this could be our first mount after creation of the filesystem, and the root dir may still have root
+		// permissions instead of the remapped root uid:gid (if user namespaces are enabled):
+		if err := os.Chown(mountpoint, rootUID, rootGID); err != nil {
```
saschagrunert, line 420: I'd love to see a test for that if possible.

asteven: I can try. Can you give me a hint how to do that?

saschagrunert: Hm, I cant find zfs integration tests in the current test-runs, but usually I would expect that we can create a new test case in one of (or a new) bats test files there: https://github.com/containers/storage/tree/master/tests

But if ZFS is not supported by the CI right now (which seems the case to me), then we should not block that PR here.

drivers/zfs/zfs.go
```diff
@@ -406,10 +406,20 @@ func (d *Driver) Get(id string, options graphdriver.MountOpts) (_ string, retErr
...
-		return "", fmt.Errorf("error modifying zfs mountpoint (%s) directory ownership: %v", mountpoint, err)
+	readWrite := true
+	for _, option := range options.Options {
+		if option == "ro" {
+			readWrite = false
+			break
+		}
+	}
+
+	if readWrite {
```
giuseppe, line 417: wouldn't this introduce a problem when namespaces are used since we skip the chown now?

I think what we need to do is:

- Mount filtering the `ro` flag
- Do the chown
- If `ro` was initially specified then remount `ro` (using `MS_REMOUNT`)

asteven: Changed as requested. Please review.

giuseppe:
could you please rebase your PR and not include any merge commit?

asteven:
Rebased as requested. Please review.

PR #610, opened 2020-05-01
Update libpod image to see if it fixes ubuntu issues

rhatdan:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -15,15 +15,20 @@ env:
     CIRRUS_CLONE_DEPTH: 50
 
     ####
-    #### Cache-image names to test with
-    ####
+    #### Cache-image names to test with (double-quotes around names are critical)
+    ###
+    FEDORA_NAME: "fedora-32"
+    PRIOR_FEDORA_NAME: "fedora-31"
+    UBUNTU_NAME: "ubuntu-19"
+    PRIOR_UBUNTU_NAME: "ubuntu-18"
+
     # GCE project where images live
     IMAGE_PROJECT: "libpod-218412"
-    _BUILT_IMAGE_SUFFIX: "libpod-6301182083727360"
-    FEDORA_CACHE_IMAGE_NAME: "fedora-32-${_BUILT_IMAGE_SUFFIX}"
-    PRIOR_FEDORA_CACHE_IMAGE_NAME: "fedora-31-${_BUILT_IMAGE_SUFFIX}"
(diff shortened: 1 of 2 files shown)


rhatdan:
@cevich is this what you intended, Does not seem to have fixed anything.

cevich:
> is this what you intended, Does not seem to have fixed anything.

Yes, perfect.  Damn...why that didn't work...(investigating)

cevich:
...okay, so I see some parts of the matrix are running and passing, but others are failing:

```
...cut...
[+0262s] ++ time bats --jobs 8 --tap .
[+0263s] 1..47
[+0263s] parallel: Warning: $HOME not set. Using /tmp.
[+0266s] cat: /tmp/bats-run-17757/parallel_output/1/stdout: No such file or directory
...
```

I don't see that complaint about `$HOME` in any of the passing tasks.  I know from past experience that Ubuntu handles shell environment setup totally differently, and in some ways incompatible with Fedora.  I've tried to duct-tape over these differences as best I can, but I'm not an Ubuntu expert and may not have not done it right :confused: 

I think the first thing I'd try here is to make `$HOME` exist and see if that fixes the issue.  Maybe this would help:

```diff
diff --git a/contrib/cirrus/lib.sh b/contrib/cirrus/lib.sh
index df199c641..ad5f295fc 100755
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -8,7 +8,7 @@ source /etc/environment  # not always loaded under all circumstances
 
 # Under some contexts these values are not set, make sure they are.
 USER="$(whoami)"
-HOME="$(getent passwd $USER | cut -d : -f 6)"
+export HOME="$(getent passwd $USER | cut -d : -f 6)"
 [[ -n "$UID" ]] || UID=$(getent passwd $USER | cut -d : -f 3)
 GID=$(getent passwd $USER | cut -d : -f 4)
 
```

cevich:
Well :hankey: what the heck.  Hmmmm, I wonder what's special about the BATS usage here, compared to other the repos.  And it use to work, then suddenly stopped :confused: ... I'll have a poke at this in more depth first thing tomorrow...adding to my list...

cevich:
@rhatdan Thanks for giving my suggestions a try, as I'm quite certain you'd rather be working on more interesting problems :smile: I'll take over this work in a new PR, also since I found a small update that's needed to `hack/get_ci_vm.sh`.  I'm investigating the problem now, hopefully it's a simple fix...(knock on wood)

cevich:
opened #616

PR #615, opened 2020-05-04
unshare: try a bind mount before copy /proc/self/exe

giuseppe:
if unshare is creating a new mount space, attempt first to create a
read only bind mount from /proc/self/exe instead of copying it to a
memfd.

The code was copied from runc/libcontainer.

It has two advantages:

- it requires less memory since we don't need a copy of the executable
  in memory.
- it is much faster ("buildah unshare" is twice faster).

### DIFF
--- a/pkg/unshare/unshare.c
+++ b/pkg/unshare/unshare.c
@@ -15,6 +15,9 @@
 #include <termios.h>
 #include <errno.h>
 #include <unistd.h>
+#include <sys/vfs.h>
+#include <sys/mount.h>
+#include <linux/limits.h>
 
 /* Open Source projects like conda-forge, want to package podman and are based
    off of centos:6, Conda-force has minimal libc requirements and is lacking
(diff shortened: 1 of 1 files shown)


pkg/unshare/unshare.c
```diff
@@ -151,30 +154,89 @@ static char **parse_proc_stringlist(const char *list) {
...
 		fprintf(stderr, "fstat(\"/proc/self/exe\"): %m\n");
+		close(fd);
 		return -1;
 	}
 	exename = basename(argv[0]);
 	mmfd = syscall(SYS_memfd_create, exename, (long) MFD_ALLOW_SEALING | MFD_CLOEXEC);
 	if (mmfd == -1) {
 		fprintf(stderr, "memfd_create(): %m\n");
-		return -1;
+		goto close_fd;
```
rhatdan, line 239: Why don't you use the autoclose fd stuff you use in crun?

giuseppe: I've only adapted the code to the existing pattern coming from `try_bindfd`.

In any case, that is a good idea, I can work on cleaning up the C code and use auto cleanup functions.  Podman can also benefit from it.

rhatdan:
Merging, and we can do the cleanup in a separate PR>

PR #616, opened 2020-05-05
Fix ubuntu testing and get_ci_vm.sh

cevich:


cevich:
@rhatdan PTAL - I added a few comments on the fix.  We should consider just completely dropping the use of `--jobs`, but I was trying to a minimal-fix here.  Even with `$JOBS=1` all the tests complete in under a minute :smile: 

@edsantiago mind peeping at the `get_ci_vm.sh` changes here?  I know you don't use it, so just a glance is fine.  I'm considering copying that change to most of the other containers-org repos. Though at some point I want to improve and centralize this script, this is just a band-aid for now.  The changes are necessary to properly handle printing of image names after you're (fantastic) simplifying of test display names (in libpod).  @rhatdan copied some of that work here.  Hopefully I did good enough commenting and typo-avoiding :grin:

hack/get_ci_vm.sh
```diff
@@ -90,17 +82,31 @@ show_usage() {
...
-import yaml
+    # Deal with both YAML and embedded shell-like substitutions in values
+    # if substitution fails, fall back to printing nakid env. var as-is.
+    python3 -c '
+import yaml,re
 env=yaml.load(open(".cirrus.yml"), Loader=yaml.SafeLoader)["env"]
-keys=[k for k in env if "ENCRYPTED" not in str(env[k])]
+rx1=re.compile(r"\$(\w+)")
+rx2=re.compile(r"\$\{(\w+)\}")
+rep=r"{\1}"
```
edsantiago, line 92: This is really clever - so clever that it took me a long while to figure it out. Would you consider adding comments to help the next maintainer?

hack/get_ci_vm.sh
```diff
@@ -90,17 +82,31 @@ show_usage() {
...
+class ReIterKey(dict):
+    def __missing__(self, key):
+        return "${0}".format(key)
```
edsantiago, line 95: Under what circumstances would this be necessary? Shouldn't it always be an error to reference an undefined variable?

cevich:
> few readability improvements if you're willing.

Exactly why I pinged you, thx :smile:

hack/get_ci_vm.sh
```diff
@@ -90,17 +82,31 @@ show_usage() {
...
-    python -c '
-import yaml
+    # Deal with both YAML and embedded shell-like substitutions in values
+    # if substitution fails, fall back to printing nakid env. var as-is.
+    python3 -c '
+import yaml,re
 env=yaml.load(open(".cirrus.yml"), Loader=yaml.SafeLoader)["env"]
-keys=[k for k in env if "ENCRYPTED" not in str(env[k])]
+rx1=re.compile(r"\$(\w+)")
+rx2=re.compile(r"\$\{(\w+)\}")
```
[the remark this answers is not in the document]
cevich: good idea.  got it.

hack/get_ci_vm.sh
```diff
@@ -90,17 +82,31 @@ show_usage() {
...
     if "ENCRYPTED" not in v:
-        print "{0}=\"{1}\"".format(k, v),
+        out[k]=rx1.sub(rep, rx2.sub(rep, v))
+for k,v in out.items():
+    print("{0}=\"{1}\"".format(k, v.format_map(out)))
     '
 }
 
+image_hints() {
+    get_env_vars | egrep '_CACHE_IMAGE_NAME' | awk -F "=" '{print $2}'
```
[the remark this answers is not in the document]
cevich: maaaaaybe...what's an fgrep....OOOohhh! Okay, didn't know that one.

hack/get_ci_vm.sh
cevich: yep, you got it boss.

hack/get_ci_vm.sh
cevich: Sadly no.  Some variables are only available by Cirrus-CI at runtime.  This script doesn't pretend to perfectly duplicate their YAML env-var-like parser.  It's perfectly fine to reference one of these runtime-only variables in the YAML for Cirrus.  Nobody can can predict if/when might be added.  I don't want that to break this script :confused: 

(Why having a more robust and unit-tested `hack/get_ci_vm.sh` script would be nice)

Hmmm...how to comment that concisely...

cevich: ...okay I think I got it.

cevich:
@edsantiago thanks for your feedback, should be all fixed up and spell-checked now :grin:

edsantiago:
/lgtm
/hold

Nice work, thank you!

cevich:
> /lgtm /hold

Won't work on this repo, no bots are watching.  I don't even think we have Bors...nope, all manual lifting here :disappointed:  Better add that to the todo list...done.

PR #618, opened 2020-05-08
Fix two issues with .dockerignore handling

yann-soubeyrand:
Some .dockerignore patterns are not handled correctly:

- `dir\` incorrectly matches `dir$`
- `**/dir` doesn't match `dir/file`

This PR should solve these two issues.


rhatdan:
@saschagrunert PTAL

PR #620, opened 2020-05-11
ioutils/writer: allow to disable fsync

giuseppe:
permit users of AtomicFileWriter to disable the fdatasync call that is
performed after the file is written.  CRI-O in particular, completely
wipes out the storage on an not clean system restart, so there is no
gain in syncing files.

### DIFF
--- a/pkg/ioutils/fswriters.go
+++ b/pkg/ioutils/fswriters.go
@@ -7,24 +7,51 @@ import (
 	"path/filepath"
 )
 
-// NewAtomicFileWriter returns WriteCloser so that writing to it writes to a
+// AtomicFileWriterOptions specifies options for creating the atomic file writer.
+type AtomicFileWriterOptions struct {
+	// NoSync specifies whether the sync call must be skipped for the file.
+	// If NoSync is not specified, the file is synced to the
+	// storage after it has been written and before it is moved to
+	// the specified path.
+	NoSync bool
+}
+
+var defaultWriterOptions AtomicFileWriterOptions = AtomicFileWriterOptions{}
+
+// SetDefaultOptions overrides the default options used when creating an
+// atomic file writer.
+func SetDefaultOptions(opts AtomicFileWriterOptions) {
+	defaultWriterOptions = opts
+}
+
+// NewAtomicFileWriterWithOpts returns WriteCloser so that writing to it writes to a
 // temporary file and closing it atomically changes the temporary file to
 // destination path. Writing and closing concurrently is not allowed.
-func NewAtomicFileWriter(filename string, perm os.FileMode) (io.WriteCloser, error) {
+func NewAtomicFileWriterWithOpts(filename string, perm os.FileMode, opts *AtomicFileWriterOptions) (io.WriteCloser, error) {
 	f, err := ioutil.TempFile(filepath.Dir(filename), ".tmp-"+filepath.Base(filename))
 	if err != nil {
 		return nil, err
 	}
-
+	if opts == nil {
+		opts = &defaultWriterOptions
+	}
 	abspath, err := filepath.Abs(filename)
 	if err != nil {
 		return nil, err
 	}
 	return &atomicFileWriter{
-		f:    f,
-		fn:   abspath,
-		perm: perm,
+		f:      f,
+		fn:     abspath,
+		perm:   perm,
+		noSync: opts.NoSync,
 	}, nil
+}
+
+// NewAtomicFileWriter returns WriteCloser so that writing to it writes to a
+// temporary file and closing it atomically changes the temporary file to
+// destination path. Writing and closing concurrently is not allowed.
+func NewAtomicFileWriter(filename string, perm os.FileMode) (io.WriteCloser, error) {
+	return NewAtomicFileWriterWithOpts(filename, perm, nil)
 }
 
 // AtomicWriteFile atomically writes data to a file named by filename.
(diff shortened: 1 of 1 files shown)


rhatdan:
@saschagrunert @nalind @TomSweeneyRedHat @mrunalp PTAL

pkg/ioutils/fswriters.go
```diff
@@ -7,26 +7,49 @@ import (
 	"path/filepath"
 )
 
-// NewAtomicFileWriter returns WriteCloser so that writing to it writes to a
+// AtomicFileWriterOptions specifies options for creating the atomic file writer.
+type AtomicFileWriterOptions struct {
+	NoSync bool
```
nalind, line 12: Document this field, please.

cgwalters:
This is a complex topic.  I am actually investigating https://github.com/openshift/machine-config-operator/issues/1897 which is a similar issue in ostree - invoking `fsync()` impacts etcd performance and needs to be batched.  For the operating system, we obviously need integrity on disk.

One trick OSTree uses btw is to store files in a "per-boot-id" directory, e.g. `mkdir -p /var/lib/containers/storage/temp/$(/proc/sys/kernel/random/boot_id)` - and those files are not fsyncd.  

But it's also quite expensive to re-pull all containers on reboot.  So I think what we should do for containers is write without `fsync()` there, and then periodically for containers that have been around a while, invoke `fsync()` on them and once successful move them to `/var/lib/containers/stable` or so.

I think though a good way to think about `fsync()` is - in many cases, it's probably OK for us to lose the last 10-15 seconds of data.  It's not OK to lose all of it.   The classic example is a text editor, but it applies everywhere - here, also things like Firefox saving your browser history.  Firefox shouldn't invoke `fsync()` on its sqlite database immediately after I go to example.com.  It should happen in a batch 10-15 or 30s after.  So the pattern to follow is "delayed fsync" IMO.

giuseppe:
>  When we're ready to do a commit, we walk over all of them `fsync()` and rename them into place in `/ostree/repo/objects`.

is a `fsync` performed for each file?  Wouldn't be cheaper to use `syncfs`?

One problem we have in containers/storage is that the image should be available immediately in the final destination (or at least when a container using it is started), otherwise it becomes very difficult to manage the overlay mounts and we can't rename files/directories if they are already used in the lower directories.

I agree we must to avoid pulling images if not necessary; the issue introduced with the CRI-O change just made easier to reproduce an underlying issue in the containers/storage that files are not synced to disk, just the metadata is not enough.

While with the metadata we see immediately the issue as containers are not created, I think it is much worse when the metadata is fully written but there is corruption in the containers images.  That is difficult to detect and I'd expect any sort of weird behaviour if files are missing or truncated.

For the cluster use case, I think we need to make sure a full `sync()` is performed by CRI-O at shutdown and then register somewhere that it happened.  On the next boot, if the file is missing then the storage is recreated, this is expensive but I think it is better to be safe that there are no corrupted images in the storage.  When CRI-O modifies the storage (for simplicity we can do it at startup in any case for now), it must make sure the file is deleted and the parent directory is `fsync()`, so we know the storage is in a "modified by not synced" state.

cgwalters:
> is a fsync performed for each file? Wouldn't be cheaper to use syncfs?

Yes.  And ostree does *also* use `syncfs` but...that's a big hammer that affects the whole system.  I'd like to avoid doing that.  (If we are going to do that ideally we can hook into some sort of coordination process)

> For the cluster use case, I think we need to make sure a full sync() is performed by CRI-O at shutdown and then register somewhere that it happened.

If we get to shutdown and the filesystem is cleanly unmounted, then everything will be consistent.  The problem is we don't *know* that...so indeed we need to do something like `syncfs && mv /path/to/storage/transient /path/to/storage/permanent` (just a directory rename) as part of service shutdown.  We are effectively doing that today with `ostree-finalize-staged.service` for ostree although we aren't fully properly delaying the sync until then.

mrunalp:
> Yes. And ostree does also use syncfs but...that's a big hammer that affects the whole system. I'd like to avoid doing that. (If we are going to do that ideally we can hook into some sort of coordination process)

Agree, this will make the problems like we are having with etcd worse if we just wait till shutdown to do the sync.

giuseppe:
> Agree, this will make the problems like we are having with etcd worse if we just wait till shutdown to do the sync.

For long running processes, I'd not expect that the entire file is not synced.  The kernel syncs inodes that are dirty in memory longer than a certain interval (by default 30 seconds on Fedora), or when the used memory crosses the maximum ratio of dirty memory (on Fedora 10%).  I'd also expect etcd to hande fsync itself.  Anyway, I don't know about the issue with etcd :) is it tracked anywhere?

mrunalp:
>  Anyway, I don't know about the issue with etcd :) is it tracked anywhere
https://github.com/openshift/machine-config-operator/issues/1897 has the relevant discussion and pointers :)
 
I think if we have a combination of periodic syncs like Colin suggested and we checkpoint data so we can recover all the synced layers we would be much better equipped to deal with unclean shutdowns or power-offs.

In the current issue, we ended up in a situation where we would have to completely wipe storage to recover and then re-pull all the containers.
Instead, if we can recover 90 - 95 % then it won't be so bad.

giuseppe:
That is definitely something we can (and should) do in CRI-O

giuseppe:
Thinking more of it, the delayed sync model is difficult to fit into containers/storage.  Ideally we would have a way to verify the images (like ostree fsck) to be sure that what we have is not corrupted.  Should we rather look into it?

mrunalp:
> Thinking more of it, the delayed sync model is difficult to fit into containers/storage. Ideally we would have a way to verify the images (like ostree fsck) to be sure that what we have is not corrupted. Should we rather look into it?

Yes, something that we can run on startup when we detect that crio didn't shutdown cleanly.

PR #621, opened 2020-05-12
bugfix for store call layerStore's Load method

zvier:
When build images high concurrency, a lot of export may happen.
To read layers, many Diff() in store.go will call frequently, and
Layers() call also occur. These two method call layerStore.Load()
with a RLock. The Load() will rebuild all layers, this may change
this layerStore's state temporary until it completely done. What
makes it terrible is in store.Diff->layerStore.Diff->r.newFileGetter
will get a simpleGetCloser with Mount or Unmount a layer. When Unmout,
a true unmout may happen with a new layer which Load() method rebuild,
but not yet load mount info, it's MountCount and MountPoint is empty,
This can cause image export faild with 'no such file or directory' error

Signed-off-by: zvier <zvier6@163.com>

### DIFF
--- a/store.go
+++ b/store.go
@@ -2742,7 +2742,11 @@ func (s *store) Diff(from, to string, options *DiffOptions) (io.ReadCloser, erro
 	}
 	for _, s := range append([]ROLayerStore{lstore}, lstores...) {
 		store := s
-		store.RLock()
+		if store == lstore {
+			store.Lock()
+		} else {
+			store.RLock()
+		}
 		if modified, err := store.Modified(); modified || err != nil {
 			if err = store.Load(); err != nil {
 				return nil, err
@@ -3002,7 +3006,11 @@ func (s *store) Layer(id string) (*Layer, error) {
 	}
 	for _, s := range append([]ROLayerStore{lstore}, lstores...) {
 		store := s
-		store.RLock()
+		if store == lstore {
+			store.Lock()
+		} else {
+			store.RLock()
+		}
 		defer store.Unlock()
 		if modified, err := store.Modified(); modified || err != nil {
 			if err = store.Load(); err != nil {


rhatdan:
Seems to be causing test failures?

zvier:
> Seems to be causing test failures?

sorry，it's ok now. PTAL ~

rhatdan:
@nalind @vrothberg @giuseppe @saschagrunert PTAL

zvier:
> Please add a comment to the checks.
> 
> I'll let @nalind be the gate keeper :)

It's not so easy to capture this error. We build images concurrently with 70 or more build request. Even so, export image error with 'no such file or directory'  is appear occasionally.  The Dockerfile is simple as follows:
```
FROM busybox:latest
RUN touch abc
```

And we also append more log for this package to find the problem. It easy to trace the call stack is: 
copy.Image()->NewImageSource()->i.store.Layer()/i.store.Diff()->layerStore.RLock()->layerStore.Load()->layerStore.Diff()->layerStore.Diff()->layerStore.newFileGetter()->layerStore.Mount()->copier.copyBlobFromStream()->copier.dest.PutBlob()->d.sendFile()->layerStore.Unmount()

With this call stack under a RLock in high concurrency build,  the MountCount of a layer can even be decreased to -1.

store.go
```diff
@@ -2929,7 +2933,11 @@ func (s *store) Layers() ([]Layer, error) {
 
 	for _, s := range lstores {
 		store := s
-		store.RLock()
+		if store == lstore {
```
nalind, line 2936: I don't think this comparison is ever true, so I don't think this hunk changes the behavior.

nalind:
Hmm, not sure I'm following the sequence here.  Both the layerStore Mount() and Unmount() methods are locking the mount counts file and saving it when they update the count for a mountpoint, and Load() resets the in-memory counts after it takes the lock.

Mounted() can also save the counts file, but only takes a read lock, which is a problem, and I could see that throwing the synchronization off.  Would that account for the behavior you're seeing?

zvier:
> Hmm, not sure I'm following the sequence here. Both the layerStore Mount() and Unmount() methods are locking the mount counts file and saving it when they update the count for a mountpoint, and Load() resets the in-memory counts after it takes the lock.
>

The locking file for Mount/Unmout we also noticed, so the prolbem is not so obviously until we append more debug logs in this file. 

Consider two build request BuidlA and BuildB access here  under a RLock. There are two typically scenario sequence can capture:

1. BuidlA is ready to Mount a layerA which get from layerStore.byid and it's address is 0xaaaaaaa1. The layerA's MountPoint and MountCount will update, but not access saveMounts() immediately . At the same time, BuildB access in Load() and rebuild all layers, but not loadMounts() immediately(even loadMounts done it doesn't matter). Then BuidlA continue to saveMounts(), it range all layerStore.layers. The layerA's address is 0xaaaaaaa2 and it's mount information will be override now, because the mount information we get from layerStore.byid which address is 0xaaaaaaa1 is discard. Then saveMounts will missing the mount information which update in Mount() method.

2. When BuildB access in Load() rebuild all layers, but not loadMounts() immediately, and the layerA's addrees 0xaaaaaaa1. BuidlA is ready to Unmout a layerA which get from layerStore.byid and it's address is 0xaaaaaaa1, that's right. But then BuidlA find the MountCount of layerA is 0, decrease it to -1 and give a true umount. If there is another build process is reading this layer and the export process will failed with 'no such file or directory' here.

store.go
zvier: Ok , that's right, I will upate this comparison.

nalind:
Okay, I think I follow what's happening now.  In `layerStore.Load()`, between the point where the `layerStore`'s `layers` field is set and the point where the lock is acquired for loading and updating the mount counts for the layers, the in-memory mount counts for the layers have essentially been zeroed, and if the `layerStore`'s `Mount()` and `Unmount()` methods are called while someone else's call to `Load()` is between those two points, they'll be starting with bad information, which confuses things.

So taking a write lock in `Diff()` is about making sure that other routines can't call `Load()` while `Diff()` is running, since it might end up calling `Mount()` or `Unmount()` as part of its operation.  Is that correct?  If so, is `Layer()` getting a similar change because containers/image also calls it?

zvier:
> Okay, I think I follow what's happening now. In `layerStore.Load()`, between the point where the `layerStore`'s `layers` field is set and the point where the lock is acquired for loading and updating the mount counts for the layers, the in-memory mount counts for the layers have essentially been zeroed, and if the `layerStore`'s `Mount()` and `Unmount()` methods are called while someone else's call to `Load()` is between those two points, they'll be starting with bad information, which confuses things.
> 
> So taking a write lock in `Diff()` is about making sure that other routines can't call `Load()` while `Diff()` is running, since it might end up calling `Mount()` or `Unmount()` as part of its operation. Is that correct? If so, is `Layer()` getting a similar change because containers/image also calls it?

Yes, and I has already update this pull. Both `store.Diff()` and `store.Layer()` was called with a Lock().

nalind:
Thanks for catching this!

Okay, so I gather that the changes in this PR are aimed at making sure that these methods don't act on, or read and return, the mount information while `Load()` has the store's information in an inconsistent state.  I'm worried that the fix is subtle enough that we could break this again later without even realizing it.  Would it be less likely to break going forward if we eliminated the window where the information is inconsistent, by taking the read lock on the mounts info earlier in `Load()`?

giuseppe:
a related PR: https://github.com/containers/storage/pull/631

zvier:
> a related PR: #631

Perfect. So based on you PR, layer store doesn't need to mount the layer when access the diff files because the driver has implement the DiffGetterDriver interface can assess files directly. And the problems about this PR also has solved naturally I think, isn't it ? @giuseppe

giuseppe:
> Perfect. So based on you PR, layer store doesn't need to mount the layer when access the diff files because the driver has implement the DiffGetterDriver interface can assess files directly. And the problems about this PR also has solved naturally I think, isn't it ? @giuseppe

I think yes.  Is there a way for you to verify it in the same testing conditions as you've tried this PR so we can be sure about it?

zvier:
> > Perfect. So based on you PR, layer store doesn't need to mount the layer when access the diff files because the driver has implement the DiffGetterDriver interface can assess files directly. And the problems about this PR also has solved naturally I think, isn't it ? @giuseppe
> 
> I think yes. Is there a way for you to verify it in the same testing conditions as you've tried this PR so we can be sure about it?

I have test it with the same condition may times.  No problem anymore, Thank you !

giuseppe:
> I have test it with the same condition may times. No problem anymore, Thank you !

are you fine to close this pR?

PR #628, opened 2020-05-16
overlay: honor ro option

giuseppe:
if "ro" is specified when mounting overlay, configure overlay without
any upper layer and workdir so the file system doesn't attempt to
write any file.

Closes: https://github.com/containers/buildah/issues/2359

Signed-off-by: Giuseppe Scrivano <giuseppe@scrivano.org>

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -809,6 +809,13 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	}
 	readWrite := true
 
+	for _, o := range options.Options {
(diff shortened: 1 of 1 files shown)


rhatdan:
@TomSweeneyRedHat @nalind @vrothberg @saschagrunert @mheon PTAL

TomSweeneyRedHat:
Changes LGTM, but some of the tests aren't hip.  Would like a @nalind head nod if possible

giuseppe:
> Changes LGTM, but some of the tests aren't hip

tests are green now

PR #631, opened 2020-05-19
overlay: implement DiffGetter

giuseppe:
if the DiffGetter interface is provided, the layer store doesn't need
to mount the layer to access the files for tar-split.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -33,6 +33,7 @@ import (
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/vbatts/tar-split/tar/storage"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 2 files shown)


giuseppe:
It is needed for: https://github.com/containers/buildah/issues/2359

rhatdan:
What about the other drivers?  Don't they at least need the interface?

giuseppe:
> What about the other drivers? Don't they at least need the interface?

it must be customized for each driver, at the moment only aufs has the interface.  I am not sure it is even possible for all the remaining drivers to access a file without mounting the layer.

rhatdan:
@nalind @saschagrunert @vrothberg @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM, but would like someone with more storage chops than I to do a head nod too.

nalind:
LGTM.  The implementations for overlay and vfs mirror the aufs one, which uses a layout for its diffs that is similar to overlay's.

PR #639, opened 2020-06-02
Add back skip_mount_home

rhatdan:
Certain workloads, we would like to eliminate the mounting of containers-storage as private.
Running containers within containers for example.

This looks like it was accidently removed in the past, since there was still partial
implementation.

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -198,6 +198,9 @@ The `storage.options.zfs` table supports the following options:
 **mountopt**=""
   Comma separated list of default options to be used to mount container images.  Suggested value "nodev". Mount options are documented in the mount(8) man page.
 
+**skip_mount_home=""**
+  Tell storage drivers to not create a PRIVATE bind mount on their home directory.
+
 **size**=""
   Maximum size of a container image.   This flag can be used to set quota on the size of container images. (format: <number>[<unit>], where unit = b (bytes), k (kilobytes), m (megabytes), or g (gigabytes))
 
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -311,6 +311,9 @@ func parseOptions(options []string) (*overlayOptions, error) {
 				return nil, fmt.Errorf("overlay: can't stat program %s: %v", val, err)
 			}
 			o.mountProgram = val
+		case "overlay2.skip_mount_home", "overlay.skip_mount_home", ".skip_mount_home":
+			logrus.Debugf("overlay: skip_mount_home=%s", val)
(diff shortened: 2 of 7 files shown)


rhatdan:
@edsantiago PTAL
This would allow us to eliminate the bind mount, anyways.

edsantiago:
Will this lead to warnings? I see this in `pkg/config/config.go:273`
```go
        if options.Overlay.SkipMountHome != "" || options.SkipMountHome != "" {
            logrus.Warn("skip_mount_home option is no longer supported, ignoring option")
        }
```
And, I realize this is a long shot, but would it make sense to give the option a positive name (`mount_home`) and hide the internal fact that it's implemented as a negative? My tiny brain has never been able to handle negative options well.

edsantiago:
OBTW the reason I found that was in looking for code that would actually trigger based on `skipMountHome`. It looks like that code is still present in `overlay/overlay.go` but I haven't actually tested it. The warning was added in 31b1ca9b906b90f32ea0814b9696a59698dea0d3.

rhatdan:
@edsantiago Thanks, I don't think I should change the name since there is some history, although I do agree that double negative names is difficult to comprehend.

Thanks for finding my PR, it made it easier to revert.  Now that we found an actual use case for the change.

rhatdan:
@edsantiago I just attempted the test for run_commands.go with this patch and the storage.conf set to skip, and podman ran successfully.

rhatdan:
@vrothberg @edsantiago @baude @giuseppe @TomSweeneyRedHat @saschagrunert PTAL

giuseppe:
I've added this feature and then removed it.  It was meant for a different reason (i.e. allow deduplication via ostree hard links).  It is a system-wide configuration and it affects every container, and I think it is also expensive to carry all these mounts in all the mount namespaces when there are many containers.

What use case are we trying to fix with it?

rhatdan:
This is blowing up podman within a container without SYS_ADMIN privs.  Just executing `podman help` commands is not joining a usernamespace and attempting to mount the storage and blowing up because of no CAP_SYS_ADMIN. 

If we don't skip the mount, then every podman command will not execute the mount and require to join the user namespace or blow up without SYS_ADMIN privs.

rhatdan:
This potentially could fix
https://github.com/containers/libpod/pull/6374

giuseppe:
ok then I guess it is fine to address it with `skip_mount_home`.  It is more immediate solution than anything I can think of.

LGTM

rhatdan:
What I don't understand though is why this is not hitting everyone?  If I run rootless podman, it either does not check to see if ~/.local/share/containers/storage is a mount point or just does not go through this code.

PR #646, opened 2020-06-18
devmapper: increase LVM PV metadatasize

mikemccracken:
The default is ~1k and is not enough to store metadata about
hundreds of LVs. This bumps it to 128M.

128M was chosen in the context of relatively large hosts, so the exact value
could be up for discussion if people are using the devicemapper driver on
small hosts.

We have experienced metadata exhaustion in other contexts with many LVs,
so this change was a proactive measure to avoid problems with many containers and images.

### DIFF
--- a/drivers/devmapper/device_setup.go
+++ b/drivers/devmapper/device_setup.go
@@ -206,7 +206,7 @@ func setupDirectLVM(cfg directLVMConfig) error {
 		cfg.ThinpMetaPercent = 1
 	}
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@rhvgoyal PTAL

rhatdan:
This should be specified in storage.conf, rather then hard coded, then different users/distributions can set their defaults.

rhvgoyal:
Agreed. This should be configurable by user. container-storage-setup had it part of config file as well where user could override the defaults.

PR #647, opened 2020-06-18
devmapper: allow devmapper devices as directlvm device

mikemccracken:
LVM allows creating a PV on top of an LV, but resolving symlinks
before checking the output of lvmdiskscan erroneously reports that an
LV is not an available device. It will resolve an LV's name into
/dev/dm-NN, which won't show up in lvmdiskscan.

Instead this patch looks for both names in lvmdiskscan.

The motivation for this was to allow using an LV instead of a whole PV as the backing device for the devicemapper driver.

### DIFF
--- a/drivers/devmapper/device_setup.go
+++ b/drivers/devmapper/device_setup.go
@@ -121,15 +121,19 @@ func checkDevHasFS(dev string) error {
 }
 
 func verifyBlockDevice(dev string, force bool) error {
-	realPath, err := filepath.Abs(dev)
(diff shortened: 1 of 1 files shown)


rhatdan:
@rhvgoyal PTAL

rhatdan:
LGTM

rhatdan:
@nalind @rhvgoyal @vrothberg @saschagrunert PTAL

mikemccracken:
great, thanks all!

PR #648, opened 2020-06-19
new interface for deleting Store object

yangfeiyu20102011:
New interface for deleting already-created Store object

When the store is not needed, it is neccessary to remove
it and release the memory

Signed-off-by: yangfeiyu <yangfeiyu20102011@163.com>

### DIFF
--- a/store.go
+++ b/store.go
@@ -3601,3 +3601,13 @@ func GetMountOptions(driver string, graphDriverOptions []string) ([]string, erro
 	}
 	return nil, nil
 }
+
+// DeleteStore deletes an already-created Store object
+func DeleteStore(s Store) {
+	for i := 0; i < len(stores); i++ {
+		if stores[i] == s {
+			stores = append(stores[:i], stores[i+1:]...)
+			return
+		}
+	}
+}


yangfeiyu20102011:
@rhatdan PTAL  Thanks!

rhatdan:
Would it make more sense to have this be

func (s store) Delete() {
}

rhatdan:
@nalind @giuseppe @saschagrunert @vrothberg PTAL

store.go
```diff
@@ -3601,3 +3601,13 @@ func GetMountOptions(driver string, graphDriverOptions []string) ([]string, erro
 	}
 	return nil, nil
 }
+
+// DeleteStore deletes an already-created Store object
+func DeleteStore(s Store) {
```
saschagrunert, line 3606: I agree we could bind it to `Store`.

giuseppe:
> Would it make more sense to have this be
> 
> func (s store) Delete() {
> }

yes I agree

rhatdan:
@yangfeiyu20102011 Are you still working on this?

rhatdan:
Replaced by https://github.com/containers/storage/pull/653

PR #650, opened 2020-06-24
Lock files should be CLOEXEC

rhatdan:
We are getting reports of SELinux failures on leaked file descriptors.

### DIFF
--- a/pkg/lockfile/lockfile_unix.go
+++ b/pkg/lockfile/lockfile_unix.go
@@ -35,9 +35,9 @@ type lockfile struct {
 // necessary.
 func openLock(path string, ro bool) (int, error) {
 	if ro {
-		return unix.Open(path, os.O_RDONLY, 0)
+		return unix.Open(path, os.O_RDONLY|unix.O_CLOEXEC, 0)
 	}
-	return unix.Open(path, os.O_RDWR|os.O_CREATE, unix.S_IRUSR|unix.S_IWUSR)
+	return unix.Open(path, os.O_RDWR|unix.O_CLOEXEC|os.O_CREATE, unix.S_IRUSR|unix.S_IWUSR)
 }
 
 // createLockerForPath returns a Locker object, possibly (depending on the platform)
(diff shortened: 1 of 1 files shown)


rhatdan:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1840666

rhatdan:
@nalind @saschagrunert @giuseppe @vrothberg @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM
but definitely want a head nod from one of the other folks on your list.

pkg/lockfile/lockfile_unix.go
```diff
@@ -35,9 +35,9 @@ type lockfile struct {
 // necessary.
 func openLock(path string, ro bool) (int, error) {
 	if ro {
-		return unix.Open(path, os.O_RDONLY, 0)
+		return unix.Open(path, os.O_RDONLY|unix.FD_CLOEXEC, 0)
```
nalind, line 38: Shouldn't this be `unix.O_CLOEXEC`?

nalind:
`openLock()` is called from two places, and the one that doesn't (`lockfile.lock()`) closes the file immediately calls `unix.CloseOnExec()`, which should no longer be necessary, so we might as well remove that.

giuseppe:
> We are getting reports of SELinux failures on leaked file descriptors.

The patch looks good and we should definitely open fds with `O_CLOEXEC` but do we know where these FDs are leaked?  Go should not leak FDs if they are not explicitely set to be inherited

pkg/lockfile/lockfile_unix.go
rhatdan: Yes, I originally looked for os.O_CLOEXEC, and later found unix.FD_CLOEXEC.  Did not look for O_CLOEXEC

rhatdan:
The linked bugzilla shows SELinux reporting on leaded file descriptors.  I would guess the unix fds are leaked unless you tell unix to not leak them.

PR #655, opened 2020-06-30
Store the pvcreate --metadatasize option in storage.conf

rhatdan:
As the number of devices increase the size of the metadata
needs to be modified.

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -138,6 +138,9 @@ The `storage.options.thinpool` table supports the following options for the `dev
     6: LogLevelInfo
     7: LogLevelDebug
 
+**metadata_size**=""
+  metadata_size is used to set the `pvcreate --metadatasize` options when creating thin devices. (Default 128k)
+
 **min_free_space**=""
   Specifies the min free space percent in a thin pool required for new device creation to succeed. Valid values are from 0% - 99%. Value 0% disables. (default: 10%)
 
--- a/drivers/devmapper/device_setup.go
+++ b/drivers/devmapper/device_setup.go
@@ -23,6 +23,7 @@ type directLVMConfig struct {
 	ThinpMetaPercent    uint64
 	AutoExtendPercent   uint64
 	AutoExtendThreshold uint64
+	MetaDataSize        string
 }
 
 var (
@@ -209,8 +210,11 @@ func setupDirectLVM(cfg directLVMConfig) error {
 	if cfg.ThinpMetaPercent == 0 {
 		cfg.ThinpMetaPercent = 1
 	}
+	if cfg.MetaDataSize == "" {
+		cfg.MetaDataSize = "128k"
+	}
 
-	out, err := exec.Command("pvcreate", "-f", cfg.Device).CombinedOutput()
+	out, err := exec.Command("pvcreate", "--metadatasize", cfg.MetaDataSize, "-f", cfg.Device).CombinedOutput()
 	if err != nil {
 		return errors.Wrap(err, string(out))
 	}
--- a/drivers/devmapper/deviceset.go
+++ b/drivers/devmapper/deviceset.go
@@ -101,6 +101,7 @@ type DeviceSet struct {
 
 	// Options
 	dataLoopbackSize      int64
+	metaDataSize          string
 	metaDataLoopbackSize  int64
 	baseFsSize            uint64
 	filesystem            string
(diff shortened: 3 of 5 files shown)


rhatdan:
Replaces https://github.com/containers/storage/pull/646

rhatdan:
@mikemccracken could you try this out and see if it satisfies your needs.
@rhvgoyal @nalind @saschagrunert @vrothberg @giuseppe @TomSweeneyRedHat PTAL

docs/containers-storage.conf.5.md
```diff
@@ -138,6 +138,9 @@ The `storage.options.thinpool` table supports the following options for the `dev
     6: LogLevelInfo
     7: LogLevelDebug
 
+**metadata_size**="1k"
+  metadata_size is used to set the `pvcreate --metadatasize` options when creating thin devices.
```
TomSweeneyRedHat, line 142: is 1k the only value accepted?  I was expecting `**metadata_size**=""` here with a description of the accepted values in the text.

TomSweeneyRedHat:
```suggestion
Specifies a custom metadata_size to be used to set the `pvcreate --metadatasize` option when creating thin devices. (default: 1k)
```

rhatdan: Nope any value accepted by pvcreate, but the default is 1k.  Or at least this is what I am told.

rhvgoyal:
@rhatdan  Default of 1K metadata size seems to be too small. container-storage-setup takes .1% of free space in volume group. So for a 100G volume group, metadata size will be 100MB. You might want to use same logic instead of a fixed size default.

rhatdan:
I don't see how we could do this in a conf file, but perhaps we could do this in the code.  If the user does not set the field, then we get the default.

rhatdan:
@rhvgoyal What does pvcreate do if there is no --metadatasize specified?

docs/containers-storage.conf.5.md
```diff
@@ -139,7 +139,7 @@ The `storage.options.thinpool` table supports the following options for the `dev
     7: LogLevelDebug
 
 **metadata_size**="1k"
```
TomSweeneyRedHat, line 141: I know it wasn't yours, but could you remove the `1k` from this line please.

mikemccracken:
> @mikemccracken could you try this out and see if it satisfies your needs.
> @rhvgoyal @nalind @saschagrunert @vrothberg @giuseppe @TomSweeneyRedHat PTAL

Yes, this config would also have worked just fine for me, and is clearly better than just some other hard coded setting.


> @rhatdan Default of 1K metadata size seems to be too small. container-storage-setup takes .1% of free space in volume group. So for a 100G volume group, metadata size will be 100MB. You might want to use same logic instead of a fixed size default.

I agree that 1K is probably too small as a default. For more context:

This is LVM PV metadata, which is different from the thin pool metadata that container-storage-setup creates. (This point might be helpful to clarify in the description). 

PV metadata requirements grow roughly with the number of LVs. Here's the description from lvm.conf:

> Volume groups with large numbers of physical or logical volumes, or volumes groups containing complex logical volume structures will need additional space for their metadata.

So you can't really base the default on the size of the PV, it depends on how many LVs get created in it. We ran into PV metadata trouble with the default value once we had between 200 and 300 LVs. (it's been a while, I don't have the exact numbers handy.) I think it's basically linear for a rough estimate. Looking at a metadata dump, most of the metadata sections seemed about the same size. Our choice of 128M is clearly overkill, but even something like 128K, which is still really small, would support more LVs than seems likely for any real system.

rhvgoyal:
@mikemccracken What's lvm pv metadata. Is this driver not about creating thin pool? If not, can you please give an idea what this driver is doing. I thought we are creating a thin pool and then carving out thin devices for containers. Why do I need to specify metadata size for regular lvm. Doesn't lvm maintain that by itself.

mikemccracken:
@rhvgoyal yes, this driver creates a thin pool, but it assumes you give it a block device, so it first creates a PV and a VG on the device you give it. So the PV it creates has the metadata for any LVs (thin or not) that get created on it, and that's where we want to avoid running out of space.

It's been a while since I looked in depth at this code, but what I recall is that the assumption of wanting a block device was pretty fundamental, which is why I did https://github.com/containers/storage/pull/647 - so you can actually hand this driver an LVM LV, and it will just create a PV on top of that LV (which LVM supports)

TomSweeneyRedHat:
LGTM

rhatdan:
Ok I just set the default to 128k and allow users/distributions to modify it.  Is the a good setting?
@rhvgoyal @mikemccracken PTAL

mikemccracken:
Sounds good to me. +1

PR #658, opened 2020-07-03
new interface for MountImage added

rhatdan:
Interface added to mount image rootfs without any container.
image can be mounted only in ReadOnly mode.

Signed-off-by: Kunal Kushwaha <kunal.kushwaha@gmail.com>

### DIFF
--- a/cmd/containers-storage/mount.go
+++ b/cmd/containers-storage/mount.go
@@ -22,12 +22,21 @@ type mountPointError struct {
 func mount(flags *mflag.FlagSet, action string, m storage.Store, args []string) int {
 	moes := []mountPointOrError{}
 	for _, arg := range args {
-		result, err := m.Mount(arg, paramMountLabel)
-		errText := ""
-		if err != nil {
-			errText = fmt.Sprintf("%v", err)
+		if paramReadOnly {
+			result, err := m.MountImage(arg, []string{"noexec", "nosuid"}, paramMountLabel)
+			errText := ""
+			if err != nil {
+				errText = err.Error()
+			}
+			moes = append(moes, mountPointOrError{arg, result, errText})
+		} else {
+			result, err := m.Mount(arg, paramMountLabel)
+			errText := ""
+			if err != nil {
+				errText = err.Error()
+			}
+			moes = append(moes, mountPointOrError{arg, result, errText})
 		}
-		moes = append(moes, mountPointOrError{arg, result, errText})
 	}
 	if jsonOutput {
 		json.NewEncoder(os.Stdout).Encode(moes)
@@ -54,8 +63,12 @@ func unmount(flags *mflag.FlagSet, action string, m storage.Store, args []string
 		mounted, err := m.Unmount(arg, force)
 		errText := ""
 		if err != nil {
-			errText = fmt.Sprintf("%v", err)
-			errors = true
+			var err1 error
+			mounted, err1 = m.UnmountImage(arg, force)
+			if err1 != nil {
+				errText = err.Error()
+				errors = true
+			}
 		} else {
 			if !mounted {
 				fmt.Printf("%s mountpoint unmounted\n", arg)
@@ -85,7 +98,7 @@ func mounted(flags *mflag.FlagSet, action string, m storage.Store, args []string
 		mounted, err := m.Mounted(arg)
 		errText := ""
 		if err != nil {
-			errText = fmt.Sprintf("%v", err)
+			errText = err.Error()
 			errors = true
 		} else {
 			if mounted > 0 {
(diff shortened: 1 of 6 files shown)


rhatdan:
Replaces https://github.com/containers/storage/pull/438

rhatdan:
Here is the implementation using this PR.
https://github.com/containers/libpod/pull/6851

rhatdan:
@vrothberg @nalind @TomSweeneyRedHat @saschagrunert @giuseppe PTAL

rhatdan:
@kunalkushwaha PTAL

rhatdan:
@mrunalp PTAL

cmd/containers-storage/mount.go
```diff
@@ -22,12 +22,21 @@ type mountPointError struct {
...
 	for _, arg := range args {
-		result, err := m.Mount(arg, paramMountLabel)
-		errText := ""
-		if err != nil {
-			errText = fmt.Sprintf("%v", err)
+		if paramReadOnly {
+			result, err := m.MountImage(arg, []string{"noexec", "nosuid"}, paramMountLabel)
+			errText := ""
+			if err != nil {
+				errText = fmt.Sprintf("%v", err)
```
saschagrunert, line 29:
```suggestion
				errText = err.Error()
```

cmd/containers-storage/mount.go
```diff
@@ -22,12 +22,21 @@ type mountPointError struct {
...
+			}
+			moes = append(moes, mountPointOrError{arg, result, errText})
+		} else {
+			result, err := m.Mount(arg, paramMountLabel)
+			errText := ""
+			if err != nil {
+				errText = fmt.Sprintf("%v", err)
```
saschagrunert, line 36:
```suggestion
				errText = err.Error()
```

utils.go
```diff
@@ -310,3 +310,18 @@ func validRootlessStoragePathFormat(path string) error {
...
+func validateMountOptions(mountOptions []string) error {
+	var Empty struct{}
+	// Add invalid options for ImageMount() here.
+	invalidOptions := map[string]struct{}{
+		"rw": Empty,
+	}
+
+	for _, opt := range mountOptions {
+		if _, ok := invalidOptions[opt]; ok {
+			return fmt.Errorf(" '%s' option not supported", opt)
```
saschagrunert, line 323:
```suggestion
			return fmt.Errorf("%q option not supported", opt)
```

cmd/containers-storage/mount.go
```diff
@@ -119,6 +128,7 @@ func init() {
 		addFlags: func(flags *mflag.FlagSet, cmd *command) {
 			flags.StringVar(&paramMountOptions, []string{"-opt", "o"}, "", "Mount Options")
 			flags.StringVar(&paramMountLabel, []string{"-label", "l"}, "", "Mount Label")
+			flags.BoolVar(&paramReadOnly, []string{"-ro", "r"}, paramReadOnly, "mount image readonly")
```
giuseppe, line 131: mount uppercase

store.go
```diff
@@ -265,6 +265,11 @@ type Store interface {
 	// Wipe removes all known layers, images, and containers.
 	Wipe() error
 
+	// MountImage mounts an image to temp folder and returns the mount point.
```
giuseppe, line 268: s/folder/directory/

cmd/containers-storage/mount.go
rhatdan: Fixed

cmd/containers-storage/mount.go
rhatdan: Fixed

cmd/containers-storage/mount.go
rhatdan: Fixed

store.go
rhatdan: Fixed

vrothberg:
Can you squash the two commits. Looks like an --amend went wrong.

store.go
```diff
@@ -2666,6 +2706,11 @@ func (s *store) Mounted(id string) (int, error) {
 func (s *store) Unmount(id string, force bool) (bool, error) {
 	if layerID, err := s.ContainerLayerID(id); err == nil {
 		id = layerID
+	} else {
+		img, err := s.Image(id)
+		if err == nil {
+			id = img.TopLayer
+		}
```
vrothberg, line 2713: Can we add an `UnmountImage` instead? At the moment, the change would require a major version bump as the behaviour is backward incompatible.

rhatdan: I originally did it that way, but I ended up duplicating a lot of code and made the callers more complex

I don't see this really as a modification of the API previously the caller would have gotten an error if they gave an image id to a umount function, now if they give the imageid, it will work. Having the Containers ID and The ImageID match would be just about impossible.

vrothberg: >  as a modification of the API previously the caller would have gotten an error if they gave an image id to a umount function, now if they give the imageid, it will work. 

That is pretty much the definition of a breaking change: different results for the same input.

The functions could still share code if we introduce a third private one.

rhatdan:
@vrothberg I added an UnmountImage and shared most of the code.

rhatdan:
@saschagrunert @nalind @TomSweeneyRedHat @vrothberg @ashley-cui @mrunalp @mheon PTAL

PR #662, opened 2020-07-09
When mounting images we have no lowers, but still need to mount

rhatdan:
Currenty overlay is mounting in the case where there are no lowers
on a read/only mount iff their is no mount_program. i

This situation might make sense for containers but in the case of images
the image should still be mounted. Having an empty layer makes little sense
if it is read/only.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -892,19 +892,6 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		}
 	}
 
-	// If the lowers list is still empty, use an empty lower so that we can still force an
-	// SELinux context for the mount.
-
-	// if we are doing a readOnly mount, and there is only one lower
-	// We should just return the lower directory, no reason to mount.
-	if !readWrite && d.options.mountProgram == "" {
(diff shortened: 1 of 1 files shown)


rhatdan:

drivers/overlay/overlay.go
```diff
@@ -898,9 +898,6 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	// if we are doing a readOnly mount, and there is only one lower
 	// We should just return the lower directory, no reason to mount.
 	if !readWrite && d.options.mountProgram == "" {
-		if len(absLowers) == 0 {
-			return path.Join(dir, "empty"), nil
-		}
 		if len(absLowers) == 1 {
```
giuseppe, line 901: do we also need to take care of the `len(absLowers) == 1` case?  Could happen that the image has two layers and we would hit this case and there is no mount happening?

rhatdan: Yes I created an image with a single lower layer, and it mounted the wrong thing.  Good catch.  I just removed both special cases and now we will always mount the image.

rhatdan:
@giuseppe @nalind PTAL

PR #663, opened 2020-07-09
Allow mounting of Non Read Write images read/only

rhatdan:
We want to block the mounting of additional stores, for
read/write, since they will not have any containers associated
with them.  But if a user is mounting an image for read/only
access then their is no reason to block the mount.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -772,7 +772,20 @@ func (r *layerStore) Mounted(id string) (int, error) {
 }
 
 func (r *layerStore) Mount(id string, options drivers.MountOpts) (string, error) {
-	if !r.IsReadWrite() {
+
+	// check whether options include ro option
+	hasReadOnlyOpt := func(opts []string) bool {
+		for _, item := range opts {
+			if item == "ro" {
+				return false
+			}
+		}
+		return true
+	}
+
+	// You are not allowed to mount layers from readonly stores if they
+	// are not mounted read/only.
+	if !r.IsReadWrite() && !hasReadOnlyOpt(options.Options) {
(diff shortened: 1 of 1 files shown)


rhatdan:

layers.go
```diff
@@ -772,7 +772,18 @@ func (r *layerStore) Mounted(id string) (int, error) {
...
-	if !r.IsReadWrite() {
+
+	// check whether options include ro option
+	hasReadWriteOpt := func(opts []string) bool {
+		for _, item := range opts {
+			if item == "ro" {
+				return false
+			}
+		}
+		return true
```
vrothberg, line 783: Can we negate the logic and call it `hasReadOnlyOpt`? There's an implicit assumption that "no ro" is "read write" which would be easier to understand the other war around.

layers.go
```diff
@@ -772,7 +772,18 @@ func (r *layerStore) Mounted(id string) (int, error) {
...
+	}
+
+	if !r.IsReadWrite() && hasReadWriteOpt(options.Options) {
```
vrothberg, line 786: Can you add a comment here that explains the same as in the commit message? Otherwise, I will likely forget by next week :)

layers.go
rhatdan: The problem with that is we end up wit an awful double negative line below

I don't like that name, because you end up with a double negative when you use the function below.
I don;t like
```
if !r.IsReadWrite() && !hasReadOnlyOpt(options.Options) {
```

vrothberg: The returns need to be inverted. I opened https://github.com/containers/storage/pull/667

Issue #664, opened 2020-07-10
Bug with residual resources in layerStore after remove an image


zvier:
Bug with residual resources in layerStore after remove an image may lead memory leak and affect other image build export failed with error messages as flows:

> ...Error reading blob sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef: error locating item named "sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef" for image with ID "7f32b3d4368beaf08fc934a1e5fb24df607d31ec0b612d65f5807527363a3af9": file does not exist 

When delete an image, related layers in layerStore only be deleted in layers, idindex, byid, byname and bymount. But bycompressedsum and byuncompressedsum still maintain the relation for these layers utill the layerStore reload.

```go
// github.com/containers/storage/layers.go
type layerStore struct {
    ...
	layers            []*Layer
	idindex           *truncindex.TruncIndex
	byid              map[string]*Layer
	byname            map[string]*Layer
	bymount           map[string]*Layer
	bycompressedsum   map[digest.Digest][]string
	byuncompressedsum map[digest.Digest][]string
    ...
}
```

This may lead memory leak. And more seriously, these residual resources may cause image export failed in some special cases. My Dockerfile is:
```Dockerfile
FROM busybox:latest AS foo
ARG USED_ARG="used_value"
RUN echo ${USED_ARG} > /root/value
RUN echo ${USED_ARG}
COPY data /src/data

FROM foo
ENV http_proxy=xxxproxy
EXPOSE 80
```

The special in this Dockerfile is the second stage not produce any actual data content, so the diffid for this layer will always be a fixed value(In my environment is 5f70bf18a...). Therefore, in layerStore.byuncompressedsum will store the relation with the same key, for example, my layerStore.byuncompressedsum is ["sha256:5f70bf18..."]->["layerID_1"] at this time.

Then, we remove the image and layers about this image will be also removed in layerStore, except bycompressedsum and byuncompressedsum. For example, the layerStore.byuncompressedsum is still ["sha256:5f70bf18..."]->["layerID_1"] at this time.

The problem is that we may build and export an image use above Dockerfile again immediately. After the image build complete, layerStore.byuncompressedsum's value is ["sha256:5f70bf18..."]->["layerID_1", "layerID_2"] because it use append operation directly when newvalue is not empty.
```go
// github.com/containers/storage/layers.go
[...]

PR #665, opened 2020-07-12
clean residual resources in layerStore when remove an image

zvier:
Fix ISSUE #664 

In some special case, residual resources in layerStore when remove an image may cause image export failed. These residual resources  also lead memory leak. So, it is nessary to clean all resources with the image remove.

Signed-off-by: zvier <zvier20@gmail.com>


rhatdan:

vrothberg:
Thanks for the fix!

nalind:
LGTM

PR #666, opened 2020-07-12
userns: fix available range with explicit idmapping

giuseppe:
when an explicit idmapping is specified, the host id must be taken
from the available range of IDs.


rhatdan:
LGTM
@saschagrunert @nalind @vrothberg @TomSweeneyRedHat PTAL

PR #667, opened 2020-07-13
layer mount: fix RO logic

vrothberg:
Fix the logic in an anon-func looking for the `ro` option to allow for
mounting images that are set as read-only.


vrothberg:
@rhatdan PTAL

rhatdan:
LGTM

PR #669, opened 2020-07-14
unshare: memoize HomeDir()

giuseppe:
do the potentially expensive user lookup function just once.


rhatdan:
LGTM

PR #671, opened 2020-07-20
archive: fix the bug of ReadSecurityXattrToTarHeader

yangfeiyu20102011:
archive: fix the bug of ReadSecurityXattrToTarHeader

The xattr of hdr will lose when the security.ima
and security.capability both have value.

Signed-off-by: yangfeiyu <yangfeiyu20102011@163.com>

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -393,13 +393,15 @@ func fillGo18FileTypeBits(mode int64, fi os.FileInfo) int64 {
 // ReadSecurityXattrToTarHeader reads security.capability, security,image
 // xattrs from filesystem to a tar header
 func ReadSecurityXattrToTarHeader(path string, hdr *tar.Header) error {
+	if hdr.Xattrs == nil {
+		hdr.Xattrs = make(map[string]string)
+	}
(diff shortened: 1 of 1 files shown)


yangfeiyu20102011:
@rhatdan PTAL,Thanks!

rhatdan:
LGTM
@nalind @vrothberg @saschagrunert @QiWang19 @giuseppe PTAL

pkg/archive/archive.go
```diff
@@ -399,7 +399,9 @@ func ReadSecurityXattrToTarHeader(path string, hdr *tar.Header) error {
 			return errors.Wrapf(err, "failed to read %q attribute from %q", xattr, path)
 		}
 		if capability != nil {
-			hdr.Xattrs = make(map[string]string)
+			if hdr.Xattrs == nil {
+				hdr.Xattrs = make(map[string]string)
+			}
```
QiWang19, line 404: Can this be moved out of the for loop before L396?

nalind:
LGTM with @QiWang19's suggested change.

pkg/archive/archive.go
yangfeiyu20102011: ok

PR #673, opened 2020-07-27
Fix leaked fd

rhatdan:
Coverity found the leaked fd.


giuseppe:
that code will never be executed, as the execve was successful.  But I am fine with merging it if it makes static analysis happy

PR #675, opened 2020-08-03
userns: make sure host id is not always 0

giuseppe:
when it finds multiple available ranges, make sure the host id is
correctly initialized.

### DIFF
--- a/userns.go
+++ b/userns.go
@@ -356,6 +356,7 @@ func findAvailableIDRange(size uint32, availableIDs, usedIDs []idtools.IDMap) ([
 			return avail[:i+1], nil
 		}
 		remaining -= uint32(avail[i].Size)
+		currentID += avail[i].Size
 	}
 
 	return nil, errors.New("could not find enough available IDs")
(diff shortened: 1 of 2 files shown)


giuseppe:
let's keep it Draft for now as I am still testing cri-o with these patches

giuseppe:
seems to work fine so far

rhatdan:
@nalind @saschagrunert @vrothberg @TomSweeneyRedHat PTAL

userns.go
```diff
@@ -356,6 +356,7 @@ func findAvailableIDRange(size uint32, availableIDs, usedIDs []idtools.IDMap) ([
 			return avail[:i+1], nil
 		}
 		remaining -= uint32(avail[i].Size)
+		currentID += avail[i].Size
```
TomSweeneyRedHat, line 359: I don't think so, but will ask.  Any chance the currentID could get too big here?

TomSweeneyRedHat:
Needs rebase and one question

userns.go
giuseppe: the sum of all avail[].Size cannot be bigger than 4294967295 and we are counting from 0.

In the worse case, the kernel will reject the user namespace configuration

giuseppe:
rebased

TomSweeneyRedHat:
LGTM 
and happy green test buttons.  Would be good to get a head nod from @nalind or @vrothberg

PR #676, opened 2020-08-04
store: support mapped layers deletion

giuseppe:
currently it is not possible to delete a mapped layer as it is always
referenced by the image.

Change it allow deleting the mapping if it is not used by any
container or other layers.

It will be useful in CRI-O as a lot of thrown away containers will be
created and the mapped images would keep growing without a way of
cleaning them up.

### DIFF
--- a/pkg/stringutils/stringutils.go
+++ b/pkg/stringutils/stringutils.go
@@ -56,11 +56,22 @@ func Truncate(s string, maxlen int) string {
 // Comparison is case insensitive
 func InSlice(slice []string, s string) bool {
 	for _, ss := range slice {
-		if strings.ToLower(s) == strings.ToLower(ss) {
+		if strings.EqualFold(s, ss) {
 			return true
 		}
 	}
 	return false
+}
+
+// RemoveFromSlice removes a string from a slice.  The string can be present
+// multiple times.  The entire slice is iterated.
+func RemoveFromSlice(slice []string, s string) (ret []string) {
+	for _, ss := range slice {
+		if !strings.EqualFold(s, ss) {
+			ret = append(ret, ss)
+		}
+	}
+	return ret
 }
 
 func quote(word string, buf *bytes.Buffer) {
--- a/store.go
+++ b/store.go
@@ -2223,16 +2223,23 @@ func (s *store) DeleteLayer(id string) error {
 		}
 		for _, layer := range layers {
 			if layer.Parent == id {
-				return ErrLayerHasChildren
+				return errors.Wrapf(ErrLayerHasChildren, "used by layer %v", layer.ID)
 			}
 		}
 		images, err := ristore.Images()
 		if err != nil {
 			return err
 		}
+
 		for _, image := range images {
-			if image.TopLayer == id || stringutils.InSlice(image.MappedTopLayers, id) {
-				return errors.Wrapf(ErrLayerUsedByImage, "Layer %v used by image %v", id, image.ID)
+			if image.TopLayer == id {
+				return errors.Wrapf(ErrLayerUsedByImage, "layer %v used by image %v", id, image.ID)
+			}
+			if stringutils.InSlice(image.MappedTopLayers, id) {
+				// No write access to the image store, fail before the layer is deleted
+				if _, ok := ristore.(*imageStore); !ok {
+					return errors.Wrapf(ErrLayerUsedByImage, "layer %v used by image %v", id, image.ID)
+				}
 			}
 		}
 		containers, err := rcstore.Containers()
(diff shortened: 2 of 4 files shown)


giuseppe:
@nalind @rhatdan @mrunalp PTAL

rhatdan:
@saschagrunert @vrothberg PTAL

pkg/stringutils/stringutils.go
```diff
@@ -55,14 +55,28 @@ func Truncate(s string, maxlen int) string {
 // InSlice tests whether a string is contained in a slice of strings or not.
 // Comparison is case insensitive
 func InSlice(slice []string, s string) bool {
+	s = strings.ToLower(s)
 	for _, ss := range slice {
-		if strings.ToLower(s) == strings.ToLower(ss) {
+		if s == strings.ToLower(ss) {
```
saschagrunert, line 60:
```suggestion
		if strings.EqualFold(s, ss) {
```

pkg/stringutils/stringutils.go
```diff
@@ -55,14 +55,28 @@ func Truncate(s string, maxlen int) string {
 // InSlice tests whether a string is contained in a slice of strings or not.
 // Comparison is case insensitive
 func InSlice(slice []string, s string) bool {
+	s = strings.ToLower(s)
```
saschagrunert, line 58:
```suggestion
```

pkg/stringutils/stringutils.go
```diff
@@ -55,14 +55,28 @@ func Truncate(s string, maxlen int) string {
...
 			return true
 		}
 	}
 	return false
 }
 
+// RemoveFromSlice removes a string from a slice.  The string can be present
+// multiple times.  The entire slice is iterated.
+func RemoveFromSlice(slice []string, s string) []string {
+	var ret []string
```
saschagrunert, line 70:
```suggestion
func RemoveFromSlice(slice []string, s string) (ret []string) {
```

pkg/stringutils/stringutils.go
```diff
@@ -55,14 +55,28 @@ func Truncate(s string, maxlen int) string {
...
+	s = strings.ToLower(s)
+	for _, ss := range slice {
+		if s != strings.ToLower(ss) {
```
saschagrunert, line 73:
```suggestion
		if strings.EqualFold(s, ss) {
```

pkg/stringutils/stringutils.go
```diff
@@ -55,14 +55,28 @@ func Truncate(s string, maxlen int) string {
...
 		}
 	}
 	return false
 }
 
+// RemoveFromSlice removes a string from a slice.  The string can be present
+// multiple times.  The entire slice is iterated.
+func RemoveFromSlice(slice []string, s string) []string {
+	var ret []string
+	s = strings.ToLower(s)
```
saschagrunert, line 71:
```suggestion
```

store.go
```diff
@@ -2241,10 +2248,23 @@ func (s *store) DeleteLayer(id string) error {
...
-				return errors.Wrapf(ErrLayerUsedByContainer, "Layer %v used by container %v", id, container.ID)
+				return errors.Wrapf(ErrLayerUsedByContainer, "layer %v used by container %v", id, container.ID)
 			}
 		}
-		return rlstore.Delete(id)
+		if err := rlstore.Delete(id); err != nil {
+			return errors.Wrapf(err, "delete layer %v", id)
+		}
+
+		if istore, ok := ristore.(*imageStore); ok {
```
saschagrunert, line 2258: Do we wanna error if `!ok`?

tests/delete-layer.bats
```diff
@@ -67,3 +67,43 @@ load helpers
...
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	lowerlayer2="$output"
+
+	run storage --debug=false create-layer -r --uidmap 0:200:100000 --gidmap 0:200:100000  $lowerlayer2
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	upperlayer="$output"
+
+        # Expect an error as both lower layers are referenced
```
saschagrunert, line 87:
```suggestion
# Expect an error as both lower layers are referenced
```
other places are affected as well.

store.go
giuseppe: it can be `!ok` and that is fine if we don't have to modify `image.MappedTopLayers`.  The check here is just to avoid iterating the images if we don't have too. We already have a check above.  I'll add a comment to make it clearer

giuseppe:
@saschagrunert thanks for the hints.  I've pushed an updated version

pkg/stringutils/stringutils.go
```diff
@@ -56,13 +56,24 @@ func Truncate(s string, maxlen int) string {
...
 		}
 	}
 	return false
 }
 
+// RemoveFromSlice removes a string from a slice.  The string can be present
+// multiple times.  The entire slice is iterated.
+func RemoveFromSlice(slice []string, s string) (ret []string) {
+	for _, ss := range slice {
+		if strings.EqualFold(s, ss) {
```
vrothberg, line 70: We need to negate the condition

store.go
saschagrunert: Thanks!

pkg/stringutils/stringutils.go
giuseppe: thanks!  That made a lot of difference..

PR #681, opened 2020-08-06
[WIP] Allow callers to pass down OmitTimestamp flag

rhatdan:
If the user passes down this flag all of the files in the archive
should be treated as being created at EPOCH.

The goal is to allow a container engine to create the same content
multiple times without having to deal with changes in timestamps.

### DIFF
--- a/drivers/aufs/aufs.go
+++ b/drivers/aufs/aufs.go
@@ -493,9 +493,9 @@ func (a *Driver) isParent(id, parent string) bool {
 
 // Diff produces an archive of the changes between the specified
 // layer and its parent layer which may be "".
-func (a *Driver) Diff(id string, idMappings *idtools.IDMappings, parent string, parentMappings *idtools.IDMappings, mountLabel string) (io.ReadCloser, error) {
+func (a *Driver) Diff(id string, idMappings *idtools.IDMappings, parent string, parentMappings *idtools.IDMappings, mountLabel string, omitTimestamp bool) (io.ReadCloser, error) {
 	if !a.isParent(id, parent) {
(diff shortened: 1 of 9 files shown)


rhatdan:

The basic idea of this patch is to allow callers to tell the storage driver to save all files as if they were creted at EPOCH.

nalind:
If this is only needed for the benefit of building images, then running the diff stream through a filter like a `buildah.tarFilterer` should work just as well, without requiring changes to this library's APIs.

rhatdan:
@nalind That looks good, where exactly do I plug this in?  IE When are the tar balls created that would use this?

nalind:
The layer diffs are generated in `containerImageRef.NewImageSource()`, and since it uses the digests of those diffs to generate config blobs, and the digests of possibly-compressed versions of those to generate manifests, that's probably the best place.

The `srcHasher` needs to compute the diffID using the modified contents, so instead of copying from an `io.TeeReader` that reads from `rc` and writes to `srcHasher.Hash()`, we need to switch things up so that the hasher gets data after it's been modified by the filter.

Looking at it, my first attempt would be to wrap the `io.WriteCloser` that we get from `archive.CompressStream()` in an `io.MultiWriter` that also writes to `srcHasher` to make sure it gets the after-modification version of the diff, and then wrap _that_ `io.Writer` using `github.com/containers/storage/pkg/ioutils.NopWriteCloser()` to get an `io.WriteCloser` that we can then wrap using `newTarFilterer` to get another `io.WriteCloser`.  We'd then copy from `rc` to the newest `io.WriteCloser` and close it before closing the compressor's `io.WriteCloser`.

rhatdan:
Seems like it would be best if you took this over.

nalind:
Sure.  Closing this one.

PR #682, opened 2020-08-06
archive: preallocate a buffer for io.Copy

giuseppe:
avoid creating a new buffer for each file in the archive but
preallocate a 1M buffer.  Another advantage of having a bigger buffer
is that we reduce the number of read and write syscalls.


TomSweeneyRedHat:
LGTM

PR #683, opened 2020-08-10
Allow any env variable for graphroot, runroot, storagepath

ashley-cui:
Previously only restricted to home, uid, and user

as per https://github.com/containers/buildah/issues/2311#issuecomment-671076880

Signed-off-by: Ashley Cui <acui@redhat.com>

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -33,33 +33,21 @@ The `storage` table supports the following options:
 **graphroot**=""
   container storage graph dir (default: "/var/lib/containers/storage")
   Default directory to store all writable content created by container storage programs.
-
-    The rootless graphroot path supports three substitutions:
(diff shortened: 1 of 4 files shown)


rhatdan:
Your removing tests, but you added none back int?  Can you add some test in this?

ashley-cui:
@rhatdan the test i removed was just to check if the path only included env vars $HOME, $UID, $USER. since we want to allow other env vars to be set, i removed the test. i'm not quite sure how to write a test for the function i changed, since it automatically looks at .config/containers/storage.conf to parse storage.conf, other than using the same trick getRootlessRuntimeDirIsolated() uses. if this is okay, i'll do the trick, if not, let me know how to test it

rhatdan:
Yes do the trick.

rhatdan:

PR #689, opened 2020-08-13
Allow users to override imagestores

rhatdan:
We have a request from a user who wants to use imagestores and
to be able to update them from podman.

Currently if I setup /var/lib/shared as an additional store
and I attempt update the share with a command like:

Error: lock "/var/lib/shared/overlay-images/images.lock" is not a read-only lock

I can not because Podman is still using the additonal store.

User expected something like:
Error: overlay: Unknown option additionalimagestore

To work.

I looked further and figured overlay.imagestore would work, but it still failed.

This PR allows users to specify additionalimagestores on the command line, and to use
more likely options to make it work.

I also cleanup some of the option parsing in overlay.go to handle overlay, overlay2, and "."
prefixes.

I checked and none of the other drivers use additionalimagestores options.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -274,22 +274,28 @@ func parseOptions(options []string) (*overlayOptions, error) {
 		if err != nil {
 			return nil, err
 		}
-		key = strings.ToLower(key)
-		switch key {
-		case ".override_kernel_check", "overlay.override_kernel_check", "overlay2.override_kernel_check":
+		trimkey := strings.ToLower(key)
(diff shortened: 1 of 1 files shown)


rhatdan:
Partially fixes: https://github.com/containers/podman/issues/7309

rhatdan:
@nalind @giuseppe @vrothberg @TomSweeneyRedHat @mheon @ashley-cui @QiWang19 PTAL

QiWang19:
PR needs rebase

rhatdan:
@mheon @nalind @TomSweeneyRedHat @QiWang19 @saschagrunert PTAL

Issue #697, opened 2020-08-20
regression in v1.23.1


vrothberg:
v1.23.1 is causing regressions in Podman's `apiv2` tests. I do not have the time to track it down but need to get work done that's currently blocked by the regression so I first revert the bump in c/image (see https://github.com/containers/image/pull/1031).

```
[+0064s] not ok 291 [40-pods] POST libpod/pods/bar/start [] : status
[+0064s] #  expected: 200
[+0064s] #    actual: 500
[+0064s]   expected: 200
```

vrothberg:
It looks like commit 7005d07fa80a causes it.

vrothberg:
```
$ POST libpod/pods/bar/start []                                          
HTTP/1.1 500 Internal Server Error                                       
Content-Type: application/json                                           
Date: Thu, 20 Aug 2020 12:11:03 GMT                                      
Content-Length: 117                                                      
                                                                         
{                                                                        
  "cause": "some containers failed",                                     
  "message": "error starting some containers: some containers failed",   
  "response": 500                                                        
}                                                                        
not ok 277 [40-pods] POST libpod/pods/bar/start [] : status              
```

vrothberg:
What I am seeing on the server side:
```
time="2020-08-20T14:27:33+02:00" level=debug msg="Received: -1"
time="2020-08-20T14:27:33+02:00" level=debug msg="Cleaning up container c6c60358e96e7ebf8ce7953644fef5a2bbb10ddeac9fd9c1b019162f59b51bc5"
``

vrothberg:
@giuseppe @nalind @rhatdan ... no idea what's going on here. The only thing I noticed is that the error only occurs when we're locking the graphlock _before_ the layer store. If we lock after, CI passes again.

rhatdan:
@zvier PTAL

vrothberg:
Note that it might very well be a bug in Podman. At least we cannot rule that out.

nalind:
> @giuseppe @nalind @rhatdan ... no idea what's going on here. The only thing I noticed is that the error only occurs when we're locking the graphlock _before_ the layer store. If we lock after, CI passes again.

That differs from the order in which they're acquired in `Shutdown()`, which would be a potential lock ordering problem.

giuseppe:
I don't think there is a regression in v1.23.1, IMO it only showed a race condition that already existed.

[...]

PR #698, opened 2020-08-21
counter: check for external umounts

giuseppe:
if a previously mounted container was unmounted
externally (e.g. through conmon cleanup for Podman containers), the
ref counter will lose track of it and report it as still mounted.

Closes: https://github.com/containers/storage/issues/697


rhatdan:
LGTM
But I want @nalind to chime in.

mheon:
This fixes a critical bug we found in the Podman REST API tests. Can we prioritize getting this in?

nalind:
This isn't going to fix everything, but it's a step in the right direction.  LGTM.

PR #700, opened 2020-08-24
Cherry-pick #698 into v1.20-stable

mheon:
As the title says - we need this in the stable branch used by Podman v2.0


rhatdan:
LGTM

PR #702, opened 2020-08-26
Start using github.com/moby/sys/mountinfo rather then local version

rhatdan:
This version of Mountinfo has been optimized for container use cases.


rhatdan:
After watching @kolyshkin Talk, I decided to look at what we are using for mountinfo.

@kolyshkin PTAL

rhatdan:
I am attempting to not break API.

kolyshkin:
The key for a seamless switch is to use type X=mountinfo.X and the same with functions.

For example, see https://github.com/moby/moby/pull/40694

TomSweeneyRedHat:
All kinds of test ugliness @rhatdan

kolyshkin:
Mocked up an alternative PR: https://github.com/containers/storage/pull/703

PR #709, opened 2020-09-03
We need to use global GraphDriverOptions for rootless environments

rhatdan:

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -198,7 +198,7 @@ func getRootlessDirInfo(rootlessUID int) (string, string, error) {
 }
 
 // getRootlessStorageOpts returns the storage opts for containers running as non root
-func getRootlessStorageOpts(rootlessUID int) (StoreOptions, error) {
+func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOptions, error) {
 	var opts StoreOptions
 
 	dataDir, rootlessRuntime, err := getRootlessDirInfo(rootlessUID)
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
LGTM

rhatdan:
Currently in rootless podman we are not using the global envionment.  We need to pass at least ignore_chown_errors down for rootless podman.

We need to make sure that non VFS options are ignored if overlay is default.

This would fix https://github.com/containers/podman/issues/7513 but I think it is risky, we want this in the long run, but for the podman-2.0.* release we should just handle the ignore_chown_errors.

utils.go
```diff
@@ -210,7 +209,7 @@ func getRootlessStorageOpts(rootlessUID int) (StoreOptions, error) {
 	opts.RootlessStoragePath = opts.GraphRoot
 	if path, err := exec.LookPath("fuse-overlayfs"); err == nil {
 		opts.GraphDriverName = "overlay"
-		opts.GraphDriverOptions = []string{fmt.Sprintf("overlay.mount_program=%s", path)}
+		opts.GraphDriverOptions = append(opts.GraphDriverOptions, fmt.Sprintf("overlay.mount_program=%s", path))
```
nalind, line 212: Does this break when the system `opts.GraphDriverName` is not "overlay"?

rhatdan: I am not sure.  I think this could cause a few issues, which is why I stated that we need to be careful, even if this is the correct thing to do.  I guess if we do the else option, we should just clear these fields.

rhatdan:
@giuseppe PTAL

utils.go
giuseppe: Could we copy these options only when `opts.GraphDriver` is also set to `overlay`?

rhatdan: Fixed

giuseppe (approved):
LGTM

rhatdan:
I am closing, since I now believe we want to move some of these flags to containtainers/common.

PR #710, opened 2020-09-04
For podman v2.0 we need to use use ignore_chown_errors field if set

rhatdan:
Since this is late in the game, it is too risky to add all system options.
We know that this is supposed to work for HPC customers, so we can just
grab this setting if set in system wide settings.

We can put the full fix into the master branch, and continue to test it
going forward.


rhatdan:
@mheon @baude @giuseppe @nalind @TomSweeneyRedHat @vrothberg PTAL

giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #713, opened 2020-09-08
For podman v2.0 we need to use use ignore_chown_errors field if set

rhatdan:
It is too risky to add all system options.
This is supposed to work for HPC customers, so we can just
grab this setting if set in system wide settings.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -198,7 +198,7 @@ func getRootlessDirInfo(rootlessUID int) (string, string, error) {
 }
 
 // getRootlessStorageOpts returns the storage opts for containers running as non root
-func getRootlessStorageOpts(rootlessUID int) (StoreOptions, error) {
+func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOptions, error) {
 	var opts StoreOptions
 
 	dataDir, rootlessRuntime, err := getRootlessDirInfo(rootlessUID)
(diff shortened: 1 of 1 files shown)


rhatdan:
I just want to get this in for now, while we continue to discuss the full PR.

This way we can release newer versions based off of this release.

My fear of the full release is additional stores can break current systems

utils.go
```diff
@@ -211,6 +211,11 @@ func getRootlessStorageOpts(rootlessUID int) (StoreOptions, error) {
 	if path, err := exec.LookPath("fuse-overlayfs"); err == nil {
 		opts.GraphDriverName = "overlay"
 		opts.GraphDriverOptions = []string{fmt.Sprintf("overlay.mount_program=%s", path)}
+		for _, o := range systemOpts.GraphDriverOptions {
+			if strings.Contains(o, "ignore_chown_errors") {
+				opts.GraphDriverOptions = append(opts.GraphDriverOptions, o)
+			}
+		}
```
TomSweeneyRedHat, line 218: Can there be multiple ignore_chow_errors that we'd capture?  if not, should we break the loop?

TomSweeneyRedHat:
One question
otherwise LGTM

utils.go
rhatdan: It is breaking now.

rhatdan:
@nalind @QiWang19 @ashley-cui @mheon @vrothberg PTAL

QiWang19:
LGTM

utils.go
```diff
@@ -207,10 +207,20 @@ func getRootlessStorageOpts(rootlessUID int) (StoreOptions, error) {
...
-	opts.RootlessStoragePath = opts.GraphRoot
+	if systemOpts.RootlessStoragePath != "" {
+		opts.RootlessStoragePath = systemOpts.RootlessStoragePath
+	} else {
+		opts.RootlessStoragePath = opts.GraphRoot
+	}
 	if path, err := exec.LookPath("fuse-overlayfs"); err == nil {
 		opts.GraphDriverName = "overlay"
 		opts.GraphDriverOptions = []string{fmt.Sprintf("overlay.mount_program=%s", path)}
+		for _, o := range systemOpts.GraphDriverOptions {
```
vrothberg, line 218: I think this deserves a comment, maybe with a `// TODO: this should go into containers.conf`

vrothberg (commented):
LGTM
If you rebase, you may drop a code comment.

PR #718, opened 2020-09-15
Bump mountinfo 0.2.0

kolyshkin:
This should considerably speed up `Mounted()` call (for non-bind mounts and for all mounts on recent kernels).

Also, a wrapper for Mounts() is not removed, as its functionality is now implemented in upstream.

**Note the (potentially) breaking change:**
Previously, calling `Mounted()` with a non-existing path as an argument resulted in an error (wrapped `ENOENT`), while now it will return "not mounted". The change is intentional -- the call is not to check if the path exists, it is to check whether it is mounted.


rhatdan (approved):
LGTM

go.mod
```diff
@@ -1,4 +1,4 @@
-go 1.15
+go 1.14
```
saschagrunert, line 1: What is the reason for this downgrade?

rhatdan: It is explained in the PR.

rhatdan:
@kolyshkin Could you rebase and we can merge.  This Vendor was done by dependabot.

PR #723, opened 2020-09-24
go.mod: fix go version

rhatdan:
Commit 1a2847751 (inadvertently?) changed go version in go.mod from 1.13
to 1.15.

A "go <version>" line in go.mod sets minimum go version required for the
package (see [1]). I don't think containers/storage suddenly requires go
1.15+.

Set it to 1.14, which is currently the oldest supported release.

[1] https://golang.org/ref/mod#go-mod-file-go

Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

PR #724, opened 2020-09-24
Fix go.mod; simplify pkg/mount.Mounted

kolyshkin:
Carry of #720 
Closes: #723 

## 1. go.mod: fix go version
    
Commit 1a2847751 (inadvertently?) changed go version in go.mod from 1.13
to 1.15.
    
A "go <version>" line in go.mod sets minimum go version required for the
package (see [1]). I don't think containers/storage suddenly requires go
1.15+.
    
Set it to 1.14, which is currently the oldest supported release.
   
[1] https://golang.org/ref/mod#go-mod-file-go

## 2. pkg/mount: rm path normalization from Mounted
    
Since moby/sys/mountinfo v0.2.0, the path does not have to be
normalized, so drop our wrapper.
    
Note this slightly change the call semantics: previously, a non-existing
path resulting in an error (wrapped ENOENT), while now it will return
"not mounted". The change is intentional -- the call is not to check if
the path exists, it is to check whether it is mounted.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-go 1.15
+go 1.14
 
 module github.com/containers/storage
 
--- a/pkg/mount/mountinfo.go
+++ b/pkg/mount/mountinfo.go
@@ -1,21 +1,13 @@
 package mount
 
 import (
-	"github.com/containers/storage/pkg/fileutils"
 	"github.com/moby/sys/mountinfo"
 )
 
 type Info = mountinfo.Info
 
+var Mounted = mountinfo.Mounted
+
 func GetMounts() ([]*Info, error) {
 	return mountinfo.GetMounts(nil)
-}
-
-// Mounted determines if a specified mountpoint has been mounted.
-func Mounted(mountpoint string) (bool, error) {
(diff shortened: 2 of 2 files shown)


rhatdan (approved):
LGTM

rhatdan:
@nalind @vrothberg @TomSweeneyRedHat @giuseppe @saschagrunert PTAL

vrothberg:
While I am okay to revert the `go version` in the module I think we need to agree that it actually means. To my understanding, it does not set a lower bound of which versions of go can **compile** the code. However, it allows for using **new features** of go tooling. For instance, go >= 1.14 will default to `-mod=vendor` if a `./vendor` dir is present. Such features seem useful to me.

vrothberg (commented):
Conditional LGTM :^)

rhatdan:
LGTM

kolyshkin:
> To my understanding, it does not set a lower bound of which versions of go can compile the code.

AFAIK this is about both tooling features as well as the code (I barely remember seeing some code-related error, alas I don't remember what it was).

PR #727, opened 2020-10-02
vendor: bump mountinfo to v0.3.1

kolyshkin:
Unfortunately it contains some breaking changes (in names),
so fix the code accordingly.


kolyshkin:
The PR from `dependabot` should be closed in favor of this one (because of breaking changes).

kolyshkin:
CI failures seem unrelated

rhatdan:
@cevich PTAL seems to be something wrong with Ubuntu images in cirrus?

cevich:
looking...whoa...WTH...I've never seen this before.  It's not the images, it's a package that changed and is failing to [upgrade during setup](https://github.com/containers/storage/blob/master/contrib/cirrus/setup.sh#L31-L32) for some reason.  These kinds of unpredictable failures are exactly why I don't like doing packaging stuffs at testing-time :cry: I think we decided to do it here since it's a low-traffic repo...and so here we are.  Sheesh, I'm going to need to dig my heals in and get dirty on this one...and let me guess...all other PRs are having the same problem :sweat: 

I'll put this at the top of my list to poke at on Monday.

cevich:
Note: if it is a wide-spread packaging problem, waiting and retrying the tests later is also a valid strategy.  It's possible that an upstream fix is needed, I dunno yet for sure.

cevich:
Let me try quick/easy fix of updating images.  There are a bunch of shell functions that need to be migrated anyway, to use their new shared versions.  Let's see if the errors reproduce in https://github.com/containers/storage/pull/732

rhatdan (approved):
LGTM

PR #730, opened 2020-10-02
layers: on delete clear also the compressed/uncompressed cache reference

fgiudici:
The storage cache is not completely cleaned when removing images.
Remove also the entries referencing layers by compressed and uncompressed sums.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -832,6 +832,27 @@ func (r *layerStore) tspath(id string) string {
 	return filepath.Join(r.layerdir, id+tarSplitSuffix)
 }
 
+func (r *layerStore) deleteInDigestMap(id string) {
+	for digest, layers := range r.bycompressedsum {
+		for i, layerID := range layers {
+			if layerID == id {
+				layers = append(layers[:i], layers[i+1:]...)
+				r.bycompressedsum[digest] = layers
+				break
+			}
+		}
+	}
+	for digest, layers := range r.byuncompressedsum {
+		for i, layerID := range layers {
+			if layerID == id {
+				layers = append(layers[:i], layers[i+1:]...)
+				r.byuncompressedsum[digest] = layers
+				break
+			}
+		}
+	}
+}
+
 func (r *layerStore) Delete(id string) error {
 	if !r.IsReadWrite() {
 		return errors.Wrapf(ErrStoreIsReadOnly, "not allowed to delete layers at %q", r.layerspath())
@@ -857,6 +878,7 @@ func (r *layerStore) Delete(id string) error {
 		if layer.MountPoint != "" {
 			delete(r.bymount, layer.MountPoint)
 		}
+		r.deleteInDigestMap(id)
 		toDeleteIndex := -1
 		for i, candidate := range r.layers {
 			if candidate.ID == id {


mtrmac:
FWIW the main branch fix is a bit more involved: https://github.com/containers/storage/pull/665 .

fgiudici:
> FWIW the main branch fix is a bit more involved: #665 .

It removes the specific layer id from the layer ids array of the r.bycompressedsum[digest] (and the same for the uncompressed map).
When the last layer is removed, the entry is left in the map with an empty array (but it doesn't harm).
I'm not able to figure out a case in which we can have multiple layer ids referenced by the same compressed digest here... anyone can help me to better understand? :-) 

For the record, also backporting #665 would fix the issue mentioned in https://github.com/containers/image/pull/1051.

mtrmac:
> I'm not able to figure out a case in which we can have multiple layer ids referenced by the same compressed digest here... anyone can help me to better understand? :-)

One case, probably reachable through c/image/storage, would be using the same layer on top of different parent layers.

The caller also might not be c/image/storage; e.g. the caller can not supply an ID, causing a random ID to be used, or perhaps the caller might _intentionally_ create a different layer from the same `CompressedDigest` (e.g. a layer with different UID/GID maps). But I’m getting very much out of my depth WRT possible/valid/expected uses of c/storage, I’ll defer to @nalind.

nalind:
> > I'm not able to figure out a case in which we can have multiple layer ids referenced by the same compressed digest here... anyone can help me to better understand? :-)
> 
> One case, probably reachable through c/image/storage, would be using the same layer on top of different parent layers.
> 
> The caller also might not be c/image/storage; e.g. the caller can not supply an ID, causing a random ID to be used, or perhaps the caller might _intentionally_ create a different layer from the same `CompressedDigest` (e.g. a layer with different UID/GID maps). But I’m getting very much out of my depth WRT possible/valid/expected uses of c/storage, I’ll defer to @nalind.

That's right - I find it useful to compare it to source control systems like git, where the same patch can be applied to different branches after they've diverged.

fgiudici:
Thanks for explaining!
So, makes much more sense to backport #665 here.

rhatdan:
I think this PR needs to be rebased.

fgiudici:
> I think this PR needs to be rebased.
Rebased, thanks!

rhatdan:
LGTM

rhatdan:
@mtrmac @nalind @vrothberg @saschagrunert PTAL

layers.go
```diff
@@ -857,6 +878,7 @@ func (r *layerStore) Delete(id string) error {
 		if layer.MountPoint != "" {
 			delete(r.bymount, layer.MountPoint)
 		}
+		r.deleteInDigestMap(id)
```
TomSweeneyRedHat, line 881: Probably outside of the scope of this PR, but is there any advantage to backport the deleteInternal() function from master where this particular function is called from there?

fgiudici: > Probably outside of the scope of this PR, but is there any advantage to backport the deleteInternal() function from master where this particular function is called from there?

Good point. Tried to look into it...
AFAICS the `deleteInternal() `just contains the core deletion operations once in `Delete()`.
The main difference  I can see in commit [bcedb54d05](https://github.com/containers/storage/commit/bcedb54d0547a9a5c0c641518a86a72efc2b19dd) on master (which introduced the function) is that now the `Load()` function calls `deleteInternal()`  in place of the full `Delete()` function.
This skips from the `Load()` the operations left in `Delete()`, which are:
* the checks if the layers are mounted
* the call to the `Save()` function

As [bcedb54d05](https://github.com/containers/storage/commit/bcedb54d0547a9a5c0c641518a86a72efc2b19dd) commit message says, this was introduced to skip the mountpoints locking in the `Save()` function, as locking already happens in `Load()`.
Looking at the code in the 1.11 branch, the locking management seems quite different. In particular the `Save()` function doesn't perform any lock.
So, my guess is that we wouldn't benefit from the `deleteInternal()` function here (moreover, backporting wouldn't be trivial).

TomSweeneyRedHat: Thanks for taking a look @fgiudici.  I was afraid that once you undid one string, too much would unravel with that.

nalind:
LGTM, merging.

PR #733, opened 2020-10-05
Fix usage of rootless_storage_path from system storage.conf file

rcasta74:
Ref: https://github.com/containers/podman/issues/7876
It makes rootless_storage_path working from /etc/containers/storage.conf.
Further it makes the same option useless from ~/.config/containers/storage.conf, so it makes it reserved only for admin use: only graphroot is meaningfull from user configuration.
This makes the option consistent with the definition in man page:
```
  rootless_storage_path="$HOME/.local/share/containers/storage"
     Storage path for  rootless  users.  By  default  the  graphroot  for  rootless  users  is  set  to
  $XDG_DATA_HOME/containers/storage,  if  XDG_DATA_HOME is set.  Otherwise $HOME/.local/share/contain‐
  ers/storage is used.  This field can be used if administrators need to change the  storage  location
   for all users.
```


rhatdan:
@giuseppe PTAL

rhatdan (approved):
I tried this out with Podman and it worked great.
LGTM
I am still seeing a file created in 
ls ~/.local/share/containers/cache/blob-info-cache-v1.boltdb 
/home/dwalsh/.local/share/containers/cache/blob-info-cache-v1.boltdb

But this is not caused by this PR.

utils.go
```diff
@@ -206,11 +206,10 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
 		return opts, err
 	}
 	opts.RunRoot = rootlessRuntime
-	opts.GraphRoot = filepath.Join(dataDir, "containers", "storage")
 	if systemOpts.RootlessStoragePath != "" {
```
giuseppe, line 209: only one question.  Should we honor the setting only when running as rootless?

rcasta74: I'm sorry, but I don't understand your concern; the above piece of code is not removed, but simply moved to the `else` statement below.

rhatdan:
https://github.com/containers/podman/pull/7971 fixes the blobcache problem I mentioned above.

utils.go
giuseppe: what I meant is that `systemOpts.RootlessStoragePath` is taken into account either if we are running as root or as rootless user, while I think it should be ignored when running as root.

Should we have instead something like `if isRootless && systemOpts.RootlessStoragePath != "" {` ?

rcasta74:
I have just realized that the test should be changed as well.
If I'm not wrong, the file `storage_test.conf` should be used to test a rootless configuration; if so the option `rootless_storage_path` with this PR became useless.
To make a proper test I think that the value of rootless_storage_path may be changed to a wrong path; the check should assert that it is not used to change the value of storageOpts.GraphRoot, while the statement 
`assert.Equal(t, storageOpts.RootlessStoragePath, expectedPath)`
can be removed.
Let me know whether to add a commit with the above suggestion.

rcasta74:
As a further change, in the function `defaultStoreOptionsIsolated` the value of `storageOpts.RootlessStoragePath` should be unset (or forced to be the same of `storageOpts.GraphRoot`) since it can be defined in a rootless configuration even if it shouln't; this prevents it to be used in wrong way elsewhere.
Let me know if it is preferable to disable it or force it to be equal to `storageOpts.GraphRoot`, even if it has not been set.
In the latter case, the statement
`
assert.Equal(t, storageOpts.RootlessStoragePath, expectedPath)
`
in the test can be kept as well, while still giving a wrong path to the property `rootless_storage_path` in the `storage_test.conf` file.

rhatdan:
LGTM
@giuseppe PTAL

giuseppe (approved):

giuseppe:
LGTM

rhatdan:
@rcasta74 @giuseppe This is now broken in Podman.  Basically the tests are failing, because it looks like storage.conf is now ignoring the value in $HOME/.config/containers/storage.conf and only paying attention to /etc/containers/storage.conf for the rootless_storage_path?

https://github.com/containers/podman/pull/8071

rcasta74:
I think I may have misunderstood the man page of storage.conf.
```
       graphroot=""
         The rootless graphroot path supports environment variable substitutions (ie. $HOME/containers/storage)

       rootless_storage_path="$HOME/.local/share/containers/storage"
         Storage path for rootless users. By default the graphroot for rootless users
         is set to $XDG_DATA_HOME/containers/storage, if XDG_DATA_HOME is set.
         Otherwise $HOME/.local/share/containers/storage is used.  This field can
         be used if administrators need to change the storage location for all users.
```
Reading the above description I was thinking that the property `graphroot` should be used to set the storage path, while the field `rootless_storage_path` can be used by `administrators` if they `need to change the storage location for all users`.
Thus with this PR I ignore it when it is found in user configuration (`$HOME/.config/containers/storage.conf`), while I honor it when it is found in system configuration (`/etc/containers/storage.conf`).
In the test code instead `rootless_storage_path` is used in the user configuration for the test:
```
configPath := filepath.Join(os.Getenv("HOME"), ".config", "containers", "storage.conf")
.
.
storageConf := []byte(fmt.Sprintf("[storage]\ndriver=%s\nrootless_storage_path=%s\n[storage.options]\nmount_program=%s", driver, rootlessStoragePath, storageOpt))
err = ioutil.WriteFile(configPath, storageConf, os.ModePerm)
```
If my interpretation of man page is correct, the test should be changed to use `graphroot` instead of `rootless_storage_path` or it should test `rootless_storage_path` into the system configuration; otherwise please let me know how these fields are supposed to be used so I can try to change the code in the proper way.

rhatdan:
I have fixed podman and the test to work correctly now, but I think this patch has broken the use of ~/.config/containers/storage.conf
The rootless_storage_path is ignored there, now.

rhatdan:
The use case of the admin changing /etc, is the far more likely case, but we should fix this as well

rcasta74:
Indeed the rootless_storage_path is ignored in ~/.config/containers/storage.conf; it was my intention when following my interpretation of man page that I explained above; I was assuming that only graphroot should be used there.
I can work to make it work again, but what to do when both graphroot and rootless_storage_path are defined in ~/.config/containers/storage.conf? Which one should take precedence in this case?
Can I assume that the desidered precedence order is the following?

1. `graphroot` defined in ~/.config/containers/storage.conf
2. `rootless_storage_path` defined in ~/.config/containers/storage.conf
3. `rootless_storage_path` defined in /etc/containers/storage.conf
4. podman default ($HOME/.local/share/containers/storage)

Or shall I have to reverse the order for the first 2 points?

rhatdan:
Yes your order looks correct.

PR #742, opened 2020-10-10
archive: fix createTarFile loses the IMA attribute

yangfeiyu20102011:
archive: fix createTarFile loses the IMA attribute

reason: Lsetxattr the xattrs in the end, because system.Chtimes and other operations
may let the file lose IMA attribute

Signed-off-by: yangfeiyu<yangfeiyu20102011@163.com>

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -694,29 +694,6 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 		}
 	}
 
-	var errors []string
-	for key, value := range hdr.Xattrs {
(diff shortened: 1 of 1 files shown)


yangfeiyu20102011:
@rhatdan

rhatdan (approved):
LGTM

rhatdan:
@saschagrunert @vrothberg @TomSweeneyRedHat @nalind PTAL

saschagrunert (approved):
Changes LGTM. Is there any way that we can add a test-case to verify that we do not break the behavior in the future again?

TomSweeneyRedHat:
LGTM
but would like a head nod from @vrothberg or @nalind

vrothberg (commented):
LGTM

PR #743, opened 2020-10-12
Tighten permissions on created directory

rhatdan:
The mount points of containers are being marked as 755 where
a lot of images now come with the default for "/", as 555.

The tighter controls on / means that a root process can not
modify the / directory without DAC_OVERRIDE.  And some consider
this better security.  Bottom line is tools like `rpm -qVf /`
are reporting failures on permissions on / in lots of RPM
based images.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1854566

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -42,6 +42,8 @@ var (
 	untar = chrootarchive.UntarUncompressed
 )
 
+const defaultPerms = os.FileMode(0555)
+
 // This backend uses the overlay union filesystem for containers
 // with diff directories for each layer.
 
@@ -571,15 +573,17 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 	if err := idtools.MkdirAllAs(path.Dir(dir), 0700, rootUID, rootGID); err != nil {
 		return err
 	}
+	perms := defaultPerms
 	if parent != "" {
 		st, err := system.Stat(d.dir(parent))
 		if err != nil {
 			return err
 		}
+		perms = os.FileMode(st.Mode())
 		rootUID = int(st.UID())
 		rootGID = int(st.GID())
 	}
-	if err := idtools.MkdirAs(dir, 0700, rootUID, rootGID); err != nil {
+	if err := idtools.MkdirAs(dir, perms, rootUID, rootGID); err != nil {
 		return err
 	}
 
@@ -604,7 +608,7 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 		}
 	}
 
-	if err := idtools.MkdirAs(path.Join(dir, "diff"), 0755, rootUID, rootGID); err != nil {
+	if err := idtools.MkdirAs(path.Join(dir, "diff"), perms, rootUID, rootGID); err != nil {
 		return err
 	}
 
@@ -847,7 +851,11 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		return "", err
 	}
 	diffN := 1
-	_, err = os.Stat(filepath.Join(dir, nameWithSuffix("diff", diffN)))
+	perms := defaultPerms
+	st, err := os.Stat(filepath.Join(dir, nameWithSuffix("diff", diffN)))
+	if err == nil {
+		perms = os.FileMode(st.Mode())
+	}
 	for err == nil {
 		absLowers = append(absLowers, filepath.Join(dir, nameWithSuffix("diff", diffN)))
 		relLowers = append(relLowers, dumbJoin(string(link), "..", nameWithSuffix("diff", diffN)))
@@ -908,7 +916,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		return "", err
 	}
 	diffDir := path.Join(dir, "diff")
-	if err := idtools.MkdirAllAs(diffDir, 0755, rootUID, rootGID); err != nil {
+	if err := idtools.MkdirAllAs(diffDir, perms, rootUID, rootGID); err != nil {
 		return "", err
 	}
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind @mtrmac @vrothberg @rhvgoyal @saschagrunert PTAL
I am not sure if I went too far.

mtrmac:
Conceptually, I would expect the permissions of the `/` directory, as visible _from inside the container_, to come from the permissions set in the corresponding layer tarball — and with a trivial test with a loopback-mounted filesystem, the permissions of the mount point are indeed set from the mounted filesystem, not from the parent.

So I don’t understand how the PR description and the code really relates; at best changing one hard-coded value for another replaces one set of images with non-matching permissions with another set. With my very minimal understanding of overlay, are _all_ of those directories even container-observable? I thought that the permissions of mount points _before the mount happens_ basically don’t matter.

Either way, I was expecting some way to apply the permissions from the layer tarball to the root of the extracted tree (or, even, some code that _removes_ something that prevents such permission application, which should be the default behavior).

This may well be correct but it is surprising to me at a first glance, and I don’t know enough about overlay or c/storage to immediately say one way or another — I’ll defer to experts.

giuseppe:
agree with @mtrmac.  We should probably set the same permissions as the rootfs in the image, otherwise switching a hardcoded value for another will silent the rpm warning but can introduce other regressions.  As you pointed out, running without `CAP_DAC_OVERRIDE` and not being able to create dir/files in the root directory.

Reinstalling `filesystem` from the container fixes the issue with `/`, but there are other errors reported by `rpm -qV`, such as  `/proc` and `/sys` having the wrong ownership, that cannot be fixed from a user namespace.
While we can think of fixing the root permissions to follow what the rootfs tarball specifies, I think these other issues must be addressed in the rpm tool itself and make it aware of running in a container

rhatdan:
@nalind Didn't we talk about this a while ago, and figured out there was an issue in / of the container.

nalind:
The root directory of a layer is frequently omitted from layer blobs in images (for example, docker.io/library/fedora), though the one in the bug report appears to be an exception to that.  In the case of overlay, if I'm reading the docs right, the permissions on the upper ("diff") directory are the ones that are visible for the mountpoint.

drivers/overlay/overlay.go
```diff
@@ -400,11 +400,11 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
...
-		_ = idtools.MkdirAs(mergedDir, 0700, rootUID, rootGID)
-		_ = idtools.MkdirAs(lower1Dir, 0700, rootUID, rootGID)
-		_ = idtools.MkdirAs(lower2Dir, 0700, rootUID, rootGID)
-		_ = idtools.MkdirAs(upperDir, 0700, rootUID, rootGID)
-		_ = idtools.MkdirAs(workDir, 0700, rootUID, rootGID)
+		_ = idtools.MkdirAs(mergedDir, 0500, rootUID, rootGID)
+		_ = idtools.MkdirAs(lower1Dir, 0500, rootUID, rootGID)
+		_ = idtools.MkdirAs(lower2Dir, 0500, rootUID, rootGID)
+		_ = idtools.MkdirAs(upperDir, 0500, rootUID, rootGID)
+		_ = idtools.MkdirAs(workDir, 0500, rootUID, rootGID)
```
nalind, line 407: This is run-time feature test code.  We delete these directories when the test is finished.  What's the purpose of this hunk?

nalind:
As an alternate approach that I haven't tested, does defaulting the permissions of the "diff" directory for a layer to the permissions of its parent layer's "diff" directory, if it has one, provide the desired result?  I'd expect that to bring us more in line the device mapper driver.

drivers/overlay/overlay.go
rhatdan: Just globally changed the permissions.

rhatdan:
@giuseppe @nalind @mtrmac PTAL, I think I got it correctly,  Tested this with Podman and it seems to work correctly.

rhatdan:
@giuseppe @mtrmac PTAL

rhatdan:
Nalin points out that most/alot of container images do not come with ./ inside of them.  

IE the tar ball contains. (Fedora comes this way)
tmp
usr
var
as opposed to (UBI8 comes this way)
./tmp
./usr
./var

For these containers need a safe default.  Fedora, Centos, RHEL have decided that the safe default for /, is 555.  So we should go with that.

This PR attempts to look at the parent directory or the directory of lower layers and use it.

vrothberg (commented):
LGTM ... Ubuntu is sick atm

rhatdan:
@giuseppe @vrothberg Please /approve this PR so testing will happen

drivers/overlay/overlay.go
```diff
@@ -571,15 +573,17 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 	if err := idtools.MkdirAllAs(path.Dir(dir), 0700, rootUID, rootGID); err != nil {
 		return err
 	}
+	perms := defaultPerms
 	if parent != "" {
 		st, err := system.Stat(d.dir(parent))
 		if err != nil {
 			return err
 		}
+		perms = os.FileMode(st.Mode())
```
mtrmac, line 582: In all instances: AFAICS `st.Mode()` is already an `os.FileMode`, so the cast is unnecessary; OTOH the value also includes the file type bits.

So AFAICS this should be `perms = st.Mode().Perms()`.

rhatdan: if I pass st.Mode() I get this error.

```
make
GO111MODULE=on go build -mod=vendor -compiler gc -tags "  "  ./cmd/containers-storage
# github.com/containers/storage/drivers/overlay
drivers/overlay/overlay.go:582:9: cannot use st.Mode() (type uint32) as type os.FileMode in assignment
make: *** [Makefile:53: containers-storage] Error 2
```
Which indicates that st.Mode() has no value of perms and needs the cast.

Note this is a system.Stat not an os.Stat.

mtrmac: Ah, my mistake.

cgwalters:
I know this is an old PR but all the relevant people are here, and this is now biting me. It took me quite a while to understand that basically container runtimes have always ignored `/` in the tar stream (if it exists, which per this PR often doesn't, but *sometimes* it does).

AI-generated summary in https://gist.github.com/cgwalters/c18c9337aa9345d763aa446cc95c7847 (but of course with a lot of guidance and tweaking from me)

I am probably going to work around this in composefs by setting the metadata for `/` to equal `/usr`...probably no one is going to ever care, but it sure is ugly.

PR #744, opened 2020-10-12
storage: add option force_mask= and permissions_xattr=

giuseppe:
force_mask sets a permission mask used for the new files and directories.

It is useful for using a NFS share for the rootless storage.  It requires this change in fuse-overlayfs:

permissions_xattr, if it is set then it for each file write its original permissions to the specified extended attribute.

Together with force_mask= it allows to use NFS as a backing store for rootless containers:

https://github.com/containers/fuse-overlayfs/pull/246

```
[storage]
  driver = "overlay"
  graphroot = "/mnt/nfs/home/storage"
  [storage.options]
    mountopt = "xattr_permissions=2"
  [storage.options.overlay]
     force_mask = "0755"
     ignore_chown_errors = "true"
```
    
### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -171,8 +171,8 @@ The `storage.options.overlay` table supports the following options:
   force_mask sets an octal permission mask that is used for the new files and directories.  The values "native", "shared", "private" are also accepted.  (default: "native")
   "private": it is equivalent to 0700.  The owner has rwx access to the files.
   "shared": it is equivalent to 0755.  The owner has rwx access to the files and everyone else can read, access and execute them.
-  It is possible to use **force_mask** only when **mount_program** is specified.
   Note: The force_mask Flag is an experimental feature, it could change in the future.
+  When "force_mask" is used then the original permission mask is stored in the "user.containers.override_stat" xattr.
 
 **mount_program**=""
   Specifies the path to a custom program to use instead of using kernel defaults
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -149,7 +149,7 @@ The `storage.options.thinpool` table supports the following options for the `dev
   Comma separated list of default options to be used to mount container images.  Suggested value "nodev". Mount options are documented in the mount(8) man page.
 
 **size**=""
-  Maximum size of a container image.   This flag can be used to set quota on the size of container images. (format: <number>[<unit>], where unit = b (bytes), k (kilobytes), m (megabytes), or g (gigabytes))
+  Maximum size of a container image.  This flag can be used to set quota on the size of container images. (format: <number>[<unit>], where unit = b (bytes), k (kilobytes), m (megabytes), or g (gigabytes))
 
 **use_deferred_deletion**=""
   Marks thinpool device for deferred deletion. If the thinpool is in use when the driver attempts to delete it, the driver will attempt to delete device every 30 seconds until successful, or when it restarts.  Deferred deletion permanently deletes the device and all data stored in the device will be lost. (default: true).
@@ -166,6 +166,13 @@ The `storage.options.overlay` table supports the following options:
 
 **ignore_chown_errors** = "false"
   ignore_chown_errors can be set to allow a non privileged user running with a  single UID within a user namespace to run containers. The user can pull and use any image even those with multiple uids.  Note multiple UIDs will be squashed down to the default uid in the container.  These images will have no separation between the users in the container. (default: false)
+
+**force_mask** = "0000|shared|private"
+  force_mask sets an octal permission mask that is used for the new files and directories.  The values "native", "shared", "private" are also accepted.  (default: "native")
+  "private": it is equivalent to 0700.  The owner has rwx access to the files.
+  "shared": it is equivalent to 0755.  The owner has rwx access to the files and everyone else can read, access and execute them.
+  It is possible to use **force_mask** only when **mount_program** is specified.
+  Note: The force_mask Flag is an experimental feature, it could change in the future.
 
 **mount_program**=""
   Specifies the path to a custom program to use instead of using kernel defaults
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"os"
 )
 
 // ThinpoolOptionsConfig represents the "storage.options.thinpool"
@@ -94,6 +95,9 @@ type OverlayOptionsConfig struct {
 	Size string `toml:"size"`
 	// Do not create a bind mount on the storage home
 	SkipMountHome string `toml:"skip_mount_home"`
+	// ForceMask indicates the permissions mask (e.g. "0755") to use for new
+	// files and directories
+	ForceMask string `toml:"force_mask"`
 }
 
 type VfsOptionsConfig struct {
@@ -128,6 +132,10 @@ type OptionsConfig struct {
 	// IgnoreChownErrors is a flag for whether chown errors should be
 	// ignored when building an image.
 	IgnoreChownErrors string `toml:"ignore_chown_errors"`
+
+	// ForceMask indicates the permissions mask (e.g. "0755") to use for new
+	// files and directories.
+	ForceMask os.FileMode `toml:"force_mask"`
 
 	// RemapUser is the name of one or more entries in /etc/subuid which
 	// should be used to set up default UID mappings.
@@ -278,6 +286,11 @@ func GetGraphDriverOptions(driverName string, options OptionsConfig) []string {
 			doptions = append(doptions, fmt.Sprintf("%s.skip_mount_home=%s", driverName, options.Overlay.SkipMountHome))
 		} else if options.SkipMountHome != "" {
 			doptions = append(doptions, fmt.Sprintf("%s.skip_mount_home=%s", driverName, options.SkipMountHome))
+		}
+		if options.Overlay.ForceMask != "" {
+			doptions = append(doptions, fmt.Sprintf("%s.force_mask=%s", driverName, options.Overlay.ForceMask))
+		} else if options.ForceMask != 0 {
+			doptions = append(doptions, fmt.Sprintf("%s.force_mask=%s", driverName, options.ForceMask))
 		}
 	case "vfs":
 		if options.Vfs.IgnoreChownErrors != "" {
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -60,6 +60,7 @@ type ApplyDiffOpts struct {
 	Mappings          *idtools.IDMappings
 	MountLabel        string
 	IgnoreChownErrors bool
+	ForceMask         *os.FileMode
 }
 
 // InitFunc initializes the storage driver.
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -93,6 +93,7 @@ type overlayOptions struct {
 	skipMountHome     bool
 	mountOptions      string
 	ignoreChownErrors bool
+	forceMask         *os.FileMode
 }
 
 // Driver contains information about the home directory and the list of active mounts that are created using this driver.
@@ -143,6 +144,9 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 
 	// check if they are running over btrfs, aufs, zfs, overlay, or ecryptfs
 	if opts.mountProgram == "" {
+		if opts.forceMask != nil {
+			return nil, errors.New("'force_mask' is supported only with 'mount_program'")
+		}
 		switch fsMagic {
 		case graphdriver.FsMagicAufs, graphdriver.FsMagicZfs, graphdriver.FsMagicOverlay, graphdriver.FsMagicEcryptfs:
 			return nil, errors.Wrapf(graphdriver.ErrIncompatibleFS, "'overlay' is not supported over %s, a mount_program is required", backingFs)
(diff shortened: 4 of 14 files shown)


rhatdan:
What is 493?

force_mask = 493

giuseppe:
> What is 493?
> 
> force_mask = 493

that is `0755`.

```
$ printf "%d" 0755
493
```
Unfortunately toml doesn't support octals.  Should we use a string or just document it?

rhatdan:
Yes it should be a string.  Humans will not understand this number. 
"0755" is much better.

giuseppe:
> Yes it should be a string. Humans will not understand this number.
> "0755" is much better.

changed to be a string.

rhatdan:
So this means every file and directory will be marked 0755?  I guess we need to talk about what you are after here?  Should we separate the files from the directories.   Would this make /etc/shadow inside of a container work readable to content in the container?

giuseppe:
> So this means every file and directory will be marked 0755? I guess we need to talk about what you are after here? Should we separate the files from the directories. Would this make /etc/shadow inside of a container work readable to content in the container?

yes, every file and directory in the container image will get the mode 0755.

I didn't want to expose too much of the fuse-overlayfs internals in containers/storage.  If we extend containers/storage to write the original permission mask (+owner uid/gid) in an extended attribute that fuse-overlayfs understands, we could keep the same privileges at mount time.  What do you think about it?

rhatdan:
I think we should have a quick meeting to discuss this.

rhatdan:
Should this be an option in storage.conf, or just something set by podman and Buildah for rootless users?

rhatdan:
The issue I have with this, is that this effects rootfull as well as rootless users.

giuseppe:
> Should this be an option in storage.conf, or just something set by podman and Buildah for rootless users?

IMO, it should be something users set explicitly and not enabled by default.  With the wrong fuse-overlayfs version containers will have the wrong permissions for the files.

I see a good use case for root as well.  We can make images available to rootless users as a rootless user can read all the files in the root storage.

With this feature we can enable a read-only storage owned by root and accessible to rootless users.

We can discuss it further at the tomorrow meeting.

rhatdan:
Ok.

docs/containers-storage.conf.5.md
```diff
@@ -167,6 +167,12 @@ The `storage.options.overlay` table supports the following options:
 **ignore_chown_errors** = "false"
   ignore_chown_errors can be set to allow a non privileged user running with a  single UID within a user namespace to run containers. The user can pull and use any image even those with multiple uids.  Note multiple UIDs will be squashed down to the default uid in the container.  These images will have no separation between the users in the container. (default: false)
 
+**force_mask** = "0000"
+  force_mask sets a octal permissions mask and those bits are set for each new file or directory created in the storage.  The final permission for the new files and directories is set to ORIGINAL_PERMISSION|FORCE_MASK. (default: 0000)
```
TomSweeneyRedHat, line 171:
```suggestion
  force_mask sets an octal permission mask and those bits are set for each new file or directory created in the storage.  The final permission for the new files and directories is set to ORIGINAL_PERMISSION|FORCE_MASK. (default: 0000)
```

docs/containers-storage.conf.5.md
```diff
@@ -167,6 +167,12 @@ The `storage.options.overlay` table supports the following options:
...
+
+**permissions_xattr** = ""
+  it specifies an extended attribute where the original permissions for the file are stored in the format UID:GID:MODE.
```
TomSweeneyRedHat, line 174:
```suggestion
  This option specifies an extended attribute where the original permissions for the file are stored in the format UID:GID:MODE.
```

mtrmac (commented):
As mentioned elsewhere, if there’s only one value of `force_mask` that works usefully, and only one value of `permissions_xattr` that works at all, I’m not a fan of making this available as individual options — _and_ requiring users to know the magic values.

Provide a higher-level `permissions_mechanism=native|fuse-overlayfs`, which instructs the code to do whatever is necessary to interoperate with fuse-overlayfs. That still allows us to support future permissions mechanisms in the future, but does not imply any assumptions about how those mechanisms work (i.e. that things are stored in an extended attribute at all, let alone using a specific format).

rhatdan:
permissions_mechanism=native|shared?

I would prefer not to hard code fuse-overlay in this as well.  Setting it shared basically means the images are all world readable.

giuseppe:
I've added the `permissions_xattr=` option only to avoid hard-coding fuse-overlayfs.  Any idea on how it can be done differently?

I like making force_mask more user friendly and support `native|shared`.  What if we support `force_mask=0xxx|native|shared` so that users can still customize the mask if needed?

rhatdan:
I am fine with the three options, but having a general user have to figure out 0755 or 0000 stinks versus native and shared.

mtrmac:
Per the other conversation, we might actually want two independent parameters (please suggest better names); “NFS-hosted” does not automatically imply “OK-to-share”.
- metadata-storage=native|simulated , which handles whether ownership+permission(+ACL) of the images are stored directly using the kernel filesystem, or only in a xattr that is interpreted by a FUSE driver.
- raw-file-permissions=0700 (private) | 0755 (shared-for-read) | 0775 (multiple-writers) | 0750 | 0770 (I ran out of naming ideas): only valid with metadata-storage=simulated, to support sharing the c/storage backing store in multi-user systems. 

(Are there any actual multi-user consumers that will benefit, or is it just theoretical at this point? Multi-user computers are increasingly rare.)

koebi001:
> (Are there any actual multi-user consumers that will benefit, or is it just theoretical at this point? Multi-user computers are increasingly rare.)

I am very interested in getting this working. In our lab we use a shared network file system and configure all hosts with slurm to better utilize our limited resources. This is especially important for machine learning topics where it gets more an more challenging to get the frameworks working with all their wide dependencies. Being able to use containers would be a big win.
Right now we have a single shared host for docker where just a few users are testing stuff. The local container storage takes already 1TB.
The only feasible way to run containers on a cluster environment is by having a shared network file system for often used container images.
Using rootless podman the only problem to solve seems to be sharing the storage. Most important the part which can be read only.

docs/containers-storage.conf.5.md
```diff
@@ -167,6 +167,15 @@ The `storage.options.overlay` table supports the following options:
 **ignore_chown_errors** = "false"
   ignore_chown_errors can be set to allow a non privileged user running with a  single UID within a user namespace to run containers. The user can pull and use any image even those with multiple uids.  Note multiple UIDs will be squashed down to the default uid in the container.  These images will have no separation between the users in the container. (default: false)
 
+**force_mask** = "0000|native|shared|private"
+  force_mask sets an octal permission mask and those bits are set for each new file or directory created in the storage.  The final permission for the new files and directories is set to ORIGINAL_PERMISSION|FORCE_MASK.  The values "native", "shared", "private" are also accepted.  (default: "native")
+  "native": it is equivalent to 0000.  The files maintain their original permission mask.
+  "private": it is equivalent to 0700.  The owner has rwx access to the files.
+  "shared": it is equivalent to 0755.  The owner has rwx access to the files and everyone else can read, access and execute them.
+
+**permissions_xattr** = ""
```
rhatdan, line 176: Should this be a constant?  Or at least document the one that fuse-overlay will use.  Not sure why users would expect more then one xattr?

rhatdan (approved):

docs/containers-storage.conf.5.md
rhatdan: I think this is useless complexity.  Just store the attributes in storage.dac.permissions or something and be done with it.

pkg/config/config.go
```diff
@@ -94,6 +95,13 @@ type OverlayOptionsConfig struct {
 	Size string `toml:"size"`
 	// Do not create a bind mount on the storage home
 	SkipMountHome string `toml:"skip_mount_home"`
+	// ForceMask indicates a mask of permissions (e.g. "0755") that are always set for
+	// created files.  The file permissions will be: ORIGINAL_PERMISSIONS | FORCE_MASK.
+	ForceMask string `toml:"force_mask"`
```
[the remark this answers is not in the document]
giuseppe: for the NFS case, the mask to use is `0700` so they won't become world readable (unless they were already so that ORing the original permission with 0700 won't change the group and world permissions).

The `ForceMask=0755` is a different use case.  That is useful to share the storage with other users.  In that cases all the files/directories must be world readable and accessible.

Do you think we should make this distinction clearer somehow?

docs/containers-storage.conf.5.md
giuseppe: I didn't want to hard code "fuse-overlayfs" in containers/storage.  I can document it better that it is expected to be used with fuse-overlayfs, and perhaps explain how fuse-overlayfs uses it

rhatdan: Why not just define a "storage" xattr and have fuse-overlyfs deal with it?

rhatdan:
In my opinion having the force-mask is all you need.  This tells containers/storage to store all of the content with that mask and to save the real owner/group/permissions in a storage defined xattr.  Once this is standardized, then we can take third party tools like fuse-overlay and use it present the owner/group/permissions to the callers.

@nalind Might prefer this to be a separate driver then to override the overlay driver.  How much work would it be to create a fuse-overlay driver rather?  Would that make more sense then to store data that the native overlay driver does not understand?

pkg/config/config.go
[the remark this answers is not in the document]
mtrmac: I’m not disputing the 0700/0755 values. But consider that the default permissions of `/` in a container are 0555, so even in the “private” setting, the file will end up with external 0755 (and an in-container 0777 file would end up 0777 external regardless of `ForceMask`). Is that OK, or should the file really end up _exactly_ 0700/0755?

docs/containers-storage.conf.5.md
giuseppe: fuse-overlayfs currently handles every `user.fuseoverlayfs*` as a private namespace, while any other xattr is not filtered out.  If we add a `storage` xattr we'd have to handle a separate namespace and not permit it inside of the containers, so `user.storage` or `user.containers` won't be available to container processes.

pkg/config/config.go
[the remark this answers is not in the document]
giuseppe: I am not sure.  Both ways are enough for the NFS storage and shared storage cases.

I thought having the `ForceMask` work as an "opposite" umask could give more flexibility in the future, but at the moment I cannot think of any other use case where it would be useful

giuseppe:
I've dropped the second option and always write `user.containers.override_stat`.  I've changed fuse-overlayfs to honor it: https://github.com/containers/fuse-overlayfs/pull/257


> @nalind Might prefer this to be a separate driver then to override the overlay driver. How much work would it be to create a fuse-overlay driver rather? Would that make more sense then to store data that the native overlay driver does not understand?

would it be ok if we restrict the use case to `mountProgram != ""` and raise an error when `force_mask` is used without `mountProgram`?

rhatdan:
Ok force the mount_program.

docs/containers-storage.conf.5.md
```diff
@@ -167,6 +167,13 @@ The `storage.options.overlay` table supports the following options:
 **ignore_chown_errors** = "false"
   ignore_chown_errors can be set to allow a non privileged user running with a  single UID within a user namespace to run containers. The user can pull and use any image even those with multiple uids.  Note multiple UIDs will be squashed down to the default uid in the container.  These images will have no separation between the users in the container. (default: false)
 
+**force_mask** = "0000|native|shared|private"
```
rhatdan, line 170: Need to mention here that it requires mount_program.

rhatdan: Update man page, then LGTM

giuseppe: fixed.  Should we mark it as experimental for now so we keep some freedom to change it if we need?

rhatdan: SGTM

mtrmac:
> I've dropped the second option and always write `user.containers.override_stat`. I've changed fuse-overlayfs to honor it: [containers/fuse-overlayfs#257](https://github.com/containers/fuse-overlayfs/pull/257

From a quick look, I can’t see anything that prevents this to be set from inside the container (I’m worried about allowing a low-privileged in-container user to escalate to a high-privileged in-container user via a set-UID executable); from a quick search I found `ovl_setxattr` which does not filter that out. Where is the security mechanism?

rhatdan:
Yes fuse-overlay will need to be changed to prevent from the setting of the file attribute.

giuseppe:
Thanks for the hint.  I've not done it previously as the user already needs write permissions to use setxattr on that file.  In any case, I've opened a PR to always block it: https://github.com/containers/fuse-overlayfs/pull/258

giuseppe:
marked as experimental :arrow_up:

docs/containers-storage.conf.5.md
```diff
@@ -167,6 +167,14 @@ The `storage.options.overlay` table supports the following options:
...
+  force_mask sets an octal permission mask and those bits are set for each new file or directory created in the storage.  The final permission for the new files and directories is set to ORIGINAL_PERMISSION|FORCE_MASK.  The values "native", "shared", "private" are also accepted.  (default: "native")
+  "native": it is equivalent to 0000.  The files maintain their original permission mask.
+  "private": it is equivalent to 0700.  The owner has rwx access to the files.
+  "shared": it is equivalent to 0755.  The owner has rwx access to the files and everyone else can read, access and execute them.
+  It is possible to use **force_mask** only when **mount_program** is specified.  It is an experimental feature.
```
rhatdan, line 175: How about `Note: The force_mask Flag is an experimental feature, it could change in the future.`

rhatdan (approved):
Can you also update the default `storage.conf` file and add the commented out force_mask flag and an explanation of what it is,

pkg/config/config.go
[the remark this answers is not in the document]
giuseppe: I've changed the semantic to mean "final permissions mask".  I guess it is easier to understand and also avoids some issues if a file in an image is world writeable and we end up opening the image to every other user.

rhatdan:
LGTM

PR #745, opened 2020-10-15
Stop excessive wrapping

rhatdan:
Golang built in functions like os.Create and others print the name of
the file system object when they fail.  Wrapping them a second time
with the file system object, makes the error message look like crap
when reported to the user.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -907,11 +907,11 @@ func (r *layerStore) ParentOwners(id string) (uids, gids []int, err error) {
 	for dir := filepath.Dir(layer.MountPoint); dir != "" && dir != string(os.PathSeparator); dir = filepath.Dir(dir) {
 		st, err := system.Stat(dir)
 		if err != nil {
-			return nil, nil, errors.Wrapf(err, "error reading ownership of directory %q", dir)
+			return nil, nil, err
 		}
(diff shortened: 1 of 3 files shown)


rhatdan:
@kolyshkin @nalind @vrothberg @saschagrunert @TomSweeneyRedHat @giuseppe @ashley-cui PTAL

layers.go
```diff
@@ -907,11 +907,11 @@ func (r *layerStore) ParentOwners(id string) (uids, gids []int, err error) {
 	for dir := filepath.Dir(layer.MountPoint); dir != "" && dir != string(os.PathSeparator); dir = filepath.Dir(dir) {
 		st, err := system.Stat(dir)
 		if err != nil {
-			return nil, nil, errors.Wrapf(err, "error reading ownership of directory %q", dir)
+			return nil, nil, err
```
saschagrunert, line 910: I think adding an additional layer of context is still valuable to the user, like:

```suggestion
			return nil, nil, errors.Wrap(err, "read directory ownership")
```

vrothberg (commented):
LGTM

saschagrunert (approved):
LGTM

PR #746, opened 2020-10-15
Drop Warn to Info for common issues

rhatdan:
We are trying to drop buildah and eventually podman to warn by default

Since these warning messages happen all the time, I believe they should be
infos.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -219,7 +219,7 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 					return nil, errors.Wrap(err, "error recording metacopy-being-used status")
 				}
 			} else {
-				logrus.Warnf("overlay test mount did not indicate whether or not metacopy is being used: %v", err)
+				logrus.Infof("overlay test mount did not indicate whether or not metacopy is being used: %v", err)
 				return nil, err
 			}
 		}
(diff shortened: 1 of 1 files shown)


drivers/overlay/overlay.go
```diff
@@ -280,7 +280,7 @@ func parseOptions(options []string) (*overlayOptions, error) {
 		trimkey = strings.TrimPrefix(trimkey, ".")
 		switch trimkey {
 		case "override_kernel_check":
-			logrus.Warnf("overlay: override_kernel_check option was specified, but is no longer necessary")
```
vrothberg, line 283: I think this can be dropped and turned into a code comment? I think it has little value to users (as they don't have to worry about it).

vrothberg (commented):
LGTM but I'd love @nalind or @giuseppe to ack on it

drivers/overlay/overlay.go
TomSweeneyRedHat: I'd rather not drop it entirely, but I'd be fine with it as a Debugf instead.

rhatdan: Dropped it down to Debugf.

rhatdan:
@saschagrunert @giuseppe @vrothberg @TomSweeneyRedHat @nalind @ashley-cui PTAL

giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #748, opened 2020-10-16
 Cleanup code to explain what is going on here. 

rhatdan:
I got fooled by this code, and attempted a PR to clean it up, that
was a mistake.  I assumed that the call should fail if the Container
call failed. But the store.Mount(id, mountlabel) handles container
mounts as well as layer mounts. This means the ID could be a container
or a different layer. When the attempt to read the container fails, we
continue assuming that the id is a layer and attempt to mount it as a layer.
If the id is a container then we grap the toplayer of the container and use
it is as an id.

This PR does not change the functionality of the code, but makes it easier
for future programmers to understand the functionality.

### DIFF
--- a/store.go
+++ b/store.go
@@ -2677,21 +2677,16 @@ func (s *store) MountImage(id string, mountOpts []string, mountLabel string) (st
 }
 
 func (s *store) Mount(id, mountLabel string) (string, error) {
-	container, err := s.Container(id)
-	var (
-		uidMap, gidMap []idtools.IDMap
-		mountOpts      []string
-	)
-	if err == nil {
-		uidMap, gidMap = container.UIDMap, container.GIDMap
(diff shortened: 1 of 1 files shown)


giuseppe:
LGTM once the error reported by lint is fixed

TomSweeneyRedHat:
LGTM
assuming happy tests which are all very red ATM

rhatdan:
@giuseppe @TomSweeneyRedHat The original code was correct, and my patch was breaking the functionality.  I have updated the code and explained above what is going on.
This Patch will not change the functionality but fixes the code so a future me does not make bad assumptions about what is going on.

TomSweeneyRedHat:
LGTM

saschagrunert (approved):
LGTM

PR #751, opened 2020-10-22
platformLchown: use os package instead

haircommander:
go 1.14 and beyond doesn't catch EINTR in the syscall package anymore. This causes manual calls to the syscall package to fail from interrupts, which is really not a failure, as the command should just be retried.

however, the os package does this for us. Luckily, there's an `os.Lchown`

unfortunately, for the other syscall functions, there will need to be a function wrapper, but I believe that should be addressed in a separate PR

Signed-off-by: Peter Hunt <pehunt@redhat.com>

### DIFF
--- a/drivers/chown_unix.go
+++ b/drivers/chown_unix.go
@@ -54,7 +54,7 @@ func platformLChown(path string, info os.FileInfo, toHost, toContainer *idtools.
 		}
 
 		// Make the change.
-		if err := syscall.Lchown(path, uid, gid); err != nil {
+		if err := os.Lchown(path, uid, gid); err != nil {
 			return fmt.Errorf("%s: chown(%q): %v", os.Args[0], path, err)
 		}
 		// Restore the SUID and SGID bits if they were originally set.


rhatdan:
LGTM
@vrothberg @nalind @saschagrunert @TomSweeneyRedHat @giuseppe PTAL

nalind:
I'm looking at `/usr/lib/golang/src/os/file_posix.go` from go 1.15.3, and I don't see any wrapping in its `Lchown()` function.  Are you sure the `os` package is handling those errors for every case?  I see some logic in helper functions used by the implementations of `OpenFile()`, `Getwd()`, and `RemoveAll()`.

haircommander:
> I'm looking at `/usr/lib/golang/src/os/file_posix.go` from go 1.15.3, and I don't see any wrapping in its `Lchown()` function. Are you sure the `os` package is handling those errors for every case? I see some logic in helper functions used by the implementations of `OpenFile()`, `Getwd()`, and `RemoveAll()`.

hmmmm, they were supposed to. I know in 1.14 there were breaking changes in os.WriteFile() that were fixed in 1.15, but maybe they didn't get all cases :'(


as a side note, I am now unable to reproduce the issue (with or without this change) where before I was reproducing regularly... I will get back to this tomorrow.

nalind:
You're right, `io/ioutil.WriteFile()` does use `os.OpenFile()`.

giuseppe (approved):
LGTM

kolyshkin:
I second the concern raised by @nalind (https://github.com/containers/storage/pull/751#issuecomment-714769377). Here's how os.Lchown looks like in golang-1.15.3:

```go
func Lchown(name string, uid, gid int) error {
        if e := syscall.Lchown(name, uid, gid); e != nil {
                return &PathError{"lchown", name, e}
        }
        return ni
```

While it's good to have it for the reason of having more context to the error, it won't fix EINTR problem.

I'll work on a PR.

PR #413, opened 2019-08-06
Don't chown on safechown if directory exists and has correct ownership 

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #754, opened 2020-10-27
idtools: skip stat if not necessary

giuseppe:
avoid the more expensive stat syscall if chown could be performed
successfully.

Fallback to stat only when chown fails.

### DIFF
--- a/pkg/idtools/idtools.go
+++ b/pkg/idtools/idtools.go
@@ -297,19 +297,27 @@ func checkChownErr(err error, name string, uid, gid int) error {
 }
 
 func SafeChown(name string, uid, gid int) error {
+	err := os.Chown(name, uid, gid)
+	if err == nil {
+		return nil
+	}
 	if stat, statErr := system.Stat(name); statErr == nil {
 		if stat.UID() == uint32(uid) && stat.GID() == uint32(gid) {
 			return nil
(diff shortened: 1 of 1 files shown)


rhatdan:
I thought this was done to speed up systems, IE Don't chown a file if it does not need it.
Now you are going to chown/write to the file system even if the chown is not necessary.

giuseppe:
> I thought this was done to speed up systems, IE Don't chown a file if it does not need it.
> Now you are going to chown/write to the file system even if the chown is not necessary.

do we have a pointer to the issue?  It seem strange the chown is slower that the stat syscall.

rhatdan:
Why would writing to disk be slowed then reading from disk?

rhatdan:
@nalind Do you remember when why we did this?

nalind:
There's no "why" in #413, but my best guess is that it was triggering metadata-only copy-ups in overlay with "metacopy=on".

giuseppe:
> There's no "why" in #413, but my best guess is that it was triggering metadata-only copy-ups in overlay with "metacopy=on".

that makes sense.  Thanks.  I'll close this PR.

@rhvgoyal do you think this is something overlay in the kernel should take care of and avoid a copyup if there is nothing was changed?

PR #757, opened 2020-10-29
chmod/chown: handle EINTR

kolyshkin:
The following failure has been observed in CI (see [1], [2]):
    
        storage-chown-by-maps: chown("xxx"): interrupted system call
    
While chown and chmod should be auto-restartable by the kernel (provided
all the signal handlers are installed with SA_RESTART flag), it looks
like it is not always the case, or there might be some exclusions,
so this should be handled.
    
Surely, the possibility of getting EINTR is amplified since Go 1.14
introduced async preemptible goroutimes (see [3]), the feature that
is implemented via frequently sending signal 22 to all threads.
    
Add and use wrappers for Chmod and Lchown that retry on EINTR.

[1] https://github.com/containers/podman/issues/8152
[2] https://github.com/cri-o/cri-o/pull/4310#issuecomment-718361022
[3] https://golang.org/doc/go1.14#runtime

Previous attempt to fix this: https://github.com/containers/storage/pull/751

### DIFF
--- a/pkg/system/chmod.go
+++ b/pkg/system/chmod.go
@@ -1,0 +1,17 @@
+package system
+
+import (
+	"errors"
+	"os"
+	"syscall"
+)
+
+func Chmod(name string, mode os.FileMode) error {
+	err := os.Chmod(name, mode)
+
+	for err != nil && errors.Is(err, syscall.EINTR) {
+		err = os.Chmod(name, mode)
+	}
+
+	return err
+}
--- a/drivers/chown_unix.go
+++ b/drivers/chown_unix.go
@@ -54,12 +54,12 @@ func platformLChown(path string, info os.FileInfo, toHost, toContainer *idtools.
 		}
 
 		// Make the change.
-		if err := os.Lchown(path, uid, gid); err != nil {
+		if err := system.Lchown(path, uid, gid); err != nil {
 			return fmt.Errorf("%s: %v", os.Args[0], err)
 		}
 		// Restore the SUID and SGID bits if they were originally set.
 		if (info.Mode()&os.ModeSymlink == 0) && info.Mode()&(os.ModeSetuid|os.ModeSetgid) != 0 {
-			if err := os.Chmod(path, info.Mode()); err != nil {
+			if err := system.Chmod(path, info.Mode()); err != nil {
 				return fmt.Errorf("%s: %v", os.Args[0], err)
 			}
 		}
--- a/drivers/chown_unix.go
+++ b/drivers/chown_unix.go
@@ -55,12 +55,12 @@ func platformLChown(path string, info os.FileInfo, toHost, toContainer *idtools.
 
 		// Make the change.
 		if err := os.Lchown(path, uid, gid); err != nil {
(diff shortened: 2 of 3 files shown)


pkg/system/chmod.go
```diff
@@ -0,0 +1,17 @@
...
+	"os"
+	"syscall"
+)
+
+func Chmod(name string, mode os.FileMode) error {
+	err := os.Chmod(name, mode)
+
+	for err != nil && errors.Is(err, syscall.EINTR) {
+		err = os.Chmod(name, mode)
+	}
```
haircommander, line 14: why not continually retry? is only one recall sufficient?

haircommander: lol I read for as if. nevermind :upside_down_face:

rhatdan:
LGTM

pkg/system/chmod.go
TomSweeneyRedHat: So glad I'm not the only one to have done that!

TomSweeneyRedHat:
All kinds of red unhappiness in the tests @kolyshkin

pkg/system/chmod.go
kolyshkin: It's probably because `for` is golang's `while`.

kolyshkin:
All the CI failures look similar:

```
[+0207s] Setting up grub-efi-amd64-signed (1.142.8+2.04-1ubuntu26.6) ...
[+0207s] mount: /var/lib/grub/esp: special device /dev/disk/by-id/google-persistent-disk-0-part15 does not exist.
[+0207s] dpkg: error processing package grub-efi-amd64-signed (--configure):
[+0207s]  installed grub-efi-amd64-signed package post-installation script subprocess returned error exit status 32
[+0207s] Errors were encountered while processing:
[+0207s]  grub-efi-amd64-signed
[+0209s] E: Sub-process /usr/bin/dpkg returned an error code (1)
[+0209s]     exit(100)
[+0209s]     Failure! Sleeping 90.720000 seconds
[+0299s] Attempt 6 of 7 (retry on non-zero exit):
```

Seems like a problem with GCS

kolyshkin:
Don't have rights to restart CI in this repo directly, and the failure is def not related to this patch, so force-pulling to restart CI.

cevich:
All: #760 should fix the Ubuntu upgrade problem (hopefully for good this time).

kolyshkin:
> All: #760 should fix the Ubuntu upgrade problem (hopefully for good this time).

Thanks @cevich, you beat me to it!

giuseppe (approved):
LGTM

Could you please rebase?

giuseppe:
@rhatdan rebased on top of origin/master. instead of merging master.  If you are fine we can merge

kolyshkin:
Ideally we should try doing it right there in golang `os` package. Let me see...

kolyshkin:
> Ideally we should try doing it right there in golang os package. Let me see...

So, this is partially* fixed in go master branch by https://go-review.googlesource.com/c/go/+/249178 but is not going to be backported (so I assume it will only make its way to go 1.16).

* it's a partial fix as e.g. (*os.File).Chmod() is not fixed

PR #758, opened 2020-10-29
system.*xattr: wrap errors into os.PathError

kolyshkin:
This way the error messages are more informative, and the use of the functions are more in-line with the standard ones (from `os). From the other side, it introduces a hassle of properly unwinding the errors (and this is why I hesitated to make this change).

Convert all the users accordingly (assuming there are no users outside of this repo).


rhatdan:
LGTM

giuseppe (approved):
LGTM

PR #764, opened 2020-11-01
Simplify ID range subtraction logic and bug fixes.

### LINKED ISSUE #763: Question about id subtraction code.
I am quite confused by the id subtraction code. Not sure if I misunderstood it, or it is actually a bug.

In https://github.com/containers/storage/blob/4188292d5709d4431a9dc1d07e144b092b087cc2/userns_test.go#L75, if we subtract range `[132768, 132768+4096)` from `[100000, 165536)`, it should be `[100000, 132768)` and `[132768+4096, 165536)`, but why does the test assert it to be `[132768, 165536-4096)`?

Similarly for https://github.com/containers/storage/blob/4188292d5709d4431a9dc1d07e144b092b087cc2/userns_test.go#L271, if `[100010, 100020)` is already in used range, why does the test asserts 100010 to be in the available range?

The next one is https://github.com/containers/storage/blob/4188292d5709d4431a9dc1d07e144b092b087cc2/userns_test.go#L214, if in the original range 11 maps to 1011, why does the test asserts it to be 1010? If `subtractContainerIDs` ignores the `HostID`, why does the test even check for this?

If we look at the code, sometimes the returned map keeps the original mapping (i.e. keep the `HostID - ContainerID` as invariant), e.g. https://github.com/containers/storage/blob/4188292d5709d4431a9dc1d07e144b092b087cc2/userns.go#L242-L243 and https://github.com/containers/storage/blob/4188292d5709d4431a9dc1d07e144b092b087cc2/userns.go#L249-L250, but not the other three places. The same holds true for `subtractContainerIDs`.

One could argue the `ContainerID` is ignored by `subtractHostIDs`, then why doesn't it always return 0 as the `ContainerID`?  It would be a lot nicer if it is consistent, i.e. either always 0, or maintain the original mapping in `avail`. What do you think?

likan999:
The logic of range subtraction [a,b)-[c,d) can be viewed as intersection
of [a,b) with (-inf, c) and [d, +inf), respectively. This makes the
logic simpler, that we no longer need to check 5 different cases.

It also fixes bugs that returns incorrect range.

Closes #763
Signed-off-by: Kan Li <likan@google.com>

### DIFF
--- a/userns.go
+++ b/userns.go
@@ -221,94 +221,71 @@ outer:
 	return size, nil
 }
 
+func minInt(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func maxInt(a, b int) int {
+	if a < b {
+		return b
+	}
+	return a
+}
+
 // subtractHostIDs return the subtraction of the range USED from AVAIL.  The range is specified
 // by [HostID, HostID+Size).
 // ContainerID is ignored.
 func subtractHostIDs(avail idtools.IDMap, used idtools.IDMap) []idtools.IDMap {
-	switch {
-	case used.HostID <= avail.HostID && used.HostID+used.Size >= avail.HostID+avail.Size:
-		return nil
-	case used.HostID <= avail.HostID && used.HostID+used.Size > avail.HostID && used.HostID+used.Size < avail.HostID+avail.Size:
-		newContainerID := avail.ContainerID + used.Size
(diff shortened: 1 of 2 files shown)


rhatdan:
@giuseppe PTAL

giuseppe (approved):
thanks for the patch.

LGTM

rhatdan:
Thanks @likan999 
LGTM

PR #765, opened 2020-11-02
Use unshare.HomeDir() when retrieving HOME environment

rhatdan:
This change will also lookup the homedir in the /etc/passwd file
if $HOME does not exists.


vrothberg (commented):
Could we get rid of `unshare.HomeDir()` entirely?

I didn't know it's existence until yesterday and fear that we'd diverge in the future as well. In case we worry about breaking the API, we could let `unshare.HomeDir` call `homedir.Get`?

rhatdan:
I can not get rid of it totally, and it is better then homedir.Get(). When we bump the version I can get rid of it.   I tried to get rid of it, but it causes a cyclical look.  Since homedir.Get() should be checking for the rootless user.

pkg/homedir/homedir_unix.go
```diff
@@ -25,17 +37,105 @@ func Key() string {
 //
 // If needing to do nss lookups, do not disable cgo or set osusergo.
 func Get() string {
-	home := os.Getenv(Key())
-	if home == "" {
-		if u, err := user.Current(); err == nil {
-			return u.HomeDir
+	homeDirOnce.Do(func() {
+		homeDir := os.Getenv(Key())
```
vrothberg, line 41: This looks dangerous as `homeDir` will always evaluate to the inner scope and the package-level variable will never be set.

pkg/homedir/homedir_unix.go
```diff
@@ -25,17 +29,97 @@ func Key() string {
...
+
+// GetDataHome returns XDG_DATA_HOME.
+// GetDataHome returns $HOME/.local/share and nil error if XDG_DATA_HOME is not set.
+//
+// See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
+func GetDataHome() (string, error) {
+	if xdgDataHome := os.Getenv("XDG_DATA_HOME"); xdgDataHome != "" {
+		return xdgDataHome, nil
+	}
+	home := os.Getenv("HOME")
```
vrothberg, line 105: Should we use `Get()` and hence `unshare...` here?

pkg/homedir/homedir_unix.go
```diff
@@ -25,17 +29,97 @@ func Key() string {
...
+
+// GetConfigHome returns XDG_CONFIG_HOME.
+// GetConfigHome returns $HOME/.config and nil error if XDG_CONFIG_HOME is not set.
+//
+// See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
+func GetConfigHome() (string, error) {
+	if xdgConfigHome := os.Getenv("XDG_CONFIG_HOME"); xdgConfigHome != "" {
+		return xdgConfigHome, nil
+	}
+	home := os.Getenv("HOME")
```
vrothberg, line 120: Should we use `Get()` and hence `unshare...` here?

vrothberg (commented):
"XDG_CONFIG_HOME" is not yet considered by `unshare`.

I still feel that the implementations vary to strongly. Any chance we can create a new `/internal` package to resolve the circular dep but consolidate the code?

pkg/homedir/homedir_unix.go
rhatdan: yes

pkg/homedir/homedir_unix.go
rhatdan: yes

rhatdan:
@vrothberg I think it is ready to go.
@giuseppe @nalind PTAL.

vrothberg (commented):
Just one thing is missing:

`unshare/HomeDir` needs to check `XDG_CONFIG_HOME` as well. The order should be:
1. XDG_
2. HOME
3. User look up

rhatdan:
I don't believe so XDG_CONFIG_HOME  and HOME are not the same.

```
    XDG_CONFIG_HOME
        Where user-specific configurations should be written (analogous to /etc).
        Should default to $HOME/.config.
```

vrothberg:
@rhatdan, you're correct. I might have ran short on caffeine...

LGTM

TomSweeneyRedHat:
@rhatdan needs a rebase....

giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #768, opened 2020-11-03
Need to load nsswitch based config outside of container

rhatdan:
Fix CVE-2019-14271 loading of nsswitch based config inside chroot under Glibc

Based on code from moby

https://github.com/moby/moby/pull/39612/files

### DIFF
--- a/pkg/chrootarchive/archive.go
+++ b/pkg/chrootarchive/archive.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"os"
+	"os/user"
 	"path/filepath"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


nalind:
When we have a parent and child process, which is how chrootarchive functions, which process are we trying to protect, and how does this change do that?  Can that information be mentioned in the commit log message?

rhatdan:
According to the CVE, 

In Docker 19.03.x before 19.03.1 linked against the GNU C Library (aka glibc), code injection can occur when the nsswitch facility dynamically loads a library inside a chroot that contains the contents of the container.

So the question would be are we in the chroot when the archive is first used.

giuseppe (approved):
LGTM

vrothberg (commented):
LGTM

PR #769, opened 2020-11-09
Bump go-winio to v0.4.15

saschagrunert:
The `go-winio/archive/tar` package has been removed, which means we now
can use the native `archive/tar` package for Windows.


rhatdan:
NICE
LGTM

PR #772, opened 2020-11-12
overlay: set permissions on "diff" correctly

nalind:
When creating a new layer that has a parent, assign the new layer's "diff" directory the same permissions as the parent layer's "diff" directory, rather than the permissions of the parent of its parent layer's "diff" directory.

Before the recent round of changes, that directory would be initialized with 0700 permissions, so layers created by new code on top of layers created by the older version would be marked 0700, and were not usable by UIDs other than 0.

Spotted in CI tests for buildah that try to use overlay to create containers on top of base images that were imported using a slightly older version of podman, for example https://cirrus-ci.com/task/6024517489262592.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -593,20 +593,15 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 	if err := idtools.MkdirAllAs(path.Dir(dir), 0700, rootUID, rootGID); err != nil {
 		return err
 	}
-	perms := defaultPerms
-	if d.options.forceMask != nil {
-		perms = *d.options.forceMask
-	}
 	if parent != "" {
 		st, err := system.Stat(d.dir(parent))
 		if err != nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@giuseppe @vrothberg @saschagrunert @TomSweeneyRedHat PTAL

vrothberg (commented):
LGTM

saschagrunert (approved):
LGTM

drivers/overlay/overlay.go
```diff
@@ -631,6 +626,18 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 		}
 	}
 
+	perms := defaultPerms
+	if d.options.forceMask != nil {
```
giuseppe, line 630: I think this should have precedence and override the parent mode

nalind: Hmm, yeah, you're right.  Changing it.

PR #775, opened 2020-11-19
Enable zstd:chunked support in containers/image

giuseppe:
a bunch of new APIs to enable zstd:chunked support in containers/image.

add a new custom variant of the zstd compression that permits to retrieve each file separately.

The idea is based on CRFS and its stargz format for having seekable and indexable tarballs.

zstd_chunked

One disadvantage of the stargz format is that a custom file is added to the tarball to store the layer metadata, and the metadata file is part of the image itself.  Clients that are not aware of the stargz format will propagate the metadata file inside of the containers.
    
The zstd compression supports embedding additional data as part of the stream that the zstd decompressor will ignore (skippable frame), so the issue above with CRFS can be solved directly within the zstd compression format.

Beside this minor advantage, zstd is much faster and compresses better than gzip, so take this opportunity to push the zstd format further.

The zstd compression is supported by the OCI image specs since August 2019: https://github.com/opencontainers/image-spec/pull/788 and has been supported by containers/image since then.

Clients that are not aware of the zstd:chunked format, won't notice any difference when handling a blob that uses the variant.

Since access to host files is required for the deduplication, the untar doesn't happen in a separate process running in a chroot, but it happens in original mount namespace.  To mitigate possible breakages, openat2(2) is used were available.

More details here: https://github.com/containers/image/pull/1084

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -705,6 +705,7 @@ func (d *Driver) Metadata(id string) (map[string]string, error) {
 // is being shutdown. For now, we just have to unmount the bind mounted
 // we had created.
 func (d *Driver) Cleanup() error {
+	_ = os.RemoveAll(d.getStagingDir())
 	return mount.Unmount(d.home)
 }
 
@@ -1490,6 +1491,10 @@ func (f fileGetNilCloser) Close() error {
 	return nil
 }
 
+func (d *Driver) getStagingDir() string {
+	return filepath.Join(d.home, "staging")
+}
+
 // DiffGetter returns a FileGetCloser that can read files from the directory that
 // contains files for the layer differences. Used for direct access for tar-split.
 func (d *Driver) DiffGetter(id string) (graphdriver.FileGetCloser, error) {
@@ -1498,6 +1503,75 @@ func (d *Driver) DiffGetter(id string) (graphdriver.FileGetCloser, error) {
 		return nil, err
 	}
 	return fileGetNilCloser{storage.NewPathFileGetter(p)}, nil
+}
+
+// CleanupStagingDirectory cleanups the staging directory.
+func (d *Driver) CleanupStagingDirectory(stagingDirectory string) error {
+	return os.RemoveAll(stagingDirectory)
+}
+
+// ApplyDiff applies the changes in the new layer using the specified function
+func (d *Driver) ApplyDiffWithDiffer(id, parent string, options *graphdriver.ApplyDiffOpts, differ graphdriver.Differ) (output graphdriver.DriverWithDifferOutput, err error) {
+	var idMappings *idtools.IDMappings
+	if options != nil {
+		idMappings = options.Mappings
+	}
+	if idMappings == nil {
+		idMappings = &idtools.IDMappings{}
+	}
+
+	applyDir := ""
+
+	if id == "" {
+		err := os.MkdirAll(d.getStagingDir(), 0700)
+		if err != nil && !os.IsExist(err) {
+			return graphdriver.DriverWithDifferOutput{}, err
+		}
+		applyDir, err = ioutil.TempDir(d.getStagingDir(), "")
+		if err != nil {
+			return graphdriver.DriverWithDifferOutput{}, err
+		}
+
+	} else {
+		var err error
+		applyDir, err = d.getDiffPath(id)
+		if err != nil {
+			return graphdriver.DriverWithDifferOutput{}, err
+		}
+	}
+
+	logrus.Debugf("Applying differ in %s", applyDir)
+
+	out, err := differ.ApplyDiff(applyDir, &archive.TarOptions{
+		UIDMaps:           idMappings.UIDs(),
+		GIDMaps:           idMappings.GIDs(),
+		IgnoreChownErrors: d.options.ignoreChownErrors,
+		WhiteoutFormat:    d.getWhiteoutFormat(),
+		InUserNS:          rsystem.RunningInUserNS(),
+	})
+	out.Target = applyDir
+	return out, err
+}
+
+// ApplyDiffFromStagingDirectory applies the changes using the specified staging directory.
+func (d *Driver) ApplyDiffFromStagingDirectory(id, parent, stagingDirectory string, diffOutput *graphdriver.DriverWithDifferOutput, options *graphdriver.ApplyDiffOpts) error {
+	if filepath.Dir(stagingDirectory) != d.getStagingDir() {
+		return fmt.Errorf("%q is not a staging directory", stagingDirectory)
+	}
+
+	diff, err := d.getDiffPath(id)
+	if err != nil {
+		return err
+	}
+	if err := os.RemoveAll(diff); err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	return os.Rename(stagingDirectory, diff)
+}
+
+// DifferTarget gets the location where files are stored for the layer.
+func (d *Driver) DifferTarget(id string) (string, error) {
+	return d.getDiffPath(id)
 }
 
 // ApplyDiff applies the new layer into a root
--- a/pkg/chunked/compression.go
+++ b/pkg/chunked/compression.go
@@ -1,0 +1,513 @@
+package chunked
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/binary"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"time"
+
+	"github.com/containers/storage/pkg/ioutils"
+	"github.com/klauspost/compress/zstd"
+	digest "github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
+	"github.com/vbatts/tar-split/archive/tar"
+)
+
+type zstdTOC struct {
+	Version int                `json:"version"`
+	Entries []zstdFileMetadata `json:"entries"`
+}
+
+type zstdFileMetadata struct {
+	Type       string            `json:"type"`
+	Name       string            `json:"name"`
+	Linkname   string            `json:"linkName,omitempty"`
+	Mode       int64             `json:"mode,omitempty"`
+	Size       int64             `json:"size"`
+	UID        int               `json:"uid"`
+	GID        int               `json:"gid"`
+	ModTime    time.Time         `json:"modtime"`
+	AccessTime time.Time         `json:"accesstime"`
+	ChangeTime time.Time         `json:"changetime"`
+	Devmajor   int64             `json:"devMajor"`
+	Devminor   int64             `json:"devMinor"`
+	Xattrs     map[string]string `json:"xattrs,omitempty"`
+	Digest     string            `json:"digest,omitempty"`
+	Offset     int64             `json:"offset,omitempty"`
+	EndOffset  int64             `json:"endOffset,omitempty"`
+
+	// Currently chunking is not supported.
+	ChunkSize   int64  `json:"chunkSize,omitempty"`
+	ChunkOffset int64  `json:"chunkOffset,omitempty"`
+	ChunkDigest string `json:"chunkDigest,omitempty"`
+}
+
+const (
+	TypeReg     = "reg"
+	TypeChunk   = "chunk"
+	TypeLink    = "hardlink"
+	TypeChar    = "char"
+	TypeBlock   = "block"
+	TypeDir     = "dir"
+	TypeFifo    = "fifo"
+	TypeSymlink = "symlink"
+)
+
+var tarTypes = map[byte]string{
+	tar.TypeReg:     TypeReg,
+	tar.TypeRegA:    TypeReg,
+	tar.TypeLink:    TypeLink,
+	tar.TypeChar:    TypeChar,
+	tar.TypeBlock:   TypeBlock,
+	tar.TypeDir:     TypeDir,
+	tar.TypeFifo:    TypeFifo,
+	tar.TypeSymlink: TypeSymlink,
+}
+
+var typesToTar = map[string]byte{
+	TypeReg:     tar.TypeReg,
+	TypeLink:    tar.TypeLink,
+	TypeChar:    tar.TypeChar,
+	TypeBlock:   tar.TypeBlock,
+	TypeDir:     tar.TypeDir,
+	TypeFifo:    tar.TypeFifo,
+	TypeSymlink: tar.TypeSymlink,
+}
+
+func getType(t byte) (string, error) {
+	r, found := tarTypes[t]
+	if !found {
+		return "", fmt.Errorf("unknown tarball type: %v", t)
+	}
+	return r, nil
+}
+
+func typeToTarType(t string) (byte, error) {
+	r, found := typesToTar[t]
+	if !found {
+		return 0, fmt.Errorf("unknown type: %v", t)
+	}
+	return r, nil
+}
+
+const (
+	manifestChecksumKey = "io.containers.zstd-chunked.manifest-checksum"
+	manifestInfoKey     = "io.containers.zstd-chunked.manifest-position"
+
+	// manifestTypeCRFS is a manifest file compatible with the CRFS TOC file.
+	manifestTypeCRFS = 1
+
+	// footerSizeSupported is the footer size supported by this implementation.
+	// Newer versions of the image format might increase this value, so reject
+	// any version that is not supported.
+	footerSizeSupported = 40
+)
+
+var (
+	// when the zstd decoder encounters a skippable frame + 1 byte for the size, it
+	// will ignore it.
+	// https://tools.ietf.org/html/rfc8478#section-3.1.2
+	skippableFrameMagic = []byte{0x50, 0x2a, 0x4d, 0x18}
+
+	zstdChunkedFrameMagic = []byte{0x47, 0x6e, 0x55, 0x6c, 0x49, 0x6e, 0x55, 0x78}
+)
+
+func isZstdChunkedFrameMagic(data []byte) bool {
+	if len(data) < 8 {
+		return false
+	}
+	return bytes.Equal(zstdChunkedFrameMagic, data[:8])
+}
+
+// readZstdChunkedManifest reads the zstd:chunked manifest from the seekable stream blobStream.  The blob total size must
+// be specified.
+// This function uses the io.containers.zstd-chunked. annotations when specified.
+func readZstdChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, annotations map[string]string) ([]byte, error) {
+	footerSize := int64(footerSizeSupported)
+	if blobSize <= footerSize {
+		return nil, errors.New("blob too small")
+	}
+
+	manifestChecksumAnnotation := annotations[manifestChecksumKey]
+	if manifestChecksumAnnotation == "" {
+		return nil, fmt.Errorf("manifest checksum annotation %q not found", manifestChecksumKey)
+	}
+
+	var offset, length, lengthUncompressed, manifestType uint64
+
+	if offsetMetadata := annotations[manifestInfoKey]; offsetMetadata != "" {
+		if _, err := fmt.Sscanf(offsetMetadata, "%d:%d:%d:%d", &offset, &length, &lengthUncompressed, &manifestType); err != nil {
+			return nil, err
+		}
+	} else {
+		chunk := ImageSourceChunk{
+			Offset: uint64(blobSize - footerSize),
+			Length: uint64(footerSize),
+		}
+		parts, errs, err := blobStream.GetBlobAt([]ImageSourceChunk{chunk})
+		if err != nil {
+			return nil, err
+		}
+		var reader io.ReadCloser
+		select {
+		case r := <-parts:
+			reader = r
+		case err := <-errs:
+			return nil, err
+		}
+		footer := make([]byte, footerSize)
+		if _, err := io.ReadFull(reader, footer); err != nil {
+			return nil, err
+		}
+
+		offset = binary.LittleEndian.Uint64(footer[0:8])
+		length = binary.LittleEndian.Uint64(footer[8:16])
+		lengthUncompressed = binary.LittleEndian.Uint64(footer[16:24])
+		manifestType = binary.LittleEndian.Uint64(footer[24:32])
+		if !isZstdChunkedFrameMagic(footer[32:40]) {
+			return nil, errors.New("invalid magic number")
+		}
+	}
+
+	if manifestType != manifestTypeCRFS {
+		return nil, errors.New("invalid manifest type")
+	}
+
+	// set a reasonable limit
+	if length > (1<<20)*50 {
+		return nil, errors.New("manifest too big")
+	}
+	if lengthUncompressed > (1<<20)*50 {
+		return nil, errors.New("manifest too big")
+	}
+
+	chunk := ImageSourceChunk{
+		Offset: offset,
+		Length: length,
+	}
+
+	parts, errs, err := blobStream.GetBlobAt([]ImageSourceChunk{chunk})
+	if err != nil {
+		return nil, err
+	}
+	var reader io.ReadCloser
+	select {
+	case r := <-parts:
+		reader = r
+	case err := <-errs:
+		return nil, err
+	}
+
+	manifest := make([]byte, length)
+	if _, err := io.ReadFull(reader, manifest); err != nil {
+		return nil, err
+	}
+
+	manifestDigester := digest.Canonical.Digester()
+	manifestChecksum := manifestDigester.Hash()
+	if _, err := manifestChecksum.Write(manifest); err != nil {
+		return nil, err
+	}
+
+	d, err := digest.Parse(manifestChecksumAnnotation)
+	if err != nil {
+		return nil, err
+	}
+	if manifestDigester.Digest() != d {
+		return nil, errors.New("invalid manifest checksum")
+	}
+
+	decoder, err := zstd.NewReader(nil)
+	if err != nil {
+		return nil, err
+	}
+	defer decoder.Close()
+
+	b := make([]byte, 0, lengthUncompressed)
+	if decoded, err := decoder.DecodeAll(manifest, b); err == nil {
+		return decoded, nil
+	}
+
+	return manifest, nil
+}
+
+func appendZstdSkippableFrame(dest io.Writer, data []byte) error {
+	if _, err := dest.Write(skippableFrameMagic); err != nil {
+		return err
+	}
+
+	var size []byte = make([]byte, 4)
+	binary.LittleEndian.PutUint32(size, uint32(len(data)))
+	if _, err := dest.Write(size); err != nil {
+		return err
+	}
+	if _, err := dest.Write(data); err != nil {
+		return err
+	}
+	return nil
+}
+
+func writeZstdChunkedManifest(dest io.Writer, outMetadata map[string]string, offset uint64, metadata []zstdFileMetadata, level int) error {
+	// 8 is the size of the zstd skippable frame header + the frame size
+	manifestOffset := offset + 8
+
+	toc := zstdTOC{
+		Version: 1,
+		Entries: metadata,
+	}
+
+	// Generate the manifest
+	manifest, err := json.Marshal(toc)
+	if err != nil {
+		return err
+	}
+
+	var compressedBuffer bytes.Buffer
+	zstdWriter, err := zstdWriterWithLevel(&compressedBuffer, level)
+	if err != nil {
+		return err
+	}
+	if _, err := zstdWriter.Write(manifest); err != nil {
+		zstdWriter.Close()
+		return err
+	}
+	if err := zstdWriter.Close(); err != nil {
+		return err
+	}
+	compressedManifest := compressedBuffer.Bytes()
+
+	manifestDigester := digest.Canonical.Digester()
+	manifestChecksum := manifestDigester.Hash()
+	if _, err := manifestChecksum.Write(compressedManifest); err != nil {
+		return err
+	}
+
+	outMetadata[manifestChecksumKey] = manifestDigester.Digest().String()
+	outMetadata[manifestInfoKey] = fmt.Sprintf("%d:%d:%d:%d", manifestOffset, len(compressedManifest), len(manifest), manifestTypeCRFS)
+	if err := appendZstdSkippableFrame(dest, compressedManifest); err != nil {
+		return err
+	}
+
+	// Store the offset to the manifest and its size in LE order
+	var manifestDataLE []byte = make([]byte, footerSizeSupported)
+	binary.LittleEndian.PutUint64(manifestDataLE, manifestOffset)
+	binary.LittleEndian.PutUint64(manifestDataLE[8:], uint64(len(compressedManifest)))
+	binary.LittleEndian.PutUint64(manifestDataLE[16:], uint64(len(manifest)))
+	binary.LittleEndian.PutUint64(manifestDataLE[24:], uint64(manifestTypeCRFS))
+	copy(manifestDataLE[32:], zstdChunkedFrameMagic)
+
+	return appendZstdSkippableFrame(dest, manifestDataLE)
+}
+
+func writeZstdChunkedStream(destFile io.Writer, outMetadata map[string]string, reader io.Reader, level int) error {
+	// total written so far.  Used to retrieve partial offsets in the file
+	dest := ioutils.NewWriteCounter(destFile)
+
+	tr := tar.NewReader(reader)
+	tr.RawAccounting = true
+
+	buf := make([]byte, 4096)
+
+	zstdWriter, err := zstdWriterWithLevel(dest, level)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if zstdWriter != nil {
+			zstdWriter.Close()
+			zstdWriter.Flush()
+		}
+	}()
+
+	restartCompression := func() (int64, error) {
+		var offset int64
+		if zstdWriter != nil {
+			if err := zstdWriter.Close(); err != nil {
+				return 0, err
+			}
+			if err := zstdWriter.Flush(); err != nil {
+				return 0, err
+			}
+			offset = dest.Count
+			zstdWriter.Reset(dest)
+		}
+		return offset, nil
+	}
+
+	var metadata []zstdFileMetadata
+	for {
+		hdr, err := tr.Next()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return err
+		}
+
+		rawBytes := tr.RawBytes()
+		if _, err := zstdWriter.Write(rawBytes); err != nil {
+			return err
+		}
+		payloadDigester := digest.Canonical.Digester()
+		payloadChecksum := payloadDigester.Hash()
+
+		payloadDest := io.MultiWriter(payloadChecksum, zstdWriter)
+
+		// Now handle the payload, if any
+		var startOffset, endOffset int64
+		checksum := ""
+		for {
+			read, errRead := tr.Read(buf)
+			if errRead != nil && errRead != io.EOF {
+				return err
+			}
+
+			// restart the compression only if there is
+			// a payload.
+			if read > 0 {
+				if startOffset == 0 {
+					startOffset, err = restartCompression()
+					if err != nil {
+						return err
+					}
+				}
+				_, err := payloadDest.Write(buf[:read])
+				if err != nil {
+					return err
+				}
+			}
+			if errRead == io.EOF {
+				if startOffset > 0 {
+					endOffset, err = restartCompression()
+					if err != nil {
+						return err
+					}
+					checksum = payloadDigester.Digest().String()
+				}
+				break
+			}
+		}
+
+		typ, err := getType(hdr.Typeflag)
+		if err != nil {
+			return err
+		}
+		xattrs := make(map[string]string)
+		for k, v := range hdr.Xattrs {
+			xattrs[k] = base64.StdEncoding.EncodeToString([]byte(v))
+		}
+		m := zstdFileMetadata{
+			Type:       typ,
+			Name:       hdr.Name,
+			Linkname:   hdr.Linkname,
+			Mode:       hdr.Mode,
+			Size:       hdr.Size,
+			UID:        hdr.Uid,
+			GID:        hdr.Gid,
+			ModTime:    hdr.ModTime,
+			AccessTime: hdr.AccessTime,
+			ChangeTime: hdr.ChangeTime,
+			Devmajor:   hdr.Devmajor,
+			Devminor:   hdr.Devminor,
+			Xattrs:     xattrs,
+			Digest:     checksum,
+			Offset:     startOffset,
+			EndOffset:  endOffset,
+
+			// ChunkSize is 0 for the last chunk
+			ChunkSize:   0,
+			ChunkOffset: 0,
+			ChunkDigest: checksum,
+		}
+		metadata = append(metadata, m)
+	}
+
+	rawBytes := tr.RawBytes()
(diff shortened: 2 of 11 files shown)


rhatdan:
Lint is not happy.

giuseppe:
could I get some feedbacks on the containers/storage side (especially about the new interfaces)?

The current implementation is feature completed, with some parts that can be added later (especially for the host files deduplication)

rhatdan:
If openat is not available, will we fall back to old method, not do the deduplicating?

giuseppe:
> If openat is not available, will we fall back to old method, not do the deduplicating?

it will still deduplicate if the underlying file system supports reflinks.  `openat2`, when available, is used for safely creating the files, otherwise it uses a less secure (and slower) fallback.

rhatdan:
@mrunalp @mtrmac @vrothberg @nalind PTAL

rhatdan:
@giuseppe What do you think about splitting out some of these PRs into separate easier to merge PRs?

giuseppe:
> @giuseppe What do you think about splitting out some of these PRs into separate easier to merge PRs?

the earlier patches are a preparation for the zstd:chunked implementation.  I am fine to split these to smaller PRs but I think it is useful first to get some feedbacks on the overall design as we might need to change the preparatory patches before they are merged

giuseppe:
moved some patches to a new PR: https://github.com/containers/storage/pull/787

pkg/zstdchunked/compression.go
```diff
@@ -0,0 +1,486 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	manifest := make([]byte, length)
+	if _, err := io.ReadFull(reader, manifest); err != nil && err != io.EOF {
+		return nil, err
+	}
+
+	manifestDigester := digest.Canonical.Digester()
```
mtrmac, line 182: This should use the algorithm from the expected digest value, instead of hard-coding `digest.Canonical`.

Also elsewhere.

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+}
+
+// findFileOnTheHost checks whether the requested file already exist on the host and copies the file content from there if possible.
+// It is currently implemented to look only at the file with the same path.  Ideally it can detect the same content also at different
+// paths.
+func findFileOnTheHost(file FileMetadata, root string, dirfd int, missingDirsMode, mode os.FileMode) (*os.File, int64, error) {
+	hostContentDigester := digest.Canonical.Digester()
+	hostContentChecksum := hostContentDigester.Hash()
+
+	sourceFile := filepath.Clean(filepath.Join("/", file.Name))
```
mtrmac, line 202: Note that this does not detect symlinks at all, e.g. `/usr/tmp`.

The whole idea of using host files still worries me quite a bit.

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+	manifestChecksum, err := digest.Parse(file.Digest)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	if checksum != manifestChecksum {
+		return nil, 0, nil
+	}
+
+	dstFile, written, err := copyFileContent(fmt.Sprintf("/proc/self/fd/%d", fd), file.Name, root, dirfd, missingDirsMode, mode)
```
mtrmac, line 239: This is Linux-only; and all it would need to become portable is to rewind and use `f`, wouldn’t it?

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+		if err != nil {
+			return err
+		}
+		if err := unix.Fsetxattr(fd, k, data, 0); err != nil {
+			return err
+		}
+	}
+
+	ts := []syscall.Timespec{timeToTimespec(metadata.AccessTime), timeToTimespec(metadata.ModTime)}
+	if _, _, err := unix.Syscall6(unix.SYS_UTIMENSAT, uintptr(fd), 0, uintptr(unsafe.Pointer(&ts[0])), 0, 0, 0); err != 0 && err != unix.ENOSYS {
```
mtrmac, line 419: Portability to non-Linux?

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+		return err
+	}
+
+	// Files must be already created with the correct mode, but we must make sure the setuid/setgid
+	// is set.
```
mtrmac, line 424: This is risky; if the file is intended to be 0770 special-user:special-group, creating files with the expected mode creates it 0770 root:current-group, possibly making it accessible to unintended users. Create files with mode 000, Chown first, _then_ Chmod.

(In general, fair bit of the extraction logic is not `zstdchunked`-specific. Couldn’t this reuse, or plug into, some existing implementation?)

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+		Flags:   flags,
+		Mode:    uint64(mode & 07777),
+		Resolve: unix.RESOLVE_IN_ROOT,
+	}
+
+	fd, err := unix.Openat2(dirfd, name, &how)
+	if err == nil {
+		return os.NewFile(uintptr(fd), name), nil
+	}
+
```
mtrmac, line 444: What’s the security guarantee this implementation provides? Whatever `unix.RESOLVE_IN_ROOT` does is essentially ignored if failures are just bypassed. Are the two implementations intended to be equivalent? Even if they are, a better-targeted error handling (just on `ENOSYS` or something like that) might be more robust.

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+	// multipart response
+	boundary, found := params["boundary"]
+	if !found {
+		return errors.Errorf("boundary param not found")
+	}
+
+	buffered := makeBufferedNetworkReader(stream, 64, 16384)
+	defer buffered.Close()
```
mtrmac, line 522: What does this buffered network reader do? Isn’t it just an expensive way to maintain kernel-side socket buffers?

It is non-obvious to me why it is worth the complexity — and if it is, why aren’t we using it elsewhere / why isn’t it a part of the standard library? What’s special about this case?

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+}
+
+type hardLinkToCreate struct {
+	dest     string
+	dirfd    int
+	mode     os.FileMode
+	metadata *FileMetadata
+}
+
+func (d *ChunkedZstdDiffer) Diff(dest string, options *archive.TarOptions) (graphdriver.DriverWithDifferOutput, error) {
```
mtrmac, line 780: Isn’t this closer to `ApplyDiff` than to `Diff`, using the existing driver terminology? (Either way it might need a unique name)

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+				mode:     mode,
+				metadata: &r,
+			})
+			continue
+		case tar.TypeSymlink:
+			if err := safeSymlink(dest, dirfd, mode, &r, options); err != nil {
+				return output, err
+			}
+			continue
+		}
```
mtrmac, line 912: `default: /* fail */ ` ?

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+		return err
+	}
+	if digester.Digest() != manifestChecksum {
+		return fmt.Errorf("checksum mismatch for %q", dest)
+	}
+	return setFileAttrs(file, mode, metadata, options)
+}
+
+func storeMissingFiles(stream io.Reader, contentType string, dest string, dirfd int, missingChunks []missingChunk, missingDirsMode os.FileMode, options *archive.TarOptions) error {
+	mediaType, params, err := mime.ParseMediaType(contentType)
```
mtrmac, line 488: Why does the result of `GetBlobAt` even _have_ a MIME type? It is surprisingly asymmetrical that the input to `GetBlobAt` is a generic []chunk, but the output is a single stream with an implied HTTP-like format that the caller must somehow parse. I’d expect one layer / piece of code to handle the formatting of requests + parsing of responses, so that the two are consistent, and abstracted (or not) at the same level.

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,977 @@
...
+	if err != nil {
+		return err
+	}
+	if !strings.HasPrefix(mediaType, "multipart/") {
+		if len(missingChunks) != 1 {
+			return errors.Errorf("invalid response, no multipart for multiple chunks requested")
+		}
+
+		for _, mf := range missingChunks[0].Files {
```
mtrmac, line 497: This loop can probably be factored out to avoid duplication.

mtrmac (commented):
A single pass, as someone unfamiliar with the details of c/storage.

While I’m happy that the file extraction code was moved out of c/image, it still feels like quite a lot of copy&paste of non-trivial code that would best be maintained in a single place. I am absolutely unqualified to recommend an improvement, though.

pkg/zstdchunked/storage.go
giuseppe: fixed

pkg/zstdchunked/storage.go
giuseppe: yes, I'll rename it

pkg/zstdchunked/storage.go
giuseppe: not sure if it can be useful elsewhere, but it makes a significant difference here so we can keep fetching data while it is being processed at the same time.  From what I could see this was a bottleneck when pulling an image from a remote registry.

pkg/zstdchunked/compression.go
```diff
@@ -0,0 +1,486 @@
...
+		Length: length,
+	}
+
+	reader, _, err := blobStream.GetBlobAt([]ImageSourceChunk{chunk})
+	if err != nil {
+		return nil, err
+	}
+
+	manifest := make([]byte, length)
+	if _, err := io.ReadFull(reader, manifest); err != nil && err != io.EOF {
```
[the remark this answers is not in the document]
giuseppe: can't remember why I needed it.  Let me drop it for now

giuseppe:
I've addressed some of the changes for now, I'll take a look at the other comments

pkg/zstdchunked/compression.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+
+	manifestDigester := digest.Canonical.Digester()
+	manifestChecksum := manifestDigester.Hash()
+	if _, err := manifestChecksum.Write(manifest); err != nil {
+		return nil, err
+	}
+
+	if manifestChecksumAnnotation := annotations[manifestChecksumKey]; manifestChecksumAnnotation == "" {
+		logrus.Warningf("Manifest checksum annotation not found.  Skipping manifest validation")
```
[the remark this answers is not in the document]
giuseppe: fixed

pkg/zstdchunked/storage.go
giuseppe: thanks for spotting it.  I am going to change the logic to always create dirs/files with mode 000

pkg/zstdchunked/storage.go
giuseppe: I was not sure whether we should support old kernels.

I've dropped the unsafe fallback and require `openat2` to be present.  It also helps to avoid another dependency

pkg/zstdchunked/storage.go
giuseppe: Since we require openat2 to be present, it won't anyway work on other systems.

pkg/zstdchunked/compression.go
```diff
@@ -0,0 +1,477 @@
+package zstdchunked
```
ktock, line 1: Can we separate out this package into a separated module? From the perspective of the pkg consumer, it would be great if they can avoid having dependency to the entier c/storage project (contians drivers, etc.).

And based on experiment with zstd:chunked at , I propose to add prefetch support based on the logic of eStargz (https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/docs/stargz-estargz.md).

ktock:
And I implemented lazy pulling of zstd:chunked! 
https://github.com/containerd/stargz-snapshotter/pull/281

giuseppe:
> And I implemented lazy pulling of zstd:chunked!
> [containerd/stargz-snapshotter#281](https://github.com/containerd/stargz-snapshotter/pull/281)

that's great news!  Thanks

pkg/zstdchunked/compression.go
giuseppe: yes, I don't think it should be an issue.

@rhatdan what do you think?

giuseppe: > And we should add prefetch support based on the logic of eStargz (https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/docs/stargz-estargz.md) for mitigating the overhead of on-demand fetching .

could we add a probabilistic score to the TOC for each file?  e.g.

```
    {
      "name": "bin/busybox",
      "type": "reg",
      "size": 833104,
      "modtime": "2019-06-12T17:52:45Z",
      "mode": 33261,
      "offset": 126,
      "NumLink": 0,
      "digest": "sha256:8b7c559b8cccca0d30d01bc4b5dc944766208a53d18a03aa8afe97252207521f",
      "chunkDigest": "sha256:8b7c559b8cccca0d30d01bc4b5dc944766208a53d18a03aa8afe97252207521f",
     "accessScore: "1.00"
    },
```

so that the client can fine tune what files must be prefetched?

From your experience with estargz, how much is it important that these most frequently accessed files are contiguous in that tarball?

ktock: Some registries doesn't support multiple Range Request so all prioritized files should be fetched by one range.

rhatdan:
@giuseppe Needs a rebase.

giuseppe:
> @giuseppe Needs a rebase.

rebased

pkg/zstdchunked/compression.go
giuseppe: thanks that makes sense.

My suggestion is to add an `accessScore` per file then have an optional pass that sorts the files in the tarball by their score.  In this way a client can decide the threshold to preload (e.g. `accessScore > 0.7`) then prefetch only the initial tarball part.
One disadvantage of such additional pass is that the diffID will be changed, so I think it should be optional so users can decide whether they want to keep the diffID or support registries that lack multiple Range Requests.

ktock: SGTM

rhatdan: I am all for breaking this into a separate REPO. Anything to make it easier to merge and move foreward.

pkg/zstdchunked/compression.go
```diff
@@ -0,0 +1,477 @@
...
+	ModTime    time.Time         `json:"modtime"`
+	AccessTime time.Time         `json:"accesstime"`
+	ChangeTime time.Time         `json:"changetime"`
+	Devmajor   int64             `json:"devMajor"`
+	Devminor   int64             `json:"devMinor"`
+	Xattrs     map[string]string `json:"xattrs,omitempty"`
+	Digest     string            `json:"digest,omitempty"`
+	Offset     int64             `json:"offset,omitempty"`
+	EndOffset  int64             `json:"endOffset,omitempty"`
+}
```
ktock, line 41: Why compression per file (not smaller chunk)? Accessing a small part of a big file ends up downloading the entire file contents?

giuseppe: yes this is something not supported yet

How does estargz do it?  Rolling checksum on the single file or does it split in a different way?

giuseppe:
moved some patches to https://github.com/containers/storage/pull/849 to simplify the review process and move forward

pkg/zstdchunked/compression.go
ktock: We have [`chunk` type of TOC Entry](https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/docs/stargz-estargz.md#the-structure). This stores the chunk's offset(absolute offset and the one relative to the top of the file), size and digest. https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/estargz/types.go#L162-L176

giuseppe: nevermind my question, I see how it is implemented.

I'll change the metadata to support this use case

giuseppe: I've added a patch (completely untested) to support files chunking.

Does the change LGTY?

ktock: Could we keep the compatibility with stargz/eStargz?

https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/estargz/types.go#L162-L176

giuseppe: do you have an example TOC file that uses chunks?

rhatdan:
@giuseppe Doesn't this need a rebase now, that the other PRs have been merged?

giuseppe:
> @giuseppe Doesn't this need a rebase now, that the other PRs have been merged?

rebased now

pkg/zstdchunked/compression.go
giuseppe: I'll drop the last commit and rewrite it to be compatible with stargz/eStargz, an example of TOC file that uses chunks would be helpful to understand better how these fields are used

ktock: Thanks! 
`sha256:cac892ca7967dfb6bd2433a1c35a5a89c1fe5373725fee80b430c8390f5171cb` of [`ghcr.io/stargz-containers/python:3.9-esgz`](https://github.com/orgs/stargz-containers/packages/container/python/911988) has "chunk" entries (please see also [our doc](https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/docs/pre-converted-images.md) for other pre-converted estargz images, if needed).

And the logic of creating TOC from tar will also help understand it: https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/estargz/estargz.go#L598-L727

---

Excerpt from TOC of `sha256:cac892ca7967dfb6bd2433a1c35a5a89c1fe5373725fee80b430c8390f5171cb`.

```json
		{
			"name": "usr/lib/gcc/x86_64-linux-gnu/8/cc1",
			"type": "reg",
			"size": 25064392,
			"modtime": "2019-04-06T14:44:55Z",
			"mode": 33261,
			"offset": 20405393,
			"NumLink": 0,
			"digest": "sha256:a0993fa7e47d8bc6abf9ec39e10cebae8929ce30b035482c70733e7effa82f39",
			"chunkSize": 4194304,
			"chunkDigest": "sha256:cf1681301b9618990b859a8a74f9fbad1740d7950ace15d6bb03609339aac7a0"
		},
		{
			"name": "usr/lib/gcc/x86_64-linux-gnu/8/cc1",
			"type": "chunk",
			"offset": 22566274,
			"NumLink": 0,
			"chunkOffset": 4194304,
			"chunkSize": 4194304,
			"chunkDigest": "sha256:981f639fb1af5031e9e38ebf2d7dd23bfcc72bbce676f8cd7d694dc12925071b"
		},
		{
			"name": "usr/lib/gcc/x86_64-linux-gnu/8/cc1",
			"type": "chunk",
			"offset": 24537576,
			"NumLink": 0,
			"chunkOffset": 8388608,
			"chunkSize": 4194304,
			"chunkDigest": "sha256:3ed7218ce02b8228487607c0039f076158062c706477d2e3a7cfe626efcf51df"
		},
...
```

giuseppe: thanks.

I see that EndOffset is not used.  Is it set implicitly to the next Offset?

pkg/zstdchunked/compression.go
```diff
@@ -0,0 +1,477 @@
...
+	return bytes.Equal(zstdChunkedFrameMagic, data[:8])
+}
+
+// ReadZstdChunkedManifest reads the zstd:chunked manifest from the seekable stream blobStream.  The blob total size must
+// be specified.
+// This function uses the io.containers.zstd-chunked. annotations when specified.
+func ReadZstdChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, annotations map[string]string) ([]byte, error) {
+	footerSize := int64(footerSizeSupported)
+	if blobSize <= footerSize {
+		return nil, fmt.Errorf("blob too small")
```
kolyshkin, line 120: nit: `errors.New`

pkg/zstdchunked/compression.go
```diff
@@ -0,0 +1,477 @@
...
+		if _, err := io.ReadFull(reader, footer); err != nil {
+			return nil, err
+		}
+
+		offset = binary.LittleEndian.Uint64(footer[0:8])
+		length = binary.LittleEndian.Uint64(footer[8:16])
+		lengthUncompressed = binary.LittleEndian.Uint64(footer[16:24])
+		manifestType = binary.LittleEndian.Uint64(footer[24:32])
+		if !isZstdChunkedFrameMagic(footer[32:40]) {
+			return nil, fmt.Errorf("invalid magic number")
```
kolyshkin, line 148: nit: `errors.New`

pkg/zstdchunked/compression.go
```diff
@@ -0,0 +1,477 @@
...
+
+	manifestDigester := digest.Canonical.Digester()
+	manifestChecksum := manifestDigester.Hash()
+	if _, err := manifestChecksum.Write(manifest); err != nil {
+		return nil, err
+	}
+
+	manifestChecksumAnnotation := annotations[manifestChecksumKey]
+	if manifestChecksumAnnotation == "" {
+		return nil, fmt.Errorf("manifest checksum annotation %q not found", manifestChecksumKey)
```
kolyshkin, line 187: I'd move this check earlier in the code just so we can fail early in case this is missing

pkg/zstdchunked/compression.go
```diff
@@ -0,0 +1,477 @@
...
+// MANIFEST_OFFSET, MANIFEST_LENGTH, MANIFEST_LENGTH_UNCOMPRESSED and CHUNKED_ZSTD_MAGIC_NUMBER are 64 bits unsigned in little endian format.
+func zstdChunkedWriterWithLevel(out io.Writer, metadata map[string]string, level int) (io.WriteCloser, error) {
+	ch := make(chan error, 1)
+	r, w := io.Pipe()
+
+	go func() {
+		defer close(ch)
+		defer r.Close()
+		defer io.Copy(ioutil.Discard, r)
+		ch <- writeZstdChunkedStream(out, metadata, r, level)
```
kolyshkin, line 465: Maybe I'm missing something, but using defers here does not look justified, as it's not any better than
```go
ch <- writeZstdChunkedStream(out, metadata, r, level)
io.Copy(ioutil.Discard, r)
r.Close()
close(ch)
```

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,983 @@
...
+	}
+	layersMetadata, layersTarget, err := getLayersCache(store)
+	if err != nil {
+		return nil, err
+	}
+
+	return &ChunkedZstdDiffer{
+		stream:         iss,
+		manifest:       manifest,
+		ctx:            ctx,
```
kolyshkin, line 161: golang context docs say

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

There's an ongoing discussion about relaxing that rule (https://github.com/golang/go/issues/22602) but so far it's still there.

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,983 @@
...
+		if err != nil {
+			return err
+		}
+		if err := unix.Fsetxattr(fd, k, data, 0); err != nil {
+			return err
+		}
+	}
+
+	ts := []syscall.Timespec{timeToTimespec(metadata.AccessTime), timeToTimespec(metadata.ModTime)}
+	if _, _, err := unix.Syscall6(unix.SYS_UTIMENSAT, uintptr(fd), 0, uintptr(unsafe.Pointer(&ts[0])), 0, 0, 0); err != 0 && err != unix.ENOSYS {
```
kolyshkin, line 427: Looks like you can use
```go
unix.UtimesNanoAt(fd, "", ts, 0)
```
instead

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,983 @@
...
+		uid, gid, mode, err := archive.GetFileOwner(dest)
+		if err == nil {
+			value := fmt.Sprintf("%d:%d:0%o", uid, gid, mode)
+			if err := unix.Setxattr(dest, containersOverrideXattr, []byte(value), 0); err != nil {
+				return output, err
+			}
+		}
+	}
+
+	dirfd, err := unix.Open(dest, unix.O_RDONLY, 0)
```
kolyshkin, line 813: Maybe add `O_PATH` (I haven't reviewed all the places where `dirfd` is used so not sure).

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,983 @@
...
+		return err
+	}
+	if err := stream.Close(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func safeMkdir(target string, dirfd int, mode os.FileMode, metadata *FileMetadata, options *archive.TarOptions) error {
```
kolyshkin, line 638: Maybe it makes sense to make a nice little package out of it (openat2 and friends).

E.g. `github.com/containers/storage/pkg/safe` (or `secure`) with methods like `Open`, `Mkdir` etc.

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,983 @@
...
+	parent := filepath.Dir(metadata.Name)
+	base := filepath.Base(metadata.Name)
+
+	parentFd := dirfd
+	if parent != "." {
+		parentFile, err := openFileUnderRoot(parent, target, dirfd, unix.O_DIRECTORY|unix.O_RDONLY, 0)
```
kolyshkin, line 644: add `O_PATH`?

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,983 @@
...
+		defer dir.Close()
+
+		dirfd = int(dir.Fd())
+	}
+
+	return unix.Mknodat(dirfd, base, mode, dev)
+}
+
+func checkChownErr(err error, name string, uid, gid int) error {
+	if e, ok := err.(*os.PathError); ok && e.Err == syscall.EINVAL {
```
kolyshkin, line 748: `if errors.Is(err, syscall.EINVAL) {`

pkg/zstdchunked/storage.go
```diff
@@ -0,0 +1,983 @@
...
+			continue
+		}
+		return dstFile, written, nil
+	}
+	return nil, 0, nil
+}
+
+// findFileOnTheHost checks whether the requested file already exist on the host and copies the file content from there if possible.
+// It is currently implemented to look only at the file with the same path.  Ideally it can detect the same content also at different
+// paths.
```
kolyshkin, line 205: Not really related to this PR, but in OpenVZ we used to have a filesystem which
 - maintained file digest in xattrs (recalculated and updated by the kernel on close() after write);
 - in read-only open() or mmap(), looked for the file with the same checksum in "master" directory and served it from the master.
 
It's a bit more complicated than that (a daemon which create "master" files, copy-on-write etc) but this helped a lot with container density.

pkg/zstdchunked/compression.go
ktock: >  Is it set implicitly to the next Offset?

Yes, it is.

https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/estargz/estargz.go#L417-L421

pkg/zstdchunked/storage.go
giuseppe: thanks, and it seems ChunkedZstdDiffer.ctx isn't even used, so I am just  going to drop it

pkg/zstdchunked/storage.go
giuseppe: I'll definitely refactor some of this code in a separate package, as some features are not really zstd related

pkg/zstdchunked/storage.go
giuseppe: I thought about improving `mlocate` to also store the file content digest on `updatedb`, but that would be a breaking change in its DB.

giuseppe:
I took care of the comments above and pushed a new version

pkg/zstdchunked/compression.go
giuseppe: thanks!

Are you fine with the additional metadata I've added?

ktock: Thanks! LGTM

giuseppe:
I don't think https://github.com/opencontainers/image-spec/issues/803 will move forward any time soon, so IMO we should move forward without it.

I can make clearer that this feature is experimental and breaks interoperability with other container engines, but at least we can start playing with it and see how much it can help with the local storage of images.

rhatdan:
SGTM

drivers/overlay/overlay.go
```diff
@@ -1182,13 +1182,74 @@ func (f fileGetNilCloser) Close() error {
...
+	}
+
+	applyDir := ""
+
+	if id == "" {
+		err := os.MkdirAll(d.getStagingDir(), 0700)
+		if err != nil && !os.IsExist(err) {
+			return graphdriver.DriverWithDifferOutput{}, err
+		}
+		applyDir, err = ioutil.TempDir(d.getStagingDir(), "")
```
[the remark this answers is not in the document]
cgwalters: https://internals.rust-lang.org/t/should-rust-programs-unwind-on-sigint/13800/11

pkg/zstdchunked/storage.go
cgwalters: This is of course very similar to how ostree works, although the digest here covers the more of the inode and not just the contents (xref https://github.com/ostreedev/ostree/issues/2259).

pkg/zstdchunked/storage.go
cgwalters: Yeah, this seems like a good thing to do as a separate PR.

pkg/zstdchunked/storage.go
giuseppe: and ostree integration is definitely something I'd like to have in future!  And that was something I was planning already long time ago but never found the time to implement when I've added https://github.com/ostreedev/ostree/pull/1443

Moving containers/storage to store the checksum for each file will make it finally possible (and we'll need a better store than a list of JSON files to store the files and iterate each of them)

rhatdan:
LGTM

drivers/overlay/overlay.go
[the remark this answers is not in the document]
giuseppe: they are cleaned up at shutdown, so if anything is left in the staging directory it will be cleaned up at some point.

It could still be a problem for long running processes like CRI-O though.

pkg/chunked/compression.go
```diff
@@ -0,0 +1,513 @@
...
+func typeToTarType(t string) (byte, error) {
+	r, found := typesToTar[t]
+	if !found {
+		return 0, fmt.Errorf("unknown type: %v", t)
+	}
+	return r, nil
+}
+
+const (
+	manifestChecksumKey = "io.containers.zstd-chunked.manifest-checksum"
```
ktock, line 98: Could we store uncompressed digest of TOC manifest to this or have another annotation for it? (https://github.com/containerd/stargz-snapshotter/pull/293#issuecomment-815657316)

giuseppe: sure, I can add more metadata if it helps you.

How will it be different than `io.containers.zstd-chunked.manifest-checksum`?

cgwalters:
I have a much higher level comment: As we all know, the original Docker format was basically "the simplest thing that could possibly work": layers of tar and gzip.  What ended up being standardized is just tweaks on that.

This work addresses *some* of those issues (zstd is way better than gzip just as a baseline, seekability, etc.).  But one thing that IMO is so key about both ostree and git is that the format is *reproducible* - there is exactly one canonical representation of a git commit and an ostree commit (unlike tar with its many variants, extensions, ability to have files in different orders, overwriting earlier files with later, starting with `./` or just ``, etc.).  

If we're going to create effectively a new container format I think it's worth considering addressing that at the same time.  Which would here probably end up as something like a "canonical tar format" - and e.g. *require* that files in the input stream are sorted by name in UTF-8 order, etc.

ktock:
> Which would here probably end up as something like a "canonical tar format"

It's already discussed in "OCIv2" image format ([discussion notes](https://hackmd.io/@cyphar/ociv2-brainstorm#Canonical-Representation-Reproducible-Image-Building), [session youtube](https://www.youtube.com/watch?v=bbTxdzbjv7I) by @cypher) which doesn't use tar at all.
IMO we should keep the initial scope of zstd:chunked (i.e. lazy pulling).

rhatdan:
I am hesitant to push this too far, since if we end up breaking installed base of container registries and container engines we will hit a huge head wind, and prevent us from improving what we have. I don't mind a grand stategy for image format 2.0, but if we can make small improvements in 1.0, I would like to do this first.

cgwalters:
Honestly I think cyphar just went off the rails a bit here - it's likely all the goals could be achieved with a canonical tar representation and hence be compatible with all existing tar parsers.

pkg/chunked/compression.go
ktock: This is for compatibility with estargz. Our snapshotter verifies TOC manifest using its uncompressed digest so it would be great if zstd:chunked contains an annotation for it. But current `io.containers.zstd-chunked.manifest-checksum` seems to contain the digest of zstd-compressed TOC manifest, right?

Except for this annotation issue, zstdchunked image (`ghcr.io/ktock/python:3.9-zstdchunked`) created using `c/storage/pkg/chunked` works fine with containerd + stargz snapshotter!

cgwalters:
> That seems hard to do with compressed artifacts

Yes; we can really only care about the uncompressed version.  Clients would need to verify the uncompressed checksum and not the compressed one.

That said, after some reflection I think "canonical tar" could be done after this in theory - they're mostly orthogonal problems.

pkg/chunked/compression.go
ktock: In terms of the annotation name, it would be great if it's prefixed by `containerd.io/snapshot/` because it allows stargz snapshotter to support zstd:chunked without adding special patches to `containerd/containerd` project.

giuseppe:
I've marked all the APIs in containers/storage and containers/image as experimental, to have the freedom of introducing breaking changes until we are happy with it.  I am planning to do more incremental improvements later, as well as possibly support stargz/estargz as well for compatibility with clients that don't support zstd yet.

The reason why I am pushing to get this merged is that we can finally get users to play with it and get some feedback, and not expect users to deal with re-vendoring containers/storage and containers/image into the container tools.

We can still improve the new format and we add tar reproducibility later.  I think the tar-split blob can be embedded in the zstd stream, in the same way the TOC metadata is added right now.

giuseppe:
fedora-32 vfs fails to start, but I guess we can probably drop tests on fedora 32

rhatdan (approved):

Romain-Geissler-1A:
Hi,

FYI, I am speaking from a docker/podman user point of view.

If I understand correctly this proposal, this is something more or less similar to the eStargz proposal, but using zstd instead, to be able to inject metadata unknown to old client impactless, where currently eStargz would result in having some unexpected files injected inside the running container. Also, if I understand correctly, this allows to share on the client side parts of files which may be found in different and unrelated images, if these files share the same hash, and that is really cool ! While this proposal seems to be a very good improvement from the "pull" side, is it also an improvement for "push" scenario, and the storage on remote registry ? I mean, let's say a file content exists already in some image of a remote registry, if I build a local image uzing this zstd format, which contains a file with the same content, when I will run "podman push", will it push the full ztd tar file, or only the parts that the registry doesn't know yet ? And on registry side, will it be able to serve the "big" zstd file as a collection of smaller concatenatd zstd tar files and thus only store split zstd tar files rather than the "big" ones ?
If this pull requests also eventually allows to improve the "push" scenario, that is super cool !

Let me explain why we in Amadeus might be interested by such a proposal (as users). We have many applications which are C++ based. We do use (for now) mainly dynamic linking, and contrary to other companies our teams are very inter-dependant: we have several layer of C++ middleware teams developing diverse library, that are eventually reused by application developers. Our applications are quite big and tend to need many dependencies from various teams, compiled .so shared libraries are often 10MB files, up to sometimes 100+MB, even after being stripped. We have thousands of these libraries in total, big applications might require a lot of them. In production images, we install part of these libraries using "RUN dnf install some-amadeus-package", but the majority of these libraries are installed a simpler way, just untaring some tar.gz file in a new OCI layer.

To develop internally, we also use Docker images, so these come with GNU C/C++ based toolchains, which also are very heavy (a single toolchain containing both gcc + LLVM tools can easily weight 1GB in total). While we do offer one "standard SDK" base image where most tools are present, some team make the choice to develop their own image, writing their own Dockerfile. Most of our tools are available as RPM packages, so they naturally write "RUN dnf install some-amadeus-packages && dnf clean all".

The fact that we have many teams over dependant on each other and that each set of dependencies in every team is different results that in the end, we have many different SDK/production images in our registries, and layer's storage diff are very rarely shared between common images, despite we have *MANY* big files in common between all these various images. Anything based on "dnf" is doomed to fail to share layers as "dnf install" will result in having unique timestamsp, but not only (ie we can't just fix it by using "podman build --timestamp"), as "dnf" internally relies on a database, which will contain different package list for every team, thus look different from a filesystem point of view, thus leading to different tar.gz sha256.

So on Amadeus side, proposals that would allow to share not only "big" layer's filesystem diff tar.gz, but really individual file content part of these tar.gz (or tar.zstd) would definitely help us save a lot of disk space and network bandwidth to download/push just the "new" files. We are concerned by the registry side, as we do host our own internal JFrog Artifactory registry, and we regularly receive mails to clean all the docker images we don't use anymore as it fills up pretty quickly.

I am curious about this one, and about the mentioned OCIv2 as well ;)

Cheers,
Romain

giuseppe:
@mtrmac @cgwalters are you fine with the current status of this PR?  Or anything against moving this forward?

cgwalters:
In rpm-ostree we have `rpm-ostree ex <subcommand>` which is a place to ship commands that are marked experimental.  Does podman have anything like this?  Basically it creates a middle ground between the otherwise vast gulf separating "code in a PR" and "code shipped to vast numbers of people via distribution stable releases".

In particular as soon as this is vendored in podman for example, then it becomes effectively frozen in that we need to support the format ~forever right?  Whereas if it requires something like `echo enable-experimental-zstd-chunked: true >> ~/.containers/storage.conf` that's a different thing.

I only skimmed the design, and barely any of the code.

mtrmac:
> @mtrmac @cgwalters are you fine with the current status of this PR? Or anything against moving this forward?

- I am still seriously worried about host file reuse. Why is it known to be secure?
- Shouldn’t @nalind review this either way? (I’m not all that familiar with the codebase, and I have only quickly skimmed it once.)

giuseppe:
> As far as scanning for matching files in the Base OS @giuseppe has agreed to make this optional and just for /usr. We can turn it off by default, and then for certain workloads users could turn it on.

it is already disabled by default in the last version of these patches and it must be enabled in the `storage.conf` configuration file.



> In particular as soon as this is vendored in podman for example, then it becomes effectively frozen in that we need to support the format ~forever right? Whereas if it requires something like `echo enable-experimental-zstd-chunked: true >> ~/.containers/storage.conf` that's a different thing.

it won't be enabled by default yet, it must be explicitly configured (haven't decided yet if through the `storage.conf` config file or through some env variable)

giuseppe:
@mtrmac @nalind @rhatdan so fine to move forward with this PR (and considering the changes needed in c/image)?

rhatdan:
Since this is experiemental, I am going to merge, so we can move forward.

cgwalters:
I randomly came across https://github.com/dragonflyoss/image-service - not sure if you all have seen it.  May be worth a "Related projects" section somewhere?

PR #776, opened 2020-11-19
Fix unshare.HomeDir to use entry in /etc/passwd

rhatdan:
We were never using the entry in the /etc/passwd file
if $HOME was not set.

### DIFF
--- a/pkg/unshare/unshare.go
+++ b/pkg/unshare/unshare.go
@@ -26,6 +26,7 @@ func HomeDir() (string, error) {
 				return
 			}
 			homeDir, homeDirErr = usr.HomeDir, nil
+			return
 		}
 		homeDir, homeDirErr = home, nil
 	})


baude:
LGTM

giuseppe (approved):
LGTM

mheon:
LGTM

pkg/unshare/unshare.go
```diff
@@ -26,6 +26,7 @@ func HomeDir() (string, error) {
 				return
 			}
 			homeDir, homeDirErr = usr.HomeDir, nil
+			return
```
nalind, line 29: LGTM

pkg/unshare/unshare.go
```diff
@@ -26,6 +26,7 @@ func HomeDir() (string, error) {
 				return
 			}
 			homeDir, homeDirErr = usr.HomeDir, nil
```
TomSweeneyRedHat, line 28: Would it be better to change the variable homeDir to just home here?  That way it would work without the return and we could avoid someone accidentally removing the return later.

TomSweeneyRedHat:
one question and it looks like some test is unhappy.

rhatdan:
I know this is correct and the fedora repo is screwed up, so I am going to force merge and move this along.

PR #780, opened 2020-12-04
Fix rootless storage.conf handling

rhatdan:
This patch handles the expansion of environment variables.

Previous fix, triggered a test failure in podman, when using
environment variable expansion in the storage paths.


rhatdan:
@giuseppe @mheon PTAL

utils.go
```diff
@@ -244,40 +244,36 @@ func DefaultStoreOptionsAutoDetectUID() (StoreOptions, error) {
...
+		_, err := os.Stat(storageConf)
+		if err != nil && !os.IsNotExist(err) {
+			return storageOpts, err
+		}
+		if err == nil && !defaultConfigFileSet {
+			defaultRootlessRunRoot := storageOpts.RunRoot
+			defaultRootlessGraphRoot := storageOpts.GraphRoot
+			storageOpts = StoreOptions{}
+			reloadConfigurationFileIfNeeded(storageConf, &storageOpts)
+			if rootless && rootlessUID != 0 {
```
QiWang19, line 264: It seems this if block can not be reached.

rhatdan:
This was a mistake. I figured out how it is supposed to work.

PR #785, opened 2020-12-11
archive: Skip FIFO creation in userns

rhafer:
handleTarTypeBlockCharFifo() would skip the creation anyway, so let's return earlier and avoid later problems with failing chown/chmod calls.

Reproducer:
1. Configure ~/.config/containers/storage.conf to use the "vfs" driver for a non-root user
2. As non-root run: `podman pull docker.io/library/openjdk:14`

It will fail with:

```
ERRO[0048] Error while applying layer: ApplyLayer exit status 1 stdout:  stderr: lchown /dev/initctl: no such file or directory 
```

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -641,16 +641,10 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 		}
 		file.Close()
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@nalind @saschagrunert @vrothberg @giuseppe PTAL

giuseppe (approved):
/lgtm

PR #787, opened 2020-12-15
layers: support BigData

giuseppe:
similarly to containers and images, add support for storing big data
also for layers, so that it is possible to store arbitrary data when
it is not feasible to embed it in the layers JSON metadata file.

Subset of: https://github.com/containers/storage/pull/775
    
### DIFF
--- a/cmd/containers-storage/layer.go
+++ b/cmd/containers-storage/layer.go
@@ -3,11 +3,84 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/mflag"
 )
+
+var (
+	paramLayerDataFile = ""
+)
+
+func listLayerBigData(flags *mflag.FlagSet, action string, m storage.Store, args []string) int {
+	layer, err := m.Layer(args[0])
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%+v\n", err)
+		return 1
+	}
+	d, err := m.ListLayerBigData(layer.ID)
+	if jsonOutput {
+		json.NewEncoder(os.Stdout).Encode(d)
+	} else {
+		for _, name := range d {
+			fmt.Printf("%s\n", name)
+		}
+	}
+	return 0
+}
+
+func getLayerBigData(flags *mflag.FlagSet, action string, m storage.Store, args []string) int {
+	layer, err := m.Layer(args[0])
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%+v\n", err)
+		return 1
+	}
+	output := os.Stdout
+	if paramLayerDataFile != "" {
+		f, err := os.Create(paramLayerDataFile)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%v\n", err)
+			return 1
+		}
+		output = f
+	}
+	b, err := m.LayerBigData(layer.ID, args[1])
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%+v\n", err)
+		return 1
+	}
+	if _, err := io.Copy(output, b); err != nil {
+		fmt.Fprintf(os.Stderr, "%+v\n", err)
+		return 1
+	}
+	output.Close()
+	return 0
+}
+
+func setLayerBigData(flags *mflag.FlagSet, action string, m storage.Store, args []string) int {
+	layer, err := m.Layer(args[0])
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%+v\n", err)
+		return 1
+	}
+	input := os.Stdin
+	if paramLayerDataFile != "" {
+		f, err := os.Open(paramLayerDataFile)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%v\n", err)
+			return 1
+		}
+		input = f
+	}
+	err = m.SetLayerBigData(layer.ID, args[1], input)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%+v\n", err)
+		return 1
+	}
+	return 0
+}
 
 func layer(flags *mflag.FlagSet, action string, m storage.Store, args []string) int {
 	matched := []*storage.Layer{}
@@ -104,6 +177,37 @@ func init() {
 			minArgs:     1,
 			addFlags: func(flags *mflag.FlagSet, cmd *command) {
 				flags.BoolVar(&jsonOutput, []string{"-json", "j"}, jsonOutput, "Prefer JSON output")
+			},
+		},
+		command{
+			names:       []string{"list-layer-data", "listlayerdata"},
+			optionsHelp: "[options [...]] layerNameOrID",
+			usage:       "List data items that are attached to a layer",
+			action:      listLayerBigData,
+			minArgs:     1,
+			maxArgs:     1,
+			addFlags: func(flags *mflag.FlagSet, cmd *command) {
+				flags.BoolVar(&jsonOutput, []string{"-json", "j"}, jsonOutput, "Prefer JSON output")
+			},
+		},
+		command{
+			names:       []string{"get-layer-data", "getlayerdata"},
+			optionsHelp: "[options [...]] layerID dataName",
+			usage:       "Get data that is attached to a layer",
+			action:      getLayerBigData,
+			minArgs:     2,
+			addFlags: func(flags *mflag.FlagSet, cmd *command) {
+				flags.StringVar(&paramLayerDataFile, []string{"-file", "f"}, paramLayerDataFile, "Write data to file")
+			},
+		},
+		command{
+			names:       []string{"set-layer-data", "setlayerdata"},
+			optionsHelp: "[options [...]] layerID dataName",
+			usage:       "Set data that is attached to a layer",
+			action:      setLayerBigData,
+			minArgs:     2,
+			addFlags: func(flags *mflag.FlagSet, cmd *command) {
+				flags.StringVar(&paramLayerDataFile, []string{"-file", "f"}, paramLayerDataFile, "Read data from file")
 			},
 		},
 	)
--- a/docs/containers-storage-get-layer-data.md
+++ b/docs/containers-storage-get-layer-data.md
@@ -1,0 +1,21 @@
+## containers-storage-get-layer-data 1 "December 2020"
+
+## NAME
+containers-storage get-layer-data - Retrieve lookaside data for a layer
+
+## SYNOPSIS
+**containers-storage** **get-layer-data** [*options* [...]] *layerID* *dataName*
+
+## DESCRIPTION
+Retrieves a piece of named data which is associated with a layer.
+
+## OPTIONS
+**-f | --file** *file*
+
+Write the data to a file instead of stdout.
+
+## EXAMPLE
+**containers-storage get-layer-data -f config.json 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824 configuration**
+
+## SEE ALSO
+containers-storage-set-layer-data(1)
--- a/docs/containers-storage-set-layer-data.md
+++ b/docs/containers-storage-set-layer-data.md
@@ -1,0 +1,21 @@
+## containers-storage-set-layer-data 1 "December 2020"
+
+## NAME
+containers-storage set-layer-data - Set lookaside data for a layer
+
+## SYNOPSIS
+**containers-storage** **set-layer-data** [*options* [...]] *layerID* *dataName*
+
+## DESCRIPTION
+Sets a piece of named data which is associated with a layer.
+
+## OPTIONS
+**-f | --file** *filename*
+
+Read the data contents from a file instead of stdin.
+
+## EXAMPLE
+**containers-storage set-layer-data -f ./config.json 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824 configuration**
+
+## SEE ALSO
+containers-storage-get-layer-data(1)
--- a/layers.go
+++ b/layers.go
@@ -117,6 +117,11 @@ type Layer struct {
 
 	// ReadOnly is true if this layer resides in a read-only layer store.
 	ReadOnly bool `json:"-"`
+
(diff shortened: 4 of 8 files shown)


giuseppe:
@nalind @rhatdan PTAL

rhatdan:
LGTM
@vrothberg @mtrmac @TomSweeneyRedHat PTAL

cmd/containers-storage/layer.go
```diff
@@ -106,5 +160,25 @@ func init() {
 				flags.BoolVar(&jsonOutput, []string{"-json", "j"}, jsonOutput, "Prefer JSON output")
 			},
 		},
+		command{
+			names:       []string{"get-layer-data", "getlayerdata"},
+			optionsHelp: "[options [...]] layerID dataName",
+			usage:       "Get data that is attached to an layer",
```
TomSweeneyRedHat, line 166: nit
```suggestion
			usage:       "Get data that is attached to a layer",
```

cmd/containers-storage/layer.go
```diff
@@ -106,5 +160,25 @@ func init() {
...
+			action:      getLayerBigData,
+			minArgs:     2,
+			addFlags: func(flags *mflag.FlagSet, cmd *command) {
+				flags.StringVar(&paramLayerDataFile, []string{"-file", "f"}, paramLayerDataFile, "Write data to file")
+			},
+		},
+		command{
+			names:       []string{"set-layer-data", "setlayerdata"},
+			optionsHelp: "[options [...]] layerID dataName",
+			usage:       "Set data that is attached to an layer",
```
TomSweeneyRedHat, line 176: nit
```suggestion
			usage:       "Set data that is attached to a layer",
```

docs/containers-storage-get-layer-data.md
```diff
@@ -0,0 +1,21 @@
+## containers-storage-get-container-data 1 "December 2020"
```
TomSweeneyRedHat, line 1: I'm a bit confused over exactly what data is being gathered.  Title of the PR says "BigData", functions about Get/Set LayerBigData, help above "data attached to a layer", and man page says "lookaside" data.

Can we be consistent through out?  I'd prefer "lookaside data" if that holds.

cmd/containers-storage/layer.go
giuseppe: thanks, fixed

cmd/containers-storage/layer.go
giuseppe: thanks, fixed!

docs/containers-storage-get-layer-data.md
giuseppe: it can be any data you want to record for a layer.  In particular, it is going to record the zstd:chunked manifest in https://github.com/containers/storage/pull/775

TomSweeneyRedHat: OK, but my comment still holds.  It appears we're calling it three or four different things inside the code and on the man pages.  Can we change the variable names/function calls to refer to lookaside data?

mtrmac (commented):
A literally one-minute impression, and design concerns like this are definitely decided by @nalind , not me:

This can make sense as an internal implementation mechanism; it’s not obvious to me that this _must_ be exposed as a raw data to external callers. If c/image is only going to consume that data through a zstd:chunked - specific API, only that zstd:chunked - specific API needs to be public.

Making raw data public like this
- effectively freezes the format of the raw data, slowing down future improvements.
- invites other uses of raw data for unexpected purposes. That can be very useful and allow for unexpected innovations, OTOH it also further cements down the concept of individual layers and makes it even harder to significantly restructure (e.g. to a file-based approach like OSTree or the chunked format).

Again, definitely up to @nalind, just airing my first impression.

docs/containers-storage-get-layer-data.md
giuseppe: BigData is the name already used for containers and images.  I think a different name can be confusing.

Should I change the comments?

TomSweeneyRedHat: @giuseppe I'm fine with whatever name you use, my main issue was you used 3 or more names for it.  So if you did this:

LayerBigData  -> keep as is
"data attached to a layer" -> "BigData attached to a layer"
"lookaside" data -> BigData  or "the layer's BigData"

I'd be good with that.

cmd/containers-storage/container.go
```diff
@@ -12,6 +12,7 @@ import (
 
 var (
 	paramContainerDataFile = ""
+	paramLayerDataFile     = ""
```
mtrmac, line 15: I’d expect this to go to `…/layer.go`.

docs/containers-storage-get-layer-data.md
mtrmac, line 1:
```suggestion
## containers-storage-get-layer-data 1 "December 2020"
```

docs/containers-storage-set-layer-data.md
```diff
@@ -0,0 +1,21 @@
+## containers-storage-set-container-data 1 "December 2020"
```
mtrmac, line 1:
```suggestion
## containers-storage-set-layer-data 1 "December 2020"
```

layers.go
```diff
@@ -278,6 +289,8 @@ func copyLayer(l *Layer) *Layer {
 		UncompressedSize:   l.UncompressedSize,
 		CompressionType:    l.CompressionType,
 		ReadOnly:           l.ReadOnly,
+		BigDataNames:       l.BigDataNames,
```
mtrmac, line 292: `copyStringSlice`?

layers.go
```diff
@@ -117,6 +117,15 @@ type Layer struct {
...
 	ReadOnly bool `json:"-"`
+
+	// BigDataNames is a list of names of data items that we keep for the
+	// convenience of the caller.  They can be large, and are only in
+	// memory when being read from or written to disk.
+	BigDataNames []string `json:"big-data-names,omitempty"`
+
+	// BigDataSizes maps the names in BigDataNames to the sizes of the data
+	// that has been stored, if they're known.
+	BigDataSizes map[string]int64 `json:"big-data-sizes,omitempty"`
```
mtrmac, line 128: Values of this are AFAICS never set.

layers.go
```diff
@@ -970,6 +985,87 @@ func (r *layerStore) SetNames(id string, names []string) error {
...
 
+func (r *layerStore) datadir(id string) string {
+	return filepath.Join(r.layerdir, id)
+}
+
+func (r *layerStore) datapath(id, key string) string {
+	return filepath.Join(r.datadir(id), makeBigDataBaseName(key))
+}
+
+func (r *layerStore) BigData(id, key string) ([]byte, error) {
```
mtrmac, line 996: How big is “big”? Does it make sense for the API to read/write using `[]byte`, or should that be a streaming interface?

layers.go
mtrmac, line 128: Just reading the API documentation, I would expect `BigDataSizes` to be included in `ImageSize`. (Otherwise it’s not even clear to me there needs to be a separate faster-to-access record of the sizes.)

mtrmac (commented):
A few more notes. I didn’t read all of the code carefully, and I completely defer to actual c/storage maintainers for all the design decisions.

layers.go
mtrmac: For a bit more context, this is probably not a concern for fairly fixed things like manifests (in image “big data”). OTOH there _has_ been a report of Raspberry Pis running out of memory, probably when loading tar-split metadata into RAM.

docs/containers-storage-get-layer-data.md
giuseppe: I am still unsure.  LayerBigData is about the API while "data attached to a layer" and "lookaside data" are used in the documentation and code comments.  Using `BigData` to explain what the Layer BigData is seems confusing:

e.g. converting:
`containers-storage get-layer-data - Retrieve lookaside data for a layer`
to:
`containers-storage get-layer-data - Retrieve bigdata for a layer`

does not seem like an improvement

layers.go
giuseppe: thanks for spotting it.  I am going to push a fixed version

giuseppe: I've only added it because it was present in the image API.  I am fine with dropping it

docs/containers-storage-get-layer-data.md
TomSweeneyRedHat: I'll let you make the final decision @giuseppe .  Not a real biggy, I just think it's a bit confusing as is.  I'll trust your judgment on the final naming.

rhatdan: I prever
 containers-storage get-layer-data - Retrieve lookaside data for a layer

layers.go
rhatdan: Let's drop it.

mtrmac: Why _shouldn’t_ the size be included in `ImageSize`?

giuseppe: > Let's drop it.

dropped and pushed a new version.

giuseppe:
is there anything more left to address?

rhatdan:
@nalind @mtrmac PTAL

mtrmac:
See the 2 points from my previous review.

layers.go
giuseppe: "big" is not really that big, the generated manifest file for Fedora is around 3M

layers.go
giuseppe: > Why _shouldn’t_ the size be included in `ImageSize`?

it is not used at the moment, we can add it if there is any user of it

giuseppe:
> See the 2 points from my previous review.

just replied to them

layers.go
mtrmac: We’ve had reports like https://github.com/containers/image/issues/495 (“hundreds of gigabytes”); and on the other side, users on Raspberry Pi with IIRC 256 MB of RAM.

I’m not yet saying that this is not good enough, right now I’m just pointing out that Fedora is _by far_ not the worst case of how far this should scale.

layers.go
mtrmac: The Podman repo suggests that `store.ImageSize` is called for “disk usage statistics” (`df`?), “inspect”, “prune” and “tree”; I didn’t trace the full call stack, it may be exposed via other features or it might be dead code.

OTOH it’s true that c/image’s `types.Image.Size()` implementation does its own calculation which does not currently benefit from extending `store.ImageSize`.

giuseppe:
@mtrmac @rhatdan anything more holding this PR?

rhatdan:
Not from me. Just waiting for @mtrmac comment.

mtrmac:
\*shrug\* I don’t have anything to add WRT the implementation; I have some concerns but they are not quite blocking.

As for the more general design review, I know very little about c/storage , that would be up to @nalind I think.

giuseppe:
how to move forward?  This is just a subset of the changes I'd like to get merged for zstd:chunked support

rhatdan:
@vrothberg @nalind PTAL so we can get this merged.

vrothberg:
Added to my todo list.  I won't make it today but promise to have a look on Monday.

Apologies for not having looked into it until now!

rhatdan (approved):
Setting approved

layers.go
vrothberg: I echo @mtrmac's concern.  Returning an `io.ReadCloser` seems like a safe(r) bet.

layers.go
```diff
@@ -970,6 +979,66 @@ func (r *layerStore) SetNames(id string, names []string) error {
...
+		return nil, errors.Wrapf(ErrInvalidBigDataName, "can't retrieve layer big data value for empty name")
+	}
+	layer, ok := r.lookup(id)
+	if !ok {
+		return nil, errors.Wrapf(ErrLayerUnknown, "error locating layer with ID %q", id)
+	}
+	return ioutil.ReadFile(r.datapath(layer.ID, key))
+}
+
+func (r *layerStore) SetBigData(id, key string, data []byte) error {
```
vrothberg, line 1001: Same here, we could turn `data` into an `io.Reader`?

layers.go
```diff
@@ -970,6 +979,66 @@ func (r *layerStore) SetNames(id string, names []string) error {
...
+	return filepath.Join(r.datadir(id), makeBigDataBaseName(key))
+}
+
+func (r *layerStore) BigData(id, key string) ([]byte, error) {
+	if key == "" {
+		return nil, errors.Wrapf(ErrInvalidBigDataName, "can't retrieve layer big data value for empty name")
+	}
+	layer, ok := r.lookup(id)
+	if !ok {
+		return nil, errors.Wrapf(ErrLayerUnknown, "error locating layer with ID %q", id)
```
vrothberg, line 996: Can we add "big data" to the error message?  That'll help to put a finger on the code in case of errors.

vrothberg: There are other cases below

layers.go
giuseppe: I am fine with it, just it will be different than all the others `BigData()/SetBigData()` functions we already have for containers and images

giuseppe: changed to use `io.Reader and io.ReadCloser`

layers.go
giuseppe: fixed

vrothberg (commented):
LGTM

layers.go
```diff
@@ -970,6 +979,77 @@ func (r *layerStore) SetNames(id string, names []string) error {
...
+
+func (r *layerStore) BigData(id, key string) (io.ReadCloser, error) {
+	if key == "" {
+		return nil, errors.Wrapf(ErrInvalidBigDataName, "can't retrieve layer big data value for empty name")
+	}
+	layer, ok := r.lookup(id)
+	if !ok {
+		return nil, errors.Wrapf(ErrLayerUnknown, "error locating layer with ID %q", id)
+	}
+	return os.Open(r.datapath(layer.ID, key))
```
mtrmac, line 998: Note, to be explicit: actual reads happen outside of the layer lock. AFAICS this is safe because removing the file does not truncate the underlying inode, and writes happen via `NewAtomicFileWriter`, i.e. don’t overwrite the original.

(Perhaps add comment to the `NewAtomicFileWriter` call below to note that the read implementation relies on it?)

giuseppe: thanks.

Added a comment

rhatdan:
LGTM

PR #794, opened 2020-12-22
idtools: handle single user mapped as root

giuseppe:
if a single user is mapped in the user namespace, handle it as root.

It is needed for running unprivileged containers with a single user available without being forced to run with euid and egid set to 0.

Needs: https://github.com/containers/fuse-overlayfs/pull/268


rhatdan:
LGTM

rhatdan:

vrothberg (approved):
LGTM

PR #795, opened 2020-12-26
Support additional layer store (patch for containers/storage)

ktock:
https://github.com/containers/podman/issues/4739

## Reconsidered the design on 2021/1/12

This enables podman to create containers using layers stored in a specified directory instead of pulling them from the registry. Leveraging this feature with remotely-mountable layers provided by stargz/zstd:chunked or CVMFS, podman can achieve lazy pulling.

### Changes in contianers/storage: https://github.com/containers/storage/pull/795

That directory is named "additional layer store" (call ALS in this doc) and has the following structure.

```
<ALS root>/base64("key1=value1")/base64("key2=value2")/.../
`-- diff
`-- info
```

`diff` directory contains the extracted layer diff contents specified by the key-value pairs.
`info` file contains `*c/storage.Layer` struct that indicates the information of this layer contents (`*c/storage.Layer.ID`, `*c/storage.Layer.Parent` and `*c/storage.Layer.Created` can be empty as it'll be filled by `c/storage.Store`).

On each pull, `c/storage.Store` searches the layer diff contents from ALS using pre-configured key-value pairs.
Each key-value pair is base64 encoded.
By default, the following `key=value` pairs can be used as elements of the path in ALS,

- `reference=<image reference>`
- `layerdigest=<digest of the compressed layer contents>`

Additionally, layer annotations (defined in the image manifest) prefixed by `containers/image/target.` can be used as well.
The prefix `containers/image/target.` will be trimmed from the key when it's used in the path on ALS.

Overlay driver supports an option to specify which key-value pair to be used and how the order should they be when `c/storage.Store` searches layers in the ALS.

```
layerstore=<ALS root directory>:<key1>:<key2>:...:<keyX>
```

In the above case, on each pull, `c/storage.Store` searches the following path in ALS,

```
<ALS root>/base64("key1=value1")/base64("key2=value2")/.../base64("keyX=valueX")/
`-- diff
`-- info
```

The underlying filesystem (e.g. stargz/zstd:chunked-based filesystem or CVMFS) should show the exploded view of the target layer diff and its information at these locations.
Example filesystem implementation (currently stargz-based) is https://github.com/ktock/stargz-snapshotter/tree/als-pool-example (this must be mounted on <ALS root>)

If the layer content is found in ALS, `c/storage.Store` creates layer using `<ALS root>/.../info` as `*c/storage.Layer` and using `<ALS root>/.../diff` as its diff directory.
So `c/image`'s copier doesn't need to pull this layer from the registry.

### Changes in containers/image: https://github.com/containers/image/pull/1109

Now `c/image`'s copier leverages this store. 
Every time this pulls an image, it first tries to reuse blobs from ALS.
That copier passes each layer's OCI annotations (key-value pairs) + the following key-value to `c/storage.Store`.

- containers/image/target.reference: The image reference of an image that contains the target layer.  This is needed for supporting registry-backed filesystems (e.g. estargz, zstd:chunked).

`*c/image.storageImageDestination.TryReusingBlob()` cannot pass image reference to `c/storage.Store` so this commit adds a new API `*c/image.storageImageDestination.TryReusingBlobWithRef() for achieving this.
When this copier successfully acquires that layer, this reuses this layer without pulling.

### Command exapmle

```
podman --storage-opt "layerstore=/tmp/storage:reference:layerdigest" pull ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz
podman --storage-opt "layerstore=/tmp/storage:reference:layerdigest" run --rm -it ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz /bin/echo 'Hello, World!'
```

In the above cases, `c/storage.Store` looks up `/tmp/storage/base64("reference=ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz")/base64(<layer digests>)/{diff, info}` in ALS.
The example filesystem implementation (https://github.com/ktock/stargz-snapshotter/tree/als-pool-example) is mounted at `/tmp/storage` shows the extracted layer at that location.
Then `rethinkdb:2.3.6-esgz` can run without pulling it from registry.

## Known limitation and request for comments

Some operations (e.g. save) requires correct value to be set to `c/storage.Layer.UncompressedSize`. This field seems to be the size of the layer but without compression (i.e. the size of the tar-archived format of that layer). For registry-backed ALS, getting this information is difficult because neither of OCI/Docker image nor registry API provides the way to get the uncompressed size of layers. We cannot get this information without actually pull and decompress the layer, which is not lazy pulling that this PR aims to.

I'll check the codebase deeper to come up with the way to get this information from somewhere or the way to safely allow `c/storage.Layer.UncompressedSize` to be *unknown* during operations. But if someone has good idea for solving this, please let me know.

cc: @siscia @giuseppe

### DIFF
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -14,6 +14,7 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/directory"
 	"github.com/containers/storage/pkg/idtools"
+	digest "github.com/opencontainers/go-digest"
 )
 
 // FsMagic unsigned id of the filesystem in use.
@@ -33,7 +34,9 @@ var (
 	// ErrPrerequisites returned when driver does not meet prerequisites.
 	ErrPrerequisites = errors.New("prerequisites for driver not satisfied (wrong filesystem?)")
 	// ErrIncompatibleFS returned when file system is not supported.
-	ErrIncompatibleFS = fmt.Errorf("backing file system is unsupported for this graph driver")
+	ErrIncompatibleFS = errors.New("backing file system is unsupported for this graph driver")
+	// ErrLayerUnknown returned when the specified layer is unknown by the driver.
+	ErrLayerUnknown = errors.New("unknown layer")
 )
 
 //CreateOpts contains optional arguments for Create() and CreateReadWrite()
@@ -117,6 +120,7 @@ type ProtoDriver interface {
 	// known to this driver.
 	Cleanup() error
 	// AdditionalImageStores returns additional image stores supported by the driver
+	// This API is experimental and can be changed without bumping the major version number.
 	AdditionalImageStores() []string
 }
 
@@ -178,6 +182,30 @@ type Capabilities struct {
 // can report on their Capabilities.
 type CapabilityDriver interface {
 	Capabilities() Capabilities
+}
+
+// AdditionalLayer reprents a layer that is stored in the additional layer store
+// This API is experimental and can be changed without bumping the major version number.
+type AdditionalLayer interface {
+	// CreateAs creates a new layer from this additional layer
+	CreateAs(id, parent string) error
+
+	// Info returns arbitrary information stored along with this layer (i.e. `info` file)
+	Info() (io.ReadCloser, error)
+
+	// Release tells the additional layer store that we don't use this handler.
+	Release()
+}
+
+// AdditionalLayerStoreDriver is the interface for driver that supports
+// additional layer store functionality.
+// This API is experimental and can be changed without bumping the major version number.
+type AdditionalLayerStoreDriver interface {
+	Driver
+
+	// LookupAdditionalLayer looks up additional layer store by the specified
+	// digest and ref and returns an object representing that layer.
+	LookupAdditionalLayer(d digest.Digest, ref string) (AdditionalLayer, error)
 }
 
 // DiffGetterDriver is the interface for layered file system drivers that
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -4,6 +4,7 @@ package overlay
 
 import (
 	"bytes"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -31,6 +32,7 @@ import (
 	"github.com/containers/storage/pkg/unshare"
 	units "github.com/docker/go-units"
 	"github.com/hashicorp/go-multierror"
+	digest "github.com/opencontainers/go-digest"
 	rsystem "github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/selinux/go-selinux"
 	"github.com/opencontainers/selinux/go-selinux/label"
@@ -94,6 +96,7 @@ const (
 
 type overlayOptions struct {
 	imageStores       []string
+	layerStores       []additionalLayerStore
 	quota             quota.Quota
 	mountProgram      string
 	skipMountHome     bool
@@ -117,6 +120,17 @@ type Driver struct {
 	supportsVolatile bool
 	usingMetacopy    bool
 	locker           *locker.Locker
+}
+
+type additionalLayerStore struct {
+
+	// path is the directory where this store is available on the host.
+	path string
+
+	// withReference is true when the store contains image reference information (base64-encoded)
+	// in its layer search path so the path to the diff will be
+	//  <path>/base64(reference)/<layerdigest>/
+	withReference bool
 }
 
 var (
@@ -397,6 +411,42 @@ func parseOptions(options []string) (*overlayOptions, error) {
 				}
 				o.imageStores = append(o.imageStores, store)
 			}
+		case "additionallayerstore":
+			logrus.Debugf("overlay: additionallayerstore=%s", val)
+			// Additional read only layer stores to use for lower paths
+			if val == "" {
+				continue
+			}
+			for _, lstore := range strings.Split(val, ",") {
+				elems := strings.Split(lstore, ":")
+				lstore = filepath.Clean(elems[0])
+				if !filepath.IsAbs(lstore) {
+					return nil, fmt.Errorf("overlay: additionallayerstore path %q is not absolute.  Can not be relative", lstore)
+				}
+				st, err := os.Stat(lstore)
+				if err != nil {
+					return nil, errors.Wrap(err, "overlay: can't stat additionallayerstore dir")
+				}
+				if !st.IsDir() {
+					return nil, fmt.Errorf("overlay: additionallayerstore path %q must be a directory", lstore)
+				}
+				var withReference bool
+				for _, e := range elems[1:] {
+					switch e {
+					case "ref":
+						if withReference {
+							return nil, fmt.Errorf("overlay: additionallayerstore config of %q contains %q option twice", lstore, e)
+						}
+						withReference = true
+					default:
+						return nil, fmt.Errorf("overlay: additionallayerstore config %q contains unknown option %q", lstore, e)
+					}
+				}
+				o.layerStores = append(o.layerStores, additionalLayerStore{
+					path:          lstore,
+					withReference: withReference,
+				})
+			}
 		case "mount_program":
 			logrus.Debugf("overlay: mount_program=%s", val)
 			if val != "" {
@@ -655,6 +705,24 @@ func (d *Driver) Metadata(id string) (map[string]string, error) {
 // we had created.
 func (d *Driver) Cleanup() error {
 	return mount.Unmount(d.home)
+}
+
+// LookupAdditionalLayer looks up additional layer store by the specified
+// digest and ref and returns an object representing that layer.
+// This API is experimental and can be changed without bumping the major version number.
+func (d *Driver) LookupAdditionalLayer(dgst digest.Digest, ref string) (graphdriver.AdditionalLayer, error) {
+	l, err := d.getAdditionalLayerPath(dgst, ref)
+	if err != nil {
+		return nil, err
+	}
+	// Tell the additional layer store that we use this layer.
+	// This will increase reference counter on the store's side.
+	// This will be decreased on Release() method.
+	notifyUseAdditionalLayer(l)
+	return &additionalLayer{
+		path: l,
+		d:    d,
+	}, nil
 }
 
 // CreateFromTemplate creates a layer with the same contents and parent as another layer.
@@ -958,6 +1026,8 @@ func (d *Driver) Remove(id string) error {
 			logrus.Debugf("Failed to remove link: %v", err)
 		}
 	}
+
+	d.releaseAdditionalLayerByID(id)
 
 	if err := system.EnsureRemoveAll(dir); err != nil && !os.IsNotExist(err) {
 		return err
@@ -1422,7 +1492,10 @@ func (f fileGetNilCloser) Close() error {
 // DiffGetter returns a FileGetCloser that can read files from the directory that
 // contains files for the layer differences. Used for direct access for tar-split.
 func (d *Driver) DiffGetter(id string) (graphdriver.FileGetCloser, error) {
-	p := d.getDiffPath(id)
+	p, err := d.getDiffPath(id)
+	if err != nil {
+		return nil, err
+	}
 	return fileGetNilCloser{storage.NewPathFileGetter(p)}, nil
 }
 
@@ -1444,7 +1517,10 @@ func (d *Driver) ApplyDiff(id, parent string, options graphdriver.ApplyDiffOpts)
 		idMappings = &idtools.IDMappings{}
 	}
 
-	applyDir := d.getDiffPath(id)
+	applyDir, err := d.getDiffPath(id)
+	if err != nil {
+		return 0, err
+	}
 
 	logrus.Debugf("Applying tar in %s", applyDir)
 	// Overlay doesn't need the parent id to apply the diff
@@ -1462,10 +1538,23 @@ func (d *Driver) ApplyDiff(id, parent string, options graphdriver.ApplyDiffOpts)
 	return directory.Size(applyDir)
 }
 
-func (d *Driver) getDiffPath(id string) string {
+func (d *Driver) getDiffPath(id string) (string, error) {
 	dir := d.dir(id)
+	return redirectDiffIfAdditionalLayer(path.Join(dir, "diff"))
+}
 
-	return path.Join(dir, "diff")
+func (d *Driver) getLowerDiffPaths(id string) ([]string, error) {
+	layers, err := d.getLowerDirs(id)
+	if err != nil {
+		return nil, err
+	}
+	for i, l := range layers {
+		layers[i], err = redirectDiffIfAdditionalLayer(l)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return layers, nil
 }
 
 // DiffSize calculates the changes between the specified id
@@ -1476,7 +1565,11 @@ func (d *Driver) DiffSize(id string, idMappings *idtools.IDMappings, parent stri
 		return d.naiveDiff.DiffSize(id, idMappings, parent, parentMappings, mountLabel)
 	}
 
-	return directory.Size(d.getDiffPath(id))
+	p, err := d.getDiffPath(id)
+	if err != nil {
+		return 0, err
+	}
+	return directory.Size(p)
 }
 
 // Diff produces an archive of the changes between the specified
@@ -1490,12 +1583,15 @@ func (d *Driver) Diff(id string, idMappings *idtools.IDMappings, parent string, 
 		idMappings = &idtools.IDMappings{}
 	}
 
-	lowerDirs, err := d.getLowerDirs(id)
+	lowerDirs, err := d.getLowerDiffPaths(id)
 	if err != nil {
 		return nil, err
 	}
 
-	diffPath := d.getDiffPath(id)
+	diffPath, err := d.getDiffPath(id)
+	if err != nil {
+		return nil, err
+	}
 	logrus.Debugf("Tar with options on %s", diffPath)
 	return archive.TarWithOptions(diffPath, &archive.TarOptions{
 		Compression:    archive.Uncompressed,
@@ -1514,8 +1610,11 @@ func (d *Driver) Changes(id string, idMappings *idtools.IDMappings, parent strin
 	}
 	// Overlay doesn't have snapshots, so we need to get changes from all parent
 	// layers.
-	diffPath := d.getDiffPath(id)
-	layers, err := d.getLowerDirs(id)
+	diffPath, err := d.getDiffPath(id)
+	if err != nil {
+		return nil, err
+	}
+	layers, err := d.getLowerDiffPaths(id)
 	if err != nil {
 		return nil, err
 	}
@@ -1626,4 +1725,140 @@ func nameWithSuffix(name string, number int) string {
 		return name
 	}
 	return fmt.Sprintf("%s%d", name, number)
+}
+
+func (d *Driver) getAdditionalLayerPath(dgst digest.Digest, ref string) (string, error) {
+	refElem := base64.StdEncoding.EncodeToString([]byte(ref))
+	for _, ls := range d.options.layerStores {
+		ref := ""
+		if ls.withReference {
+			ref = refElem
+		}
+		target := path.Join(ls.path, ref, dgst.String())
+		// Check if all necessary files exist
+		for _, p := range []string{
+			filepath.Join(target, "diff"),
+			filepath.Join(target, "info"),
+			// TODO(ktock): We should have an API to expose the stream data of this layer
+			//              to enable the client to retrieve the entire contents of this
+			//              layer when it exports this layer.
+		} {
+			if _, err := os.Stat(p); err != nil {
+				return "", errors.Wrapf(graphdriver.ErrLayerUnknown,
+					"failed to stat additional layer %q: %v", p, err)
+			}
+		}
+		return target, nil
+	}
+
+	return "", errors.Wrapf(graphdriver.ErrLayerUnknown,
+		"additional layer (%q, %q) not found", dgst, ref)
+}
+
+func (d *Driver) releaseAdditionalLayerByID(id string) {
+	if al, err := ioutil.ReadFile(path.Join(d.dir(id), "additionallayer")); err == nil {
+		notifyReleaseAdditionalLayer(string(al))
+	} else if !os.IsNotExist(err) {
+		logrus.Warnf("unexpected error on reading Additional Layer Store pointer %v", err)
+	}
+}
+
+// additionalLayer represents a layer in Additional Layer Store.
+type additionalLayer struct {
+	path        string
+	d           *Driver
+	releaseOnce sync.Once
+}
+
+// Info returns arbitrary information stored along with this layer (i.e. `info` file).
+// This API is experimental and can be changed without bumping the major version number.
+func (al *additionalLayer) Info() (io.ReadCloser, error) {
+	return os.Open(filepath.Join(al.path, "info"))
+}
+
+// CreateAs creates a new layer from this additional layer.
+// This API is experimental and can be changed without bumping the major version number.
+func (al *additionalLayer) CreateAs(id, parent string) error {
+	// TODO: support opts
+	if err := al.d.Create(id, parent, nil); err != nil {
+		return err
+	}
+	dir := al.d.dir(id)
+	diffDir := path.Join(dir, "diff")
+	if err := os.RemoveAll(diffDir); err != nil {
+		return err
+	}
+	// tell the additional layer store that we use this layer.
+	// mark this layer as "additional layer"
+	if err := ioutil.WriteFile(path.Join(dir, "additionallayer"), []byte(al.path), 0644); err != nil {
+		return err
+	}
+	notifyUseAdditionalLayer(al.path)
+	return os.Symlink(filepath.Join(al.path, "diff"), diffDir)
+}
+
+// Release tells the additional layer store that we don't use this handler.
+// This API is experimental and can be changed without bumping the major version number.
+func (al *additionalLayer) Release() {
+	// Tell the additional layer store that we don't use this layer handler.
+	// This will decrease the reference counter on the store's side, which was
+	// increased in LookupAdditionalLayer (so this must be called only once).
+	al.releaseOnce.Do(func() {
+		notifyReleaseAdditionalLayer(al.path)
+	})
+}
+
+// notifyUseAdditionalLayer notifies Additional Layer Store that we use the specified layer.
+// This is done by creating "use" file in the layer directory. This is useful for
+// Additional Layer Store to consider when to perform GC. Notification-aware Additional
+// Layer Store must return ENOENT.
+func notifyUseAdditionalLayer(al string) {
+	if !path.IsAbs(al) {
+		logrus.Warnf("additionallayer must be absolute (got: %v)", al)
+		return
+	}
+	useFile := path.Join(al, "use")
+	f, err := os.Create(useFile)
+	if os.IsNotExist(err) {
+		return
+	} else if err == nil {
+		f.Close()
+		if err := os.Remove(useFile); err != nil {
+			logrus.Warnf("failed to remove use file")
+		}
+	}
+	logrus.Warnf("unexpected error by Additional Layer Store %v during use; GC doesn't seem to be supported", err)
+}
+
+// notifyReleaseAdditionalLayer notifies Additional Layer Store that we don't use the specified
+// layer anymore. This is done by rmdir-ing the layer directory. This is useful for
+// Additional Layer Store to consider when to perform GC. Notification-aware Additional
+// Layer Store must return ENOENT.
+func notifyReleaseAdditionalLayer(al string) {
+	if !path.IsAbs(al) {
+		logrus.Warnf("additionallayer must be absolute (got: %v)", al)
+		return
+	}
+	// tell the additional layer store that we don't use this layer anymore.
+	err := unix.Rmdir(al)
+	if os.IsNotExist(err) {
+		return
+	}
+	logrus.Warnf("unexpected error by Additional Layer Store %v during release; GC doesn't seem to be supported", err)
+}
+
+// redirectDiffIfAdditionalLayer checks if the passed diff path is Additional Layer and
+// returns the redirected path. If the passed diff is not the one in Additional Layer
+// Store, it returns the original path without changes.
+func redirectDiffIfAdditionalLayer(diffPath string) (string, error) {
+	if ld, err := os.Readlink(diffPath); err == nil {
+		// diff is the link to Additional Layer Store
+		if !path.IsAbs(ld) {
+			return "", fmt.Errorf("linkpath must be absolute (got: %q)", ld)
+		}
+		diffPath = ld
+	} else if err.(*os.PathError).Err != syscall.EINVAL {
+		return "", err
+	}
+	return diffPath, nil
 }
--- a/layers.go
+++ b/layers.go
@@ -250,6 +250,11 @@ type LayerStore interface {
 	// LoadLocked wraps Load in a locked state. This means it loads the store
 	// and cleans-up invalid layers if needed.
 	LoadLocked() error
+
+	// PutAdditionalLayer creates a layer using the diff contained in the additional layer
+	// store.
+	// This API is experimental and can be changed without bumping the major version number.
+	PutAdditionalLayer(id string, parentLayer *Layer, names []string, aLayer drivers.AdditionalLayer) (layer *Layer, err error)
 }
 
 type layerStore struct {
@@ -608,6 +613,58 @@ func (r *layerStore) SetFlag(id string, flag string, value interface{}) error {
 
 func (r *layerStore) Status() ([][2]string, error) {
 	return r.driver.Status(), nil
+}
+
+func (r *layerStore) PutAdditionalLayer(id string, parentLayer *Layer, names []string, aLayer drivers.AdditionalLayer) (layer *Layer, err error) {
+	if duplicateLayer, idInUse := r.byid[id]; idInUse {
+		return duplicateLayer, ErrDuplicateID
+	}
+	for _, name := range names {
+		if _, nameInUse := r.byname[name]; nameInUse {
+			return nil, ErrDuplicateName
+		}
+	}
+
+	parent := ""
+	if parentLayer != nil {
+		parent = parentLayer.ID
+	}
+
+	info, err := aLayer.Info()
+	if err != nil {
+		return nil, err
+	}
+	defer info.Close()
+	layer = &Layer{}
+	if err := json.NewDecoder(info).Decode(layer); err != nil {
+		return nil, err
+	}
+	layer.ID = id
+	layer.Parent = parent
+	layer.Created = time.Now().UTC()
+
+	if err := aLayer.CreateAs(id, parent); err != nil {
+		return nil, err
+	}
+
+	// TODO: check if necessary fields are filled
+	r.layers = append(r.layers, layer)
+	r.idindex.Add(id)
+	r.byid[id] = layer
+	for _, name := range names { // names got from the additional layer store won't be used
+		r.byname[name] = layer
+	}
+	if layer.CompressedDigest != "" {
+		r.bycompressedsum[layer.CompressedDigest] = append(r.bycompressedsum[layer.CompressedDigest], layer.ID)
+	}
+	if layer.UncompressedDigest != "" {
+		r.byuncompressedsum[layer.CompressedDigest] = append(r.byuncompressedsum[layer.CompressedDigest], layer.ID)
+	}
+	if err := r.Save(); err != nil {
+		r.driver.Remove(id)
+		return nil, err
+	}
+	return copyLayer(layer), nil
 }
 
 func (r *layerStore) Put(id string, parentLayer *Layer, names []string, mountLabel string, options map[string]string, moreOptions *LayerOptions, writeable bool, flags map[string]interface{}, diff io.Reader) (layer *Layer, size int64, err error) {
--- a/store.go
+++ b/store.go
@@ -2,6 +2,7 @@ package storage
 
 import (
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -489,6 +490,30 @@ type Store interface {
 
 	// GetDigestLock returns digest-specific Locker.
 	GetDigestLock(digest.Digest) (Locker, error)
+
+	// LayerFromAdditionalLayerStore searches layers from the additional layer store and
+	// returns the object for handling this. Note that this hasn't been stored to this store
+	// yet so this needs to be done through PutAs method.
+	// Releasing AdditionalLayer handler is caller's responsibility.
+	// This API is experimental and can be changed without bumping the major version number.
+	LookupAdditionalLayer(d digest.Digest, imageref string) (AdditionalLayer, error)
+}
+
+// AdditionalLayer reprents a layer that is contained in the additional layer store
+// This API is experimental and can be changed without bumping the major version number.
+type AdditionalLayer interface {
+	// PutAs creates layer based on this handler, using diff contents from the additional
+	// layer store.
+	PutAs(id, parent string, names []string) (*Layer, error)
+
+	// UncompressedDigest returns the uncompressed digest of this layer
+	UncompressedDigest() digest.Digest
+
+	// CompressedSize returns the compressed size of this layer
+	CompressedSize() int64
+
+	// Release tells the additional layer store that we don't use this handler.
+	Release()
 }
 
 type AutoUserNsOptions = types.AutoUserNsOptions
@@ -3132,6 +3157,91 @@ func (s *store) Layer(id string) (*Layer, error) {
 		}
 	}
 	return nil, ErrLayerUnknown
+}
+
+func (s *store) LookupAdditionalLayer(d digest.Digest, imageref string) (AdditionalLayer, error) {
+	adriver, ok := s.graphDriver.(drivers.AdditionalLayerStoreDriver)
+	if !ok {
+		return nil, ErrLayerUnknown
+	}
+
+	al, err := adriver.LookupAdditionalLayer(d, imageref)
+	if err != nil {
+		if errors.Is(err, drivers.ErrLayerUnknown) {
+			return nil, ErrLayerUnknown
+		}
+		return nil, err
+	}
+	info, err := al.Info()
+	if err != nil {
+		return nil, err
+	}
+	defer info.Close()
+	var layer Layer
+	if err := json.NewDecoder(info).Decode(&layer); err != nil {
+		return nil, err
+	}
+	return &additionalLayer{&layer, al, s}, nil
+}
+
+type additionalLayer struct {
+	layer   *Layer
+	handler drivers.AdditionalLayer
+	s       *store
+}
+
+func (al *additionalLayer) UncompressedDigest() digest.Digest {
+	return al.layer.UncompressedDigest
+}
+
+func (al *additionalLayer) CompressedSize() int64 {
+	return al.layer.CompressedSize
+}
+
+func (al *additionalLayer) PutAs(id, parent string, names []string) (*Layer, error) {
+	rlstore, err := al.s.LayerStore()
+	if err != nil {
+		return nil, err
+	}
+	rlstore.Lock()
+	defer rlstore.Unlock()
+	if modified, err := rlstore.Modified(); modified || err != nil {
+		if err = rlstore.Load(); err != nil {
+			return nil, err
+		}
+	}
+	rlstores, err := al.s.ROLayerStores()
+	if err != nil {
+		return nil, err
+	}
+
+	var parentLayer *Layer
+	if parent != "" {
+		for _, lstore := range append([]ROLayerStore{rlstore}, rlstores...) {
+			if lstore != rlstore {
+				lstore.RLock()
+				defer lstore.Unlock()
+				if modified, err := lstore.Modified(); modified || err != nil {
+					if err = lstore.Load(); err != nil {
+						return nil, err
+					}
+				}
+			}
+			parentLayer, err = lstore.Get(parent)
+			if err == nil {
+				break
+			}
+		}
+		if parentLayer == nil {
+			return nil, ErrLayerUnknown
+		}
+	}
+
+	return rlstore.PutAdditionalLayer(id, parentLayer, names, al.handler)
+}
+
+func (al *additionalLayer) Release() {
+	al.handler.Release()
 }
 
 func (s *store) Image(id string) (*Image, error) {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -122,6 +122,13 @@ type OptionsConfig struct {
 	// for shared image content
 	AdditionalImageStores []string `toml:"additionalimagestores"`
 
+	// AdditionalLayerStores is the location of additional read/only
+	// Layer stores.  Usually used to access Networked File System
+	// for shared image content
+	// This API is experimental and can be changed without bumping the
+	// major version number.
+	AdditionalLayerStores []string `toml:"additionallayerstores"`
+
 	// Size
 	Size string `toml:"size"`
 
--- a/types/options.go
+++ b/types/options.go
@@ -300,6 +300,9 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 	for _, s := range config.Storage.Options.AdditionalImageStores {
 		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.imagestore=%s", config.Storage.Driver, s))
 	}
+	for _, s := range config.Storage.Options.AdditionalLayerStores {
+		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.additionallayerstore=%s", config.Storage.Driver, s))
+	}
 	if config.Storage.Options.Size != "" {
 		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.size=%s", config.Storage.Driver, config.Storage.Options.Size))
 	}


drivers/overlay/overlay.go
```diff
@@ -499,16 +520,24 @@ func (d *Driver) Status() [][2]string {
 // Metadata returns meta data about the overlay driver such as
 // LowerDir, UpperDir, WorkDir and MergeDir used to store data.
 func (d *Driver) Metadata(id string) (map[string]string, error) {
+	metadata := make(map[string]string)
+	found := false
+	if additionalLayer := d.additionalLayer(id); additionalLayer != "" {
+		metadata = make(map[string]string)
```
rhatdan, line 526: This line is not needed, since it happens two lines above.

drivers/overlay/overlay.go
```diff
@@ -595,11 +624,12 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts) (retErr
...
-		if err != nil {
+		if err == nil {
+			rootUID = int(st.UID())
+			rootGID = int(st.GID())
+		} else if pLayer := d.additionalLayer(parent); pLayer == "" {
 			return err
-		}
-		rootUID = int(st.UID())
-		rootGID = int(st.GID())
+		} // TODO: rootUID & rootGID of additional layer?
```
rhatdan, line 632: You are ignoring the error, Should you at least log it?

drivers/overlay/overlay.go
```diff
@@ -734,19 +776,43 @@ func (d *Driver) dir(id string) string {
 	return newpath
 }
 
+// additionalLayer returns zero value if not found
+func (d *Driver) additionalLayer(id string) (dirPath string) {
+	for _, p := range d.AdditionalLayerStores() {
+		dirPath = path.Join(p, id)
+		if _, err := os.Stat(dirPath); err == nil {
+			return
```
rhatdan, line 784: I think you should do a `return dirpath` here, making the code easier to read.

layers.go
```diff
@@ -1480,3 +1485,87 @@ func (r *layerStore) TouchedSince(when time.Time) bool {
...
+		if strings.HasPrefix(k, targetPrefix) {
+			name := base64.StdEncoding.EncodeToString(
+				[]byte(strings.TrimPrefix(k, targetPrefix) + "=" + v))
+			if strings.HasPrefix(name, "rootfs.") {
+				return nil, fmt.Errorf("annotation must not have \"rootfs.\" prefix")
+			}
+			kvs = append(kvs, name)
+		}
+	}
+	target := filepath.Join(append([]string{r.layerdir, ".pool"}, append(kvs, "rootfs."+id)...)...)
```
mtrmac, line 1558: `annotations` is a map, which makes its iteration order unspecified, making the `target` value unstable. Besides, making this work depend on order of annotations in a manifest is unnecessarily brittle.

mtrmac (commented):
Just a drive-by comment; I don’t yet understand how this works at all.

layers.go
mtrmac: Also, accessing the filesystem _at all_ if any of the required information is missing worries me conceptually. Given the base64, it’s _probably_ not a security risk, but why toy with fire?

ktock: > Also, accessing the filesystem at all if any of the required information is missing worries me conceptually.

Cloud you elaborate you concern? `accessing the filesystem if any of the required information is missing` should simply fail and fallback to the normal pull procedure.

mtrmac: The container image is, in general, not trusted by the consumer, i.e. it is a possible way to attack the container runtime.

Thus, every image metadata field and feature has to be evaluated as a possible way for a malicious attacker to manipulate the container runtime; given the attacker too much flexibility is always a bit of a risk (that has to be somehow weighted against utility of features, sure).

---

Having read a bit more about the design, I think the core issue that the reference + layer digest that are currently passed via `WithAnnotations` should not be a part of the _image format_ at all! Gathering this data and passing it through to `searchLayer` is an internal implementation detail of the “pull” process, entirely local to the pull operation; there is AFAICS no reason for that a data to show up in manifests uploaded to registries.

It seems to me that the relevant data is available, or can be collected, by `c/image/copy` before calling `TryReusingBlob`, and that should happen by just adding more well-specified parameters to (some future variant of) `TryReusingBlob`, not by adding an opaque `map[string]string` data transfer mechanism just because that avoids new parameters, and especially not by using a data transfer mechanism that results in long-term externally-observable manifest changes.

(Sure, that would imply incompatible changes to `types.ImageDestination`, which is a hassle and requires rather more process. That’s common to all the recently proposed image reuse / on-demand loading proposals for c/image, and while I appreciate that having to deal with that on top of the complexity of actually implementing the image reuse / on-demand loading is an extra obstacle, it is something that just needs to happen for the c/image feature set to grow — and it will happen sooner or later.)

ktock:
Request for comments (in the above PR description)

> Known limitation and request for comments
> 
> Some operations (e.g. save) requires correct value to be set to c/storage.Layer.UncompressedSize. This field seems to be the size of the layer but without compression (i.e. the size of the tar-archived format of that layer). For registry-backed ALS, getting this information is difficult because neither of OCI/Docker image nor registry API provides the way to get the uncompressed size of layers. We cannot get this information without actually pull and decompress the layer, which is not lazy pulling that this PR aims to.
> 
> I'll check the codebase deeper to come up with the way to get this information from somewhere or the way to safely allow c/storage.Layer.UncompressedSize to be unknown during operations. But if someone has good idea for managing this, please let me know.

layers.go
ktock: > It seems to me that the relevant data is available, or can be collected, by c/image/copy before calling TryReusingBlob, and that should happen by just adding more well-specified parameters to (some future variant of) TryReusingBlob, not by adding an opaque map[string]string data transfer mechanism just because that avoids new parameters, and especially not by using a data transfer mechanism that results in long-term externally-observable manifest changes.

Fixed this.
Now `*c/image.storageImageDestination.TryReusingBlobWithRef()` recieves reference through its parameter, not through anontations.

ktock:
Can we move this forward?

drivers/driver.go
```diff
@@ -149,6 +149,19 @@ type LayerIDMapUpdater interface {
 	SupportsShifting() bool
 }
 
+// AdditionalLayerStore is the directory where exploded layer diff is stored.
+// Each layer can be accessed at:
+// Path/base64(RequiredAnnotations[0]=<value>)/.../base64(RequiredAnnotations[len(RequiredAnnotations)]=<value>)/
```
giuseppe, line 154: why do we need the annotations as part of the path?

What is the disadvantage of using directly the layer digest as specified in the manifest?

ktock: @giuseppe I initally thought we can use annotations for content verification: passing the digest of TOC from the runtime to the filesystem for ensuring that the (lazily) pulled contents is ones expected by the runtime. But I currently think we can let the filesystem do all these tasks (i.e. fetching manifest, getting TOC digest and verifying each chunk all on the filesystem's side).
Should we limit the possible path elements to `reference` and `layerdigest`?

Additionally, I thought we don't need `key=` in the part of the path because the order of `value` is deterministic (specified through driver's config `layerstore=<ALS root directory>:<key1>:<key2>:...:<keyX>`). So the path can be the following instead:
```
<ALS root>/base64("value1")/base64("value2")/.../base64("valueX")/
`-- diff
`-- info
```

WDYT?

drivers/overlay/overlay.go
```diff
@@ -316,6 +317,33 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+		case "layerstore", "additionallayerstore":
+			logrus.Debugf("overlay: layerstore=%s", val)
+			// Additional read only layer stores to use for lower paths
+			if val == "" {
+				continue
+			}
+			for _, store := range strings.Split(val, ",") {
+				elems := strings.Split(val, ":")
+				if len(elems) < 2 {
+					return nil, fmt.Errorf("overlay: layerstore config must be <path>:<anntation1>:<anntation2>:...:<anntationX>")
```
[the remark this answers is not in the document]
ktock: Fixed this.

store.go
```diff
@@ -488,6 +489,15 @@ type Store interface {
 
 	// GetDigestLock returns digest-specific Locker.
 	GetDigestLock(digest.Digest) (Locker, error)
+
+	// LayerFromAdditionalLayerStore searches layers from the additional layer store and
+	// returns the Layer obeject. Note that this hasn't been stored to this store yet so
```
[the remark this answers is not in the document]
ktock: Fixed this.

store.go
```diff
@@ -3076,6 +3086,83 @@ func (s *store) Layer(id string) (*Layer, error) {
...
+func (s *store) LayerFromAdditionalLayerStore(d digest.Digest, labels map[string]string) (*Layer, error) {
+	adriver, ok := s.graphDriver.(drivers.AdditionalLayerStoreDriver)
+	if !ok {
+		return nil, ErrLayerUnknown
+	}
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	for _, ls := range adriver.AdditionalLayerStores() {
+		labels[defaultAdditionalLayerStoreDigestLabel] = d.String()
```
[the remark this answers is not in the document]
ktock: Fixed this.

drivers/driver.go
giuseppe: > @giuseppe I initally thought we can use annotations for content verification: passing the digest of TOC from the runtime to the filesystem for ensuring that the (lazily) pulled contents is ones expected by the runtime. But I currently think we can let the filesystem do all these tasks (i.e. fetching manifest, getting TOC digest and verifying each chunk all on the filesystem's side).

I think this is a good idea.  I didn't know of why you were doing it.

My only comment is about storing this kind of metadata somewhere else (maybe just a json file?), so that in future we can extend it if we need, and that will simplify the path to just store the digest without any annotation.

```
<ALS root>/checksum-for-the-layer/
-- diff
-- info
-- info/metadata
```

That would work in a similar way to what I am doing with the Layers Big Data PR.

WDYT?

ktock: @giuseppe Thanks for the suggestion. Please let me clarify my understanding about `info/metadata`. Which of the following did you mean?

A) The runtime writes some additional information (TOC digest, in this case) to `info/metadata` then access to `diff`. The filesystem reads `info/metadata` written by the runtime, does some additional tasks (verifying TOC, in this case) and shows the exploded diff to `diff` directory.

B) The filesystem shows `info/metadata` containing some additional information (TOC digest, in this case) to the runtime. The runtime does some tasks (verifying TOC, in this case) before it uses `diff` directory. (In this case, image reference still needs to be contained in the path for enabling the filesystem to query layer data to the registry)

giuseppe: Should the runtime store such information?   Would it not be the file system responsibility to fetch the TOC digest and do the verification?

I was thinking how the file system could expose this metadata to the runtime, if the runtime needs to access it (and we could probably look at storing it in a format compatible with the SetBigData PR I am working on)

ktock: > Would it not be the file system responsibility to fetch the TOC digest and do the verification?

I agree with you. We should make verification filesystem's responsibility.

> I was thinking how the file system could expose this metadata to the runtime, if the runtime needs to access it (and we could probably look at storing it in a format compatible with the SetBigData PR I am working on)

Currently I don't immediately come up with the usecase where the fileystem needs to expose metadata (because if we  do the verification on the filesystem's side, we don't need to expose TOC digest to the runtime anymore). But if we'll need to do this, I think we can make it compatible to layer BigData in the way you mentioned in https://github.com/containers/storage/pull/795#discussion_r560063580.

giuseppe: thanks.

Yes, so I think we can avoid exposing metadata for now.  If it is ever required we can add it later.

ktock:
@giuseppe Fixed the path design based on https://github.com/containers/storage/pull/795#discussion_r560063580.

The default layer search path on ALS:
```
<ALS root>/<layer-digest>/
-- diff
-- info
```

- `diff` shows the extracted layer diff.
- `info` contains `*c/storage.Layer` struct.

When `ref` option is specified for ALS through `layerstore` option (e.g. `"/tmp/storage:ref"`), the image reference is contained to that path as well. This is needed for registry-backed file systems (eStargz/zstd:chunked) that require image references for querying the layer contents.

```
<ALS root>/<image-reference>/<layer-digest>
-- diff
-- info
```

Now we can pull and run the image with reusing layer from ALS. The example implementation of the filesystem for ALS is https://github.com/ktock/stargz-snapshotter/tree/als-pool-example.

The remaining limitation is that ALS layers currently cannot be exported (i.e. saved or pushed) because:
- It's difficult to reproduce the tar blob that has the exact same digest written in the manifest (".tar-split.gz" files aren't provided by ALS) from the `diff` directory.
- Registry-backed filesystem cannot fill `c/storage.Layer.UncompressedSize` of `info` file (as described in "Known limitation and request for comments" section in the above PR description).
  - Possible workaround is to contain `UncompressedSize` in the eStargz/zstd:chunked image as an OCI Annotation.

The current workaround for this limitation:
- Podman: When we export layers using podman, we need to remove the image and re-pull this without `--layerstore` option.
- CRI-O: This won't be a matter for CRI-O because we don't export layers on Kubernetes.

I think we can work on this issue on the following PRs but please tell me if we should work on them in this PR as well.

ktock:
Can we move this forward?

drivers/overlay/overlay.go
```diff
@@ -1200,8 +1252,26 @@ func (d *Driver) ApplyDiff(id, parent string, options graphdriver.ApplyDiffOpts)
 
 func (d *Driver) getDiffPath(id string) string {
 	dir := d.dir(id)
+	diffPath := path.Join(dir, "diff")
+	if ld, err := os.Readlink(diffPath); err == nil {
```
giuseppe, line 1256: how would a relative symlink work in this case?  Do we need to raise an error if the symlink is not absolute?

drivers/overlay/overlay.go
```diff
@@ -1200,8 +1252,26 @@ func (d *Driver) ApplyDiff(id, parent string, options graphdriver.ApplyDiffOpts)
...
+	layers, err := d.getLowerDirs(id)
+	if err != nil {
+		return nil, err
+	}
+	for i, l := range layers {
+		if ld, err := os.Readlink(l); err == nil {
+			// `diff` of additional layer store is symlink to the actual `diff` directory.
+			layers[i] = ld
+		}
+	}
```
giuseppe, line 1273: same here, how would we handle relative symlinks?

giuseppe (commented):
I usually experiment with new features touching containers/image and containers/storage in skopeo.

Do you have any example on how the new API is going to be used?

EDIT: we don't need to support any existing file system, it is enough to create the remote directory manually, as long as we have a clear idea of how it can be used

drivers/overlay/overlay.go
ktock: fixed to allow only absolute paths.

drivers/overlay/overlay.go
ktock: fixed to allow only absolute paths.

ktock:
I list examples of the current patch with c/storage-related commands.

## Manually preparing additional layer stores

We can manually create an additional layers store using something like the following script:

```bash
#!/bin/bash

set -euo pipefail

ORG="${1}"
STORE="${2}"

if [ "${1}" == "--ref" ] ; then
  ORG="${2}"
  STORE="${3}/$(echo -n ${2} | base64)"
fi

OCI="$(mktemp -d)"
skopeo copy docker://${ORG} oci://${OCI}
cat ${OCI}/blobs/sha256/$(cat ${OCI}/index.json | jq -r '.manifests[0].digest' | sed 's/sha256://') \
  | jq -r '.layers[].digest' | sed 's/sha256://' | while read DGST ; do
      mkdir -p ${STORE}/${DGST}/diff && \
        tar -xf ${OCI}/blobs/sha256/${DGST} -C ${STORE}/${DGST}/diff && \
        cat <<EOF > ${STORE}/${DGST}/info
{
  "compressed-diff-digest": "sha256:${DGST}",
  "compressed-size": $(stat --printf="%s" ${OCI}/blobs/sha256/${DGST}),
  "diff-digest": "sha256:$(cat ${OCI}/blobs/sha256/${DGST} | gunzip | sha256sum | sed -E 's/([^ ]*).*/\1/g')",
  "diff-size": $(cat ${OCI}/blobs/sha256/${DGST} | gunzip | wc -c),
  "compression": 2
}
EOF
done
```

Here we prepare two stores.
The first one is ALS with default path (`<ALSroot>/<digest>`) and the second one contains the image reference in the path (`<ALSroot>/base64(<reference>)/<digest>`).

```
/create_store.sh ghcr.io/stargz-containers/postgres:13.1-esgz /tmp/teststore
/create_store.sh --ref ghcr.io/stargz-containers/postgres:13.1-esgz /tmp/teststore_ref
```

### config for default layer store:

```
cat <<EOF > /etc/containers/storage.conf
[storage]
driver = "overlay"

[storage.options]
additionallayerstores = ["/tmp/teststore"]
EOF
```

### config for the layer store with "ref" option:

```
cat <<EOF > /etc/containers/storage.conf
[storage]
driver = "overlay"

[storage.options]
additionallayerstores = ["/tmp/teststore_ref:ref"]
EOF
```

## Running c/storage, c/image -based tools with the additional layer stores

For both of above configurations, we can use these stores with cri-o, podman, skopeo, etc.

### CRI-O

```
# crio
# time crictl pull ghcr.io/stargz-containers/postgres:13.1-esgz
...
real	0m5.502s
user	0m0.021s
sys	0m0.004s
```

### Podman

NOTE: Export-related commands (e.g. `podman save`) are not supported.

```
# podman system prune -a
# time podman pull ghcr.io/stargz-containers/postgres:13.1-esgz
...
real	0m4.433s
user	0m0.222s
sys	0m0.137s
# podman run --rm -it ghcr.io/stargz-containers/postgres:13.1-esgz /bin/echo 'Hello, World!'
Hello, World!
# podman save ghcr.io/stargz-containers/postgres:13.1-esgz > /tmp/test.tar
Error: unable to save "ghcr.io/stargz-containers/postgres:13.1-esgz": error copying image to the remote destination: Error writing blob: error happened during read: Digest did not match, expected sha256:0167446b81cb115e90d668017ed409b449aee3a8fade44c7a96d36ebb0937ca5, got sha256:0a45da8cdff8388828644ad2052e91595733e558067b60b9f4bd290035964193
```

### Skopeo

NOTE: Export-related commands (e.g. `scopeo copy` from storage to somewhere) are not supported.

```
# skopeo copy docker://ghcr.io/stargz-containers/postgres:13.1-esgz containers-storage:myimage
...
real	0m4.335s
user	0m0.216s
sys	0m0.138s
# skopeo inspect containers-storage:myimage
# skopeo inspect --config containers-storage:myimage
# skopeo copy containers-storage:myimage oci:///tmp/ocidump
FATA[0004] Error writing blob: error happened during read: Digest did not match, expected sha256:0167446b81cb115e90d668017ed409b449aee3a8fade44c7a96d36ebb0937ca5, got sha256:0a45da8cdff8388828644ad2052e91595733e558067b60b9f4bd290035964193 
# ./bin/skopeo delete containers-storage:myimage
```

## Mounting and using PoC filesystem

https://github.com/ktock/stargz-snapshotter/tree/als-pool-example/cmd/registry-storage

The following mounts current PoC filesystem (esgz-based) to `/tmp/teststore_ref`.
NOTE: the implementation is highly experimental

```
# git clone -b als-pool-example https://github.com/ktock/stargz-snapshotter /tmp/registry-storage
# cd /tmp/registry-storage
# make registry-storage
# mkdir /tmp/registry
# ./out/registry-storage /tmp/registry
```

This mounted store can be used in the same ways as manually-created stores.
Registry-backed stores (esgz/zstd:chunked-based ones) require image reference in the path (thus `ref` option is used here) for querying diff contents to the registry API that requires image ref string in addition to the digest.

```
# cat <<EOF > /etc/containers/storage.conf
[storage]
driver = "overlay"

[storage.options]
additionallayerstores = ["/tmp/registry:ref"]
EOF
# crio
# crictl pull ghcr.io/stargz-containers/python:3.7-esgz
```

giuseppe:
thanks!

Do I need any patch for Podman?  I've tried the suggestion above but "podman pull" still pulls the entire image

ktock:
> Do I need any patch for Podman?

No patch is needed to c/podman. It just need to be compiled with the patched c/image (https://github.com/containers/image/pull/1109) and the patched c/storage (https://github.com/containers/storage/pull/795) so go.mod and Makefile need to be modified a bit.

giuseppe:
> No patch is needed to c/podman. It just need to be compiled with the patched c/image ([containers/image#1109](https://github.com/containers/image/pull/1109)) and the patched c/storage (#795) so go.mod and Makefile need to be modified a bit.

thanks, so it seems the :ref version doesn't work for me, while the version without :ref works

```
$ ls /tmp/teststore_ref/
'Z2hjci5pby9zdGFyZ3otY29udGFpbmVycy9wb3N0Z3JlczoxMy4xLWVzZ3o='

$ grep -B1 additionallayerstores ~/.config/containers/storage.conf 
  [storage.options]
  additionallayerstores = ["/tmp/teststore_ref:ref"]
```

giuseppe:
I'll play more with it and update with more details, it is probably a mistake on my side

giuseppe:
sorry for the noise, the issue was on my side.  It works fine now.

Do you think containers/storage should signal somehow that the layer is not used anymore?  A `rmdir` would not cause the directory to go away when it is empty, but a FUSE file system could catch the syscall and decide whether to delete or not the layer.

What do you think?

ktock:
@giuseppe I agree with the notification can be done using rmdir. `c/storage/drivers/overlay.Remove()` will be a good place to do this.

However, the layer directory possibly is shared (symlinked) by multiple storages. So isn't it hard to determine whether it's safe to cleanup this only with "cleanup" notification?
I think another type of notification is also needed when the driver starts to use a layer (e.g. on `c/storage/drivers/overlay.CreateFromDirectory()`). Or, xattr might be used as a reference counter.

WDYT?

drivers/driver.go
mtrmac: > We should make verification filesystem's responsibility.

So how is that going to work? The data available to `PutFromAdditionalLayerStore` is a single digest; does the filesystem read all of the layer to verify the digest? If I understand correctly, that would defeat the entire purpose.

mtrmac:
> It's difficult to reproduce the tar blob that has the exact same digest written in the manifest (".tar-split.gz" files aren't provided by ALS) from the diff directory.

If the magic backing filesystem can read blobs on-demand, can’t it just read all of it and provide that stream?

---

The lack of `UncompressedSize` is somewhat of an issue anyway, if `TryReusingBlob` returns incorrect data, it could show up in that incorrect data recorded in c/storage image manifests.

layers.go
```diff
@@ -602,6 +606,66 @@ func (r *layerStore) Status() ([][2]string, error) {
...
+	var diffDir string
+	var layerInfo *Layer
+	for _, ls := range adriver.AdditionalLayerStores() {
+		if layerInfo, diffDir, err = searchLayer(ls, d, imageref); err == nil {
+			break
+		} else if err != ErrLayerUnknown {
+			return nil, err
+		}
+	}
+
```
mtrmac, line 624: The code can get here without setting `layerInfo` and `diffDir`, e.g. if there are no ALS.

layers.go
```diff
@@ -602,6 +606,66 @@ func (r *layerStore) Status() ([][2]string, error) {
...
+		}
+		names[name] = layer
+	}
+	// TODO: check if necessary fields are filled
+	r.layers = append(r.layers, layer)
+	r.idindex.Add(id)
+	r.byid[id] = layer
+	r.byname = names
+	r.bycompressedsum = compressedsums
+	r.byuncompressedsum = uncompressedsums
```
mtrmac, line 661: AFAICS this discards a lot of data about all the other layers. How does that work??!

store.go
```diff
@@ -3682,3 +3753,30 @@ func (s *store) Free() {
 		}
 	}
 }
+
+func searchLayer(ls drivers.AdditionalLayerStore, d digest.Digest, ref string) (*Layer, string, error) {
+	var elems []string
+	if ls.WithReference {
+		elems = append(elems, base64.StdEncoding.EncodeToString([]byte(ref)))
+	}
+	elems = append(elems, d.Encoded())
```
mtrmac, line 3762: Doesn’t this discard the algorithm ID? Why is that safe?

store.go
```diff
@@ -3682,3 +3753,30 @@ func (s *store) Free() {
...
+	target := filepath.Join(append([]string{ls.Path}, elems...)...)
+	if _, err := os.Stat(filepath.Join(target, "diff")); err == nil {
```
mtrmac, line 3764: The filesystem accesses here feel weird to me.

If it is the generic code that defines the layout and file format, and the drivers must accept absolute paths to arbitrary locations, why is the driver in charge of maintaining `driver.AdditionalLayerStores()`  at all, and it isn’t a generic c/storage option?

Alternatively, shouldn’t the driver be fully in charge of the filesystem layout, so that the generic code only calls “find”/“use” operations provided by the driver?

I’m leaning towards the second view — why does the generic code care about a `diff` subdirectory at all, when that’s an overlay-driver-specific concept?

mtrmac (commented):
Most importantly, how does digest verification happen?

(Reminder: I know very little about c/storage .)

ktock:
@mtrmac 

Thanks for the comments.

> If the magic backing filesystem can read blobs on-demand, can’t it just read all of it and provide that stream?

The filesystem provides the extracted view of the layer so the runtime needs to walk the directory tree and re-archive/compress them.
In this case, isn't it hard to provide the layer blob that is exactly same as the one contained in the image (e.g. the order of entries can't be guaranteed)?

The possible idea to solve this to introduce `/<ALS>/<digest>/stream` file where the filesystem provides the exact layer blob of that image.
I'll add this if this idea LGTY.

> Most importantly, how does digest verification happen?

Content verification is done in the filesystem's side (please see also discussion in https://github.com/containers/storage/pull/795#discussion_r561702466 ).
The filesystem verifies each lazily-fetched chunk and provides the verified one to the client.

drivers/overlay/overlay.go
```diff
@@ -316,6 +317,36 @@ func parseOptions(options []string) (*overlayOptions, error) {
 				}
 				o.imageStores = append(o.imageStores, store)
 			}
+		case "layerstore", "additionallayerstore":
```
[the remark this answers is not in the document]
ktock: Fixed this (picked `additionallayerstore`).

drivers/overlay/overlay.go
```diff
@@ -316,6 +317,36 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+				}
+				if !st.IsDir() {
+					return nil, fmt.Errorf("overlay: layerstore path %q must be a directory", store)
+				}
+				var withReferece bool
+				for _, e := range elems[1:] {
+					if e == "ref" {
+						withReferece = true
+					}
+				}
```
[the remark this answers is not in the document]
ktock: Fixed to return an error on that case.

drivers/overlay/overlay.go
```diff
@@ -316,6 +317,36 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+					return nil, fmt.Errorf("overlay: layerstore path %q is not absolute.  Can not be relative", store)
+				}
+				st, err := os.Stat(store)
+				if err != nil {
+					return nil, fmt.Errorf("overlay: can't stat layerStore dir %s: %v", store, err)
+				}
+				if !st.IsDir() {
+					return nil, fmt.Errorf("overlay: layerstore path %q must be a directory", store)
+				}
+				var withReferece bool
```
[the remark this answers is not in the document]
ktock: Fixed this.

layers.go
ktock: Added a check to avoid this case.

layers.go
ktock: Fixed this.

store.go
ktock: Fixed this to preserve the algorithm ID.

store.go
ktock: > Alternatively, shouldn’t the driver be fully in charge of the filesystem layout, so that the generic code only calls “find”/“use” operations provided by the driver?

Thanks for the suggestion. I think it's possible by adding something like the following API to the driver. I'll add this if this idea LGTY.

- `CreateFromAdditionalLayerStore(id, parent string) error` : creates a layer using the diff contents from the additional layer store.
- `AdditionalLayerInfo(d digest.Digest, ref string) (io.ReadCloser, error)` : provides `info` file contents. The store will decode this to `*c/storage.Layer` struct.

mtrmac:
> > Most importantly, how does digest verification happen?
> 
> Content verification is done in the filesystem's side (please see also discussion in [#795 (comment)](https://github.com/containers/storage/pull/795#discussion_r561702466) ).
> The filesystem verifies each lazily-fetched chunk and provides the verified one to the client.

Against what are the chunks verified? Where does the trust start? #775 , consuming (if I understand it correctly) the same format, needs to rely on
```go
	manifestChecksumKey = "io.containers.zstd-chunked.manifest-checksum"
 	manifestInfoKey     = "io.containers.zstd-chunked.manifest-position"
```

layers.go
mtrmac: It can still happen if `searchLayer` never finds anything.

layers.go
mtrmac: Looks better now (but needs a review by someone who knows c/storage, i.e. not me).

store.go
mtrmac: I’ll defer to c/storage maintainers. (Another design concern: Should the API, in advance, account for multiple formats of the on-demand-created data?)

giuseppe:
> @giuseppe I agree with the notification can be done using rmdir. `c/storage/drivers/overlay.Remove()` will be a good place to do this.
> 
> However, the layer directory possibly is shared (symlinked) by multiple storages. So isn't it hard to determine whether it's safe to cleanup this only with "cleanup" notification?
> I think another type of notification is also needed when the driver starts to use a layer (e.g. on `c/storage/drivers/overlay.CreateFromDirectory()`). Or, xattr might be used as a reference counter.
> 
> WDYT?

yes I think we also need a way to increase the reference counter (what syscall can we use for it?  mkdir?)  The xattr idea LGTM, although that would be an implementation detail for the file system.

How is garbage collection implemented in the containerd remote snapshotter?

layers.go
ktock: Moved the logic into overlay driver (https://github.com/containers/storage/blob/05a4820e7372d0faba1d2c09ef78fd1226b495f4/drivers/overlay/overlay.go#L1490-L1513) and fixed this.

store.go
ktock: Moved the directory-structure-aware logic into overlay driver. The driver's API is format-agnostic I believe.

ktock:
@mtrmac 

>>> Most importantly, how does digest verification happen?

>> Content verification is done in the filesystem's side (please see also discussion in #795 (comment) ).

>> The filesystem verifies each lazily-fetched chunk and provides the verified one to the client.

> Against what are the chunks verified?

I'm thinking about that the filesystem exposes each layer's TOC digest (i.e. `io.containers.zstd-chunked.manifest-checksum`) through `info` file or the big data file. Thus:

Each chunk is verified by the filesystem, using the digest written in the TOC
The digest of this TOC is exposed to `c/storage` through `info` or the big data file and `c/storage` verifies this by making sure that this is the same as the one written in the (already verified) OCI manifest.

WDYT?

@giuseppe

> How is garbage collection implemented in the containerd remote snapshotter?

Containerd has its own GC system on top of its central metadata DB and remote snapshotter relies on it (https://github.com/containerd/containerd/blob/master/docs/garbage-collection.md).
During GC, containerd queries snapshotter to clean up (remove/unmount) some snapshots and the snapshotter does so.

-----

Memo about other TODOs

- Add `stream` API to ALS for exposing the layer tar (https://github.com/containers/storage/pull/795#issuecomment-773259555)

giuseppe:
> > How is garbage collection implemented in the containerd remote snapshotter?
> 
> Containerd has its own GC system on top of its central metadata DB and remote snapshotter relies on it (https://github.com/containerd/containerd/blob/master/docs/garbage-collection.md).
> During GC, containerd queries snapshotter to clean up (remove/unmount) some snapshots and the snapshotter does so.

thanks.  Do you think it make sense to just use `mkdir` as opposed to `rmdir` to signal the file system to increase the ref counter?

giuseppe:
sorry I've clicked on "Update branch" by mistake.  I've pushed again the old commit

ktock:
@giuseppe Mkdir isn't propagated to FUSE daemon if the directory already exists. So, instead, used `create` on `/<ALS>/.../<layerdigest>/use` file as the signal to increase the ref counter. Filesystems that care the reference counter must fail to `create` on `/<ALS>/.../<layerdigest>/use` and increment the reference counter.

giuseppe:
> @giuseppe Mkdir isn't propagated to FUSE daemon if the directory already exists. So, instead, used `create` on `/<ALS>/.../<layerdigest>/use` file as the signal to increase the ref counter. Filesystems that care the reference counter must fail to `create` on `/<ALS>/.../<layerdigest>/use` and increment the reference counter.

thanks, it make sense.

@mtrmac do you have any blocking issue with this PR?

ktock:
Rebased. Can we move this forward?

giuseppe:
@mtrmac @rhatdan @nalind PTAL

giuseppe:
how can we move forward with this one?

ktock:
@giuseppe Added support of lazy pulling of zstd:chunked to the current additional layer store implementation (https://github.com/ktock/stargz-snapshotter/tree/als-pool-example).
The following benchmarking result measured on GitHub Actions (`Standard_DS2_v2` instance at `eastus2`) shows additional layer store improves pull performance. CRI-O should also show a similar result (not measured yet).

I'll work on experiment on a richer instance with larger number of samples.

https://github.com/ktock/stargz-snapshotter/actions/runs/623653929

result

drivers/driver.go
giuseppe: The file system starts with the TOC and its digest (through an annotation).  Once the TOC is verified, each file is referenced from there and can be verified separately.

@ktock any further details you'd like to add?

giuseppe: nevermind my last question, I see you added a comment with more details

drivers/overlay/overlay.go
```diff
@@ -1486,3 +1595,78 @@ func nameWithSuffix(name string, number int) string {
...
+		if ls.withReference {
+			elems = append(elems, refelem)
+		}
+		elems = append(elems, dgst.String())
+		target := filepath.Join(append([]string{ls.path}, elems...)...)
+		// Check if all necessary files exist
+		for _, p := range []string{
+			filepath.Join(target, "diff"),
+			filepath.Join(target, "info"),
+			// filepath.Join(target, "stream"), // TODO
```
giuseppe, line 1612: let's drop this dead code

giuseppe:
@mtrmac are you fine with the PR as it is now?

giuseppe:
@ktock could you please add a comment on each new public function saying the API is still experimental and that can be changed without bumping the major version number for containers/storage?

drivers/overlay/overlay.go
```diff
@@ -370,6 +384,38 @@ func parseOptions(options []string) (*overlayOptions, error) {
 				}
 				o.imageStores = append(o.imageStores, store)
 			}
+		case "additionallayerstore":
+			logrus.Debugf("overlay: additionallayerstore=%s", val)
+			// Additional read only layer stores to use for lower paths
+			if val == "" {
+				continue
+			}
+			for _, store := range strings.Split(val, ",") {
```
rhatdan, line 393: Nit, can you call this a lstore rather then a store.

drivers/overlay/overlay.go
```diff
@@ -370,6 +384,38 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+				if err != nil {
+					return nil, fmt.Errorf("overlay: can't stat additionallayerstore dir %s: %v", store, err)
+				}
+				if !st.IsDir() {
+					return nil, fmt.Errorf("overlay: additionallayerstore path %q must be a directory", store)
+				}
+				var withReference bool
+				for _, e := range elems[1:] {
+					if e == "ref" {
+						withReference = true
```
rhatdan, line 409: TO be anal here, we could return an error `if withReference == true`

drivers/overlay/overlay.go
```diff
@@ -594,6 +640,23 @@ func (d *Driver) Cleanup() error {
...
+func (d *Driver) LookupAdditionalLayer(dgst digest.Digest, ref string) (graphdriver.AdditionalLayer, error) {
+	l, err := d.getAdditionalLayerPath(dgst, ref)
+	if err != nil {
+		return nil, err
+	}
+	// tell the additional layer store that we use this layer.
+	if f, err := os.Create(path.Join(l, "use")); err == nil {
+		f.Close()
+	}
+	return &additionalLayer{
```
rhatdan, line 654: Should we warn here if their is an error?  If the Error is not ENOEXIST?

drivers/overlay/overlay.go
```diff
@@ -886,6 +949,12 @@ func (d *Driver) Remove(id string) error {
 		}
 	}
 
+	al, err := ioutil.ReadFile(path.Join(dir, "additionallayer"))
+	if err == nil {
```
rhatdan, line 953: Should we warn here if their is an error? If the Error is not ENOEXIST?

drivers/overlay/overlay.go
```diff
@@ -1321,10 +1396,34 @@ func (d *Driver) ApplyDiff(id, parent string, options graphdriver.ApplyDiffOpts)
 	return directory.Size(applyDir)
 }
 
-func (d *Driver) getDiffPath(id string) string {
+func (d *Driver) getDiffPath(id string) (string, error) {
 	dir := d.dir(id)
+	diffPath := path.Join(dir, "diff")
+	if ld, err := os.Readlink(diffPath); err == nil {
```
rhatdan, line 1402: What do we do if this err!=nil?

drivers/overlay/overlay.go
```diff
@@ -1321,10 +1396,34 @@ func (d *Driver) ApplyDiff(id, parent string, options graphdriver.ApplyDiffOpts)
...
+}
 
-	return path.Join(dir, "diff")
+func (d *Driver) getLowerDiffPaths(id string) ([]string, error) {
+	layers, err := d.getLowerDirs(id)
+	if err != nil {
+		return nil, err
+	}
+	for i, l := range layers {
+		if ld, err := os.Readlink(l); err == nil {
```
rhatdan, line 1418: Should we report error if err !=nil?

drivers/overlay/overlay.go
```diff
@@ -1486,3 +1595,80 @@ func nameWithSuffix(name string, number int) string {
...
+			elems = append(elems, refelem)
+		}
+		elems = append(elems, dgst.String())
+		target := filepath.Join(append([]string{ls.path}, elems...)...)
+		// Check if all necessary files exist
+		for _, p := range []string{
+			filepath.Join(target, "diff"),
+			filepath.Join(target, "info"),
+			// TODO(ktock): We should have an API to expose the stream data of this layer
+			//              to enable the client to retrieve the entier contents of this
```
rhatdan, line 1613: nit: s/entier/entire/g

drivers/overlay/overlay.go
```diff
@@ -1486,3 +1595,80 @@ func nameWithSuffix(name string, number int) string {
...
+			//              layer when it exports this layer.
+		} {
+			if _, err := os.Stat(p); err != nil {
+				return "", graphdriver.ErrLayerUnknown
```
rhatdan, line 1617: Can you wrap the error.

drivers/overlay/overlay.go
```diff
@@ -1486,3 +1595,80 @@ func nameWithSuffix(name string, number int) string {
...
+	return os.Open(filepath.Join(al.path, "info"))
+}
+
+// CreateAs creates a new layer from this additional layer.
+func (al *additionalLayer) CreateAs(id, parent string) error {
+	targetDiffDir := filepath.Join(al.path, "diff")
+	if _, err := os.Stat(targetDiffDir); err != nil {
+		return err
+	}
+	if ld, err := os.Readlink(targetDiffDir); err == nil {
```
rhatdan, line 1643: Should we return or warn if the err!=nil?  If it does not exists?

drivers/overlay/overlay.go
```diff
@@ -1486,3 +1595,80 @@ func nameWithSuffix(name string, number int) string {
...
+	// mark this layer as "additional layer"
+	if err := ioutil.WriteFile(path.Join(dir, "additionallayer"), []byte(al.path), 0644); err != nil {
+		return err
+	}
+	diffDir := path.Join(dir, "diff")
+	if err := os.RemoveAll(diffDir); err != nil {
+		return err
+	}
+	// tell the additional layer store that we use this layer.
+	if f, err := os.Create(path.Join(al.path, "use")); err == nil {
```
rhatdan, line 1662: Should we report err!=nil?

store.go
```diff
@@ -489,6 +490,29 @@ type Store interface {
...
+	// LayerFromAdditionalLayerStore searches layers from the additional layer store and
+	// returns the object for handling this. Note that this hasn't been stored to this store
+	// yet so this needs to be done through PutAs method.
+	// Releasing AdditionalLayer handler is caller's responsibility.
+	LookupAdditionalLayer(d digest.Digest, imageref string) (AdditionalLayer, error)
+}
+
+// AdditionalLayer reprents a layer that is contained in the additional layer store
+type AdditionalLayer interface {
+
```
rhatdan, line 503: Nit: Remove blank line.

rhatdan (approved):

drivers/driver.go
```diff
@@ -34,6 +35,8 @@ var (
 	ErrPrerequisites = errors.New("prerequisites for driver not satisfied (wrong filesystem?)")
 	// ErrIncompatibleFS returned when file system is not supported.
 	ErrIncompatibleFS = fmt.Errorf("backing file system is unsupported for this graph driver")
+	// ErrLayerUnknown returned when the specified layer is unknown by the driver.
+	ErrLayerUnknown = fmt.Errorf("unknown layer")
```
kolyshkin, line 39: nit: use `errors.New` as there's no `%`-style formatting in here.

Feel free to also fix the definition right above (in a separate commit).

drivers/overlay/overlay.go
```diff
@@ -114,6 +117,17 @@ type Driver struct {
...
 }
 
+type additionalLayerStore struct {
+
+	// path is the directory where this store is available on the host.
+	path string
+
+	// withReference is true when the store contains image reference information (base64-encoded)
+	// in its layer search path so the path to the diff will be:
+	// `<path>/base64(reference)/<layerdigest>/`
```
kolyshkin, line 127: nit: markdown formatting in golang comments does not make sense.

drivers/overlay/overlay.go
```diff
@@ -370,6 +384,38 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+			}
+			for _, store := range strings.Split(val, ",") {
+				elems := strings.Split(val, ":")
+				store = filepath.Clean(elems[0])
+				if !filepath.IsAbs(store) {
+					return nil, fmt.Errorf("overlay: additionallayerstore path %q is not absolute.  Can not be relative", store)
+				}
+				st, err := os.Stat(store)
+				if err != nil {
+					return nil, fmt.Errorf("overlay: can't stat additionallayerstore dir %s: %v", store, err)
```
kolyshkin, line 401: nit: the error from `os.Stat()` already includes the path being stat'ed, so no need to explicitly add it again.

drivers/overlay/overlay.go
```diff
@@ -370,6 +384,38 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+				}
+				if !st.IsDir() {
+					return nil, fmt.Errorf("overlay: additionallayerstore path %q must be a directory", store)
+				}
+				var withReference bool
+				for _, e := range elems[1:] {
+					if e == "ref" {
```
kolyshkin, line 408: Perhaps switch/case would look better here?

drivers/overlay/overlay.go
```diff
@@ -370,6 +384,38 @@ func parseOptions(options []string) (*overlayOptions, error) {
...
+						withReference = true
+					} else {
+						return nil, fmt.Errorf("overlay: additionallayerstore config %q contains unkown option %q", store, e)
```
kolyshkin, line 411: nit: typo: unkown -> unknown

drivers/overlay/overlay.go
```diff
@@ -1321,10 +1396,34 @@ func (d *Driver) ApplyDiff(id, parent string, options graphdriver.ApplyDiffOpts)
...
 }
 
-func (d *Driver) getDiffPath(id string) string {
+func (d *Driver) getDiffPath(id string) (string, error) {
 	dir := d.dir(id)
+	diffPath := path.Join(dir, "diff")
+	if ld, err := os.Readlink(diffPath); err == nil {
+		// `diff` of additional layer store is symlink to the actual `diff` directory.
+		if !path.IsAbs(ld) {
+			return "", fmt.Errorf("path of the diff directory must be absolute but %q", ld)
```
kolyshkin, line 1405: nit: 's/but %q/(got %q)/`

or something like this

drivers/overlay/overlay.go
```diff
@@ -1321,10 +1396,34 @@ func (d *Driver) ApplyDiff(id, parent string, options graphdriver.ApplyDiffOpts)
...
+func (d *Driver) getLowerDiffPaths(id string) ([]string, error) {
+	layers, err := d.getLowerDirs(id)
+	if err != nil {
+		return nil, err
+	}
+	for i, l := range layers {
+		if ld, err := os.Readlink(l); err == nil {
+			// `diff` of additional layer store is symlink to the actual `diff` directory.
+			if !path.IsAbs(ld) {
+				return nil, fmt.Errorf("path of the diff directory must be absolute but %q", ld)
```
kolyshkin, line 1421: ditto

drivers/overlay/overlay.go
```diff
@@ -1486,3 +1595,80 @@ func nameWithSuffix(name string, number int) string {
 	}
 	return fmt.Sprintf("%s%d", name, number)
 }
+
+func (d *Driver) getAdditionalLayerPath(dgst digest.Digest, ref string) (string, error) {
+	refelem := base64.StdEncoding.EncodeToString([]byte(ref))
```
kolyshkin, line 1600: nit: `refElem`

drivers/overlay/overlay.go
```diff
@@ -1486,3 +1595,80 @@ func nameWithSuffix(name string, number int) string {
...
+	for _, ls := range d.options.layerStores {
+		var elems []string
+		if ls.withReference {
+			elems = append(elems, refelem)
+		}
+		elems = append(elems, dgst.String())
+		target := filepath.Join(append([]string{ls.path}, elems...)...)
```
kolyshkin, line 1607: This code is not super readable (to me). What about something like
```go
ref := ""
if ls.withReference {
   ref = refElem
}
target := path.Join(ls.path, ref, dgst.String())
```
instead? (untested)

drivers/overlay/overlay.go
kolyshkin: We have already checked that it exists above (but haven't checked permissions).

Perhaps it's better to remove `os.Stat` above and use this to return all errors except EINVAL (i.e. when the argument is not a symlink).

drivers/overlay/overlay.go
```diff
@@ -1486,3 +1595,80 @@ func nameWithSuffix(name string, number int) string {
...
+	// tell the additional layer store that we use this layer.
+	if f, err := os.Create(path.Join(al.path, "use")); err == nil {
+		f.Close()
+	}
+	return os.Symlink(targetDiffDir, diffDir)
+}
+
+// Release tells the additional layer store that we don't use this handler.
+func (al *additionalLayer) Release() {
+	al.closeOnce.Do(func() {
```
kolyshkin, line 1670: 1. Why is this called `closeOnce` it the `Rmdir` (but not `Close`) is wrapped into it?

2. I don't understand why the syncOnce here? Is repeated `Rmdir` can cause some harm?

layers.go
```diff
@@ -610,6 +614,58 @@ func (r *layerStore) Status() ([][2]string, error) {
 	return r.driver.Status(), nil
 }
 
+func (r *layerStore) PutAdditionalLayer(id string, parentLayer *Layer, names []string, aLayer drivers.AdditionalLayer) (layer *Layer, err error) {
+	if duplicateLayer, idInUse := r.byid[id]; idInUse {
```
kolyshkin, line 618: (I know this is not something that you've added but) should a map access be protected with a mutex?

layers.go
```diff
@@ -610,6 +614,58 @@ func (r *layerStore) Status() ([][2]string, error) {
...
+		return duplicateLayer, ErrDuplicateID
+	}
+	for _, name := range names {
+		if _, nameInUse := r.byname[name]; nameInUse {
```
kolyshkin, line 622: ditto

store.go
```diff
@@ -3134,6 +3158,92 @@ func (s *store) Layer(id string) (*Layer, error) {
...
+	}
+	rlstores, err := al.s.ROLayerStores()
+	if err != nil {
+		return nil, err
+	}
+
+	var parentLayer *Layer
+	if parent != "" {
+		for _, s := range append([]ROLayerStore{rlstore}, rlstores...) {
+			lstore := s
```
kolyshkin, line 3220: What is the meaning of this (given the fact you're not using `s` anywhere else below)?

drivers/overlay/overlay.go
ktock: Please see also: https://github.com/containers/storage/pull/795#issuecomment-771594370

c/storage sends notification when it starts to use a layer and when it releases that layer. Additional layer store can internally manage reference counter based on these notifications and can perform GC when nobody refers to a layer. Creating `<Store>/<layerdigest>/use` file notifies that this layer is used and rmdir-ing `<Store>/<layerdigest>` notifies that this layer is no longer used.

The mentioned `Rmdir` decreases the reference counter originally increased by *`(c/storage/drivers/overlay).Driver.LookupAdditionalLayer()`. Calling this twice sends notification twice, which breaks reference management on the Additional Layer Store's side.

layers.go
ktock: `*layerStore` is proetected by `*layerStore.Lock()` and `*layerStore.Unlock()`.

ktock:
@rhatdan @kolyshkin 

Thanks for your review.
Fixed except where I commented.

---

@giuseppe @mtrmac

The followings are remaining Known Limitations (https://github.com/containers/storage/pull/795#issuecomment-773259555)
Please let me know if the following should be included in this PR or can be separated (following-up) PRs.

- Layers in Additional Layer Store cannot be exported (i.e. saved or pushed) because:
  - It's difficult to reproduce the tar blob that has the exact same digest written in the manifest (".tar-split.gz" files aren't provided by ALS) from the extracted layer diff directory.
  - Workarounds:
    - Podman: When we export layers using podman, we need to remove the image and re-pull this without `additionallayerstore` config.
    - CRI-O: This won't be an issue for CRI-O because we don't export layers on Kubernetes.
  - Possible solution:
    - introducing `/<store>/<digest>/stream` file where the filesystem provides the original (archived) layer blob of that image.

- Runtime cannot verify lazy-pulled chunks
  - Though runtime (Podman/CRI-O) currently passes layer digest to Additional Layer Store for querying a layer, registry-backed store requires manifest digest or TOC digest for verifying each chunk.
    - https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/docs/verification.md
  - Possible solution A: Passing manifest digest from c/storage to Additional Layer Store via path element.
    - Adding `manifestdigest` option to `additionallayerstore` config. This makes c/storage search the following path on Additional Layer Store.
      - `/<store>/<manifestdigest>/<layerdigest>/diff`
    - Additional Layer Store can use this digest for verifying each chunk.
  - Possible solution B: Passing manifest digest from Additional Layer Store to c/storage via big data file or something.
    - c/storage makes sure that the passed manifest digest is the expected one.

giuseppe:
@kolyshkin are you fine with the last version?

ktock:
Can we move this forward?

giuseppe:
@kolyshkin can you please take another look?

ktock:
How is this going?
Are there anything should be fixed?

rhatdan:
LGTM

giuseppe (approved):
LGTM

giuseppe:
@ktock merged!

ktock:
Thanks! :tada:

chvish:
Hey @ktock, this is super cool. I have been looking for something like this for a while now.

Quick question: Does the demo from above still work? I have been trying to get podman to pull an image from the ALS, but it stills pulls from the remote registry even when the ALS is properly configured.
```
$ ./bin/podman info -f json | jq .store.graphOptions
{
  "overlay.additionallayerstore": "/tmp/teststore",
  "overlay.mount_program": {
    "Executable": "/usr/bin/fuse-overlayfs",
    "Package": "fuse-overlayfs: /usr/bin/fuse-overlayfs",
    "Version": "fusermount3 version: 3.9.0\nfuse-overlayfs: version 0.7.6\nFUSE library version 3.9.0\nusing FUSE kernel interface version 7.31"
  }
}

```

ktock:
@chvish Are you talking about https://github.com/containers/storage/pull/795#issuecomment-771330580 ? The layout has been updated since then. Could you try the following script instead:

```shell
#!/bin/bash

set -euo pipefail

ORG="${1}"
STORE="${2}"

if [ "${1}" == "--ref" ] ; then
    ORG="${2}"
    STORE="${3}/$(echo -n ${2} | base64)"
fi

OCI="$(mktemp -d)"
skopeo copy docker://${ORG} oci://${OCI}
cat ${OCI}/blobs/sha256/$(cat ${OCI}/index.json | jq -r '.manifests[0].digest' | sed 's/sha256://') \
    | jq -r '.layers[].digest' | while read DGST ; do
    ENCODED_DGST=$(echo -n ${DGST} | sed 's/sha256://')
    mkdir -p ${STORE}/${DGST}/diff
    tar -xf ${OCI}/blobs/sha256/${ENCODED_DGST} -C ${STORE}/${DGST}/diff
    touch ${STORE}/${DGST}/use
    cat <<EOF > ${STORE}/${DGST}/info
{
  "compressed-diff-digest": "${DGST}",
  "compressed-size": $(stat --printf="%s" ${OCI}/blobs/sha256/${ENCODED_DGST}),
  "diff-digest": "sha256:$(cat ${OCI}/blobs/sha256/${ENCODED_DGST} | gunzip | sha256sum | sed -E 's/([^ ]*).*/\1/g')",
  "diff-size": $(cat ${OCI}/blobs/sha256/${ENCODED_DGST} | gunzip | wc -c),
  "compression": 2
}
EOF
done
```

chvish:
@ktock Works like a charm thanks. 
Do we have any ETA on when Podman gets a release with the changes in `container/storage` and `containers/image`?

ktock:
@chvish Thank you for trying. CRI-O will also support this after https://github.com/cri-o/cri-o/pull/4850 is merged. So please try it too if you're interested in. The necessary configuration should be the same as Podman.

@rhatdan Thank you for starting the new release cycle!

giuseppe:
> Runtime cannot verify lazy-pulled chunks
> 
> * Though runtime (Podman/CRI-O) currently passes layer digest to Additional Layer Store for querying a layer, registry-backed store requires manifest digest or TOC digest for verifying each chunk.
>   
>   * https://github.com/containerd/stargz-snapshotter/blob/v0.4.1/docs/verification.md
> * Possible solution A: Passing manifest digest from c/storage to Additional Layer Store via path element.
>   
>   * Adding `manifestdigest` option to `additionallayerstore` config. This makes c/storage search the following path on Additional Layer Store.
>     
>     * `/<store>/<manifestdigest>/<layerdigest>/diff`
>   * Additional Layer Store can use this digest for verifying each chunk.
> * Possible solution B: Passing manifest digest from Additional Layer Store to c/storage via big data file or something.
>   
>   * c/storage makes sure that the passed manifest digest is the expected one.

I think what we can do is to change the `LookupAdditionalLayer` API to accept also the TOC Digest, so that the c/image change looks like:

```diff
diff --git a/storage/storage_dest.go b/storage/storage_dest.go
index b15c9c3b..7a56fb55 100644
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -367,7 +367,7 @@ func (s *storageImageDestination) tryReusingBlobAsPending(blobDigest digest.Dige
 
 	if options.SrcRef != nil {
 		// Check if we have the layer in the underlying additional layer store.
-		aLayer, err := s.imageRef.transport.store.LookupAdditionalLayer(blobDigest, options.SrcRef.String())
+		aLayer, err := s.imageRef.transport.store.LookupAdditionalLayer(blobDigest, options.SrcRef.String(), options.TOCDigest)
 		if err != nil && !errors.Is(err, storage.ErrLayerUnknown) {
 			return false, private.ReusedBlob{}, fmt.Errorf(`looking for compressed layers with digest %q and labels: %w`, blobDigest, err)
 		} else if err == nil {
```

and then propagate it down, until we can do something like, e.g.:

```diff
$ git diff
diff --git a/drivers/overlay/overlay.go b/drivers/overlay/overlay.go
index b2f0e7a94..95a67bc15 100644
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -2475,6 +2475,9 @@ func (al *additionalLayer) CreateAs(id, parent string) error {
        if err := os.WriteFile(path.Join(dir, "additionallayer"), []byte(al.path), 0o644); err != nil {
                return err
        }
+       if err := os.WriteFile(path.Join(dir, "toc-digest"), []byte(al.tocDigest), 0o644); err != nil {
+               return err
+       }
        notifyUseAdditionalLayer(al.path)
        return os.Symlink(filepath.Join(al.path, "diff"), diffDir)
 }
```

what do you think?

mtrmac:
AFAICS nothing ever documented that this is related to TOC digests. So, from the c/image side, I’d prefer passing down the full raw value of `annotations`, which is both:
- compatible with non-TOC users of ALS
- forward-compatible with new TOC / compression formats (and with formats c/image is otherwise uninterested in handling, where using the `TOCDigest` field might be unclear or ambiguous)

And I wonder about compatibility with ALS backend which don’t support this `toc-digest` file.

But I’ll leave both design aspects to c/storage experts.

ktock:
@giuseppe  SGTM, thanks for your suggestion!

In addition to the suggested change, Additional Layer Store implementation needs be fixed to expose the "actual" TOC digest (acquired from the layer data, etc.) via `info` file of the layer (c/storage accesses to it via `*additionalLayer.Info()`). Then c/storage  should check if the actual TOC digest (passed from Additional Layer Store) == the expected digest (passed from c/image) and make sure that Additional Layer Store uses the expected TOC.
Or, rather than directly using TOC digest, it's also ok to use the manifest digests that contains the TOC digest as a manifest annotation.

If either of the above design is ok to you, I'm willing to implement these changes.

giuseppe:
> Then c/storage should check if the actual TOC digest (passed from Additional Layer Store) == the expected digest (passed from c/image) to make sure that Additional Layer Store uses the expected TOC.

when would c/storage check for that though?  Before using the layer?

I think that the FUSE file system must fail at runtime with EIO if the provided TOC is different than what could be validated.

ktock:
> I think that the FUSE file system must fail at runtime with EIO if the provided TOC is different than what could be validated.

 How about the following design:

Introduce a new FUSE file path `<mountpoint>/base64(imageref)/<layerdigest>/toc-digest` for each layer on Additional Layer Store FUSE.

1. [`(drivers/overlay).*Driver.LookupAdditionalLayer()`](https://github.com/containers/storage/blob/fa36725dcbe22ce848a99f6345617a87c79e1574/drivers/overlay/overlay.go#L890) writes TOC Digest (expected digest passed from c/image)  to `toc-digest` after confirmed the layer exists in the store (after [`getAdditionalLayerPath()` invocation](https://github.com/containers/storage/blob/fa36725dcbe22ce848a99f6345617a87c79e1574/drivers/overlay/overlay.go#L891)).
2. Additional Layer Store checks the provied(expected) toc-digest against the actual one. If not matched, reading `<mountpoint>/base64(imageref)/<layerdigest>/info` will return EIO since this step.
3. `(drivers/overlay).*Driver.LookupAdditionalLayer()` checks if reading on `<mountpoint>/base64(imageref)/<layerdigest>/info` returns EIO. If EIO, it propagates a proper error to the caller expecting fallback to the non-lazy behaviour. (possibly with a warning log)

Other changes to `c/image` (`/storage/storage_dest.go`) are the same as you suggested at https://github.com/containers/storage/pull/795#issuecomment-2004290352

mtrmac:
Thinking about this more… from the c/image + c/storage side, every layer only has _one_ identifier; it’s _either_ the compressed blob digest, or the layer digest.

And that identifier is used for deduplication, and it is security-relevant.

So, this needs to be a rather more invasive change.

Historically, ALS was always identifying / deduplicating the layer using the (compressed) blob digest.

If the ALS is _actually_ constructing trust based on the TOC digest, and the compressed blob digest is not relevant / not verified, then the layer **must not** be identified / deduplicated based on the compressed digest (because an attacker could cause a TOC-based pull with a non-matching compressed digest X, causing future ordinary “victim” pulls of digest X to be deduplicated with the attacker’s layer).

---

So, I think we need the `LookupAdditionalLayer` API to
- already have all lookup inputs (compressed digest, TOC digest if known)
- look up _independently_ based on one or the other digest
  - it must not look up by compressed digest and then “attach” the TOC digest to the compressed digest if the compressed digest is never validated. (Looking up by the _pair_ of (compressed digest, TOC digest) would be acceptable if it never matched other layers where the TOC digest value is not set or does not match, but in that case using the compressed digest seems to have no benefit)
- return to the caller an identification of whether the layer was identified by compressed digest / uncompressed digest / TOC digest (or possibly only 2 of the 3), so that `AdditionalLayer.PutAs` can be called with an appropriate layer ID.
   - In particular, `AdditionalLayer.UncompressedDigest` must not be returning unverified data; we rely on it being correct for security.

PR #796, opened 2020-12-30
When metacopy is not supported, warn and proceed

ashokponkumar:
Signed-off-by: Ashok Pon Kumar <ashokponkumar@gmail.com>

### DIFF
--- a/drivers/overlay/check.go
+++ b/drivers/overlay/check.go
@@ -145,7 +145,7 @@ func doesMetacopy(d, mountOpts string) (bool, error) {
 		opts = fmt.Sprintf("%s,%s", opts, data)
 	}
 	if err := unix.Mount("overlay", filepath.Join(td, "merged"), "overlay", uintptr(flags), opts); err != nil {
-		return false, errors.Wrapf(err, "failed to mount overlay for metacopy check with %q options", mountOpts)
+		logrus.Warnf("failed to mount overlay for metacopy check with %q options : %v", mountOpts, err)
 	}
 	defer func() {
 		if err := unix.Unmount(filepath.Join(td, "merged"), 0); err != nil {


drivers/overlay/check.go
```diff
@@ -145,7 +145,7 @@ func doesMetacopy(d, mountOpts string) (bool, error) {
 		opts = fmt.Sprintf("%s,%s", opts, data)
 	}
 	if err := unix.Mount("overlay", filepath.Join(td, "merged"), "overlay", uintptr(flags), opts); err != nil {
-		return false, errors.Wrapf(err, "failed to mount overlay for metacopy check with %q options", mountOpts)
+		logrus.Warnf("failed to mount overlay for metacopy check with %q options : %v", mountOpts, err)
```
rhatdan, line 148: Does this actually fix it?  IE Is the option removed when this happens?  So that a future mount will not use the metacopy=on option?

b00ga:
I got here from containers/podman#8118

Posting here to provide additional context/info and to not keep adding to a closed ticket.

I'm using the kubic libcontainers OBS repo. They just upgraded from podman 2.2 to 3.0 and I started receiving these errors:
```
Error: Cannot connect to the Podman socket, make sure there is a Podman REST API service running.: failed to mount overlay for metacopy check with "nodev,metacopy=off" options: invalid argument
```

Guessing this PR may have just given me a warning and not broke podman on my system as current behavior did.
Mentioned thread states to set metacopy=off but wanted to observe that I get the same error. Needed to remove metacopy option altogether to start working again (so just `mountopt = "nodev"` in _/etc/containers/storage.conf_.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.5 LTS
Release:	18.04
Codename:	bionic
$ uname -r
4.15.0-135-generic
```

rhatdan:
Right the error is that the metacopy option is not supported by your kernel.  But changing the check from an error to a warning is just pushing the problem down the road, to where the actual mount happens, and fails.

donngchao:
So I try to just comment the line "nodev,metacopy=on" in the  /etc/containers/storage.conf file in my linux machine.

rhatdan:
Sure.

PR #797, opened 2021-01-04
Let STORAGE_DRIVER env be used even after getting from configfile

chmouel:
Currently when we have a storage driver set in storage.conf and we have a
STORAGE_DRIVER environement variable, the value from the storage.conf file is
used. While it would be expected to use the value from the environement variable
instead of the storage.conf (which make it easy to override in kubernetes env).

This goes in line of how we do it for STORAGE_OPT later in the same function
where the env variable here overrides the value form the file.

Signed-off-by: Chmouel Boudjnah <chmouel@redhat.com>

### DIFF
--- a/store.go
+++ b/store.go
@@ -3522,11 +3522,11 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 		fmt.Printf("Failed to parse %s %v\n", configFile, err.Error())
 		return
 	}
-	if os.Getenv("STORAGE_DRIVER") != "" {
-		config.Storage.Driver = os.Getenv("STORAGE_DRIVER")
-	}
 	if config.Storage.Driver != "" {
 		storeOptions.GraphDriverName = config.Storage.Driver
+	}
+	if os.Getenv("STORAGE_DRIVER") != "" {
+		config.Storage.Driver = os.Getenv("STORAGE_DRIVER")
 	}
 	if storeOptions.GraphDriverName == "" {
 		logrus.Errorf("The storage 'driver' option must be set in %s, guarantee proper operation.", configFile)


chmouel:
I am not sure if the CI error are legitimate or not, it should fail on the fedora-* if my patch broke something, isnt it ?

giuseppe (approved):
LGTM

store.go
```diff
@@ -3522,12 +3522,12 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
...
 		return
 	}
-	if os.Getenv("STORAGE_DRIVER") != "" {
-		config.Storage.Driver = os.Getenv("STORAGE_DRIVER")
-	}
 	if config.Storage.Driver != "" {
 		storeOptions.GraphDriverName = config.Storage.Driver
 	}
+	if os.Getenv("STORAGE_DRIVER") != "" {
+		config.Storage.Driver = os.Getenv("STORAGE_DRIVER")
```
sbose78, line 3529: 👍

rhatdan: This does not change anything, as far as I can see.


In the old code config.Storage.Driver will be set to os.Getenv("STORAGE_DRIVER") if it was set, and then the 
if config.Storage.Driver will not == "".

rhatdan:
I do not believe this is changing any behaviour but maybe the code is easier to understand.

store.go
chmouel: @rhatdan ah thanks, i didn't understand the code properly, would setting `storeOptions.GraphDriverName` instead works ?

rhatdan: No I am fine with the change from a readability point of view. but I wanted to know if someone found a real bug, because I don't believe this fixes anything.

chmouel: The bug is when you set the environment variable `STORAGE_DRIVER`, it doesn't take precedence on the configuration file, is it expected? 

`STORAGE_OPTS`  seems to works differently where the STORAGE_OPTS env are added to the already storage opts set from the file  https://github.com/containers/storage/blob/3ef97f424/store.go#L3600-L3604

rhatdan: If that is true, then it will still be true after this PR, since it changes nothing.

giuseppe: I misread the original code.  With this change `storeOptions.GraphDriverName` won't be set, so I think the old code is correct

rhatdan:
Well this code is clearer what is going on, so I like it, but you need to rebase your PR so we can get it merged.

giuseppe:
> Well this code is clearer what is going on, so I like it, but you need to rebase your PR so we can get it merged.

it will need to change `storeOptions.GraphDriverName` as well though, as it is not currently done

chmouel:
@giuseppe yes you are right, I have misread the code too, this supposed to do the right thing, 

- read storage.conf in config struct
- if os getenv STORAGE_DRIVER, set config.storage.Driver with it
- if config.Storage.Driver is set storeOptions.GraphDriverName with that value

I'll need to do some more testing to figure out why are not experiencing this behavior on our testing.

@pradeepitm12 @nikhil-thomas did we make sure we test the latest buildah in our tekton task ?

pradeepitm12:
> @giuseppe yes you are right, I have misread the code too, this supposed to do the right thing,
> 
> * read storage.conf in config struct
> * if os getenv STORAGE_DRIVER, set config.storage.Driver with it
> * if config.Storage.Driver is set storeOptions.GraphDriverName with that value
> 
> I'll need to do some more testing to figure out why are not experiencing this behavior on our testing.
> 
> @pradeepitm12 @nikhil-thomas did we make sure we test the latest buildah in our tekton task ?

I tested with buildah:8.3 and buildah:latest in our tekton task, but the default storage diver is set as overlay
```
cat /etc/containers/storage.conf
# This file is is the configuration file for all tools
# that use the containers/storage library.
# See man 5 containers-storage.conf for more information
# The "container storage" table contains all of the server options.
[storage]

# Default Storage Driver
driver = "overlay"

```

rhatdan:
This looks like a rootless issue.  If you do the same command as root, it works.

$ STORAGE_DRIVER=vfs buildah info | grep DriverName
        "GraphDriverName": "overlay",
$ sudo STORAGE_DRIVER=vfs buildah info | grep DriverName
        "GraphDriverName": "vfs",

chmouel:
> This looks like a rootless issue. If you do the same command as root, it works.

this is how we want to run the containers, is it an expected behavior ?

rhatdan:
Need the attached patch, I believe
[patch.txt](https://github.com/containers/storage/files/5776439/patch.txt)

rhatdan:
I opened a replacement PR for this one, PTAL.

PR #800, opened 2021-01-06
Use STORAGE_DRIVER environment variable in rootless mode

rhatdan:
Currently we ignore STORAGE_DRIVER enviroment variable in
rootless mode, always forcing it to be Overlay if fuse-overlay
is installed of vfs.

This patch will respect both the STORAGE_DRIVER and STORAGE_OPT
environment variable if set in rootless mode.

### DIFF
--- a/store.go
+++ b/store.go
@@ -3522,10 +3522,11 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 		fmt.Printf("Failed to parse %s %v\n", configFile, err.Error())
 		return
 	}
+	if config.Storage.Driver != "" {
+		storeOptions.GraphDriverName = config.Storage.Driver
+	}
 	if os.Getenv("STORAGE_DRIVER") != "" {
 		config.Storage.Driver = os.Getenv("STORAGE_DRIVER")
-	}
-	if config.Storage.Driver != "" {
 		storeOptions.GraphDriverName = config.Storage.Driver
 	}
 	if storeOptions.GraphDriverName == "" {
--- a/utils.go
+++ b/utils.go
@@ -211,18 +211,27 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
 	} else {
 		opts.GraphRoot = filepath.Join(dataDir, "containers", "storage")
 	}
-	if path, err := exec.LookPath("fuse-overlayfs"); err == nil {
-		opts.GraphDriverName = "overlay"
-		opts.GraphDriverOptions = []string{fmt.Sprintf("overlay.mount_program=%s", path)}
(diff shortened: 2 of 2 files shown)


rhatdan:
Replaces https://github.com/containers/storage/pull/797

chmouel:
cool! thank you,

rhatdan:
@chmouel Could you take this patched utils.go and test it against your buildah to make sure it works the way you want.

rhatdan:
@giuseppe @vrothberg @nalind @TomSweeneyRedHat @saschagrunert  PTAL

store.go
```diff
@@ -3522,12 +3522,12 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
...
 		return
 	}
-	if os.Getenv("STORAGE_DRIVER") != "" {
-		config.Storage.Driver = os.Getenv("STORAGE_DRIVER")
-	}
 	if config.Storage.Driver != "" {
 		storeOptions.GraphDriverName = config.Storage.Driver
 	}
+	if os.Getenv("STORAGE_DRIVER") != "" {
+		config.Storage.Driver = os.Getenv("STORAGE_DRIVER")
```
giuseppe, line 3529: I think it needs to set `storeOptions.GraphDriverName` as the old code was doing

utils.go
```diff
@@ -211,18 +211,30 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
...
 	}
+	opts.GraphDriverName = os.Getenv("STORAGE_DRIVER")
 	if path, err := exec.LookPath("fuse-overlayfs"); err == nil {
-		opts.GraphDriverName = "overlay"
-		opts.GraphDriverOptions = []string{fmt.Sprintf("overlay.mount_program=%s", path)}
-		for _, o := range systemOpts.GraphDriverOptions {
-			if strings.Contains(o, "ignore_chown_errors") {
-				opts.GraphDriverOptions = append(opts.GraphDriverOptions, o)
-				break
+		if opts.GraphDriverName == "" {
```
giuseppe, line 216: do we need to check at all for `fuse-overlayfs` if the graphdriver is set and has a value different than "overlay"?

store.go
rhatdan: Fixed.

utils.go
rhatdan: No, so I cleaned this checking up.

giuseppe (approved):
LGTM

rhatdan:
@TomSweeneyRedHat @nalind @saschagrunert @vrothberg @mheon PTAL and Merge.

vrothberg (commented):
LGTM

PR #806, opened 2021-01-16
Fix FreeBSD suport

mateuszkwiatkowski:
I added `pkg/archive/archive_freebsd.go` which is copy of `pkg/archive/archive_unix.go` with only one modification as FreeBSD requires `dev` to be `uint64` since 12.0: https://github.com/golang/go/commit/dc6eb200dd59dfedaa5259565b8ef1aa96a39271
 I'm not experienced with Go build system and I'm not sure how to avoid this code duplication - I'm open to suggestion.

Ref https://github.com/containers/skopeo/issues/1163

### DIFF
--- a/pkg/archive/archive_freebsd.go
+++ b/pkg/archive/archive_freebsd.go
@@ -38,7 +38,6 @@ func CanonicalTarNameForPath(p string) (string, error) {
 
 // chmodTarEntry is used to adjust the file permissions used in tar header based
 // on the platform the archival is done.
-
 func chmodTarEntry(perm os.FileMode) os.FileMode {
 	return perm // noop for unix as golang APIs provide perm bits correctly
 }
--- a/pkg/archive/archive_freebsd.go
+++ b/pkg/archive/archive_freebsd.go
@@ -1,4 +1,4 @@
-// +build !windows,!freebsd
+// +build freebsd
 
 package archive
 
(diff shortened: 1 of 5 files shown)


mateuszkwiatkowski:
I compiled skopeo with these modifications successfully on Linux, macOS and FreeBSD so I think they're correct. I extracted few images on FreeBSD and it worked like a charm!
Please let me know if anything else is required to have this PR merged.

rhatdan:
LGTM
But could you rebase and repush to get this to pass CI system.

mateuszkwiatkowski:
Hi @mateuszkwiatkowski, 
I've just rebased against master, so PTAL. Thanks!

rhatdan (approved):
@vrothberg @giuseppe @nalind @TomSweeneyRedHat PTAL

pkg/archive/archive_freebsd.go
```diff
@@ -0,0 +1,126 @@
...
+// CanonicalTarNameForPath returns platform-specific filepath
+// to canonical posix-style path for tar archival. p is relative
+// path.
+func CanonicalTarNameForPath(p string) (string, error) {
+	return p, nil // already unix-style
+}
+
+// chmodTarEntry is used to adjust the file permissions used in tar header based
+// on the platform the archival is done.
+
```
vrothberg, line 41:
```suggestion
```

vrothberg (commented):
Just one nit, but LGTM

Note that I have no way to test it. Thanks for your work!

pkg/archive/archive_freebsd.go
rhatdan: Not sure what this suggested change is?

vrothberg: To remove the empty line.

mateuszkwiatkowski: Empty line removed.

vrothberg:
Thanks, @mateuszkwiatkowski!

Could you rebase the commits? Then we can merge.

rhatdan:
LGTM

PR #810, opened 2021-01-22
Improve project quota to support querying disk usage

haircommander:
To convert to CRI stats for CRI-O, we need to improve the performance of finding the disk stats. The current implementation walks the `/diff` directory and adds up the size and inodes.

Luckily, with XFS and project quotas, we can simply query the value.

This PR adds a couple of things, broken into smaller commits:
- updates pkg/directory to have a function GetDiskUsage, which does what directory.Size did, but also counts the inodes
- Add a similar GetDiskUsage function to the projectquota package, but this one queries the quota instead of walking
- Adds a ReadWriteDiskUsage endpoint for all drivers, most of which simply call directory.GetDiskUsage
- updates overlay to call projectquota.GetDiskUsage (if we successfully initialized our quota object)

### DIFF
--- a/drivers/aufs/aufs.go
+++ b/drivers/aufs/aufs.go
@@ -482,6 +482,21 @@ func (a *Driver) Put(id string) error {
 	return err
 }
 
+// ReadWriteDiskUsage returns the disk usage of the writable directory for the ID.
+// For AUFS, it queries the mountpoint for this ID.
+func (a *Driver) ReadWriteDiskUsage(id string) (*directory.DiskUsage, error) {
+	a.locker.Lock(id)
+	defer a.locker.Unlock(id)
+	a.pathCacheLock.Lock()
+	m, exists := a.pathCache[id]
+	if !exists {
+		m = a.getMountpoint(id)
+		a.pathCache[id] = m
+	}
+	a.pathCacheLock.Unlock()
+	return directory.Usage(m)
+}
+
 // isParent returns if the passed in parent is the direct parent of the passed in layer
 func (a *Driver) isParent(id, parent string) bool {
 	parents, _ := getParentIDs(a.rootPath(), id)
--- a/drivers/quota/projectquota.go
+++ b/drivers/quota/projectquota.go
@@ -56,6 +56,7 @@ import (
 	"path/filepath"
 	"unsafe"
 
+	"github.com/containers/storage/pkg/directory"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
@@ -196,17 +197,37 @@ func setProjectQuota(backingFsBlockDev string, projectID uint32, quota Quota) er
 
 // GetQuota - get the quota limits of a directory that was configured with SetQuota
 func (q *Control) GetQuota(targetPath string, quota *Quota) error {
+	d, err := q.fsDiskQuotaFromPath(targetPath)
+	if err != nil {
+		return err
+	}
+	quota.Size = uint64(d.d_blk_hardlimit) * 512
+	return nil
+}
+
+// GetDiskUsage - get the current disk usage of a directory that was configured with SetQuota
+func (q *Control) GetDiskUsage(targetPath string, usage *directory.DiskUsage) error {
+	d, err := q.fsDiskQuotaFromPath(targetPath)
+	if err != nil {
+		return err
+	}
+	usage.Size = int64(d.d_bcount) * 512
+	usage.InodeCount = int64(d.d_icount)
+
+	return nil
+}
+
+func (q *Control) fsDiskQuotaFromPath(targetPath string) (C.fs_disk_quota_t, error) {
+	var d C.fs_disk_quota_t
 
 	projectID, ok := q.quotas[targetPath]
 	if !ok {
-		return fmt.Errorf("quota not found for path : %s", targetPath)
+		return d, fmt.Errorf("quota not found for path : %s", targetPath)
 	}
 
 	//
 	// get the quota limit for the container's project id
 	//
-	var d C.fs_disk_quota_t
-
 	var cs = C.CString(q.backingFsBlockDev)
 	defer C.free(unsafe.Pointer(cs))
 
@@ -214,12 +235,11 @@ func (q *Control) GetQuota(targetPath string, quota *Quota) error {
 		uintptr(unsafe.Pointer(cs)), uintptr(C.__u32(projectID)),
 		uintptr(unsafe.Pointer(&d)), 0, 0)
 	if errno != 0 {
-		return fmt.Errorf("Failed to get quota limit for projid %d on %s: %v",
+		return d, fmt.Errorf("Failed to get quota limit for projid %d on %s: %v",
 			projectID, q.backingFsBlockDev, errno.Error())
 	}
-	quota.Size = uint64(d.d_blk_hardlimit) * 512
 
-	return nil
+	return d, nil
 }
 
 // getProjectID - get the project id of path on xfs
--- a/drivers/btrfs/btrfs.go
+++ b/drivers/btrfs/btrfs.go
@@ -27,6 +27,7 @@ import (
 	"unsafe"
 
 	graphdriver "github.com/containers/storage/drivers"
+	"github.com/containers/storage/pkg/directory"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/mount"
 	"github.com/containers/storage/pkg/parsers"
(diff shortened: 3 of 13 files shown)


haircommander:
This PR is nearly ready, I haven't yet shown it actually does what I want (I have XFS in a VM, but I'm getting ENOSYS on the SYS_QUOTACTL syscalls, even though the kernel was compiled with CONFIG_QUOTA :upside_down_face: )

I'll try testing differently on monday

drivers/quota/projectquota.go
```diff
@@ -222,6 +223,32 @@ func (q *Control) GetQuota(targetPath string, quota *Quota) error {
 	return nil
 }
 
+// GetDiskUsage - get the current disk usage of a directory that was configured with SetQuota
+func (q *Control) GetDiskUsage(targetPath string, usage *directory.DiskUsage) error {
+	projectID, ok := q.quotas[targetPath]
+	if !ok {
+		return fmt.Errorf("quota not found for path : %s", targetPath)
```
rhatdan, line 230: This should be a defined error constant, I believe.

drivers/quota/projectquota.go
```diff
@@ -222,6 +223,32 @@ func (q *Control) GetQuota(targetPath string, quota *Quota) error {
...
+	var d C.fs_disk_quota_t
+
+	var cs = C.CString(q.backingFsBlockDev)
+	defer C.free(unsafe.Pointer(cs))
+
+	_, _, errno := unix.Syscall6(unix.SYS_QUOTACTL, C.Q_XGETPQUOTA,
+		uintptr(unsafe.Pointer(cs)), uintptr(C.__u32(projectID)),
+		uintptr(unsafe.Pointer(&d)), 0, 0)
+	if errno != 0 {
+		return fmt.Errorf("Failed to get quota limit for projid %d on %s: %v",
```
rhatdan, line 243: Should be a constant error.

rhatdan (approved):

drivers/quota/projectquota.go
haircommander: I've refactored to not duplicate this code

haircommander:
this is still currently untested, but I am curious about feedback, PTAL @rhatdan @giuseppe @nalind

nalind:
I like the general organization.  I think the device mapper driver should be able to query the size of a snapshot volume without having to mount it and walk its contents, but it's been a while since I looked at that code.

drivers/aufs/aufs.go
```diff
@@ -482,6 +482,20 @@ func (a *Driver) Put(id string) error {
...
+// For AUFS, it queries the mountpoint for this ID.
+func (a *Driver) ReadWriteDiskUsage(id string) (*directory.DiskUsage, error) {
+	a.locker.Lock(id)
+	defer a.locker.Unlock(id)
+	m, exists := a.pathCache[id]
+	if !exists {
+		m = a.getMountpoint(id)
+		a.pathCache[id] = m
+	}
+	a.pathCacheLock.Unlock()
```
giuseppe, line 495: doesn't seem we take this lock?

giuseppe (commented):
LGTM once tested :-)

drivers/aufs/aufs.go
haircommander: it sure doesn't :)

haircommander:
> I like the general organization. I think the device mapper driver should be able to query the size of a snapshot volume without having to mount it and walk its contents, but it's been a while since I looked at that code.

unfortunately, it doesn't seem to track inode number, so we'd still have to walk the fs. If you want, I can use that optimization and use something other than filepath.Walk that doesn't stat each file, but rather just finds each inode, which would make it perform better, but that would also increase the complexity quite a bit

drivers/aufs/aufs.go
haircommander: fixed

rhatdan:
Do you want to go back to the filesystem team to see if they have any ideas.

Wrote quick go program to attempt to beat the "du" or "find" and have lossed.

```
package main

import (
	"fmt"
	"io/ioutil"
	"os"
	"path/filepath"

	"github.com/sirupsen/logrus"
)

func fileCount(path string) int {
	i := 0
	files, err := ioutil.ReadDir(path)
	if err != nil {
		logrus.Debugf("Failed ReadDir %v", err)
		return 0
	}
	i += len(files)
	for _, file := range files {
		if file.IsDir() {
			i += fileCount(filepath.Join(path, file.Name()))
		}
	}
	return i
}

func main() {
	fmt.Println(fileCount(os.Args[1]))
}
```
```
$ time du -s --inodes /usr 
449537	/usr

real	0m1.388s
user	0m0.425s
sys	0m0.946s
$ time find /usr | wc -l
461397

real	0m0.924s
user	0m0.388s
sys	0m0.614s
$ time /tmp/t /usr 
461396

real	0m2.000s
user	0m1.113s
sys	0m1.221s
```

haircommander:
I actually think most of the overhead is from go's runtime. Running with these two versions:

1.go
```
package main

import (
	"fmt"
	"os"
	"path/filepath"
	"syscall"
)

// DiskUsage is a structure that describes the disk usage (size and inode count)
// of a particular directory.
type DiskUsage struct {
	Size       int64
	InodeCount int64
}

// Usage walks a directory tree and returns its total size in bytes and the number of inodes.
func Usage(dir string) (usage *DiskUsage, err error) {
	usage = &DiskUsage{}
	data := make(map[uint64]struct{})
	err = filepath.Walk(dir, func(d string, fileInfo os.FileInfo, err error) error {
		if err != nil {
			// if dir does not exist, Usage() returns the error.
			// if dir/x disappeared while walking, Usage() ignores dir/x.
			if os.IsNotExist(err) && d != dir {
				return nil
			}
			return err
		}

		if fileInfo == nil {
			return nil
		}

		// Check inode to handle hard links correctly
		inode := fileInfo.Sys().(*syscall.Stat_t).Ino
		// inode is not a uint64 on all platforms. Cast it to avoid issues.
		if _, exists := data[uint64(inode)]; exists {
			return nil
		}

		// inode is not a uint64 on all platforms. Cast it to avoid issues.
		data[uint64(inode)] = struct{}{}

		// Count the unique inode
		usage.InodeCount++

		// Ignore directory sizes
		if fileInfo.IsDir() {
			return nil
		}

		usage.Size += fileInfo.Size()

		return nil
	})
	return
}

func main() {
	usage, err := Usage(os.Args[1])
	if err != nil {
		panic(err)
	}
	fmt.Println(usage.Size, usage.InodeCount)
}
```

2.go
```
package main

import (
	"fmt"
	"os"
	"os/exec"
	"strings"
)

type DiskUsage struct {
	Size       string
	InodeCount string
}

func execDiskUsageCommand(args []string) (string, error) {
	out, err := exec.Command("du", args...).CombinedOutput()
	if err != nil {
		return "", err
	}
	parts := strings.Fields(string(out))
	if len(parts) != 2 {
		return "", fmt.Errorf("expected two fields in du output got %s", len(parts))
	}
	return parts[0], nil
}

func Usage(path string) (*DiskUsage, error) {
	var err error
	usage := &DiskUsage{}
	usage.InodeCount, err = execDiskUsageCommand([]string{"-s", "--inodes", path})
	if err != nil {
		return nil, err
	}

	usage.Size, err = execDiskUsageCommand([]string{"-s", "-b", path})
	if err != nil {
		return nil, err
	}

	return usage, nil
}

func main() {
	usage, err := Usage(os.Args[1])
	if err != nil {
		panic(err)
	}
	fmt.Println(usage.Size, usage.InodeCount)
}
```

and using multitime to compare gets:
``` $ multitime -n 10 sudo go run 1.go /usr
12205302206 392193
12205302206 392193
12205302206 392193
12205302206 392193
12205302206 392193
12205302206 392193
12205302206 392193
12205302206 392193
12205302206 392193
12205302206 392193
===> multitime results
1: sudo go run 1.go /usr
            Mean        Std.Dev.    Min         Median      Max
real        2.241       0.029       2.196       2.235       2.288       
user        1.319       0.033       1.257       1.322       1.380       
sys         1.315       0.036       1.258       1.318       1.381       
[pehunt@localhost tmp]
 $ multitime -n 10 sudo go run 2.go /usr
12421779902 392193
12421779902 392193
12421779902 392193
12421779902 392193
12421779902 392193
12421779902 392193
12421779902 392193
12421779902 392193
12421779902 392193
12421779902 392193
===> multitime results
1: sudo go run 2.go /usr
            Mean        Std.Dev.    Min         Median      Max
real        2.739       0.018       2.716       2.736       2.782       
user        1.095       0.029       1.037       1.094       1.155       
sys         1.749       0.041       1.672       1.743       1.833       
```

rhatdan:
LGTM
@giuseppe @nalind @vrothberg @TomSweeneyRedHat PTAL

vrothberg (commented):
LGTM

PR #811, opened 2021-01-25
RFC: Support managing images as an unpriv user on macOS

slp:
Support managing images as an unprivileged user on macOS, by storing
user ownership and file mode bits as extended attributes. This is
mainly intended to be used on libkrun-based lightweight VMs, where its
virtio-fs implementation reads those attributes and translates them
for the Guest.

In this implementation, the decision to either store those bits as
either extended attributes for virtio-fs or actual permission and
ownerships values on the filesystem, is done by checking in-place if
the current operating system is "darwin" and uid != 0. This way, we
avoid the need to extend the interface by adding a new configuration
option. On the other hand, if it's found to be useful for other
systems, we should implmement a proper tunable.

Signed-off-by: Sergio Lopez <slp@redhat.com>

### DIFF
--- a/drivers/driver_darwin.go
+++ b/drivers/driver_darwin.go
@@ -1,15 +1,14 @@
-// +build !linux,!windows,!freebsd,!solaris,!darwin
-
 package graphdriver
 
 var (
-	// Slice of drivers that should be used in an order
+	// Slice of drivers that should be used in order
 	priority = []string{
-		"unsupported",
+		"vfs",
 	}
 )
 
 // GetFSMagic returns the filesystem id given the path.
 func GetFSMagic(rootpath string) (FsMagic, error) {
+	// Note it is OK to return FsMagicUnsupported on Windows.
 	return FsMagicUnsupported, nil
 }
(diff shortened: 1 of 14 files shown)


rhatdan:
@giuseppe PTAL
This seems pretty similar to what @giuseppe is doing for handling rootless containers on an NFS based homedir.  We should coordinate the name-ing of the xattrs?

giuseppe:
> This seems pretty similar to what @giuseppe is doing for handling rootless containers on an NFS based homedir. We should coordinate the name-ing of the xattrs?

yes this is similar to what we do on NFS with https://github.com/containers/storage/pull/744

@slp would that work for you?

slp:
> > This seems pretty similar to what @giuseppe is doing for handling rootless containers on an NFS based homedir. We should coordinate the name-ing of the xattrs?
> 
> yes this is similar to what we do on NFS with #744
> 
> @slp would that work for you?

We need more changes to detect rootless in Darwin, but certainly #744 gets us closer. I'm going to rework this PR around `forceMask` to reduce the number of changes.

Thanks!

rhatdan:
@slp Are you still interested in this PR or should we close?

slp:
@rhatdan This PR needs a complete rework, and I honestly don't know when I'll be able to do that, so let's close this one for the moment.

giuseppe (approved):
LGTM

PR #812, opened 2021-02-01
store: support volatile containers

giuseppe:
add a new class of containers that are not guaranteed to survive a crash.  The advantage of such containers is that storage can be
optimized to skip some synchronizations with the underlying storage.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -313,6 +313,9 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
 	if options.MountOpts != nil {
 		options.Flags["MountOpts"] = append([]string{}, options.MountOpts...)
 	}
+	if options.Volatile {
+		options.Flags["Volatile"] = true
+	}
 	names = dedupeNames(names)
 	for _, name := range names {
 		if _, nameInUse := r.byname[name]; nameInUse {
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -53,6 +53,10 @@ type MountOpts struct {
 	UidMaps []idtools.IDMap // nolint: golint
 	GidMaps []idtools.IDMap // nolint: golint
 	Options []string
+
+	// Volatile specifies whether the container storage can be optimized
+	// at the cost of not syncing all the dirty files in memory.
+	Volatile bool
 }
 
 // ApplyDiffOpts contains optional arguments for ApplyDiff methods.
--- a/drivers/overlay/check.go
+++ b/drivers/overlay/check.go
@@ -163,3 +163,40 @@ func doesMetacopy(d, mountOpts string) (bool, error) {
 	}
 	return metacopy != nil, nil
 }
+
+// doesVolatile checks if the filesystem supports the "volatile" mount option
+func doesVolatile(d string) (bool, error) {
+	td, err := ioutil.TempDir(d, "volatile-check")
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		if err := os.RemoveAll(td); err != nil {
+			logrus.Warnf("Failed to remove check directory %v: %v", td, err)
+		}
+	}()
+
+	if err := os.MkdirAll(filepath.Join(td, "lower"), 0755); err != nil {
+		return false, err
+	}
+	if err := os.MkdirAll(filepath.Join(td, "upper"), 0755); err != nil {
+		return false, err
+	}
+	if err := os.Mkdir(filepath.Join(td, "work"), 0755); err != nil {
+		return false, err
+	}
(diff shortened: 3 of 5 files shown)


giuseppe:
@nalind @rhatdan how can we expose this setting to Podman/Buildah?  Should we enable it by default with Buildah containers?

rhatdan:
I think enabling this by default for buildah makes sense.  

What is the use case for this on Podman?  `podman run --rm` ?

I think we need to have the fix for cleaning up bad containers easily though.

rhatdan:
podman mount could probably use this flag as well.

giuseppe:
The patch for Buildah is trivial:

```
diff --git a/new.go b/new.go
index 2ee86dd1..b95e680a 100644
--- a/new.go
+++ b/new.go
@@ -359,6 +359,7 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
                coptions := storage.ContainerOptions{
                        LabelOpts:        options.CommonBuildOpts.LabelOpts,
                        IDMappingOptions: newContainerIDMappingOptions(options.IDMappingOptions),
+                       Volatile:         true,
                }
                container, err = store.CreateContainer("", []string{tmpName}, imageID, "", "", &coptions)
                if err == nil {
```

I think this option must be enabled by default for Buildah and CRI-O (do we even want a way to disable it?).

For Podman instead it must be an explicit setting except when `--rm` is used: `podman run --volatile`?

rhatdan:
Have you thought about how we clean this up, if we get corrupt containers?  Can we easily remove them?

giuseppe:
> Have you thought about how we clean this up, if we get corrupt containers? Can we easily remove them?

containers can be removed in the usual fashion (e.g. podman rm).  Or do you think they should be removed automatically on shutdown?

I'd leave users the choice on such containers, for example it is safe to run volatile containers, stop them and run manually (sync/syncfs).

rhatdan (approved):
LGTM

rhatdan:
But will the tools be smart enough to discover bad containers?

rhatdan:

vrothberg (commented):
LGTM, I leave merging up to you in case you want to continue talking about the UX in the tools

rhatdan:
@giuseppe rebase and we can merge.

giuseppe:
I am changing idea about "volatile" and start to think it should be a mount opt, so it can be more easily used with `--storage-opt`

giuseppe:
rebased

cgwalters:
Did anything else happen with this?  ISTM that e.g. `podman run --rm` for example should use this by default, because the processes and filesystem are defined to be lifecycled together.  

For Kubernetes where that's the default, crio should just start doing this everywhere right?

giuseppe:
Both Podman, Buildah and CRI-O are taking advantage of it on kernels where volatile is supported.  CRI-O is doing it for every container, while Podman only when `--rm` is used

nalind:
How are you checking for it?

cgwalters:
> How are you checking for it?

```
[core@cosa-devsh ~]$ rpm-ostree status
State: idle
Deployments:
● fedora:fedora/x86_64/coreos/testing-devel
                   Version: 36.20220602.dev.0 (2022-06-02T15:24:13Z)
                    Commit: 12d5c0e6c587826e2e78ae6b72981f4d3f7609755883f994701d486831d75d83
              GPGSignature: (unsigned)
[core@cosa-devsh ~]$ rpm -q podman
podman-4.1.0-1.fc36.x86_64
[core@cosa-devsh ~]$ podman run --rm -ti busybox sleep 1h&
[core@cosa-devsh ~]$ podman unshare
Failed to connect to bus: Operation not permitted
[root@cosa-devsh ~]# findmnt |grep overlay
│ └─/var/home/core/.local/share/containers/storage/overlay /dev/vda4[/ostree/deploy/fedora-coreos/var/home/core/.local/share/containers/storage/overlay]                         xfs        rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota
```

I'd expect to see `volatile` in there.  I don't see the `incompat/volatile` in the storage directories.

OK, I see this bit https://github.com/containers/podman/blob/13cdf862e6dbf31902405e4204a5bd87f62a9759/pkg/specgenutil/specgen.go#L832 which seems like it should be setting it...

nalind:
(The overlay directory itself is being bind mounted with different propagation options.)

cgwalters:
> That's an xfs mount, not overlay. 

Right sorry.

> Try "podman run --rm fedora-minimal findmnt /".
> (The overlay directory itself is being bind mounted with different propagation options.)

Hum...interesting.  OK you're right, I do see `volatile` there.  Thanks.

PR #815, opened 2021-02-05
overlay: force metacopy=on for naivediff

giuseppe:
make sure the metacopy=on mount option is specified when it was
present in the d.options.mountOptions.

It solves a problem where the NaiveDiffer doesn't specify
"metacopy=on" when mounting a different layer and the kernel refuses
to load file attributes (even for root), failing with the error:

kernel: overlayfs: refusing to follow metacopy origin for ...

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -125,6 +125,15 @@ func init() {
 	graphdriver.Register("overlay2", Init)
 }
 
+func hasMetacopyOption(opts []string) bool {
+	for _, s := range opts {
+		if s == "metacopy=on" {
+			return true
+		}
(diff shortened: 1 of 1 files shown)


rhatdan (approved):
LGTM

rhatdan:
Was there a specific bug that this addresses?

@nalind @vrothberg @TomSweeneyRedHat @mheon @saschagrunert PTAL

giuseppe:
> Was there a specific bug that this addresses?

it is rhbz#1924562

rhatdan:
@rhvgoyal PTAL

vrothberg (commented):
LGTM

PR #816, opened 2021-02-09
overlay: support native rootless mounts

giuseppe:
tested on Fedora Rawhide with kernel
5.11.0-0.rc6.20210204git61556703b610.144.fc34.x86_64

There is currently an issue when context= is specified and the mount
fails with EACCES.  We can drop it once it works in the kernel.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -28,6 +28,7 @@ import (
 	"github.com/containers/storage/pkg/mount"
 	"github.com/containers/storage/pkg/parsers"
 	"github.com/containers/storage/pkg/system"
+	"github.com/containers/storage/pkg/unshare"
 	units "github.com/docker/go-units"
 	rsystem "github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/selinux/go-selinux/label"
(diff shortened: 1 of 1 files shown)


giuseppe:
I am already in touch with Vivek and Miklos about the kernel issue

rhatdan:
What happens if the kernel does not support overlay, do we still fall back to using fuse-overlayfs if it is installed?

rhatdan (approved):
LGTM

giuseppe:
> What happens if the kernel does not support overlay, do we still fall back to using fuse-overlayfs if it is installed?

we don't have that check in place here, we do it in containers/common.

I've dropped the workaround for the mount label since there is a patch upstream

rhatdan:
@vrothberg @saschagrunert @nalind @mheon @baude @umohnani8 PTAL

saschagrunert (approved):
LGTM

drivers/overlay/overlay.go
```diff
@@ -1023,6 +1024,11 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	} else if d.options.mountOptions != "" {
 		opts = fmt.Sprintf("%s,%s", d.options.mountOptions, opts)
 	}
+
+	if d.options.mountProgram == "" && unshare.IsRootless() {
+		opts = fmt.Sprintf("%s,userxattr", opts)
```
AkihiroSuda, line 1029: Shouldn't this be set only on kernel >= 5.11?

Ubuntu and Debian have been supporting rootless overlayfs and do not want userxattr opt here (for kernel < 5.11).

giuseppe: we don't support the Ubuntu and Debian custom patch for rootless overlay.  The main reason is that on old kernels it is not allowed to create whiteout files without an overlay mount and that would complicate how image pulls are performed

rhatdan: We also don't want kernel version checks, since these features can/will be back ported to RHEL8 and Centos Stream.

PR #822, opened 2021-02-16
Call recreateSymlinks when not found during Readlink

umohnani8:
If the symlinks under the l directory are lost during a
sudden reboot, we can hit an error when grabbing the lower
directories and reading the symlinks before they are recreated
in the Get function. Add a call to recreateSymlinks in this
code path as well.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1921128

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -812,8 +812,22 @@ func (d *Driver) getLowerDirs(id string) ([]string, error) {
 		for _, s := range strings.Split(string(lowers), ":") {
 			lower := d.dir(s)
 			lp, err := os.Readlink(lower)
+			// if the link does not exist, we lost the symlinks during a sudden reboot.
+			// Let's go ahead and recreate those symlinks.
 			if err != nil {
-				return nil, err
+				if os.IsNotExist(err) {
+					logrus.Warnf("Can't read link %q because it does not exist. Going through storage to recreate the missing symlinks.", lower)
+					if err := d.recreateSymlinks(); err != nil {
+						return nil, fmt.Errorf("error recreating the missing symlinks: %v", err)
+					}
+					// let's call Readlink on lower again now that we have recreated the missing symlinks
+					lp, err = os.Readlink(lower)
+					if err != nil {
+						return nil, err
+					}
+				} else {
+					return nil, err
+				}
 			}
 			lowersArray = append(lowersArray, path.Clean(d.dir(path.Join("link", lp))))
 		}


umohnani8:
@rhatdan @nalind PTAL

rhatdan (approved):
LGTM

drivers/overlay/overlay.go
```diff
@@ -812,7 +812,19 @@ func (d *Driver) getLowerDirs(id string) ([]string, error) {
 		for _, s := range strings.Split(string(lowers), ":") {
 			lower := d.dir(s)
 			lp, err := os.Readlink(lower)
-			if err != nil {
+			// if the link does not exist, we lost the symlinks during a sudden reboot.
+			// Let's go ahead and recreate those symlinks.
+			if lp == "" && os.IsNotExist(err) {
```
nalind, line 817: Please restructure this a bit to check for `err != nil` instead of `lp == ""`.

umohnani8: Fixed

drivers/overlay/overlay.go
```diff
@@ -812,7 +812,19 @@ func (d *Driver) getLowerDirs(id string) ([]string, error) {
 		for _, s := range strings.Split(string(lowers), ":") {
 			lower := d.dir(s)
 			lp, err := os.Readlink(lower)
-			if err != nil {
+			// if the link does not exist, we lost the symlinks during a sudden reboot.
```
rhatdan, line 815: Nit, but could you resturcture this to be just 
```
if err != nil {
          if os.IsNotExist(err) {   
                   ...
          } else {
                  return nil, err
          }
}
```

umohnani8: Yup, fixed

rhatdan:
LGTM

rhatdan:

vrothberg (commented):
LGTM

giuseppe:
Should we at least warn that the storage is corrupted and files might be missing or are partially stored?

umohnani8:
@giuseppe there is a warning that we can't find the symlinks so going to restore those https://github.com/containers/storage/pull/822/files#diff-4b0197157e354c66e154f5cdc6fea71c4e43732b4bad32d8652f23bd588e51a1R819

giuseppe:
> @giuseppe there is a warning that we can't find the symlinks so going to restore those https://github.com/containers/storage/pull/822/files#diff-4b0197157e354c66e154f5cdc6fea71c4e43732b4bad32d8652f23bd588e51a1R819

I am just worried that users will get the impression this would fix the storage corruptions but there could be other issues that we are not able to detect yet, such as missing or incomplete files in the pulled images.  Maybe a warning saying the safest is probably to completely wipe the storage if possible to not incur in weird errors?

vrothberg:
SGTM

umohnani8:
Follow up with warning in https://github.com/containers/storage/pull/833

PR #827, opened 2021-02-19
Call recreateSymlinks when not found during Readlink

vrothberg:
If the symlinks under the l directory are lost during a
sudden reboot, we can hit an error when grabbing the lower
directories and reading the symlinks before they are recreated
in the Get function. Add a call to recreateSymlinks in this
code path as well.

Backported-by: Valentin Rothberg <rothberg@redhat.com>


rhatdan:
LGTM

PR #832, opened 2021-02-22
pkg/lockfile: fix a race and re-enable unit tests

nalind:
Fix a race where the state lock in a Unixy lockfile structure wasn't being used to protect access to all of the structure's state fields, fixing #831.

Fix usage of sync/atomic in the corresponding unit tests to not mix atomic adds and reads with non-atomic reads, something which the race detector complained about.

Start running unit tests again in CI.  It was not intentionally that we stopped doing that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ NATIVETAGS :=
 AUTOTAGS := $(shell ./hack/btrfs_tag.sh) $(shell ./hack/libdm_tag.sh)
 BUILDFLAGS := -tags "$(AUTOTAGS) $(TAGS)" $(FLAGS)
 GO ?= go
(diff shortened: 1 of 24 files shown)


rhatdan:
LGTM
But tests ...

rhatdan (approved):
LGTM

rhatdan:
@vrothberg @giuseppe @saschagrunert @umohnani8 PTAL

vrothberg (commented):
Changes LGTM (really nice work, @nalind)

Do we need to cherry-pick all commits to the stable branch for 8.4?

giuseppe (approved):
LGTM

rhatdan:
I think we need to at least cherry-pick the lock fix.

PR #833, opened 2021-02-24
Add warning about possible storage corruption

umohnani8:

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -816,7 +816,7 @@ func (d *Driver) getLowerDirs(id string) ([]string, error) {
 			// Let's go ahead and recreate those symlinks.
 			if err != nil {
 				if os.IsNotExist(err) {
-					logrus.Warnf("Can't read link %q because it does not exist. Going through storage to recreate the missing symlinks.", lower)
+					logrus.Warnf("Can't read link %q because it does not exist. A storage corruption might have occurred, attempting to recreate the missing symlinks. It might be best wipe the storage to avoid further errors due to storage corruption.", lower)
 					if err := d.recreateSymlinks(); err != nil {
 						return nil, fmt.Errorf("error recreating the missing symlinks: %v", err)
 					}


umohnani8:
@rhatdan @giuseppe @vrothberg PTAL

drivers/overlay/overlay.go
```diff
@@ -816,7 +816,7 @@ func (d *Driver) getLowerDirs(id string) ([]string, error) {
 			// Let's go ahead and recreate those symlinks.
 			if err != nil {
 				if os.IsNotExist(err) {
-					logrus.Warnf("Can't read link %q because it does not exist. Going through storage to recreate the missing symlinks.", lower)
+					logrus.Warnf("Can't read link %q because it does not exist. A storage corruption might have occurred, we are going to attempt to recreate the missing symlinks. If possible, it is best to wipe the storage to avoid other errors due to storage corruption.", lower)
```
rhatdan, line 819: logrus.Warnf("Can't read link %q because it does not exist. A storage corruption might have occurred, attempting to recreate the missing symlinks. It might be best wipe the storage to avoid further errors due to storage corruption.", lower)

TomSweeneyRedHat: Would the general user know how to wipe storage?  Plus, even thought possibly true, I'm not real hip about saying "we might have had a storage corruption".

TomSweeneyRedHat: I'd prefer something like:  
```
				logrus.Warnf("Can't read link %q because it does not exist.  It is recommend to run {command(s)} to avoid potential storage corruption.", lower)
```

rhatdan: I would just remove it, Just warn and let it go.  This is storage and only Podman has a way to wipe out storage.  Skopeo and Buildah do not, and I don't want to add an error message to storage that talks about podman.

umohnani8: Fixed to @rhatdan's suggestion and I agree with Dan, just warn and let it go.

rhatdan:
LGTM

rhatdan (approved):

giuseppe (approved):
/lgtm

PR #834, opened 2021-02-25
Move storageOpts structures into types subdir to shrink bindings.

rhatdan:
Currently when we build podman bindings we are pulling in the entire
storage libraries, even though we only need a few structures and
functions.

Testing with the following program

```
package main

import (
	"fmt"
	"github.com/containers/storage/types"
)

func main() {
	fmt.Println(types.GetRootlessRuntimeDir(0))
}
```

Removing types above gives me compile size of the the program

du -s t.old t.new
9640	t.before
3232	t.after

Currently these functions are being vendored into
containers/common/pkg/config, which leads to large size in podman-remote
and podman bindings.


vrothberg (commented):
LGTM
linter is not yet happy

types/errors.go
```diff
@@ -0,0 +1,56 @@
+package types
+
+import (
+	"errors"
+)
+
+var (
```
TomSweeneyRedHat, line 7: These really need to be alphabetized.

errors.go
```diff
@@ -1,56 +1,56 @@
 package storage
 
 import (
-	"errors"
+	"github.com/containers/storage/types"
 )
 
 var (
```
TomSweeneyRedHat, line 7: alphabetize me please!

types/utils.go
```diff
@@ -0,0 +1,201 @@
...
+				return tmpPerUserDir, nil
+			}
+		} else if isRootlessRuntimeDirOwner(tmpPerUserDir, env) {
+			return tmpPerUserDir, nil
+		}
+	}
+
+	homeDir := env.homedirGet()
+	if homeDir == "" {
+		return "", errors.New("neither XDG_RUNTIME_DIR not temp dir nor HOME was set non-empty")
```
TomSweeneyRedHat, line 101:
```suggestion
		return "", errors.New("neither XDG_RUNTIME_DIR nor temp dir nor HOME was set non-empty")
```

TomSweeneyRedHat (commented):
Couple nits, otherwise LGTM

PR #837, opened 2021-02-26
Add more mount information to errors

rhatdan:
Currently it is difficult to diagnose mount failures in overlay storage.
This issue is we are not sure of all of the mount options used when
attempting to mount the file system.

This PR adds missing data.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1159,7 +1159,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		}
 		mountData = label.FormatMountLabel(opts, options.MountLabel)
 		if len(mountData) > pageSize {
-			return "", fmt.Errorf("cannot mount layer, mount label too large %d", len(mountData))
+			return "", fmt.Errorf("cannot mount layer, mount label %q too large %d > page size %d", options.MountLabel, len(mountData), pageSize)
 		}
 		mountFunc = func(source string, target string, mType string, flags uintptr, label string) error {
 			return mountFrom(d.home, source, target, mType, flags, label)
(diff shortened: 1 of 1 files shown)


rhatdan:
The goal is to help with https://github.com/containers/podman/issues/9507

rhatdan:
@edsantiago  PTAL

drivers/overlay/overlay.go
```diff
@@ -1169,7 +1169,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	flags, data := mount.ParseOptions(mountData)
 	logrus.Debugf("overlay: mount_data=%s", mountData)
 	if err := mountFunc("overlay", mountTarget, "overlay", uintptr(flags), data); err != nil {
-		return "", fmt.Errorf("error creating overlay mount to %s: %v", mountTarget, err)
+		return "", fmt.Errorf("error creating overlay mount to %s, mount_data=%s: %v", mountTarget, mountData, err)
```
edsantiago, line 1172: Will `mountData` ever include colons or spaces? If so would `%q` make those instances easier to read? Sorry for not testing on my end; I don't have much time today.

drivers/overlay/overlay.go
```diff
@@ -1159,7 +1159,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		}
 		mountData = label.FormatMountLabel(opts, options.MountLabel)
 		if len(mountData) > pageSize {
-			return "", fmt.Errorf("cannot mount layer, mount label too large %d", len(mountData))
+			return "", fmt.Errorf("cannot mount layer, mount label %s too large %d", options.MountLabel, len(mountData))
```
edsantiago, line 1162: If I were seeing this error as a user my first question would be, "what is the limit?" Is this a likely error condition? If so, it might be polite to include something like "length (%d) must be <= %d", ..., len(mountData), pageSize)

edsantiago (commented):
/me likes anything that makes error messages more user-friendly, thank you!

TomSweeneyRedHat:
@rhatdan needs a rebase and I concur with @edsantiago 's questions.

drivers/overlay/overlay.go
rhatdan: This is a very unlikely message.

TomSweeneyRedHat:
LGTM

vrothberg (commented):
LGTM

I clicked on rebase. Let's merge when it turns green again.

giuseppe (approved):
LGTM

the issue with https://github.com/containers/podman/issues/9507 seems to be at unmount time though, should we improve logging there as well?

EDIT: nevermind my suggestion, there is not much to improve there and the podman issue is caused by mount

PR #840, opened 2021-03-01
overlay: fix path to incompat/volatile

giuseppe:
overlay creates the directory "$workdir/work/incompat/volatile", not
"$workdir/incompat/volatile".

commit ffe1eb2df5e0ecac7ce794a1925dece6890579c5 introduced the issue.

Closes: https://github.com/containers/podman/issues/9507

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1094,9 +1094,11 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		}
 	}()
 
+	workdir := path.Join(dir, "work")
+
 	var opts string
 	if readWrite {
-		opts = fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s", strings.Join(absLowers, ":"), diffDir, path.Join(dir, "work"))
+		opts = fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s", strings.Join(absLowers, ":"), diffDir, workdir)
 	} else {
 		opts = fmt.Sprintf("lowerdir=%s:%s", diffDir, strings.Join(absLowers, ":"))
 	}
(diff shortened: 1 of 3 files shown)


edsantiago:
Is there any way to add a regression test?

drivers/overlay/overlay.go
```diff
@@ -1167,6 +1159,18 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		mountTarget = path.Join(id, "merged")
 	}
 	flags, data := mount.ParseOptions(mountData)
+
+	// overlay has a check in place to prevent mounting the same file system twice
+	// if volatile was already specified.
+	err = os.RemoveAll(filepath.Join(workdir, "work/incompat/volatile"))
```
rhatdan, line 1165: Is this correct?  It is going to remove $ID/work/work/incompat/volatile?

giuseppe: yes, I got confused in the first version as it removed "$ID/work/incompat/volatile"

rhatdan: Should we just fix what we created rather then what we remove?

giuseppe:
> Is there any way to add a regression test?

added a test

drivers/overlay/overlay.go
giuseppe: this file is created by the overlay driver in the kernel, we have no control over it: https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html?highlight=overlayfs#volatile-mount:

```
When overlay is mounted with “volatile” option, the directory “$workdir/work/incompat/volatile” is created. During next mount, overlay checks for this directory and refuses to mount if present. This is a strong indicator that user should throw away upper and work directories and create fresh one. In very limited cases where the user knows that the system has not crashed and contents of upperdir are intact, The “volatile” directory can be removed.
```

rhatdan (approved):
LGTM

rhatdan:
@edsantiago @vrothberg @TomSweeneyRedHat @umohnani8 @saschagrunert PTAL

umohnani8:
LGTM

edsantiago:
@giuseppe have you built podman with this and run the `diff` system test?

I don't have an easy way to test this, but we will know if it works because podman fedora gating tests will start passing (they are failing right now).

TomSweeneyRedHat:
LGTM
Can you spin up a WIP PR in Podman to see if tests clear?

edsantiago:
A WIP isn't going to help anything because we don't run rawhide in CI

giuseppe:
> @giuseppe have you built podman with this and run the `diff` system test?

yes, I've vendored containers/storage in podman and now "podman diff" works again

edsantiago:
In that case, LGTM!

PR #841, opened 2021-03-01
Inherit system storage driver in rootless configurations

jwillikers:
Fixes containers/podman#9547

The storage driver set in `/etc/containers/storage.conf` was ignored for rootless configurations.
Rootless configurations would use overlay or vfs by default.
If the `STORAGE_DRIVER` environment was variable set, this value would be used instead.

This commit changes this behavior to inherit the driver set in `/etc/containers/storage.conf`.
To inherit the driver in rootless configurations, the driver must be valid in a rootless context.
The valid rootless drivers are btrfs, overlay, and vfs at this time.
To remain consistent with previous behavior, the `STORAGE_DRIVER` environment variable supersedes any driver set in `/etc/containers/storage.conf` and does not need to be a valid rootless driver.

buildah's documentation for the `--storage-driver` command [here](https://github.com/containers/buildah/blob/master/docs/buildah.md#options) will need to be updated to reflect these changes.

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -149,6 +149,17 @@ type StoreOptions struct {
 	AutoNsMaxSize uint32 `json:"auto_userns_max_size,omitempty"`
 }
 
+// isRootlessDriver returns true if the given storage driver is valid for containers running as non root
+func isRootlessDriver(driver string) bool {
+	validDrivers := map[string]bool{
+		"btrfs":    true,
+		"overlay":  true,
+		"overlay2": true,
+		"vfs":      true,
+	}
+	return validDrivers[driver]
+}
+
 // getRootlessStorageOpts returns the storage opts for containers running as non root
 func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOptions, error) {
 	var opts StoreOptions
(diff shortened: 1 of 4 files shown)


types/options.go
```diff
@@ -149,6 +149,14 @@ type StoreOptions struct {
 	AutoNsMaxSize uint32 `json:"auto_userns_max_size,omitempty"`
 }
 
+// isRootlessDriver returns true if the given storage driver is valid for containers running as non root
+func isRootlessDriver(driver string) bool {
+	validDrivers := map[string]bool{
+		"btrfs": true,
```
rhatdan, line 155: Shouldn't vfs and overlayfs be also in this list?

rhatdan:
Could you squash your commits down to one.

rhatdan:
@giuseppe PTAL

types/options.go
jwillikers: Yes, sorry, I was trying to figure out the build error with that change.

jwillikers:
Squashed.

rhatdan (approved):

rhatdan:
LGTM

jwillikers:
I just added a few quick unit tests.

TomSweeneyRedHat:
LGTM
but definitely want a head nod from @giuseppe

giuseppe (approved):

PR #844, opened 2021-03-03
Add dependabot.yml configuration file

saschagrunert:
We now follow the official [dependabot guide][0] for this repository,
which should work with vendoring of go modules, too.

[0]: https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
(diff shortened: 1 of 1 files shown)


rhatdan (approved):
Well see how well this works.
LGTM
@vrothberg PTAL

vrothberg (commented):
LGTM

I am bit afraid of "daily" but in the end that may be easier to tackle than N PRs on a Monday morning :^)

PR #847, opened 2021-03-08
overlay: allow to reset mount_program

giuseppe:
if the mount_program is set to the empty string then clear its value
and not use a mount program instead of failing.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -397,7 +397,7 @@ func parseOptions(options []string) (*overlayOptions, error) {
 			if val != "" {
 				_, err := os.Stat(val)
 				if err != nil {
-					return nil, fmt.Errorf("overlay: can't stat program %s: %v", val, err)
+					return nil, errors.Wrapf(err, "overlay: can't stat program %q", val)
 				}
 			}
 			o.mountProgram = val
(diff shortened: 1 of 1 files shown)


drivers/overlay/overlay.go
```diff
@@ -394,9 +394,11 @@ func parseOptions(options []string) (*overlayOptions, error) {
 			}
 		case "mount_program":
 			logrus.Debugf("overlay: mount_program=%s", val)
-			_, err := os.Stat(val)
-			if err != nil {
-				return nil, fmt.Errorf("overlay: can't stat program %s: %v", val, err)
+			if val != "" {
+				_, err := os.Stat(val)
+				if err != nil {
+					return nil, fmt.Errorf("overlay: can't stat program %s: %v", val, err)
```
saschagrunert, line 400: non-blocking nit since we're here
```suggestion
					return nil, errors.Wrapf(err, "overlay: can't stat program %s", val)
```

saschagrunert (approved):
LGTM

drivers/overlay/overlay.go
giuseppe: thanks!  I've added a new patch with:

```
				return nil, errors.Wrapf(err, "overlay: can't stat program %q", val)
```

saschagrunert (approved):

rhatdan:
LGTM
Although I am not sure why the tests are failing.  Refired.

giuseppe:
tests are green now

PR #851, opened 2021-03-10
chown: ignore both pkg/system.EOPNOTSUPP and pkg/system.ErrNotSupportedPlatform

nalind:
Where we ignored a pkg/system.ErrNotSupportedPlatform error from pkg/system.Lgetxattr(), also ignore ENOTSUP/EOPNOTSUPP, as we already do elsewhere.  Should fix https://github.com/containers/podman/issues/9687.

### DIFF
--- a/drivers/chown_unix.go
+++ b/drivers/chown_unix.go
@@ -3,6 +3,7 @@
 package graphdriver
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"syscall"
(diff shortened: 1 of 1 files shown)


drivers/chown_unix.go
```diff
@@ -49,7 +50,7 @@ func platformLChown(path string, info os.FileInfo, toHost, toContainer *idtools.
 	}
 	if uid != int(st.Uid) || gid != int(st.Gid) {
 		cap, err := system.Lgetxattr(path, "security.capability")
-		if err != nil && err != system.ErrNotSupportedPlatform {
+		if err != nil && !errors.Is(err, system.EOPNOTSUPP) && err != system.ErrNotSupportedPlatform {
```
rhatdan, line 53: LGTM

rhatdan (approved):

giuseppe (approved):
LGTM

flyn-org:
@nalind, silly question from someone who is new to container-related development: how does this change make its way back to podman to fix #9687? The reason I ask is that the podman repository contains a copy of containers/storage. Does the podman build process pull this code from here, or does it use its local copy? Why is there a copy in the podman repository?

rhatdan:
@flyn-org We will cut a release of containers/storage and then in podman we will vendor in (Copy/Update) its code with the newly released code.

PR #853, opened 2021-03-11
Return contiguous user-ns ID mapping ranges.

### LINKED ISSUE #852: getAutoUserns() returns non consecutive usern-ns ID mappings
I am seeing that getAutoUserns() (in store.go) returns non consecutive user-ns ID mappings in scenarios where it should have returned a continuous mapping.

For example, when I call this function to request allocation of 65536 UIDs from host ID range starting at 296608, where mapping {ContainerID: 0, HostID: 296608, Size:1} is passed in as a "used UID", the function returns the following mappings:

```
{ContainerID:0 HostID:296608 Size:1}
{ContainerID:1 HostID:296610 Size:65535} 
```

Note how host ID 296609 was skipped, even though no other container is mapped to that host ID.  Ideally the second range would have started at HostID 296609.

I looked at the code, and it seems function `findAvailableRange()` is the real culprit. It's not returning the correct range in some cases. 

For example, adding this simple sub-test (resembling the example above) to unit test TestFindAvailableIDRange() triggers the problem:

```golang
   avail = []idtools.IDMap{                                                                                                                                                                                                                                                                                                   
      {                                                                                                                                                                                                                                                                                                                       
         ContainerID: 0,                                                                                                                                                                                                                                                                                                      
         HostID:      296608,                                                                                                                                                                                                                                                                                                 
         Size:        65536,                                                                                                                                                                                                                                                                                                  
      },                                                                                                                                                                                                                                                                                                                      
   }                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
   used = []idtools.IDMap{                                                                                                                                                                                                                                                                                                    
      {                                                                                                                                                                                                                                                                                                                       
         ContainerID: 0,                                                                                                                                                                                                                                                                                                      
         HostID:      296608,                                                                                                                                                                                                                                                                                                 
         Size:        1,                                                                                                                                                                                                                                                                                                      
      },                                                                                                                                                                                                                                                                                                                      
   }                                                                                                                                                                                                                                                                                                                          
   ret, err = findAvailableIDRange(65535, avail, used)                                                                                                                                                                                                                                                                        
```

This returns the range `{0 296609 65535}`, instead of range `{1 296609 65535}` as expected (i.e., the ContainerID is incorrect).

As a result, availableUIDs in line 416 below has the incorrect ContainerID, and this in turn causes the call to `subtractAll()` in line 424 to bump up the host ID to 296610, thereby incorrectly skipping 296609.

```
 416 │    availableUIDs, availableGIDs, err = findAvailableRange(size, size, availableUIDs, availableGIDs, usedUIDs, usedGIDs)                                                                                                                                                                                                
 417 │    if err != nil {                                                                                                                                                                                                                                                                                                     
 418 │       return nil, nil, err                                                                                                                                                                                                                                                                                             
 419 │    }                                                                                                                                                                                                                                                                                                                   
 420 │                                                                                                                                                                                                                                                                                                                        
 421 │    // We need to make sure the specified container IDs are also dropped from the automatic                                                                                                                                                                                                                             
 422 │    // namespaces we have found.                                                                                                                                                                                                                                                                                        
 423 │    if len(options.AdditionalUIDMappings) > 0 {                                                                                                                                                                                                                                                                         
 424 │       availableUIDs = subtractAll(availableUIDs, options.AdditionalUIDMappings, false)                                                                                                                                                                                                                                 
 425 │    }  
```

ctalledo:
This commit fixes a bug that was causing function getAutoUserns() (in store.go)
to return non-contigoius user-ns host ID mappings in scenarios where it should
have returned a contiguous mapping.

For example, prior to this change, calling this function to request allocation
of 65536 UIDs from host ID range starting at 296608, where mapping {ContainerID:
0, HostID: 296608, Size:1} is passed in as a "used UID", caused the function
to return the following non-contiguous mappings:

{ContainerID:0 HostID:296608 Size:1}
{ContainerID:1 HostID:296610 Size:65535}

Note how host ID 296609 was skipped, even though no other container is mapped to
that host ID. Ideally the second range would have started at HostID 296609.

After this change, the function returns the following (correct) mappings:

{ContainerID:0 HostID:296608 Size:1}
{ContainerID:1 HostID:296609 Size:65535}

I wrote a unit test in userns_test.go to cover this case. Prior to this
change the unit test fails, with this change it now passes. I also
made some minor improvements to other tests in userns_test.go.

Closes #852.

Signed-off-by: Cesar Talledo <ctalledo@nestybox.com>

### DIFF
--- a/userns.go
+++ b/userns.go
@@ -323,24 +323,26 @@ func findAvailableIDRange(size uint32, availableIDs, usedIDs []idtools.IDMap) ([
 	}
 	avail = subtractAll(avail, usedIDs, true)
 
-	currentID := 0
-	remaining := size
-	// We know the size for each intervals, let's adjust the ContainerID for each
-	// of them.
-	for i := 0; i < len(avail); i++ {
-		avail[i].ContainerID = currentID
-		if uint32(avail[i].Size) >= remaining {
-			avail[i].Size = int(remaining)
-			return avail[:i+1], nil
+	if len(avail) > 0 {
+		currentID := avail[0].ContainerID
+		remaining := size
+		// We know the size for each intervals, let's adjust the ContainerID for each
+		// of them.
+		for i := 0; i < len(avail); i++ {
+			avail[i].ContainerID = currentID
+			if uint32(avail[i].Size) >= remaining {
+				avail[i].Size = int(remaining)
+				return avail[:i+1], nil
+			}
+			remaining -= uint32(avail[i].Size)
+			currentID += avail[i].Size
 		}
-		remaining -= uint32(avail[i].Size)
-		currentID += avail[i].Size
 	}
 
 	return nil, errors.New("could not find enough available IDs")
 }
 
-// findAvailableRange returns both the list of UIDs and GIDs ranges that are not
+// findAvailableRange returns both the list of host UIDs and GIDs ranges that are not
 // currently used by other containers.
 // It is a wrapper for findAvailableIDRange.
 func findAvailableRange(sizeUID, sizeGID uint32, availableUIDs, availableGIDs, usedUIDs, usedGIDs []idtools.IDMap) ([]idtools.IDMap, []idtools.IDMap, error) {
@@ -413,7 +415,20 @@ func (s *store) getAutoUserNS(id string, options *types.AutoUserNsOptions, image
 	// mapping
 	usedUIDs = append(usedUIDs, options.AdditionalUIDMappings...)
 	usedGIDs = append(usedGIDs, options.AdditionalGIDMappings...)
-	availableUIDs, availableGIDs, err = findAvailableRange(size, size, availableUIDs, availableGIDs, usedUIDs, usedGIDs)
+
+	// To find available IDs, we must first subtract from the requested size the
+	// combined size of the all used mappings
+	findUIDSize := requestedSize
+	findGIDSize := requestedSize
+
+	for _, u := range usedUIDs {
+		findUIDSize -= uint32(u.Size)
+	}
+	for _, u := range usedGIDs {
+		findGIDSize -= uint32(u.Size)
+	}
+
+	availableUIDs, availableGIDs, err = findAvailableRange(findUIDSize, findGIDSize, availableUIDs, availableGIDs, usedUIDs, usedGIDs)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -426,5 +441,6 @@ func (s *store) getAutoUserNS(id string, options *types.AutoUserNsOptions, image
 	if len(options.AdditionalGIDMappings) > 0 {
 		availableGIDs = subtractAll(availableGIDs, options.AdditionalGIDMappings, false)
 	}
+
 	return append(availableUIDs, options.AdditionalUIDMappings...), append(availableGIDs, options.AdditionalGIDMappings...), nil
 }
--- a/userns_test.go
+++ b/userns_test.go
@@ -21,6 +21,7 @@ func TestSubtractHostID(t *testing.T) {
 	ret := subtractHostIDs(avail, used)
 	assert.Equal(t, len(ret), 1)
 	assert.Equal(t, avail.HostID, ret[0].HostID)
+	assert.Equal(t, avail.ContainerID, ret[0].ContainerID)
 	assert.Equal(t, avail.Size, ret[0].Size)
 
 	used = idtools.IDMap{
@@ -31,16 +32,7 @@ func TestSubtractHostID(t *testing.T) {
 	ret = subtractHostIDs(avail, used)
 	assert.Equal(t, len(ret), 1)
 	assert.Equal(t, avail.HostID, ret[0].HostID)
-	assert.Equal(t, avail.Size, ret[0].Size)
-
-	used = idtools.IDMap{
-		ContainerID: 0,
-		HostID:      0,
-		Size:        65536,
-	}
-	ret = subtractHostIDs(avail, used)
-	assert.Equal(t, len(ret), 1)
-	assert.Equal(t, avail.HostID, ret[0].HostID)
+	assert.Equal(t, avail.ContainerID, ret[0].ContainerID)
 	assert.Equal(t, avail.Size, ret[0].Size)
 
 	used = idtools.IDMap{
@@ -51,6 +43,7 @@ func TestSubtractHostID(t *testing.T) {
 	ret = subtractHostIDs(avail, used)
 	assert.Equal(t, len(ret), 1)
 	assert.Equal(t, ret[0].HostID, avail.HostID+4096)
+	assert.Equal(t, ret[0].ContainerID, avail.ContainerID+4096)
 	assert.Equal(t, ret[0].Size, avail.Size-4096)
 
 	used = idtools.IDMap{
@@ -61,6 +54,7 @@ func TestSubtractHostID(t *testing.T) {
 	ret = subtractHostIDs(avail, used)
 	assert.Equal(t, len(ret), 1)
 	assert.Equal(t, ret[0].HostID, avail.HostID)
+	assert.Equal(t, ret[0].ContainerID, avail.ContainerID)
 	assert.Equal(t, ret[0].Size, avail.Size-4096)
 
 	used = idtools.IDMap{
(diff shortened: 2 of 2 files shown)


giuseppe:
thanks for the PR!

We don't have an easy way to test this with CRI-O, but we can use Podman.

If you re-vendor containers/storage in Podman, you can run `podman run --userns auto ...` to use the feature.

If I try to vendor this PR, with Podman I get now:

```
$ bin/podman run --userns=auto --rm -ti alpine sh
Error: container uses ID mappings ([]specs.LinuxIDMapping{specs.LinuxIDMapping{ContainerID:0x0, HostID:0x1, Size:0x0}}), but doesn't map UID 0
```

giuseppe (changes requested):
podman tests are failing

giuseppe:
does the following patch solve the issue you've seen (we'd also need to adapt a couple of tests)?

```
diff --git a/userns.go b/userns.go
index 3b0f24fcd..c72a026ad 100644
--- a/userns.go
+++ b/userns.go
@@ -254,7 +254,7 @@ func subtractHostIDs(avail idtools.IDMap, used idtools.IDMap) []idtools.IDMap {
        // Intersection of [avail.HostID, availEnd) and [usedEnd, +inf) is [newStart, availEnd).
        if newStart := maxInt(avail.HostID, usedEnd); newStart < availEnd {
                out = append(out, idtools.IDMap{
-                       ContainerID: newStart + avail.ContainerID - avail.HostID,
+                       ContainerID: avail.ContainerID,
                        HostID:      newStart,
                        Size:        availEnd - newStart,
                })
@@ -282,7 +282,7 @@ func subtractContainerIDs(avail idtools.IDMap, used idtools.IDMap) []idtools.IDM
        if newStart := maxInt(avail.ContainerID, usedEnd); newStart < availEnd {
                out = append(out, idtools.IDMap{
                        ContainerID: newStart,
-                       HostID:      newStart + avail.HostID - avail.ContainerID,
+                       HostID:      avail.HostID,
                        Size:        availEnd - newStart,
                })
        }
```

ctalledo:
> podman tests are failing

Got it, thanks.

> does the following patch solve the issue you've seen (we'd also need to adapt a couple of tests)?

Let me try and will get back soon.

likan999:
> does the following patch solve the issue you've seen (we'd also need to adapt a couple of tests)?
> 
> ```
> diff --git a/userns.go b/userns.go
> index 3b0f24fcd..c72a026ad 100644
> --- a/userns.go
> +++ b/userns.go
> @@ -254,7 +254,7 @@ func subtractHostIDs(avail idtools.IDMap, used idtools.IDMap) []idtools.IDMap {
>         // Intersection of [avail.HostID, availEnd) and [usedEnd, +inf) is [newStart, availEnd).
>         if newStart := maxInt(avail.HostID, usedEnd); newStart < availEnd {
>                 out = append(out, idtools.IDMap{
> -                       ContainerID: newStart + avail.ContainerID - avail.HostID,
> +                       ContainerID: avail.ContainerID,
>                         HostID:      newStart,
>                         Size:        availEnd - newStart,
>                 })
> @@ -282,7 +282,7 @@ func subtractContainerIDs(avail idtools.IDMap, used idtools.IDMap) []idtools.IDM
>         if newStart := maxInt(avail.ContainerID, usedEnd); newStart < availEnd {
>                 out = append(out, idtools.IDMap{
>                         ContainerID: newStart,
> -                       HostID:      newStart + avail.HostID - avail.ContainerID,
> +                       HostID:      avail.HostID,
>                         Size:        availEnd - newStart,
>                 })
>         }
> ```

Hey, this patch totally messes up the mapping between container id and host id. In the HEAD version, it maintains the invariant that the differences between host ids and container ids are unchanged after the subtract. This is one of the bug the PR #764 fixed. Don't get the bug back.

userns.go
```diff
@@ -323,24 +323,26 @@ func findAvailableIDRange(size uint32, availableIDs, usedIDs []idtools.IDMap) ([
...
-	remaining := size
-	// We know the size for each intervals, let's adjust the ContainerID for each
-	// of them.
-	for i := 0; i < len(avail); i++ {
-		avail[i].ContainerID = currentID
-		if uint32(avail[i].Size) >= remaining {
-			avail[i].Size = int(remaining)
-			return avail[:i+1], nil
+	if len(avail) > 0 {
+		currentID := avail[0].ContainerID
```
likan999, line 327: This won't work if the original `availableIDs[0].ContainerID != 0`

giuseppe:
thanks for the review.

> Hey, this patch totally messes up the mapping between container id and host id. In the HEAD version, it maintains the invariant that the differences between host ids and container ids are unchanged after the subtract

with this invariant though we are losing one ID as reported by @ctalledo.  Do you have any suggestions on how to address that?

userns.go
```diff
@@ -413,7 +415,20 @@ func (s *store) getAutoUserNS(id string, options *types.AutoUserNsOptions, image
 	// mapping
 	usedUIDs = append(usedUIDs, options.AdditionalUIDMappings...)
 	usedGIDs = append(usedGIDs, options.AdditionalGIDMappings...)
-	availableUIDs, availableGIDs, err = findAvailableRange(size, size, availableUIDs, availableGIDs, usedUIDs, usedGIDs)
+
+	// To find available IDs, we must first subtract from the requested size the
```
likan999, line 419: If I read the code above correctly, usedUIDs and usedGIDs are taken from existing containers. How other containers should affect the number of ids allocated for this container?

giuseppe: the issue happens because when we subtract (subtractAll with host=false) the IDs already pre-allocated inside the container the HostID is also incremented:

```
{ContainerID:0 HostID:296608 Size:1}  <- this was preallocated, through options.AdditionalUIDMappings
{ContainerID:1 HostID:296610 Size:65535}  <- when we subtract the preallocated range the HostID that was previously 296609 is also changed.
```

The number of IDs is unchanged, but it is preferable to use `{ContainerID:1 HostID:296609 Size:65535}` so we don't leave a hole in the available IDs

likan999: Not sure if I understand the issues correctly, but I have a simpler version of findAvailableIDRange (#856) that passes the unit tests in this commit. Current implementation seems overly complicated. Please let me know if #856 correctly addresses the issues.

ctalledo: Thanks again for the review and apologies for the late reply. Did not mean to break anything on the PR request, which is why I asked about integration tests to vet it (given that the unit tests were passing). Let me look into PR #856.

ctalledo: Looks like PR #856 is not a proper fix either (see my comments there where I described the mapping under 3 scenarios and why some look incorrect). 

For completeness, here are the same scenarios with this current PR (#853): 

1) "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :

{ContainerID:0 HostID:296608 Size:1}
{ContainerID:1 HostID:296609 Size:65535}    <<< LOOKS GOOD

2) Same as before, but change "run_as_user: 296609":

{ContainerID:0 HostID:296609 Size:1}
{ContainerID:1 HostID:296610 Size:65534}     <<< LOOKS GOOD

3) Same as before, but change "run_as_user: 296610":

{ContainerID:0 HostID:296610 Size:1}
{ContainerID:1 HostID:296609 Size:1}            <<< ERROR: HOST-ID DISCONTINUITY
{ContainerID:2 HostID:296611 Size:65533}

This last error was not addressed by this PR. It's a legal mapping, but will create problems for the lower-level runtimes since the mapping of host-IDs is not contiguous. 

Ideally all mappings would be a continuous range of container IDs starting at 0 mapped to a continues range of host IDs starting at whatever host ID is available for a given range size. For example, the last mapping should be {ContainerID: 0 HostID: 296610 Size: 65536}.

giuseppe: > {ContainerID:0 HostID:296610 Size:1}
> {ContainerID:1 HostID:296609 Size:1} <<< ERROR: HOST-ID DISCONTINUITY
> {ContainerID:2 HostID:296611 Size:65533}

Why is there discontinuity in this case?

ctalledo: > Why is there discontinuity in this case?

There is discontinuity in the sense that Container ID 0 maps to HostID 296610, so one expects Container ID 1 to map to 296611, and so on. But Container ID 1 maps to 296609 instead. To be more precise, the mapping is continuous but not linear. This is legal, but unnecessary and will likely cause all sorts of problems for lower-level runtimes.

giuseppe: I think in this case it is necessary and it cannot be linear.

If you specify `run_as_user: 296610` then that is mapped to root in the container, but since 296609 is available, then it is also mapped inside the container to the first unused ID.

ctalledo: Thanks, got it. Non-linearity in the container ID -> host ID mappings will cause extra complexity though. For example, in the Sysbox runtime we use a filesystem called shiftfs, which only works with linear mappings. I've not looked into the recent fs ID mapped mounts mechanism yet, so I don't know if that's affected. 

Also, even if we keep the non-linear mappings, CRI-O should have chosen to map container ID 1 to host ID 296608 (rather than 296609) since that was the first available host ID.

ctalledo:
Closing this PR since @likan999 has an alternative fix in PR #856.

PR #856, opened 2021-03-12
Rework autons ID mapping generation.

likan999:
This implements the algorithm proposed in
https://github.com/containers/storage/issues/852#issuecomment-798954173,
which is:
1. find available IDs from subuid/subgid file; by subtracting the used
   IDs (from other containers) as well as additional IDs, we get the IDs
   available to allocate;
2. target ID range is [0, requestedSize), subtract the additional IDs;
3. allocate IDs from range in step 1; the number to allocate is the
   number of IDs in step 2;
4. generate a mapping from IDs in step 3 to the ones in step 2.

Closes: https://github.com/containers/storage/issues/852

Signed-off-by: Kan Li <likan@google.com>

### DIFF
--- a/idset.go
+++ b/idset.go
@@ -1,0 +1,220 @@
+package storage
+
+import (
+	"github.com/containers/storage/pkg/idtools"
+	"github.com/google/go-intervals/intervalset"
+	"github.com/pkg/errors"
+)
+
+// idSet represents a set of integer IDs. It is stored as an ordered set of intervals.
+type idSet struct {
+	set *intervalset.ImmutableSet
+}
+
+func newIDSet(intervals []interval) *idSet {
+	s := intervalset.Empty()
+	for _, i := range intervals {
+		s.Add(intervalset.NewSet([]intervalset.Interval{i}))
+	}
+	return &idSet{set: s.ImmutableSet()}
+}
+
+// getHostIDs returns all the host ids in the id map.
+func getHostIDs(idMaps []idtools.IDMap) *idSet {
+	var intervals []interval
+	for _, m := range idMaps {
+		intervals = append(intervals, interval{start: m.HostID, end: m.HostID + m.Size})
+	}
+	return newIDSet(intervals)
+}
+
+// getContainerIDs returns all the container ids in the id map.
+func getContainerIDs(idMaps []idtools.IDMap) *idSet {
+	var intervals []interval
+	for _, m := range idMaps {
+		intervals = append(intervals, interval{start: m.ContainerID, end: m.ContainerID + m.Size})
+	}
+	return newIDSet(intervals)
+}
+
+// subtract returns the subtraction of `s` and `t`. `s` and `t` are unchanged.
+func (s *idSet) subtract(t *idSet) *idSet {
+	if s == nil || t == nil {
+		return s
+	}
+	return &idSet{set: s.set.Sub(t.set)}
+}
+
+// union returns the union of `s` and `t`. `s` and `t` are unchanged.
+func (s *idSet) union(t *idSet) *idSet {
+	if s == nil {
+		return t
+	} else if t == nil {
+		return s
+	}
+	return &idSet{set: s.set.Union(t.set)}
+}
+
+// Methods to iterate over the intervals of the idSet. intervalset doesn't provide one :-(
+
+// iterator to idSet. Returns nil if iteration finishes.
+type iteratorFn func() *interval
+
+// cancelFn must be called exactly once unless iteratorFn returns nil, otherwise go routine might
+// leak.
+type cancelFn func()
+
+func (s *idSet) iterator() (iteratorFn, cancelFn) {
+	if s == nil {
+		return func() *interval { return nil }, func() {}
+	}
+	cancelCh := make(chan byte)
+	dataCh := make(chan interval)
+	go func() {
+		s.set.Intervals(func(ii intervalset.Interval) bool {
+			select {
+			case <-cancelCh:
+				return false
+			case dataCh <- ii.(interval):
+				return true
+			}
+		})
+		close(dataCh)
+	}()
+	iterator := func() *interval {
+		i, ok := <-dataCh
+		if !ok {
+			return nil
+		}
+		return &i
+	}
+	return iterator, func() { close(cancelCh) }
+}
+
+// size returns the total number of ids in the ID set.
+func (s *idSet) size() int {
+	var size int
+	iterator, cancel := s.iterator()
+	defer cancel()
+	for i := iterator(); i != nil; i = iterator() {
+		size += i.length()
+	}
+	return size
+}
+
+// findAvailable finds the `n` ids from `s`.
+func (s *idSet) findAvailable(n int) (*idSet, error) {
+	var intervals []intervalset.Interval
+	iterator, cancel := s.iterator()
+	defer cancel()
+	for i := iterator(); n > 0 && i != nil; i = iterator() {
+		i.end = minInt(i.end, i.start+n)
+		intervals = append(intervals, *i)
+		n -= i.length()
+	}
+	if n > 0 {
+		return nil, errors.New("could not find enough available IDs")
+	}
+	return &idSet{set: intervalset.NewImmutableSet(intervals)}, nil
+}
+
+// zip creates an id map from `s` (host ids) and container ids.
+func (s *idSet) zip(container *idSet) []idtools.IDMap {
+	hostIterator, hostCancel := s.iterator()
+	defer hostCancel()
+	containerIterator, containerCancel := container.iterator()
+	defer containerCancel()
+	var out []idtools.IDMap
+	for h, c := hostIterator(), containerIterator(); h != nil && c != nil; {
+		if n := minInt(h.length(), c.length()); n > 0 {
+			out = append(out, idtools.IDMap{
+				ContainerID: c.start,
+				HostID:      h.start,
+				Size:        n,
+			})
+			h.start += n
+			c.start += n
+		}
+		if h.IsZero() {
+			h = hostIterator()
+		}
+		if c.IsZero() {
+			c = containerIterator()
+		}
+	}
+	return out
+}
+
+// interval represents an interval of integers [start, end). Note it is allowed to have
+// start >= end, in which case it is treated as an empty interval. It implements interface
+// intervalset.Interval.
+type interval struct {
+	// Start of the interval (inclusive).
+	start int
+	// End of the interval (exclusive).
+	end int
+}
+
+func (i interval) length() int {
+	return maxInt(0, i.end-i.start)
+}
+
+func (i interval) Intersect(other intervalset.Interval) intervalset.Interval {
+	j := other.(interval)
+	return interval{start: maxInt(i.start, j.start), end: minInt(i.end, j.end)}
+}
+
+func (i interval) Before(other intervalset.Interval) bool {
+	j := other.(interval)
+	return !i.IsZero() && !j.IsZero() && i.end < j.start
+}
+
+func (i interval) IsZero() bool {
+	return i.length() <= 0
+}
+
+func (i interval) Bisect(other intervalset.Interval) (intervalset.Interval, intervalset.Interval) {
+	j := other.(interval)
+	if j.IsZero() {
+		return i, interval{}
+	}
+	// Subtracting [j.start, j.end) is equivalent to the union of intersecting (-inf, j.start) and
+	// [j.end, +inf).
+	left := interval{start: i.start, end: minInt(i.end, j.start)}
+	right := interval{start: maxInt(i.start, j.end), end: i.end}
+	return left, right
+}
+
+func (i interval) Adjoin(other intervalset.Interval) intervalset.Interval {
+	j := other.(interval)
+	if !i.IsZero() && !j.IsZero() && (i.end == j.start || j.end == i.start) {
+		return interval{start: minInt(i.start, j.start), end: maxInt(i.end, j.end)}
+	}
+	return interval{}
+}
+
+func (i interval) Encompass(other intervalset.Interval) intervalset.Interval {
+	j := other.(interval)
+	switch {
+	case i.IsZero():
+		return j
+	case j.IsZero():
+		return i
+	default:
+		return interval{start: minInt(i.start, j.start), end: maxInt(i.end, j.end)}
+	}
+}
+
+func minInt(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func maxInt(a, b int) int {
+	if a < b {
+		return b
+	}
+	return a
+}
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/Microsoft/go-winio v0.4.17-0.20210211115548-6eac466e5fa3
 	github.com/Microsoft/hcsshim v0.8.15
 	github.com/docker/go-units v0.4.0
+	github.com/google/go-intervals v0.0.2
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/klauspost/compress v1.11.13
 	github.com/klauspost/pgzip v1.2.5
--- a/go.sum
+++ b/go.sum
@@ -258,6 +258,8 @@ github.com/google/go-cmp v0.5.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/
 github.com/google/go-cmp v0.5.1/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.2 h1:X2ev0eStA3AbceY54o37/0PQ/UWqKEiiO2dKL5OPaFM=
 github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/go-intervals v0.0.2 h1:FGrVEiUnTRKR8yE04qzXYaJMtnIYqobR5QbblK3ixcM=
+github.com/google/go-intervals v0.0.2/go.mod h1:MkaR3LNRfeKLPmqgJYs4E66z5InYjmCjbbr4TQlcT6Y=
 github.com/google/gofuzz v1.0.0/go.mod h1:dBl0BpW6vV/+mYPU4Po3pmUjxk6FQPldtuIdl/M65Eg=
 github.com/google/gofuzz v1.1.0/go.mod h1:dBl0BpW6vV/+mYPU4Po3pmUjxk6FQPldtuIdl/M65Eg=
 github.com/google/martian v2.1.0+incompatible/go.mod h1:9I4somxYTbIHy5NJKHRl3wXiIaQGbYVAs8BPL6v8lEs=
--- a/idset_test.go
+++ b/idset_test.go
@@ -1,0 +1,924 @@
+package storage
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/containers/storage/pkg/idtools"
+	"github.com/google/go-intervals/intervalset"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMinMaxInt(t *testing.T) {
+	assert.Equal(t, 5, minInt(5, 8))
+	assert.Equal(t, 3, minInt(10, 3))
+	assert.Equal(t, 2, minInt(2, 2))
+	assert.Equal(t, 8, maxInt(5, 8))
+	assert.Equal(t, 10, maxInt(10, 3))
+	assert.Equal(t, 2, maxInt(2, 2))
+}
+
+func allIntervals(s *idSet) []interval {
+	iterator, cancel := s.iterator()
+	defer cancel()
+	var out []interval
+	for i := iterator(); i != nil; i = iterator() {
+		out = append(out, *i)
+	}
+	return out
+}
+
+func idSetsEqual(x, y *idSet) bool {
+	return reflect.DeepEqual(allIntervals(x), allIntervals(y))
+}
+
+func TestNewIDSet(t *testing.T) {
+	tests := []struct {
+		name      string
+		intervals []interval
+		want      *idSet
+	}{
+		{
+			"Nil",
+			nil,
+			&idSet{set: intervalset.NewImmutableSet([]intervalset.Interval{})},
+		},
+		{
+			"Empty",
+			[]interval{},
+			&idSet{set: intervalset.NewImmutableSet([]intervalset.Interval{})},
+		},
+		{
+			"OneValidRange",
+			[]interval{{3, 4}},
+			&idSet{set: intervalset.NewImmutableSet([]intervalset.Interval{interval{3, 4}})},
+		},
+		{
+			"OneEmptyRange",
+			[]interval{{3, 3}},
+			&idSet{set: intervalset.NewImmutableSet([]intervalset.Interval{})},
+		},
+		{
+			"OneInvalidRange",
+			[]interval{{3, 2}},
+			&idSet{set: intervalset.NewImmutableSet([]intervalset.Interval{})},
+		},
+		{
+			"TwoRanges",
+			[]interval{{1, 4}, {6, 8}},
+			&idSet{set: intervalset.NewImmutableSet([]intervalset.Interval{interval{1, 4}, interval{6, 8}})},
+		},
+		{
+			"SecondRangeEmpty",
+			[]interval{{1, 4}, {6, 6}},
+			&idSet{set: intervalset.NewImmutableSet([]intervalset.Interval{interval{1, 4}})},
+		},
+		{
+			"SecondRangeInvalid",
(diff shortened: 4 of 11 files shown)


rhatdan:
@giuseppe PTAL

ctalledo:
Hi @likan999, thanks for the PR.

Unfortunately I think this patch does not look correct either. For example, in my host, user `containers` is assigned  subuid range 296608:655360 in `/etc/subuid`. With this patch, crictl + CRIO generate these mappings:

1) "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :

{ContainerID:0 HostID:296608 Size:1}
{ContainerID:1 HostID:296609 Size:65536}      <<< INVALID SIZE (expect 65535)

2) Same as before, but change "run_as_user: 296609":

{ContainerID:0 HostID:296609 Size:1}
{ContainerID:2 HostID:296610 Size:65535}      <<< CONTAINER ID DISCONTINUITY (expect ContainerID 1).

2) Same as before, but change "run_as_user: 296610":

{ContainerID:0 HostID:296610 Size:1}              
{ContainerID:1 HostID:296609 Size:1}              <<< HOST-ID  DISCONTINUITY
{ContainerID:3 HostID:296611 Size:65534}      <<< CONTAINER-ID DISCONTINUITY


Apart from this patch, I still don't quite understand why "run_as_user" refers to the host ID (as opposed to the container ID), as discussed in issue #852. But even if it does, we should return continuous container ID mappings in all cases, ideally mapped to a set of continuous host ID mappings.  Gaps in either the container ID range or host ID range will cause extra complexity to the lower-level runtimes (e.g., fsuid ID mappings, etc.)

ctalledo:
FYI, with top-of-tree master (i.e., without this fix), here is the what I see on the same 3 scenarios in the prior comment:

 1. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :

{ContainerID:0 HostID:296608 Size:1}
{ContainerID:1 HostID:296610 Size:65535}     <<< HOST-ID DISCONTINUITY  (expect HOST ID 296609) (issue #852)

2. Same as before, but change "run_as_user: 296609":

{ContainerID:0 HostID:296609 Size:1}
{ContainerID:1 HostID:296610 Size:65535}       <<< LOOKS GOOD 

3. Same as before, but change "run_as_user: 296610":

{ContainerID:0 HostID:296610 Size:1}        
{ContainerID:1 HostID:296609 Size:1}                <<<< HOST-ID DISCONTINUITY
{ContainerID:2 HostID:296611 Size:65534}       

Ideally the last set of ranges should have been equivalent to: {ContainerID: 0 HostID: 296610: Size: 65536}

likan999:
> Hi @likan999, thanks for the PR.
> 
> Unfortunately I think this patch does not look correct either. For example, in my host, user `containers` is assigned subuid range 296608:655360 in `/etc/subuid`. With this patch, crictl + CRIO generate these mappings:
> 
> 1. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :
> 
> {ContainerID:0 HostID:296608 Size:1}
> {ContainerID:1 HostID:296609 Size:65536} <<< INVALID SIZE (expect 65535)
> 
> 1. Same as before, but change "run_as_user: 296609":
> 
> {ContainerID:0 HostID:296609 Size:1}
> {ContainerID:2 HostID:296610 Size:65535} <<< CONTAINER ID DISCONTINUITY (expect ContainerID 1).
> 
> 1. Same as before, but change "run_as_user: 296610":
> 
> {ContainerID:0 HostID:296610 Size:1}
> {ContainerID:1 HostID:296609 Size:1} <<< HOST-ID DISCONTINUITY
> {ContainerID:3 HostID:296611 Size:65534} <<< CONTAINER-ID DISCONTINUITY
> 
> Apart from this patch, I still don't quite understand why "run_as_user" refers to the host ID (as opposed to the container ID), as discussed in issue #852. But even if it does, we should return continuous container ID mappings in all cases, ideally mapped to a set of continuous host ID mappings. Gaps in either the container ID range or host ID range will cause extra complexity to the lower-level runtimes (e.g., fsuid ID mappings, etc.)

Can you give steps to reproduce? At least the unit tests in #853 are passing in this PR.

ctalledo:
> Can you give steps to reproduce? At least the unit tests in #853 are passing in this PR. More specifically, what are the parameters of getAutoUserNS and what should be the expected output? I don't have the crictl at hand so I can't reproduce on my side.

The best way to reproduce is to install crictl and use it in conjunction with CRI-O. I am using the following jsons:

alpine-pod.json:

```json
{                                                                                                                                                                                                                                                                                                                             
    "metadata": {                                                                                                                                                                                                                                                                                                             
        "name": "alpine-sandbox",                                                                                                                                                                                                                                                                                             
        "namespace": "default",                                                                                                                                                                                                                                                                                               
        "attempt": 1,                                                                                                                                                                                                                                                                                                         
        "uid": "pod"                                                                                                                                                                                                                                                                                                   
    },                                                                                                                                                                                                                                                                                                                        
    "annotations": {                                                                                                                                                                                                                                                                                                          
        "io.kubernetes.cri-o.userns-mode": "auto:size=65536"                                                                                                                                                                                                                                                                  
    },                                                                                                                                                                                                                                                                                                                        
    "log_directory": "/tmp",                                                                                                                                                                                                                                                                                                  
    "linux": {                                                                                                                                                                                                                                                                                                                
        "security_context": {                                                                                                                                                                                                                                                                                                 
            "run_as_user": {                                                                                                                                                                                                                                                                                                  
                "value": 296608                                                                                                                                                                                                                                                                                               
            }                                                                                                                                                                                                                                                                                                                 
        }                                                                                                                                                                                                                                                                                                                     
    }                                                                                                                                                                                                                                                                                                                         
} 
```

alpine-container.json

```json
{                                                                                                                                                                                                                                                                                                                             
  "metadata": {                                                                                                                                                                                                                                                                                                               
      "name": "alpine"                                                                                                                                                                                                                                                                                                 
  },                                                                                                                                                                                                                                                                                                                          
  "image":{                                                                                                                                                                                                                                                                                                                   
      "image": "alpine"                                                                                                                                                                                                                                                                                       
  },                                                                                                                                                                                                                                                                                                                          
  "command": [                                                                                                                                                                                                                                                                                                                
      "tail", "-f", "/dev/null"                                                                                                                                                                                                                                                                                               
  ],                                                                                                                                                                                                                                                                                                                          
  "log_path":"container.log",                                                                                                                                                                                                                                                                                                 
  "linux": {                                                                                                                                                                                                                                                                                                                  
  }                                                                                                                                                                                                                                                                                                                           
}     
```

And the following crictl command to start the pod:

```
$ crictl run alpine-container.json alpine-pod.json
```

I then check the mappings that get generated inside CRI-O (e.g., using a the debugger).

likan999:
@ctalledo while I setup the environment, could you test this current version and see if it works (to any potential problems faster)? If it returns incorrect result, could you record the parameters of the function `getAutoUserNSIDMappings`? Thanks.

ctalledo:
> @ctalledo while I setup the environment, could you test this current version and see if it works (to any potential problems faster)? If it returns incorrect result, could you record the parameters of the function `getAutoUserNSIDMappings`? Thanks.

Hi @likan999, sure, would be happy to. Will do this later today and get back to you.

ctalledo:
Hi @likan999,

I tested the latest patch you have for this PR and here are my findings:

1) "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 0" :
{ContainerID:0 HostID:0 Size:1}
{ContainerID:1 HostID:296608 Size:65535}    <<< OK

2) "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :
{ContainerID:0 HostID:0 Size:1}    <<< ERROR (expect HostID 296608)
{ContainerID:1 HostID:296608 Size:65535}    <<< ERROR (expect HostID 296609)


Unfortunately, the debugger is not much help in printing the parameters to getAutoUserNSIDMappings() because some of values are hiding behind several levels of pointers:

```
(dlv) args
size = 65536
availableUIDs = ("*github.com/containers/storage.idSet")(0xc0000105c8)
availableGIDs = ("*github.com/containers/storage.idSet")(0xc000010628)
usedUIDMappings = []github.com/containers/storage/pkg/idtools.IDMap len: 0, cap: 0, nil
usedGIDMappings = []github.com/containers/storage/pkg/idtools.IDMap len: 0, cap: 0, nil
additionalUIDMappings = []github.com/containers/storage/pkg/idtools.IDMap len: 1, cap: 1, [...]
additionalGIDMappings = []github.com/containers/storage/pkg/idtools.IDMap len: 1, cap: 1, [...]
```

It seems like the patch fixes the bug in #852, but it does break "run_as_user" settings for values other than 0. Per @giuseppe, "run_as_user" should refer to the host ID mapping for the container's root user. Thus, whenever "run_as_user = X", we expect a mapping of {ContainerID: 0 HostID: X Size: 1}, followed by other mappings for the remainder of the requested range (e.g., {ContainerID: 1 HostID: X+1 Size: 65535}.

Let me know if I can help in any other way please.

likan999:
> Hi @likan999,
> 
> I tested the latest patch you have for this PR and here are my findings:
> 
> 1. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 0" :
>    {ContainerID:0 HostID:0 Size:1}
>    {ContainerID:1 HostID:296608 Size:65535}    <<< OK
> 2. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :
>    {ContainerID:0 HostID:0 Size:1}    <<< ERROR (expect HostID 296608)
>    {ContainerID:1 HostID:296608 Size:65535}    <<< ERROR (expect HostID 296609)
> 
> Unfortunately, the debugger is not much help in printing the parameters to getAutoUserNSIDMappings() because some of values are hiding behind several levels of pointers:
> 
> ```
> (dlv) args
> size = 65536
> availableUIDs = ("*github.com/containers/storage.idSet")(0xc0000105c8)
> availableGIDs = ("*github.com/containers/storage.idSet")(0xc000010628)
> usedUIDMappings = []github.com/containers/storage/pkg/idtools.IDMap len: 0, cap: 0, nil
> usedGIDMappings = []github.com/containers/storage/pkg/idtools.IDMap len: 0, cap: 0, nil
> additionalUIDMappings = []github.com/containers/storage/pkg/idtools.IDMap len: 1, cap: 1, [...]
> additionalGIDMappings = []github.com/containers/storage/pkg/idtools.IDMap len: 1, cap: 1, [...]
> ```
> 
> It seems like the patch fixes the bug in #852, but it breaks "run_as_user" settings for values other than 0. Per @giuseppe, "run_as_user" should refer to the host ID mapping for the container's root user. Thus, whenever "run_as_user = X", we expect a mapping of {ContainerID: 0 HostID: X Size: 1}, followed by other mappings for the remainder of the requested range (e.g., {ContainerID: 1 HostID: X+1 Size: 65535}.
> 
> Let me know if I can help in any other way please.

Thanks @ctalledo , it seems dlv supports pointer dereferencing, could you print contents of the arguments by dereferencing the pointers? Thanks. https://github.com/go-delve/delve/blob/master/Documentation/cli/expr.md

ctalledo:
Hi @likan999,

For the failing case:

```
"io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :

{ContainerID:0 HostID:0 Size:1} <<< ERROR (expect HostID 296608)
{ContainerID:1 HostID:296608 Size:65535} <<< ERROR (expect HostID 296609)
```

The arguments are:

size = 65536
availableUIDs = [{start: 296608, end: 951968}]
usedUIDMappings = []
additionalUIDMappings = {ContainerID: 0, HostID: 0, Size: 1},

likan999:
> Hi @likan999,
> 
> For the failing case:
> 
> ```
> "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :
> 
> {ContainerID:0 HostID:0 Size:1} <<< ERROR (expect HostID 296608)
> {ContainerID:1 HostID:296608 Size:65535} <<< ERROR (expect HostID 296609)
> ```
> 
> The arguments are:
> 
> size = 65536
> availableUIDs = [{start: 296608, end: 951968}]
> usedUIDMappings = []
> additionalUIDMappings = {ContainerID: 0, HostID: 0, Size: 1},

This is interesting. The additionalUIDMappings maps 0 to 0. So is this case broken with HEAD, or is it working on HEAD but broken by this PR? It looks like with HEAD it also maps 0 to 0: https://github.com/containers/storage/blob/9d6cee014f3aa74800ee23c053018a99e5215322/userns.go#L429

@giuseppe @ctalledo Now the questions are:
1. Why is additionalUIDMappings 0->0, instead of 296608->0, in this case?
2. Is this library the one that should handle `run_as_user=X`, or is it the caller of this library who should set up the additionalUIDMappings to 296608->0 based on `run_as_user=X`? I don't see an option that specifies `run_as_user`, so how is this library supposed to know? (By saying "this library" I mean this repo, more specifically, the `CreateContainer` method).
3. Or is `{ContainerID:0 HostID:0 Size:1}` really wrong? In the past, I looked at how my container's user namespace is set up, and I had a vague impression that there are two levels of namespace mappings. Is it possible that in some cases, in an outer namespace it maps X to 0; while in an inner namespace it maps 0 to 0, which is where this library is called?

ctalledo:
Hi @likan999 ,

> Why is additionalUIDMappings 0->0, instead of 296608->0, in this case?

Good point; should be 296608->0. I can't answer since I am not familiar enough with the code.

> Is this library the one that should handle run_as_user=X, or is it the caller of this library who should set up the additionalUIDMappings to 296608->0 based on run_as_user=X? I don't see an option that specifies run_as_user, so how is this library supposed to know? (By saying "this library" I mean this repo, more specifically, the CreateContainer method).

I believe it's the caller of the library (CRI-O in my case), but @giuseppe can confirm.

> Or is {ContainerID:0 HostID:0 Size:1} really wrong? In the past, I looked at how my container's user namespace is set up, and I had a vague impression that there are two levels of namespace mappings. Is it possible that in some cases, in an outer namespace it maps X to 0; while in an inner namespace it maps 0 to 0, which is where this library is called?

No, there is no user-ns nesting in this setup.

giuseppe:
> Hi @likan999 ,
> 
> > Why is additionalUIDMappings 0->0, instead of 296608->0, in this case?
> 
> Good point; should be 296608->0. I can't answer since I am not familiar enough with the code.

yes, when run_as_user it is specified, CRI-O creates the additional mapping

> Something must have changed recently thought (either in this patch or in upstream CRI-O) because in my earlier comment, I indicated than when the pod spec had "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" , I got the following mapping (which is incorrect but does have 296608->0 at least):
> 
> ```
> {ContainerID:0 HostID:296608 Size:1}
> {ContainerID:1 HostID:296609 Size:65536} <<< INVALID SIZE (expect 65535)
> ```
> 
> > Is this library the one that should handle run_as_user=X, or is it the caller of this library who should set up the additionalUIDMappings to 296608->0 based on run_as_user=X? I don't see an option that specifies run_as_user, so how is this library supposed to know? (By saying "this library" I mean this repo, more specifically, the CreateContainer method).
> 
> I believe it's the caller of the library (CRI-O in my case), but @giuseppe can confirm.
> 
> > Or is {ContainerID:0 HostID:0 Size:1} really wrong? In the past, I looked at how my container's user namespace is set up, and I had a vague impression that there are two levels of namespace mappings. Is it possible that in some cases, in an outer namespace it maps X to 0; while in an inner namespace it maps 0 to 0, which is where this library is called?
> 
> No, there is no user-ns nesting in this setup.

If CRI-O creates the mapping 0 -> 0 then that is an issue in CRI-O.  I'll have to verify that.

Were you seeing the same behavior also without re-vendoring this version of containers/storage?

giuseppe:
@ctalledo can you share the .yaml file you've used?

I've tried with:

```
apiVersion: v1
kind: ReplicationController
metadata:
  name: sleep
spec:
  replicas: 1
  selector:
    app: sleep
  template:
    metadata:
      name: sleep
      labels:
        app: sleep
      annotations:
        io.kubernetes.cri-o.userns-mode: "auto:keep-id=true"
    spec:
      securityContext:
        runAsUser: 1002
        runAsGroup: 1003
        fsGroup: 1001
      containers:
      - name: sleep
        command: ["/bin/sh", "-c", "while true; do id; sleep 100000; done"]
        image: alpine
```

and I get both with master and this PR (applied to CRI-O commit: 50b6b5c87598f751468c07fd70d3c3b30fbc9701):

```
# kubectl logs pod/sleep-b8q6s
uid=1002(1002) gid=1003 groups=1001

# cat /proc/158461/uid_map 
         0     265535       1002
      1003     266537      64533
      1002       1002          1

# cat /proc/158461/gid_map 
         0     265534       1001
      1002     266535          1
      1004     266536      64532
      1003       1003          1
      1001       1001          1
```

I've also added this patch to containers/storage:

```
diff --git a/vendor/github.com/containers/storage/store.go b/vendor/github.com/containers/storage/store.go
index 9c08eda69..79b62bf82 100644
--- a/vendor/github.com/containers/storage/store.go
+++ b/vendor/github.com/containers/storage/store.go
@@ -1261,6 +1261,8 @@ func (s *store) CreateContainer(id string, names []string, image, layer, metadat
 
        if options.AutoUserNs {
                var err error
+               fmt.Printf("Request UIDs %v\n", options.AutoUserNsOpts.AdditionalUIDMappings)
+               fmt.Printf("Request GIDs %v\n", options.AutoUserNsOpts.AdditionalGIDMappings)
                options.UIDMap, options.GIDMap, err = s.getAutoUserNS(id, &options.AutoUserNsOpts, cimage)
                if err != nil {
                        return nil, err
```

and I get:
```
Request UIDs [{1002 1002 1}]
Request GIDs [{1001 1001 1}]    
```

Do you get anything different?

ctalledo:
Hi @likan999, @giuseppe ,

My apologies, I think I reported incorrect results in my prior comment. The reason for my mistake is that I had added the "run_as_user" on the crictl's container spec rather than in the pod-spec (I was mislead by [this CRI-O test example](https://github.com/cri-o/cri-o/blob/master/test/testdata/container_config.json)).

Anyway, I re-did the manual testing and things look better now.  Below are the results:

With top of tree CRI-O + containers/storage with this patch:

"io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 0" :
{0 0 1}
{1 296608 65535}    <<<  Better than top-of-tree (starts with host ID 296608 rather than 296609)

"io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :
{0 296608 1}
{1 296609 65535}     <<< Looks good (fixes #852).

"io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296609" :
{0 296609 1}
{1 296608 1}             <<< Legal mapping, but breaks linearity on the host ID
{2 296610 65534}     <<< Looks good

In the last example, the mappings are legal but notice that they are non-linear on the host-ID range. I argue this will cause problems for lower level runtimes (e.g., we use shiftfs in [sysbox-runc](https://github.com/nestybox/sysbox), and shiftfs requires linear mappings, such as {0 296609 65536}).

For comparison, with top of tree CRI-O without this patch:

"io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 0" :
{0 0 1}
{1 296609 65535} 

"io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :
{0 296608 1}
{1 296610 65535}     <<< BUG: issue #852 

"io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296609" :
{0 296609 1}
{1 296610 65535}     <<< Linear mapping

Notice that in the last case, top-of-tree CRIO does generate linear mappings. 

I am not asking that the linearity issue be fixed in this PR, just noting it.

likan999:
> Hi @likan999, @giuseppe ,
> 
> My apologies, I think I reported incorrect results in my prior comment. The reason for my mistake is that I had added the "run_as_user" on the crictl's container spec rather than in the pod-spec (I was mislead by [this CRI-O test example](https://github.com/cri-o/cri-o/blob/master/test/testdata/container_config.json)).
> 
> Anyway, I re-did the manual testing and things look better now. Below are the results:
> 
> With top of tree CRI-O + containers/storage with this patch:
> 
>     1. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 0" :
>        {0 0 1}
>        {1 296608 65535}    <<<  Better than top-of-tree (starts with host ID 296608 rather than 296609)
> 
>     2. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :
>        {0 296608 1}
>        {1 296609 65535}     <<< Looks good (fixes #852).
> 
>     3. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296609" :
>        {0 296609 1}
>        {1 296608 1}             <<< Legal mapping, but breaks linearity on the host ID
>        {2 296610 65534}     <<< Looks good
> 
> 
> In the last example, the mappings are legal but notice that they are non-linear on the host-ID range. I argue this will cause problems for lower level runtimes (e.g., we use shiftfs in [sysbox-runc](https://github.com/nestybox/sysbox), and shiftfs requires linear mappings, such as {0 296609 65536}).
> 
> For comparison, with top of tree CRI-O without this patch:
> 
>     1. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 0" :
>        {0 0 1}
>        {1 296609 65535}
> 
>     2. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296608" :
>        {0 296608 1}
>        {1 296610 65535}     <<< BUG: issue #852
> 
>     3. "io.kubernetes.cri-o.userns-mode": "auto:size=65536", "run_as_user: 296609" :
>        {0 296609 1}
>        {1 296610 65535}     <<< Linear mapping
> 
> 
> Notice that in the last case, top-of-tree CRIO does generate linear mappings.
> 
> I am not asking that the linearity issue be fixed in this PR, just noting it.

Thanks for confirming it works. Sure for the 3rd case, we can probably improve the allocation algorithm in the future. Right now it just takes the lowest N available IDs; potentially we can try finding N consecutive IDs. However, this may needs more thoughts, as it is unclear how to define linearity: if the container ids are not contiguous, then is linearity defined as host IDs being contiguous, or they have the same difference to the corresponding container IDs, or something else.

ctalledo:
> Thanks for confirming it works. 

Sure @likan999 , thank you for implementing the fix.

> for the 3rd case, we can probably improve the allocation algorithm in the future. Right now it just takes the lowest N available IDs; potentially we can try finding N consecutive IDs. However, this may needs more thoughts, as it is unclear how to define linearity: if the container ids are not contiguous, then is linearity defined as host IDs being contiguous, or they have the same difference to the corresponding container IDs, or something else.

Yes, the topic of linearity for ID mappings is not something this PR should fix. In my mind, container IDs are always linear, and it would be best if they also map to a linear and consecutive host ID range (e.g., container ID 0 -> 65535 maps to host IDs X + (X+65535)). Otherwise it creates more complexity for the lower level runtimes (shiftfs is one example of something that requires such linearity).

kevinlee999:
@giuseppe mind reviewing this PR?

likan999:
@giuseppe @ctalledo do you mind reviewing the PR?

idset.go
```diff
@@ -0,0 +1,214 @@
...
+	start int
+	// End of the interval (exclusive).
+	end int
+}
+
+func (i interval) length() int {
+	return i.end - i.start
+}
+
+type idSet struct {
```
ctalledo, line 22: Suggest adding a brief description of what idSet represents.

idset.go
```diff
@@ -0,0 +1,214 @@
...
+			h = hostIterator()
+		}
+		if c.IsZero() {
+			c = containerIterator()
+		}
+	}
+	return out
+}
+
+func (i interval) Intersect(other intervalset.Interval) intervalset.Interval {
```
ctalledo, line 164: Suggestion: group type `interval` and all it's methods on one part of this file, and type `idSet` and all its methods on another part of the file. Currently they are interleaved which makes it harder to read.

giuseppe (approved):
LGTM

ctalledo (commented):
In general this looks good, though I am not super-familiar with the code. In particular, the changes to `getAutoUserNSIDMappings` look good to me. 

I would suggest adding unit tests for the newly introduced `interval` and `idSet` types, given that they each have several methods on which higher level functions rely.

Also, please get a review from a developer that is more familiar with this code that I am (likely @giuseppe).

likan999:
> In general this looks good, though I am not super-familiar with the code. In particular, the changes to `getAutoUserNSIDMappings` look good to me.
> 
> I would suggest adding unit tests for the newly introduced `interval` and `idSet` types, given that they each have several methods on which higher level functions rely.
> 
> Also, please get a review from a developer that is more familiar with this code that I am (likely @giuseppe).

Done addressing the comments: added test cases, moved it around, etc. PTAL.

ctalledo (approved):
Looks great, thanks for reorganizing the idset.go file and for the writing thorough unit tests for it.

rhatdan:
LGTM

PR #860, opened 2021-03-22
Use an xz library instead of shelling out to xz for decompression

nalind:
When decompressing layers compressed with `xz`, use a library (`github.com/ulikunitz/xz`, already used by the image library) rather than shelling out to the `xz` CLI.


mheon:
Thanks! LGTM

rhatdan (approved):
LGTM

TomSweeneyRedHat:
Addresses CVE-2021-20291

PR #861, opened 2021-03-23
[release-1.24] Use an xz library instead of shelling out to xz for decompression

haircommander:
When decompressing layers compressed with xz, use a library (github.com/ulikunitz/xz, already used by the image library) rather than shelling out to the xz CLI.


haircommander:
ah crap

PR #862, opened 2021-03-23
[release-1.24] Use an xz library instead of shelling out to xz for decompression

haircommander:
When decompressing layers compressed with xz, use a library (github.com/ulikunitz/xz, already used by the image library) rather than shelling out to the xz CLI.


TomSweeneyRedHat:
@haircommander all kinds of unhappy red tests

nalind:
LGTM

contrib/cirrus/setup.sh
```diff
@@ -12,23 +12,17 @@ cd $GOSRC
...
-        $SHORT_APTGET update  # Fetch latest package metadata
-        [[ -z "$DEBS_HOLD" ]] || \
-            apt-mark hold $DEBS_HOLD
-        $LONG_APTGET upgrade # install latest packages
-        [[ -z "$DEBS_REQUIRED" ]] || \
-            $SHORT_APTGET -q install $DEBS_REQUIRED
         [[ -z "$DEBS_CONFLICTING" ]] || \
             $SHORT_APTGET -q remove $DEBS_CONFLICTING
+        $SHORT_APTGET -q update
+        $SHORT_APTGET -q install zstd
```
cevich, line 25: I think installing `zstd` at runtime is not necessary if you tell `.cirrus.yml` to use images with ID `c6374915067609088` (F32, F33 + U2004, U2010).

cevich: Ya, [just update this line.](https://github.com/containers/storage/blob/ff9149bcd8a6d2419ba7811fe9477d664c31a55f/.cirrus.yml#L28)

Runtime package installs nearly always end badly...eventually :D

nalind: Right, this is a backport from a PR that didn't want to have to wait for that.

cevich: Oh okay, if it's a backport that's fine.  I'm okay either way then.

rhatdan:
LGTM

PR #867, opened 2021-03-31
    ReloadConfigurationFile should Reset storage options

rhatdan:
Currently in Podman if we reset the CONTAINERS_STORAGE_CONF path, we get
the graph driver options from the original config file as well as the
options from the override path. This PR resets the storageconf to the
initial state when called multiple times.

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -278,6 +278,9 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 		fmt.Printf("Failed to parse %s %v\n", configFile, err.Error())
 		return
 	}
+
+	// Clear storeOptions of previos settings
+	*storeOptions = StoreOptions{}
 	if config.Storage.Driver != "" {
 		storeOptions.GraphDriverName = config.Storage.Driver
 	}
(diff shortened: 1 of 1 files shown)


rhatdan:
Basically I am trying to not use /var/lib/shared as an addiitonal store.

CONTAINERS_STORAGE_CONF=/etc/containers/storage.conf.shared ./bin/podman --root /var/lib/shared pull alpine

saschagrunert (approved):
LGTM

rhatdan:
@nalind @vrothberg @giuseppe PTAL

nalind:
How much of `storeOptions`'s previous value before `ReloadConfigurationFile()` was called is retained here?  If the answer is "none of it", it'd be much simpler to clear the whole structure once at the start of the function.

types/options.go
```diff
@@ -288,15 +289,20 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 	if storeOptions.GraphDriverName == "" {
 		logrus.Errorf("The storage 'driver' option must be set in %s, guarantee proper operation.", configFile)
```
nalind, line 290: Hey, that reminds me: the example config we provide in this repository triggers this message, but it is often used as-is by third-party packaging, because why wouldn't it?

rhatdan: Would you want to change our default in storage.conf to "overlay", so that this is not an issue?

rhatdan:
I don't believe we want to retain anything, so just resetting storeOptions to init value.

types/options.go
nalind: That should work, I think.

types/options.go
```diff
@@ -278,6 +278,9 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 		fmt.Printf("Failed to parse %s %v\n", configFile, err.Error())
 		return
 	}
+
+	// Clear storeOptions of previos settings
+	storeOptions = StoreOptions{}
```
giuseppe, line 283: should this be `storeOptions = &StoreOptions{}` ?

rhatdan: Yup

rhatdan:
@nalind @TomSweeneyRedHat @giuseppe @umohnani8 @saschagrunert @vrothberg PTAL

rhatdan:
I want to get this in so we can cut a release.

vrothberg (commented):
LGTM

saschagrunert (approved):
LGTM

PR #870, opened 2021-04-02
Set default to overlay from storage.conf

rhatdan:
If users do not set the storage driver in their storage.conf, tools that
use containers/storage will print a warning. Since we recommend that
users should use "overlay" driver by default, we should set the settings
in our default storage.conf. This ways distros that just grab the
storage.conf in the storage library will work without the warning.

### DIFF
--- a/storage.conf
+++ b/storage.conf
@@ -5,7 +5,7 @@
 [storage]
 
 # Default Storage Driver, Must be set for proper operation.
-driver = ""
+driver = "overlay"
 
 # Temporary storage location
 runroot = "/run/containers/storage"


rhatdan:
@nalind PTAL

nalind:
LGTM

TomSweeneyRedHat:
Change LGTM
but tests aren't hip

TomSweeneyRedHat:
@ypu @ddarrah we may need to adjust some of the RHEL tests for this.  Thoughts?

ypu:
Hi @TomSweeneyRedHat  As overlayfs is the default storage driver from code this should be not influence of testing in rhel a lot for now. But it is good to have a record for such change in upstream from QE side. Thanks a lot for your reminding.

PR #872, opened 2021-04-06
rootless overlay: use user.* instead of trusted.*

giuseppe:
unprivileged users cannot use the trusted.* xattrs.  Since for rootless we always mount overlay with userxattr, we can just check if
running in rootless mode and use user.* instead of trusted.*.

Closes: https://github.com/containers/podman/issues/9936

### DIFF
--- a/drivers/copy/copy_linux.go
+++ b/drivers/copy/copy_linux.go
@@ -25,6 +25,7 @@ import (
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/pools"
 	"github.com/containers/storage/pkg/system"
+	"github.com/containers/storage/pkg/unshare"
 	rsystem "github.com/opencontainers/runc/libcontainer/system"
 	"golang.org/x/sys/unix"
 )
@@ -291,6 +292,10 @@ func doCopyXattrs(srcPath, dstPath string) error {
 				return err
 			}
 		}
+	}
+
+	if unshare.IsRootless() {
+		return nil
 	}
 
 	// We need to copy this attribute if it appears in an overlay upper layer, as
--- a/pkg/archive/archive_linux.go
+++ b/pkg/archive/archive_linux.go
@@ -12,6 +12,10 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+func getOverlayOpaqueXattrName() string {
+	return GetOverlayXattrName("opaque")
+}
+
 func GetWhiteoutConverter(format WhiteoutFormat, data interface{}) TarWhiteoutConverter {
 	if format == OverlayWhiteoutFormat {
 		if rolayers, ok := data.([]string); ok && len(rolayers) > 0 {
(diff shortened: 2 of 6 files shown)


drivers/copy/copy_linux.go
```diff
@@ -297,5 +297,8 @@ func doCopyXattrs(srcPath, dstPath string) error {
 	// this function is used to copy those. It is set by overlay if a directory
 	// is removed and then re-created and should not inherit anything from the
 	// same dir in the lower dir.
-	return copyXattr(srcPath, dstPath, "trusted.overlay.opaque")
+	if err := copyXattr(srcPath, dstPath, "trusted.overlay.opaque"); err != nil {
```
rhatdan, line 300: This seems like you are still doing both.  I don't see a check for rootless versus rootfull?  If I am running rootless, will the first copyXattr fail and return an error and never get to the second?

pkg/archive/archive_linux.go
```diff
@@ -9,9 +9,17 @@ import (
 
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/system"
+	"github.com/containers/storage/pkg/unshare"
 	"golang.org/x/sys/unix"
 )
 
+func getWhiteoutXattrName() string {
```
rhatdan, line 16: Shouldn't you use the getXattrName function?

drivers/copy/copy_linux.go
giuseppe: the first copy wouldn't be performed as the xattr is missing.

I thought there could be cases where the copy runs as root and misses to copy the `user.*` xattr if present.

I can simplify it and check `unshare.IsRootless()`

pkg/archive/archive_linux.go
giuseppe: it is in a different package

rhatdan: Sure, but should we make it public and use it?

giuseppe: the best place seems `drivers/overlay`.  That will cause to have a dependency on it from `pkg/archive` and `drivers/copy`.

Do you have any suggestions?  `pkg/unshare`?

rhatdan: I was thinking just put it in archive?

giuseppe: I've tried in archive but it looks out of place.

What do you think about the current implementation and having the dependency only for `pkg/archive`?

rhatdan (approved):
LGTM

rhatdan:
@mheon @nalind PTAL

nalind:
If we're continuing to extract and apply layer diffs using the overlay-aware logic rather than the naive logic, how are we handling whiteouts?

giuseppe:
Native overlay allows device whiteouts also in a user namespace.  The special whiteout device can be created without capabilities

nalind:
Okay, so that was new in 5.7.

mheon:
LGTM

TomSweeneyRedHat:
LGTM and happy green test buttons.

vrothberg (commented):
LGTM

PR #879, opened 2021-04-22
Validate selinux label before attempting to use it

rhatdan:
Certain platforms might not have container_file_t defined, so we need to
check if the label is defined before attempting to use it.

Discussion is going on in: https://github.com/containers/podman/issues/9682


rhatdan:
@bigon PTAL

rhatdan:
@nalind @vrothberg @giuseppe @umohnani8 @ashley-cui PTAL

giuseppe (approved):
LGTM

nalind:
If the label doesn't validate, does this also prevent us from trying to set labels when mounting overlay filesystems?

rhatdan:
It will prevent you from using that label.  This was brought up for a distribution that had selinux enabled but did not have container-selinux policy.  On these systems the containers would have SELinux separation disabled.  In the case of this test, though we would not even support overlay file systems for disabled separated containers.

rhatdan:
@vrothberg @saschagrunert @nalind @TomSweeneyRedHat @ashley-cui @umohnani8 PTAL

saschagrunert (approved):
LGTM

PR #881, opened 2021-04-26
Allow users to tag images in read/only image stores

rhatdan:
Fixes:  https://github.com/containers/podman/issues/9412

### DIFF
--- a/store.go
+++ b/store.go
@@ -2115,6 +2115,37 @@ func (s *store) SetNames(id string, names []string) error {
 		return ristore.SetNames(id, deduped)
 	}
 
+	// Check is id refers to a RO Store
+	ristores, err := s.ROImageStores()
+	if err != nil {
+		return err
+	}
+	for _, s := range ristores {
+		store := s
+		store.RLock()
+		defer store.Unlock()
+		if modified, err := store.Modified(); modified || err != nil {
+			if err = store.Load(); err != nil {
+				return err
+			}
+		}
+		if i, err := store.Get(id); err == nil {
+			// Unlock R/O lock and lock with R/W lock
+			// Previous defer.Unlock() will free the new lock.
+			ristore.Unlock()
+			ristore.Lock()
+			if len(deduped) > 1 {
+				// Do not want to create image name in R/W storage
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
Alll sorts of test unhappiness @rhatdan

rhatdan:
@nalind @giuseppe @vrothberg @TomSweeneyRedHat @umohnani8 @saschagrunert PTAL

giuseppe (approved):
LGTM

saschagrunert (approved):
LGTM

store.go
```diff
@@ -2115,6 +2115,37 @@ func (s *store) SetNames(id string, names []string) error {
...
+		}
+		if i, err := store.Get(id); err == nil {
+			// Unlock R/O lock and lock with R/W lock
+			// Previous defer.Unlock() will free the new lock.
+			ristore.Unlock()
+			ristore.Lock()
+			if len(deduped) > 1 {
+				// Do not want to create image name in R/W storage
+				deduped = deduped[1:]
+			}
```
nalind, line 2140: Can someone walk me through what's going on from line 2135 to 2140?
Why is the lock we take on 2136 read-write if the lock taken at line 2107 was read-only?
Why do we drop the first (and only the first) name of the possibly many names that the image record in the read-only store has attached to it when adding a record to the read-write store that will include the new name?  What happens when someone calls us without exactly the set of names in `i.Names`?

rhatdan: The first name  is the existing image, the follow on names are the names being tagged.
Since we did not find the first name in the Read/Write store, but only in the Read/Only store, we want to add the new names to the readwrite store.  We need the read/write lock now, so close the existing read/only lock and create a read/write lock, and then create images with the new name.

nalind: The existing image can have had multiple names attached to it, and there's no requirement that `SetNames()` include more names than were already set on the image.  That sort of explains the `> 1` test for me, but it sounds like you're making an assumption here that isn't correct.
The read-write store was already locked for writing on line 2107.  I still don't see the point of releasing the lock on line 2135 only to reacquire it on the very next line.

PR #882, opened 2021-04-27
idtools: add support for libsubid

giuseppe:
when building with cgo, add support for libsubid to read the
additional sub IDs for the user instead of parsing the /etc/sub?id
files.


rhatdan:
Thanks @giuseppe When do you think libsubuid will show up?

giuseppe:
> Thanks @giuseppe When do you think libsubuid will show up?

I am not sure about that, AFAIK there wasn't even a release upstream yet.

@alexey-tikhonov do you know when a new shadow-utils can land in Fedora?

alexey-tikhonov:
It is actually in Fedora 35 already: https://bodhi.fedoraproject.org/updates/FEDORA-2021-8483b52301

(but most probably will need an update when https://github.com/shadow-maint/shadow/issues/325 is fixed)

giuseppe:
thanks!  Sorry I've missed it before.

I've tested the current patch with the package available for F35 and it works fine

rhatdan:
What do we need to do to get this in and able to work on distros without libsubid?

giuseppe:
> What do we need to do to get this in and able to work on distros without libsubid?

Probably using a build tag would be the easiest and we can use it in Rawhide

alexey-tikhonov:
Btw, there were several changes in libsubid (including https://github.com/shadow-maint/shadow/commit/3d670ba7ed58f910f1395569ea14cd39fd046cc7)
We will rebuild package in Rawhide soon (probably after shadow-utils upstream release)

rhatdan:
Can you add something like:
```
cat hack/libsubid_tag.sh
#!/usr/bin/env bash
if test $(${GO:-go} env GOOS) != "linux" ; then
	exit 0
fi
tmpdir="$PWD/tmp.$RANDOM"
mkdir -p "$tmpdir"
trap 'rm -fr "$tmpdir"' EXIT
cc -o "$tmpdir"/libdm_tag -ldevmapper -x c - > /dev/null 2> /dev/null << EOF
#include <"shadow/subid.h">
int main() {
        subid_range *ranges;
	err := get_subuid_ranges("root", &ranges);
	return 0;
}
EOF
if test $? -eq 0 ; then
	echo libsubid
fi
```

rhatdan:
Also modify the Makefile with

AUTOTAGS := $(shell ./hack/btrfs_tag.sh) $(shell ./hack/libdm_tag.sh) $(shell ./hack/libsubid_tag.sh)

rhatdan:
Perhaps we should make the Tag a negative as well so it will always get compiled in unless users specify don't use it.

giuseppe:
switched the logic

hack/libsubid_tag.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+if test $(${GO:-go} env GOOS) != "linux" ; then
+	exit 0
```
rhatdan, line 3: This should also do 
 	echo no_libsubid

rhatdan:
One issue, and then LGTM
Is this still a WIP, or something we can merge now?

hack/libsubid_tag.sh
giuseppe: amended

giuseppe:
>  Is this still a WIP, or something we can merge now?

it is a WIP because the libsubid API were not released yet.

@alexey-tikhonov is there any risk the API will change again or can we merge this PR?

rhatdan:
LGTM

rhatdan:
It would be nice to get a version of Podman into Rawhide that used libsubid, to make sure all of the plumbing works.

rhatdan:
@alexey-tikhonov PTAL

pkg/idtools/idtools_supported.go
```diff
@@ -0,0 +1,60 @@
...
+{
+    return ranges[i];
+}
+*/
+import "C"
+
+func readSubid(username string, isUser bool) (ranges, error) {
+	var ret ranges
+	if username == "ALL" {
+		return nil, fmt.Errorf("username ALL not supported")
```
saschagrunert, line 25:
```suggestion
		return nil, errors.New("username ALL not supported")
```

saschagrunert (approved):
Just a non-blocking nit, otherwise LGTM

pkg/idtools/idtools_supported.go
giuseppe: thanks, fixed now

alexey-tikhonov:
> it is a WIP because the libsubid API were not released yet.
> 
> @alexey-tikhonov is there any risk the API will change again or can we merge this PR?

To the best of my knowledge, I don't anticipate `libsubid` API changes in a near future. (And this is already merged in C9S: https://gitlab.com/redhat/centos-stream/rpms/shadow-utils/-/commit/fbf9d3a3ea5ea94a96eb747034c5ca9a25095633)

But that's true that there is no upstream release yet and I'm not authoritative source here.
@hallyn, what's your current plan with regards to upstream release of shadow-utils?



> @alexey-tikhonov PTAL

Looks good to me.

alexey-tikhonov:
JFYI: https://github.com/shadow-maint/shadow/releases/tag/v4.9

ikerexxe:
I'm working on rebasing Fedora rawhide to the new version of shadow-utils. Once it is ready I'll let you know.

rhatdan:
Awesome news, Are you planning on releasing this to Fedora 34?  We will need this released/backported to RHEL8.6, also.

alexey-tikhonov:
> Awesome news, Are you planning on releasing this to Fedora 34? We will need this released/backported to RHEL8.6, also.

RHEL8.6 - sure, Fedora 34 - not planned at the moment.

rhatdan:
@giuseppe Now that libsubuid is in Rawhide, can we move this out of draft?

giuseppe:
rebased

ikerexxe:
Rawhide update is ready https://bodhi.fedoraproject.org/updates/FEDORA-2021-3466bdabc1

PR #885, opened 2021-04-28
Log expected rootless overlay mount failures as debug level

rhatdan:
Most linux kernels do not support overlay mounts in rootless mode,
we should not be reporting this as an error, but drop it to debug
level.

Fixes: https://github.com/containers/podman/issues/10153

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -620,7 +620,7 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 		if len(flags) < unix.Getpagesize() {
 			err := unix.Mount("overlay", mergedDir, "overlay", 0, flags)
 			if err == nil {
-				logrus.Errorf("overlay test mount with multiple lowers failed, but succeeded with a single lower")
+				logrus.StandardLogger().Logf(logLevel, "overlay test mount with multiple lowers failed, but succeeded with a single lower")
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe @saschagrunert @vrothberg @nalind @TomSweeneyRedHat @umohnani8 PTAL

umohnani8:
LGTM

saschagrunert (approved):

drivers/overlay/overlay.go
```diff
@@ -620,7 +620,7 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 		if len(flags) < unix.Getpagesize() {
 			err := unix.Mount("overlay", mergedDir, "overlay", 0, flags)
 			if err == nil {
-				logrus.Errorf("overlay test mount with multiple lowers failed, but succeeded with a single lower")
+				logrus.StandardLogger().Logf(logLevel, "overlay test mount with multiple lowers failed, but succeeded with a single lower")
```
vrothberg, line 623: I am a bit confused. Why not use `.Debugf`?

giuseppe: logLevel is set to `logrus.ErrorLevel` for root and `logrus.DebugLevel` for rootless, so it is still an error when running as root

giuseppe (approved):
LGTM

PR #886, opened 2021-04-29
Update to F34 and U2104

cevich:
Also modernize main library to resemble what's used in other
containers-org repositories.  Lastly, update hack/get_ci_vm.sh to use
the new shared container image.

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #889, opened 2021-05-06
Switch from ffjson to json-iterator

nalind:
Per https://github.com/containers/storage/pull/886#issuecomment-833607777, replace our uses of ffjson with json-iterator.


cevich:
Holy crap! Thanks @nalind !

Let me try rebasing my PR onto this to give it some runtime with the newer VM images...

cevich:
[With F34, I see one thing that appears like the tests simply need some adjusting.](https://cirrus-ci.com/task/6223544364302336?logs=build_and_test#L115-L107).  I think that can be addressed in #886 so this LGTM (though I really don't fully understand the complete details).

nalind:
Hmm, looks like Go 1.16's `path/filepath.Match()` is rejecting as syntactically invalid a test case that 1.15 accepted.  Will take a closer look.

rhatdan:
@nalind Will this cause an upgrade problem?

rhatdan:
LGTM Depending on the question.
@vrothberg @saschagrunert @giuseppe PTAL

giuseppe (approved):
LGTM

nalind:
> @nalind Will this cause an upgrade problem?

I would be surprised.  Which logic we call into to encode or decode structures should be an implementation detail that our consumers already don't have to care about outside of vendoring the deps, and the logic that reads and writes encoded data from and to files isn't changing.

TomSweeneyRedHat:
Just to be sure, could we spin up a new c/storage and get it merged upstream into Buildah and Podman early next week?

rhatdan:
Yup, just waiting for the dependabots to fire up.

PR #890, opened 2021-05-07
Added support for CONTAINERS_STORAGE_CONF override

molar:
Following the discussion from [here](https://github.com/containers/skopeo/issues/1272) 
The `DefaultStoreOptions` now checks if the environment variable `CONTAINERS_STORAGE_CONF` is set and uses that path 
for the configuration file. This is in line with how podman does it.

Please let me know your thoughts

### DIFF
--- a/types/default_override_test.conf
+++ b/types/default_override_test.conf
@@ -1,0 +1,11 @@
+[storage]
+
+# Default Storage Driver
+driver = ""
+
+# Primary Read/Write location of container storage
+graphroot = "environment_override_graphroot"
+
+# Storage path for rootless users
+#
+rootless_storage_path = "environment_override_rootless_storage_path"
--- a/types/utils_test.go
+++ b/types/utils_test.go
@@ -272,3 +272,25 @@ func TestDefaultStoreOpts(t *testing.T) {
 	assert.Equal(t, storageOpts.GraphRoot, expectedPath)
 	assert.Equal(t, storageOpts.RootlessStoragePath, expectedPath)
 }
+
+func TestStorageConfOverrideEnvironmentDefaultConfigFileRootless(t *testing.T) {
+	os.Setenv("CONTAINERS_STORAGE_CONF", "default_override_test.conf")
+	defer os.Unsetenv("CONTAINERS_STORAGE_CONF")
+	defaultFile, err := DefaultConfigFile(true)
+
+	expectedPath := "default_override_test.conf"
+
+	assert.NilError(t, err)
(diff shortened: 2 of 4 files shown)


rhatdan:
Code looks good, but the tests don't like it.

rhatdan:
LGTM
@nalind @TomSweeneyRedHat @giuseppe @vrothberg @ashley-cui @umohnani8 PTAL

types/default_override_test.conf
```diff
@@ -0,0 +1,11 @@
...
+
+# Default Storage Driver
+driver = ""
+
+# Primary Read/Write location of container storage
+graphroot = "environment_override_graphroot"
+
+# Storage path for rootless users
+#
+rootless_storage_path = "environment_override_rootless_storage_path"
```
vrothberg, line 11: Is this file meant to be used in a test?

vrothberg:
Can you squash the 5 commits into 1?

TomSweeneyRedHat:
LGTM
assuming happy tests and @vrothberg 's comments are addressed.

types/default_override_test.conf
molar: Yes it was, but calling the `DefaultStoreOptions` function from the tests seems to break all tests.
Can you help with a better way of testing this? Else i can remove it

vrothberg: I would imagine a test that would use [os.Setenv()](https://golang.org/pkg/os/#Setenv) and query for the expected values of the loaded storage conf.

types/utils_test.go
```diff
@@ -272,3 +272,25 @@ func TestDefaultStoreOpts(t *testing.T) {
...
+	expectedPath := "default_override_test.conf"
+
+	assert.NilError(t, err)
+	assert.Equal(t, default_file, expectedPath)
+}
+
+func TestStorageConfOverrideEnvironmentDefaultConfigFileRoot(t *testing.T) {
+	os.Setenv("CONTAINERS_STORAGE_CONF", "default_override_test.conf")
+	defer os.Unsetenv("CONTAINERS_STORAGE_CONF")
+	default_file, err := DefaultConfigFile(false)
```
rhatdan, line 290: defaultFile

rhatdan:
Could you also squash your commits?

rhatdan:
@molar Are you still working on this, or would you prefer I take it over?

giuseppe (approved):
restarted the tests, now they are green

LGTM

PR #892, opened 2021-05-07
TestMatch: handle cases where NewPatternMatcher catches syntax errors

nalind:
With Go 1.16 and later, `pkg/fileutils.NewPatternMatcher()` will flag patterns with bad syntax immediately.  With older compilers, it would have returned a matcher object which flagged syntax errors only when the matcher was actually used to check a path.  Update tests to accept either behavior.


rhatdan (approved):
LGTM

TomSweeneyRedHat:
LGTM
and happy green test buttons.

PR #894, opened 2021-05-11
Expand the scope of transaction in the process of deleting device

gaohuatao-1:
When "docker load $image" and "docker rmi $image" commands are repeatedly executed in the background, the dockerd daemon process is killed. As a result, the DM device where the image resides may be unavailable. The image can be queried, but the container fails to be run. After function “devices.issueDiscard(info)” is executed and before function "devices.deleteTransaction(info, syncDelete)" is executed, at this point, dockerd daemon's withdrawal would result in dm device discarded. Howerver, the dm device is not deleted at the same time.

### DIFF
--- a/drivers/devmapper/deviceset.go
+++ b/drivers/devmapper/deviceset.go
@@ -2010,14 +2010,7 @@ func (devices *DeviceSet) markForDeferredDeletion(info *devInfo) error {
 }
 
 // Should be called with devices.Lock() held.
-func (devices *DeviceSet) deleteTransaction(info *devInfo, syncDelete bool) error {
-	if err := devices.openTransaction(info.Hash, info.DeviceID); err != nil {
-		logrus.Debugf("devmapper: Error opening transaction hash = %s deviceId = %d", "", info.DeviceID)
-		return err
-	}
-
-	defer devices.closeTransaction()
-
+func (devices *DeviceSet) deleteDeviceNoLock(info *devInfo, syncDelete bool) error {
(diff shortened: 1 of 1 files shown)


gaohuatao-1:
When  "docker load $image" and "docker rmi $image" commands are repeatedly executed in the background, the dockerd daemon process is killed. As a result, the DM device where the image resides may be unavailable. The image can be queried, but the container fails to be run. After function “devices.issueDiscard(info)” is executed and before function "devices.deleteTransaction(info, syncDelete)" is executed,   at this point, dockerd daemon's withdrawal would result in dm device discarded. Howerver, the dm device is not deleted at the same time.

gaohuatao-1:
> Could you add
> 
> ```
> When "docker load $image" and "docker rmi $image" commands are repeatedly executed in the background, the dockerd daemon process is killed. As a result, the DM device where the image resides may be unavailable. The image can be queried, but the container fails to be run. After function “devices.issueDiscard(info)” is executed and before function "devices.deleteTransaction(info, syncDelete)" is executed, at this point, dockerd daemon's withdrawal would result in dm device discarded. Howerver, the dm device is not deleted at the same time.
> ```
> 
> To you PR description?

ok

rhatdan:
Looks like devicemapper tests are failing?

rhatdan (approved):
LGTM

rhatdan:
@saschagrunert @vrothberg @giuseppe @TomSweeneyRedHat @nalind PTAL

giuseppe (approved):
LGTM

Issue #880, opened 2021-04-22
panic: runtime error: index out of range


DavidHaltinner:
@vrothberg asked that I bring this issue over here from cri-o. Original posted issue: https://github.com/cri-o/cri-o/issues/4777

**Description**
I have had about a half dozen crashes of cri-o in the last few months, on a few different versions.
Crio crashes with a "panic: runtime error: index out of range" and a stack trace which I will paste below.
I am utilizing OKD, 4.7 currently, but issue occured under 4.6 as well, maybe 4.5 but my memory is bad.
Current crio is 1.20.0.
There does not seem to anything that I can see that triggers it, it appears random. In today's issue, directly before, podman removes an image, and hyperkube "Failed to create an existing container".
I did also note that just before this latest panic, podman had removed an image: Apr 19 08:06:34 my.server.com podman[3091098]: 2021-04-19 08:06:34.307610175 -0500 CDT m=+249.235712493 image remove a6d081b7d335d71b74c7084079956b1fadb9e2e3edbe12d686756bc368f952fc image-registry.openshift-image-registry.svc:5000/graylog/graylog-custom@sha256:2b34cb10953329487e00b9cdb20a43f99b2db5decc2268326509714899e39bf8

**Steps to reproduce the issue:**
1. Run OKD using crio
2. Wait? No determinable trigger to me
3.

**Describe the results you received:**
No panics should occur.

**Describe the results you expected:**
Crio to continue to run normally.

**Additional information you deem important (e.g. issue happens only occasionally):**
The crio service logs from the journal are being attached from the crash.
[crio.txt](https://github.com/cri-o/cri-o/files/6336851/crio.txt)

**Output of `crio --version`:**

```
INFO[0000] Starting CRI-O, version: 1.20.0, git: d388528dbed26b93c5bc1c89623607a1e597aa57(clean) 
crio version 1.20.0
Version:       1.20.0
GitCommit:     d388528dbed26b93c5bc1c89623607a1e597aa57
GitTreeState:  clean
BuildDate:     2020-12-11T17:42:02Z
GoVersion:     go1.15.5
Compiler:      gc
Platform:      linux/amd64
Linkmode:      dynamic

```

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Virtual machines running on Nutanix AOS.

DavidHaltinner:
Another crash over the weekend.
This was also proceeded by podman removing an image.
Logs
[...]

PR #898, opened 2021-05-12
store: replace Modified+Load with ReloadIfChanged

giuseppe:
simplify the pattern if store.Modified() => store.Load() with a single function call store.ReloadIfChanged().
    
In addition to the code simplification, it solves a race condition that when only a RLock() is hold and that multiple goroutines could
run the sequence store.Modified() and store.Load().  In this case only one goroutine would see store.Modified( and run store.Load().
Other goroutines could access the old data while the goroutine that saw the change modifies the store data.

Closes: https://github.com/containers/storage/issues/880

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/containers/storage/pkg/idtools"
@@ -113,6 +114,7 @@ type containerStore struct {
 	byid       map[string]*Container
 	bylayer    map[string]*Container
 	byname     map[string]*Container
+	loadMut    sync.Mutex
 }
 
 func copyContainer(c *Container) *Container {
@@ -614,4 +616,15 @@ func (r *containerStore) TouchedSince(when time.Time) bool {
 
 func (r *containerStore) Locked() bool {
 	return r.lockfile.Locked()
+}
+
+func (r *containerStore) ReloadIfChanged() error {
+	r.loadMut.Lock()
+	defer r.loadMut.Unlock()
+
+	modified, err := r.Modified()
+	if err == nil && modified {
+		return r.Load()
+	}
+	return nil
 }
--- a/images.go
+++ b/images.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/containers/storage/pkg/ioutils"
@@ -152,6 +153,7 @@ type imageStore struct {
 	byid     map[string]*Image
 	byname   map[string]*Image
 	bydigest map[digest.Digest][]*Image
+	loadMut  sync.Mutex
 }
 
 func copyImage(i *Image) *Image {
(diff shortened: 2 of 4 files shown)


rhatdan (approved):
Nice Fix
LGTM

nalind:
The diagnosis at https://github.com/containers/storage/issues/880#issuecomment-839760238 looks good, but I'm missing how this keeps one goroutine that's using the data without having taken the `loadMut` mutex from having the slice it's iterating over replaced, while it's running and unaware, by another goroutine that has the mutex.

giuseppe:
> The diagnosis at [#880 (comment)](https://github.com/containers/storage/issues/880#issuecomment-839760238) looks good, but I'm missing how this keeps one goroutine that's using the data without having taken the `loadMut` mutex from having the slice it's iterating over replaced, while it's running and unaware, by another goroutine that has the mutex.

that is protected by the other locking mechanisms we have.  A goroutine that is accessing the data must already hold a RO or a RW lock.  A goroutine checks for `ReloadIfChanged()` while holding the RO lock.  No changes could have happened to the storage between a goroutine got the RO lock and it accesses the data because another goroutine/process needed to get the RW lock before it could make any change.  Since a goroutine is already holding a RO lock (thus blocking other writers), it means that any other goroutine getting into the `ReloadIfChanged()` code block won't hit the `r.Modified()` condition.

The issue I've seen only happens when there are multiple goroutines checking `r.Modified()` together and not syncing on the `Load()` that only one goroutine does.

| Goroutine 1 | Process 2 |  Goroutine 3 |
| -------------- | --------- | ---------- |
| GetROLock() | | |
| ReloadIfChanged() [might trigger a `Load()`] | GetRWLock() |  |
| [use the store data]  | [wait]              | GetROLock() |
| [use the store data]  | [wait]              | **ReloadIfChanged()** |

**ReloadIfChanged()** should never cause a `Load()` because goroutine 1 already checked while holding the RO lock thus blocking `Process 2`.

One possible race condition though is that we attempt the `Load` also on errors:

```
	modified, err := r.Modified()
	if err != nil || modified {
		return r.Load()
	}
```
A temporary error could cause the race as it calls `Load()` while there are other goroutines running.  I've pushed a new version where I change the condition to `err == nil && modified`

giuseppe:
does the explanation above make sense?  Are there other cases where a race can happen?

nalind:
I was thinking of the case where one thread is calling a method like store.Images(), which takes a read lock, but another is in a method like store.ImageSize(), which also takes a read lock, and one of them alters the data the object is holding while the other is iterating through it.  It's not exclusive to this pair of methods, but that's a window I don't see us closing here.

giuseppe:
> I was thinking of the case where one thread is calling a method like store.Images(), which takes a read lock, but another is in a method like store.ImageSize(), which also takes a read lock, and one of them alters the data the object is holding while the other is iterating through it. It's not exclusive to this pair of methods, but that's a window I don't see us closing here.

That is the case I've tried to explain with the table.

If one thread is calling `store.Images()` while it holds a read lock, it already went through the `ReloadIfChanged()` check.  Holding the RO lock prevents other processes to modify the store.  Another thread that gets the RO lock and gets into `ReloadIfChanged()`, will either 1) wait that for the first thread to complete `ReloadIfChanged()`, then `Modified()` returns false 2) don't block on the lock but `Modified()` still returns false so it doesn't reload the store while the first thread is already accessing its data.
This assumption is based on the fact that any reload is made while holding the RO so other threads/processes cannot modify the store in the meanwhile.  The store could be modified only when there are no RO locks in place.

nalind:
Hmm, okay, that makes sense.  LGTM.

vrothberg:
@haircommander, that should fix the issue in CRI-O

PR #899, opened 2021-05-12
Remove unlock/lock caused by Incorrect assumption

rhatdan:
I thought the previos lock was a read/only lock, which is why I was
freeing and taking the read/write lock.  Since previous lock was
read/write, no reason for this code.

### DIFF
--- a/store.go
+++ b/store.go
@@ -2129,10 +2129,6 @@ func (s *store) SetNames(id string, names []string) error {
 			}
 		}
 		if i, err := store.Get(id); err == nil {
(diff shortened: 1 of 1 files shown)


nalind:
LGTM, as far as it goes.

rhatdan:

rhatdan:
@vrothberg @saschagrunert @giuseppe @TomSweeneyRedHat PTAL

giuseppe (approved):
LGTM

PR #907, opened 2021-05-17
archive/overlay: ignore failures from nested whiteouts

tych0:
See comment for the situation that arises, we've seen this with images
generated by very old dockers (and possibly others?).

I added a test for both syscall.ENOTDIR and unix.ENOTDIR even though they
*should be* aliases to be extra defensive (syscall.Stat_t and unix.Stat_t
were aliases for a while, but now they are not), but I can drop it if
necessary.

Signed-off-by: Tycho Andersen <tycho@tycho.pizza>

### DIFF
--- a/pkg/archive/changes.go
+++ b/pkg/archive/changes.go
@@ -121,6 +121,9 @@ func isENOTDIR(err error) bool {
 	if err == nil {
 		return false
 	}
+	if err == syscall.ENOTDIR {
+		return true
+	}
 	if perror, ok := err.(*os.PathError); ok {
 		if errno, ok := perror.Err.(syscall.Errno); ok {
 			return errno == syscall.ENOTDIR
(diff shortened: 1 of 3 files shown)


pkg/archive/changes.go
```diff
@@ -121,6 +122,9 @@ func isENOTDIR(err error) bool {
 	if err == nil {
 		return false
 	}
+	if err == syscall.ENOTDIR || err == unix.ENOTDIR {
```
giuseppe, line 125: are there cases where `syscall.ENOTDIR != unix.ENOTDIR`?

tych0: See the commit message, I've had problems with syscall.* and unix.* diverging in the past. But somehow this doesn't build in your CI, so I'll just go ahead and drop it :)

giuseppe (approved):
restarted the failing tests, now everything seems green.

LGTM

rhatdan:
LGTM

PR #913, opened 2021-05-20
overlay: check for unix.ENOTSUP

giuseppe:
if lgetxattr(..., "user.overlay.metacopy") fails with EOPNOTSUPP, it
means metacopy is not supported for the current user.

Prefer this runtime check instead of testing unshare.IsRootless(), so
that metacopy support will be detected in case it will be enabled in a
user namespace in future.


rhatdan (approved):
LGTM

rhatdan:
@vrothberg @saschagrunert @nalind @ashley-cui @umohnani8 @TomSweeneyRedHat PTAL
@giuseppe Should we hold up rc2 for this?

PR #916, opened 2021-05-21
Add Dev, Nlink, and Ino to pkg/system/Stat struct

rhatdan:
podman needs to look at the Device, Inode and number of
links of a file, but needs this to work on Windows and
Darwin.  We are using this information to determine if
a file is a hard link whe tarring it up.

Add new function FileInfoToStatT to transate a FileInfor into
the StatT struct.

### DIFF
--- a/pkg/system/lstat_windows.go
+++ b/pkg/system/lstat_windows.go
@@ -10,5 +10,5 @@ func Lstat(path string) (*StatT, error) {
 		return nil, err
 	}
 
-	return fromStatT(&fi)
+	return FileInfoToStatT(fi)
 }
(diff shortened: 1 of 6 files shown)


rhatdan:
@nalind @vrothberg @giuseppe @saschagrunert @TomSweeneyRedHat PTAL

rhatdan:
Lets wait for the podman PR to work before merging this.  Since I have had a few changes to this found via podman.

vrothberg (commented):
LGTM

rhatdan:
Turns out @nalind does not like this, since the Windows version is returning bogus data,  IE Nlink(), Dev(), Ino() always return 0, and this is wrong.  So will implement everything in Podman.

Issue #917, opened 2021-05-21
`podman system df` hangs


Hi-Angel:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Command `podman system df` never finishes.

**Steps to reproduce the issue:**

1. Execute `podman system df`

**Describe the results you received:**

It never finishes. I let it run in background for 1 hours and 5 minutes, and then interrupted it.

**Describe the results you expected:**

It would return immediately with some information.

**Additional information you deem important (e.g. issue happens only occasionally):**

Running `strace podman system df` shows it hangs on some futex:

```
[…]
openat(AT_FDCWD, "/etc/containers/containers.conf", O_RDONLY|O_CLOEXEC) = 7
epoll_ctl(4, EPOLL_CTL_ADD, 7, {events=EPOLLIN|EPOLLOUT|EPOLLRDHUP|EPOLLET, data={u32=1486363312, u64=140103319558832}}) = -1 EPERM (Operation not permitted)
epoll_ctl(4, EPOLL_CTL_DEL, 7, 0xc00055921c) = -1 EPERM (Operation not permitted)
fstat(7, {st_mode=S_IFREG|0644, st_size=16329, ...}) = 0
read(7, "# The containers configuration f"..., 16330) = 16329
read(7, "", 1)                          = 0
close(7)                                = 0
openat(AT_FDCWD, "/home/constantine/.config/containers/containers.conf", O_RDONLY|O_CLOEXEC) = 7
epoll_ctl(4, EPOLL_CTL_ADD, 7, {events=EPOLLIN|EPOLLOUT|EPOLLRDHUP|EPOLLET, data={u32=1486363312, u64=140103319558832}}) = -1 EPERM (Operation not permitted)
epoll_ctl(4, EPOLL_CTL_DEL, 7, 0xc00055921c) = -1 EPERM (Operation not permitted)
fstat(7, {st_mode=S_IFREG|0644, st_size=38, ...}) = 0
read(7, "[engine]\ndetach_keys=\"ctrl-q,ctr"..., 512) = 38
read(7, "", 474)                        = 0
close(7)                                = 0
newfstatat(AT_FDCWD, "/home/constantine/.config/cni/net.d", 0xc00039c378, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/run/user/1000/bus", {st_mode=S_IFSOCK|0666, st_size=0, ...}, 0) = 0
[...]

PR #918, opened 2021-05-21
overlay: enable native diff for fuse-overlayfs

giuseppe:
NaiveDiff was enabled since the initial implementation to be
conservative.  There is not really any good reason for using it
instead of native diff.

Switch to native diff for better performance.

Closes: https://github.com/containers/storage/issues/917


rhatdan:
Looks like the testing does not like your change?

giuseppe:
> Looks like the testing does not like your change?

tests are green

rhatdan (approved):
LGTM

rhatdan:
@nalind @saschagrunert @vrothberg PTAL

TomSweeneyRedHat:
LGTM

PR #919, opened 2021-05-22
Avoid failure when umount an unmounted mountpoint

gaohuatao-1:
Check if the mountpoint is mounted when unmount it to avoid failure.
If user manually run the umount command before it, the function
UnmountDevice returns an error, Although this error dose not cause the
container deletion process fail for the reason that the return value of
UnmountDevice function is not processed. However, the ERROR logs in the
log system are misleading.


rhatdan (approved):
LGTM

rhatdan:
@giuseppe @vrothberg @saschagrunert @nalind @umohnani8 @ashley-cui PTAL

drivers/devmapper/deviceset.go
```diff
@@ -2445,8 +2445,10 @@ func (devices *DeviceSet) UnmountDevice(hash, mountPath string) error {
 	defer devices.Unlock()
 
 	logrus.Debugf("devmapper: Unmount(%s)", mountPath)
-	if err := mount.Unmount(mountPath); err != nil {
-		return err
+	if ok, _ := Mounted(mountPath); ok {
+		if err := mount.Unmount(mountPath); err != nil {
+			return err
+		}
```
giuseppe, line 2451: this introduces a race.

Let's do the opposite, first we try to unmount and in case of errors check if it is mounted (if it is mounted then return the Unmount error)

giuseppe (changes requested):

drivers/devmapper/deviceset.go
gaohuatao-1: Obviously, your proposal is better, and I've changed it as you suggested.

giuseppe (approved):
LGTM

PR #920, opened 2021-05-25
sync write link and lower files to avoid file corruption caused by poweroff

gaohuatao-1:
When the os is powered off during file writing, the scenario is possible happen that the file exists but it's content is empty.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -25,6 +25,7 @@ import (
 	"github.com/containers/storage/pkg/directory"
 	"github.com/containers/storage/pkg/fsutils"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/locker"
 	"github.com/containers/storage/pkg/mount"
 	"github.com/containers/storage/pkg/parsers"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@giuseppe PTAL

rhatdan:
@nalind PTAL

giuseppe:
This is not robust as image files could have been left unsynced.  I suggest to look at what CRI-O does

gaohuatao-1:
> This is not robust as image files could have been left unsynced. I suggest to look at what CRI-O does.
> 
> I'd like we avoid additional `fsync` calls as they slow down I/O and they usually just hide the problem of a corrupted storage.

gaohuatao-1:
@giuseppe sorry, i closed it accidentally

giuseppe:
> @giuseppe Yes, your opinion is reasonable, but it may cause functional problem

the problem I see is that while this gives the impression the storage is not corrupted (and it is an expensive I/O operation), the stored images are likely corrupted.  So you'll end up with random errors at runtime when there are running containers.

gaohuatao-1:
> > @giuseppe Yes, your opinion is reasonable, but it may cause functional problem
> 
> the problem I see is that while this gives the impression the storage is not corrupted (and it is an expensive I/O operation), the stored images are likely corrupted. So you'll end up with random errors at runtime when there are running containers.

This operation does cause I/O to slow down. I recommend to close this PR

PR #926, opened 2021-05-27
store: fix graphLock reload

giuseppe:
use s.graphLock.Modified() instead of s.graphLock.TouchedSince().
    
TouchedSince() seems to fail in some cases when comparing the inode mtime with the current time.
    
Avoid such kind of problems in a critical code path as store.mount(), as we must be sure there is already a driver home mount present, otherwise it the next process may cover the container mount with the home mount.

Closes: https://github.com/containers/podman/issues/10454

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -626,5 +626,5 @@ func (r *containerStore) ReloadIfChanged() error {
 	if err == nil && modified {
 		return r.Load()
 	}
-	return nil
+	return err
 }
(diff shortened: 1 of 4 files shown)


vrothberg (commented):
LGTM

rhatdan (approved):
LGTM

edsantiago:
Followup: I applied this  as a patch to  my podman tree this morning, built, started a reproducer. Just got back home, 298 iterations, no failures. LGTM. Thank you @giuseppe .

PR #932, opened 2021-06-02
canUseShifting can segfault

rhatdan:
Fixes: https://github.com/containers/podman/issues/10535

### DIFF
--- a/store.go
+++ b/store.go
@@ -937,7 +937,7 @@ func (s *store) ContainerStore() (ContainerStore, error) {
 }
 
 func (s *store) canUseShifting(uidmap, gidmap []idtools.IDMap) bool {
-	if !s.graphDriver.SupportsShifting() {
+	if s.graphDriver == nil || !s.graphDriver.SupportsShifting() {
 		return false
 	}
 	if uidmap != nil && !idtools.IsContiguous(uidmap) {
@@ -2668,6 +2668,10 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
 		s.lastLoaded = time.Now()
 	}
 
+	if options.UidMaps != nil || options.GidMaps != nil {
+		options.DisableShifting = !s.canUseShifting(options.UidMaps, options.GidMaps)
+	}
+
 	if rlstore.Exists(id) {
 		return rlstore.Mount(id, options)
 	}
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe @edsantiago PTAL

edsantiago:
LGTM, in principle, but do you have any idea what could've led to `s.graphDriver` being uninitialized in this particular instance?

rhatdan:
No idea, @giuseppe needs to look at this, This change just makes it safer.

giuseppe:
what do you think if we just move the check later in `mount()` where we are sure `s.graph` is not nil?

```
diff --git a/store.go b/store.go
index 759407c63..8331fafde 100644
--- a/store.go
+++ b/store.go
@@ -2666,6 +2666,10 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
                s.lastLoaded = time.Now()
        }
 
+       if options.UidMaps != nil || options.GidMaps != nil {
+               options.DisableShifting = !s.canUseShifting(options.UidMaps, options.GidMaps)
+       }
+
        if rlstore.Exists(id) {
        }
@@ -2706,7 +2710,6 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
                                options.Volatile = v.(bool)
                        }
                }
-               options.DisableShifting = !s.canUseShifting(container.UIDMap, container.GIDMap)
        }
        return s.mount(id, options)
 }
```

Otherwise we may end up using shifting even if the driver doesn't support it

rhatdan:
@giuseppe Looking at this, I am not sure this should be handled at the store level at all.  I think we should just remove this and do it at the driver level. Is it your intention to allow users in their storage.conf to specify to disable shifting?

giuseppe:
> @giuseppe Looking at this, I am not sure this should be handled at the store level at all. I think we should just remove this and do it at the driver level. Is it your intention to allow users in their storage.conf to specify to disable shifting?

I think the store needs to know when shifting is supported by the underlying driver.  When it is not supported it needs to create a copy of the image with the shifted IDs.

rhatdan:
@giuseppe PTAL

giuseppe (approved):
LGTM

rhatdan:
@saschagrunert @vrothberg PTAL

vrothberg (commented):
LGTM

PR #934, opened 2021-06-03
overlay: make userxattr,metacopy=on debug message

giuseppe:
currently the kernel doesn't support both userxattr and metacopy=on
options.  When they are specified the mount fails with EINVAL.

Downgrade the warning message to debug level when running as
rootless.


rhatdan:
LGTM
@giuseppe @saschagrunert @TomSweeneyRedHat @nalind PTAL

saschagrunert (approved):
LGTM

PR #941, opened 2021-06-12
Fix cancel deferred remove bug

gaohuatao-1:
When cancel the deferred removal, if the device is already gone,
continue. According to the original logic, if the device does not exist,
an error is reported.


rhatdan:
LGTM

rhatdan (approved):

PR #943, opened 2021-06-16
Reload layer storage if layers.json got externally modified

saschagrunert:
This helps long running processes like CRI-O to determine changes to the
local storage, while handling corrupted images automatically.

The corresponding fix in [Podman][0] handles corrupt layers by reloading
the image. This does not work for CRI-O, because it will not reload the
storage on subsequent calls of pullImage if no storage modification has
been done.

[0]: https://github.com/containers/podman/pull/10637/commits/b4bd886fcc01c99b1c8aa5800710b1c31bae8ae5

Refers to https://github.com/cri-o/cri-o/pull/5005

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -271,21 +271,22 @@ type LayerStore interface {
 }
 
 type layerStore struct {
-	lockfile          Locker
-	mountsLockfile    Locker
-	rundir            string
-	driver            drivers.Driver
-	layerdir          string
-	layers            []*Layer
-	idindex           *truncindex.TruncIndex
-	byid              map[string]*Layer
-	byname            map[string]*Layer
-	bymount           map[string]*Layer
-	bycompressedsum   map[digest.Digest][]string
-	byuncompressedsum map[digest.Digest][]string
-	uidMap            []idtools.IDMap
-	gidMap            []idtools.IDMap
-	loadMut           sync.Mutex
+	lockfile           Locker
+	mountsLockfile     Locker
+	rundir             string
+	driver             drivers.Driver
+	layerdir           string
+	layers             []*Layer
+	idindex            *truncindex.TruncIndex
+	byid               map[string]*Layer
+	byname             map[string]*Layer
+	bymount            map[string]*Layer
+	bycompressedsum    map[digest.Digest][]string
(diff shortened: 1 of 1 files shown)


saschagrunert:
The added integration test in https://github.com/cri-o/cri-o/pull/5005 is now passing:
https://github.com/cri-o/cri-o/pull/5005/files#diff-0b5b5d1619dddec727d1e347a5e9777d71b98ec5f35034690eb8b5693cf5ab4bR305-R327

I don't think that I can reproduce it with the containers-storage binary since the store shuts-down after the invocation.

layers.go
```diff
@@ -1757,7 +1758,19 @@ func (r *layerStore) Modified() (bool, error) {
...
+	// reload the storage in any case.
+	info, err := os.Stat(r.layerspath())
+	if err != nil && !os.IsNotExist(err) {
+		return false, errors.Wrap(err, "stat layers file")
+	}
+	if info != nil {
+		tmodified = info.ModTime() != r.layerspathModified
+		r.layerspathModified = info.ModTime()
+	}
+
```
rhatdan, line 1772: What happens in the case where the layerspath did not exist?

saschagrunert: In that case `info == nil` and we stick to the previous behavior (`tmodified` will be false). This is fine because we will load/create the layerspath in any case afterwards.

rhatdan:
LGTM

layers.go
giuseppe: could we add an early return for `if lmodified || mmodified {return true}` so that we avoid a Stat if that condition is already true?

layers.go
```diff
@@ -1757,7 +1758,19 @@ func (r *layerStore) Modified() (bool, error) {
...
+	return lmodified || mmodified || tmodified, nil
```
giuseppe, line 1773: then here we could just return tmodified

layers.go
saschagrunert: Yep, fixed as suggested :+1:

layers.go
saschagrunert: Fixed with previous push.

giuseppe (approved):
LGTM

PR #945, opened 2021-06-22
Vendor in opencontainers/runc v1.0.0

rhatdan:
The RunningInUserNS Function was removed from libcontainer and is now
in contianerd.


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/Microsoft/go-winio v0.5.0
 	github.com/Microsoft/hcsshim v0.8.17
+	github.com/containerd/containerd v1.5.1
```
vrothberg, line 9: I'd rather not add a dependency on `containerd`. Great code but also a big code base. Was `RunningInUserNS` removed from runc?

rhatdan:
@giuseppe @saschagrunert @vrothberg @nalind @TomSweeneyRedHat PTAL

go.mod
saschagrunert: It think it's still available in 
```
	"github.com/opencontainers/runc/libcontainer/userns"
```

rhatdan: Ok, maybe I missed it.

rhatdan: I grepped through an older version of runc

vrothberg (approved):
LGTM, thanks @rhatdan

giuseppe (approved):
LGTM

PR #948, opened 2021-06-22
overlay: fix check for rootless native diff

giuseppe:
make sure "userxattr" is specified, otherwise the "user.overlay."
xattrs are not treated specially and copied up as for the root case.

Closes: https://github.com/containers/storage/issues/947


rhatdan:
LGTM

vrothberg (approved):
LGTM

StarpTech:
Hi, I run into a similar issue. Is it possible that the detection is still buggy? https://github.com/containers/podman/discussions/15515

PR #952, opened 2021-06-25
Use /run/user/UID in rootless mode if writable

rhatdan:
Other parts of the code are using this directory, so we end up
with us creating an empty directory.

I don't see a reason why we would just use this directory only if
the init program is systemd?

Fixes: https://github.com/containers/podman/issues/10782

### DIFF
--- a/types/utils.go
+++ b/types/utils.go
@@ -2,7 +2,6 @@ package types
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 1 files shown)


tobwen:
LGTM - tested and works on my system.

rhatdan:

giuseppe (approved):
LGTM

types/utils.go
```diff
@@ -75,12 +74,9 @@ func getRootlessRuntimeDirIsolated(env rootlessRuntimeDirEnvironment) (string, e
 		return runtimeDir, nil
 	}
 
-	initCommand, err := ioutil.ReadFile(env.getProcCommandFile())
```
mtrmac, line 78: - This should have had an extra test case for the new behavior
- After that , `getProcCommandFile` is now unused apart from tests trying to affect it; so drop it, and then remove test cases from the duplicates.

mtrmac (commented):
It’s fairly easy to find that all of this function was originally created _primarily_ for this check, in https://github.com/containers/storage/pull/568 . Why is that no longer a relevant concern?

https://github.com/containers/podman/issues/10782 explicitly says that Systemd is being used, so this change _should not_ have made a difference for that bug report. Why is this fixing that bug report, then? Is it even known what has actually gone wrong in that report?

types/utils.go
mtrmac: (This is assuming that the change is the right thing to do; whether that’s the case is needs to be decided first, and that seems to be an unresolved question so far.)

PR #966, opened 2021-07-16
User are reporting potential issues with using overlay2

rhatdan:
We should just switch to overlay right away rather then
handling the driver code lower in the stack.


types/options.go
```diff
@@ -190,6 +190,10 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
 	if driver := os.Getenv("STORAGE_DRIVER"); driver != "" {
 		opts.GraphDriverName = driver
 	}
+	// override users selection of 'overlay2' with 'overlay'
+	if opts.GraphDriverName == "overlay2" {
```
TomSweeneyRedHat, line 194: lint is barfing, it wants a constant for "overlay2"

TomSweeneyRedHat:
@rhatdan looks like you need a rebase

types/options.go
```diff
@@ -190,6 +194,10 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
 	if driver := os.Getenv("STORAGE_DRIVER"); driver != "" {
 		opts.GraphDriverName = driver
 	}
+	// override users selection of 'overlay2' with 'overlay'
```
TomSweeneyRedHat, line 197: Might want to leave behind a comment as to why the override is happening.

rhatdan:
This is a Potentially breaking change, and should be held off until 4.0 of Podman.

vrothberg (commented):
@rhatdan, I think it's time to get this one in.

PR #967, opened 2021-07-16
overlay: check if we can mknod() kernel whiteout

nalind:
When checking if we can use the kernel's overlay filesystem, verify that we can create whiteouts in the format the kernel expects (character devices with major and minor of 0), and if we fail to do that, don't try to use the kernel's overlay filesystem.  I think this is a "we can mount kernel overlay, but not everything else works" case that pops up among the gating test errors in https://bugzilla.redhat.com/show_bug.cgi?id=1955166.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -613,6 +613,10 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 		if unshare.IsRootless() {
 			flags = fmt.Sprintf("%s,userxattr", flags)
 		}
+		if err := syscall.Mknod(filepath.Join(upperDir, "whiteout"), syscall.S_IFCHR|0600, int(unix.Mkdev(0, 0))); err != nil {
+			logrus.Debugf("unable to create kernel-style whiteout: %v", err)
+			return supportsDType, errors.Wrapf(err, "unable to create kernel-style whiteout")
+		}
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
Although it is good that we saw this in RHEL8.5 or we would not have known the kernel was not fully updated.

rhatdan (approved):

rhatdan:
@giuseppe @vrothberg @saschagrunert PTAL

TomSweeneyRedHat:
LGTM, but definitely want another head nod here from one of the folks @rhatdan pointed at.

drivers/overlay/overlay.go
```diff
@@ -613,6 +613,10 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 		if unshare.IsRootless() {
 			flags = fmt.Sprintf("%s,userxattr", flags)
 		}
+		if err := syscall.Mknod(filepath.Join(upperDir, "whiteout"), syscall.S_IFCHR|0600, int(unix.Mkdev(0, 0))); err != nil {
+			logrus.Debugf("unable to create kernel-style whiteout: %v", err)
+			return supportsDType, errors.Wrapf(err, "unable to create kernel-style whiteout")
```
saschagrunert, line 618: nit: Wondering for what we need the debug log if we return an error in any case.

saschagrunert (approved):

PR #969, opened 2021-07-22
Creating fifo files while non root should be supported

rhatdan:

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -645,10 +645,13 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 		}
 		file.Close()
 
-	case tar.TypeBlock, tar.TypeChar, tar.TypeFifo:
+	case tar.TypeBlock, tar.TypeChar:
 		if inUserns { // cannot create devices in a userns
+			logrus.Debugf("Tar: Can't create device %v while running in user namespace", path)
 			return nil
 		}
+		fallthrough
+	case tar.TypeFifo:
 		// Handle this is an OS-specific way
 		if err := handleTarTypeBlockCharFifo(hdr, path); err != nil {
 			return err
--- a/pkg/archive/archive_unix.go
+++ b/pkg/archive/archive_unix.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/system"
-	"github.com/opencontainers/runc/libcontainer/userns"
 	"golang.org/x/sys/unix"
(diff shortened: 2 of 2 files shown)


pkg/archive/archive_unix.go
```diff
@@ -88,8 +89,9 @@ func minor(device uint64) uint64 {
 // handleTarTypeBlockCharFifo is an OS-specific helper function used by
 // createTarFile to handle the following types of header: Block; Char; Fifo
 func handleTarTypeBlockCharFifo(hdr *tar.Header, path string) error {
-	if userns.RunningInUserNS() {
-		// cannot create a device if running in user namespace
+
+	if userns.RunningInUserNS() && hdr.Typeflag != tar.TypeFifo {
+		logrus.Warnf("Tar: Can't create device %v while running in user namespace", path)
```
nalind, line 94: I'm not certain that we want to start warning about things we don't do when we're in a user namespace.
Looking for other instances where we did or didn't, it looks like we'll also need to tweak line 649 in archive.go, and maybe we end up dropping this block.

rhatdan: Ok I will drop this to debug.

nalind: Probably going to need to touch https://github.com/containers/storage/blob/main/pkg/archive/archive.go#L649 also.

pkg/archive/archive.go
```diff
@@ -645,8 +645,14 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
...
+	case tar.TypeFifo:
+		// Handle this is an OS-specific way
+		if err := handleTarTypeBlockCharFifo(hdr, path); err != nil {
+			return err
+		}
+	case tar.TypeBlock, tar.TypeChar:
 		if inUserns { // cannot create devices in a userns
+			logrus.Debugf("Tar: Can't create device %v while running in user namespace", path)
 			return nil
 		}
```
nalind, line 657:
```suggestion
		}
		fallthrough
	case tar.TypeFifo:
```
would avoid the duplication, but it's not a blocker.

nalind:
One suggestion, but LGTM either way.

pkg/archive/archive_unix.go
mbhangui: +1 for the warning. This issue hit me because I was running my app in user namespace and I couldn't figure out the problem why my fifos were'nt there.

rhatdan:
@saschagrunert @vrothberg @giuseppe PTAL

TomSweeneyRedHat:
@rhatdan whiteout is redding out

PR #972, opened 2021-07-23
Split pkg/chunked.ZstdCompressor into a separate subpackage

mtrmac:
`containers/image/pkg/compression` depends on that function, so it's highly undesirable for it to drag in all of containers/storage.

So, move it into a separate subpackage, along with its exclusive dependency tree.

Code that is called both from the new `pkg/chunked/compressor` and the original `pkg/chunked` has been moved into `pkg/chunked/internal` , so that we don't expose any new public API. That move was made purely mechanically without any concern for conceptual consistency, to minimize the size of the dependency (and not to spend time on that until the concept is proven to be useful).

Should not change behavior.

### DIFF
--- a/pkg/chunked/compression.go
+++ b/pkg/chunked/compression.go
@@ -2,71 +2,28 @@ package chunked
 
 import (
 	"bytes"
-	"encoding/base64"
 	"encoding/binary"
-	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
-	"time"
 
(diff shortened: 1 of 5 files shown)


rhatdan:
/approve
LGTM

rhatdan:
@giuseppe PTAL

rhatdan:
@vrothberg @saschagrunert @nalind PTAL

giuseppe (approved):
LGTM

PR #975, opened 2021-07-27
Fix handling of quota on volumes

rhatdan:
This patch fixes the handling of inodes and sizes, currently if
user sets indoes and sizes together, quota is only set on sizes.

Second problem with quota is that we have to have unigue projectids
for each directory. Originally container/storage only did quota on
rootfs, now we want to support it on volumes as well. We need to be
able to get unigue projectids for these two different parent
directories. The added function, attempts to maintain at least 10,000
unigue id's based on the inode of the parent directory. I know that this
is not perfect and we have a potential for overlay.  If you have a
better algorythm, I would love to use it.

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -263,12 +263,44 @@ The semanage command above tells SELinux to setup the default labeling of `NEWST
 
 Now all new content created in these directories will automatically be created with the correct label.
 
-## SEE ALSO
-`semanage(8)`, `restorecon(8)`, `mount(8)`, `fuse-overlayfs(1)`
+## QUOTAS
+
+Container storage implements `XFS project quota controls` for overlay storage
+containers and volumes. The directory used to store the containers must be an
+`XFS` file system and be mounted with the `pquota` option.
+
+Example /etc/fstab entry:
+```
+/dev/podman/podman-var /var xfs defaults,x-systemd.device-timeout=0,pquota 1 2
+```
+
+Container storage generates project ids for each container and builtin volume, but these project ids need to be unique for the XFS file system.
+
+The xfs_quota tool can be used to assign a project id to the storage driver directory, e.g.:
+
+```
+echo 100000:/var/lib/containers/storage/overlay >> /etc/projects
+echo 200000:/var/lib/containers/storage/volumes >> /etc/projects
+echo storage:100000 >> /etc/projid
+echo volumes:200000 >> /etc/projid
+xfs_quota -x -c 'project -s storage volumes' /<xfs mount point>
+```
+
+In the example above, the storage directory project id will be used as a "start offset"
+and all containers will be assigned larger project ids (e.g. >= 100000).
+Then the volumes directory project id will be used as a "start offset"
+and all volumes will be assigned larger project ids (e.g. >= 200000).
+This is a way to prevent xfs_quota management from conflicting with containers/storage.
 
 ## FILES
 
 Distributions often provide a `/usr/share/containers/storage.conf` file to define default storage configuration. Administrators can override this file by creating `/etc/containers/storage.conf` to specify their own configuration. The storage.conf file for rootless users is stored in the `$XDG_CONFIG_HOME/containers/storage.conf` file.  If `$XDG_CONFIG_HOME` is not set then the file `$HOME/.config/containers/storage.conf` is used.
+
+/etc/projects - XFS persistent project root definition
+/etc/projid -  XFS project name mapping file
+
+## SEE ALSO
+`semanage(8)`, `restorecon(8)`, `mount(8)`, `fuse-overlayfs(1)`, `xfs_quota(8)`, `projects(5)`, `projid(5)`
 
 ## HISTORY
 May 2017, Originally compiled by Dan Walsh <dwalsh@redhat.com>
--- a/drivers/quota/projectquota.go
+++ b/drivers/quota/projectquota.go
@@ -52,14 +52,19 @@ import "C"
 import (
 	"fmt"
 	"io/ioutil"
+	"math"
+	"os"
 	"path"
 	"path/filepath"
+	"syscall"
 	"unsafe"
 
 	"github.com/containers/storage/pkg/directory"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
+
+const projectIDsAllocatedPerQuotaHome = 10000
 
 // Quota limit params - currently we only control blocks hard limit and inodes
 type Quota struct {
@@ -75,23 +80,48 @@ type Control struct {
 	quotas            map[string]uint32
 }
 
+// Attempt to generate a unigue projectid.  Multiple directories
+// per file system can have quota and they need a group of unique
+// ids. This function attempts to allocate at least projectIDsAllocatedPerQuotaHome(10000)
+// unique projectids, based on the inode of the basepath.
+func generateUniqueProjectID(path string) (uint32, error) {
+	fileinfo, err := os.Stat(path)
+	if err != nil {
+		return 0, err
+	}
+	stat, ok := fileinfo.Sys().(*syscall.Stat_t)
+	if !ok {
+		return 0, fmt.Errorf("Not a syscall.Stat_t %s", path)
+
+	}
+	projectID := projectIDsAllocatedPerQuotaHome + (stat.Ino*projectIDsAllocatedPerQuotaHome)%(math.MaxUint32-projectIDsAllocatedPerQuotaHome)
+	return uint32(projectID), nil
+}
+
 // NewControl - initialize project quota support.
 // Test to make sure that quota can be set on a test dir and find
 // the first project id to be used for the next container create.
 //
 // Returns nil (and error) if project quota is not supported.
 //
-// First get the project id of the home directory.
+// First get the project id of the basePath directory.
 // This test will fail if the backing fs is not xfs.
 //
 // xfs_quota tool can be used to assign a project id to the driver home directory, e.g.:
-//    echo 999:/var/lib/containers/storage/overlay >> /etc/projects
-//    echo storage:999 >> /etc/projid
-//    xfs_quota -x -c 'project -s storage' /<xfs mount point>
+//    echo 100000:/var/lib/containers/storage/overlay >> /etc/projects
+//    echo 200000:/var/lib/containers/storage/volumes >> /etc/projects
+//    echo storage:100000 >> /etc/projid
+//    echo volumes:200000 >> /etc/projid
+//    xfs_quota -x -c 'project -s storage volumes' /<xfs mount point>
 //
-// In that case, the home directory project id will be used as a "start offset"
-// and all containers will be assigned larger project ids (e.g. >= 1000).
-// This is a way to prevent xfs_quota management from conflicting with containers/storage.
+// In the example above, the storage directory project id will be used as a
+// "start offset" and all containers will be assigned larger project ids
+// (e.g. >= 100000). Then the volumes directory project id will be used as a
+// "start offset" and all volumes will be assigned larger project ids
+// (e.g. >= 200000).
+// This is a way to prevent xfs_quota management from conflicting with
+// containers/storage.
+
 //
 // Then try to create a test directory with the next project id and set a quota
 // on it. If that works, continue to scan existing containers to map allocated
@@ -105,8 +135,15 @@ func NewControl(basePath string) (*Control, error) {
 	if err != nil {
 		return nil, err
 	}
-	minProjectID++
+	if minProjectID == 0 {
+		// Indicates the storage was never initialized
+		// Generate a unique range of Projectids for this basepath
+		minProjectID, err = generateUniqueProjectID(basePath)
+		if err != nil {
+			return nil, err
+		}
 
+	}
 	//
 	// create backing filesystem device node
 	//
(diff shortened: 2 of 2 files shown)


rhatdan:
@mtrmac @nalind @vrothberg @rhvgoyal PTAL

drivers/quota/projectquota.go
```diff
@@ -129,7 +151,7 @@ func NewControl(basePath string) (*Control, error) {
 
 	q := Control{
 		backingFsBlockDev: backingFsBlockDev,
-		nextProjectID:     minProjectID + 1,
+		nextProjectID:     minProjectID,
```
mtrmac, line 154: `minProjectID` was already used for something just above. Why is it OK to reuse it here? It might well be, but the commit message / PR description doesn’t say that it is, or even that it is changing this.

drivers/quota/projectquota.go
```diff
@@ -141,6 +163,12 @@ func NewControl(basePath string) (*Control, error) {
 		return nil, err
 	}
 
+	if q.nextProjectID == 1 {
```
mtrmac, line 166: What does this mean? The project ID is computed by `getProjectID` +` findNextProjectID` + some increments. Under which conditions would it end up 1?

This might well be correct, but please comment this to clearly document the intent and logic instead of just using a magic `1`.

Also, isn’t this too late, if `setProjectQuota` was already called on `minProjectID`?

drivers/quota/projectquota.go
```diff
@@ -75,6 +78,25 @@ type Control struct {
...
+		return 0, fmt.Errorf("Not a syscall.Stat_t %s", path)
+
+	}
+	projectID := uint32(((stat.Ino * minProjectids) % math.MaxUint32))
+	return projectID, nil
+}
+
 // NewControl - initialize project quota support.
 // Test to make sure that quota can be set on a test dir and find
 // the first project id to be used for the next container create.
```
mtrmac, line 102: This documentation needs updating.

drivers/quota/projectquota.go
```diff
@@ -75,6 +78,25 @@ type Control struct {
 	quotas            map[string]uint32
 }
 
+// Attempt to generate a unigue projectid.  Multiple directories
+// per file system can have quota and they need a group of unigue
+// ids. This function attempts to allocate at least minPorjectIDS(10000)
+// unigue projectids, based on the inode of the basepath.
+func generateUnigueProjectID(path string) (uint32, error) {
+	const minProjectids = 10000
```
mtrmac, line 86: If this is a global tuning parameter, make it more prominent, at least a top-level constant on the top of the file.

And isn’t it more a `projectIDsAllocatedPerQuotaHome` or something like that? At least it’s more of a “max”ProjectIDs than a “min”projectIDs.

---

AFAICS this doesn’t even _try_ to prevent overlap. Shouldn’t it?

drivers/quota/projectquota.go
```diff
@@ -75,6 +78,25 @@ type Control struct {
...
+	fileinfo, err := os.Stat(path)
+	if err != nil {
+		return 0, err
+	}
+	stat, ok := fileinfo.Sys().(*syscall.Stat_t)
+	if !ok {
+		return 0, fmt.Errorf("Not a syscall.Stat_t %s", path)
+
+	}
+	projectID := uint32(((stat.Ino * minProjectids) % math.MaxUint32))
```
mtrmac, line 96: Are the project IDs supposed to be unique across the whole OS? (I have no idea, but the existence of `/etc/projects` and `/etc/projid` seems to suggest so.) If they are supposed to be OS-global, how does this ensure non-c/storage users don’t conflict?

---

There already is a `findNextProjectID` mechanism. Why does this do something so very different?

mtrmac (commented):
> Originally container/storage only did quota on rootfs, now we want to support it on volumes as well. 

Is that a forthcoming PR? Right now I only can see a single overlay caller. Assuming so…

---

(I know nothing about project quotas, I haven’t now taken the time to learn, I’m blindly assuming that the current implementation is correct and does what it says.) As an absolute newbie to this, the current “take an allocated value and just keep incrementing” approach already feels like a hack^Wcompromise, and layering it under another unreliable compromise feels unnecessarily unreliable, especially for a security mechanism.

As an initial absolutely uneducated guess, why not make the project allocator a per-c/storage.Store state (it might even keep its current initialization mechanism of starting with a project ID associated with the driver directory), and give out single project IDs for both the existing overlay-driven projects, and the new (volume?) ones from that allocator, instead of trying to blindly partition space that isn’t ours at all?

drivers/quota/projectquota.go
mtrmac: At the very least, this should be something like `projectID := minProjectID + (stat.Ino * minProjectids) % (math.MaxUint32 - minProjectID)`, with `minProjectID` determined by the caller, so that this does not wrap and overlap the likely-allocated values [0, 1, 2, …] for those systems where the administrators intentionally set a high `minProjectID` by allocating one.

rhatdan:
https://github.com/containers/podman/pull/10973 uses containers/storage/drivers/quota on volumes.

drivers/quota/projectquota.go
rhatdan: It will only be 1 if no quota has ever been set.  setProjectQuota above is checking to see if quota is supported.

drivers/quota/projectquota.go
rhatdan: I am not sure of the best way to do this. This was just a stab at it, hoping someone would have a silver bullet.

drivers/quota/projectquota.go
rhatdan: Done

drivers/quota/projectquota.go
mtrmac: Ah, the `setProjectQuota` above seems to be a no-op.

drivers/quota/projectquota.go
mtrmac: Then please make `getProjectID` document that it returns 0 (without error) on a non-quota directory, and add a comment to this condition as well.

mtrmac: … the `setProjectQuota` call above _is_ well-documented, my bad for not noticing.

drivers/quota/projectquota.go
rhatdan: Updated and added information to the man page.

drivers/quota/projectquota.go
rhatdan: I reworked it and if the user configures the system correctly then the overflow IDs will no longer be a problem. Less hacky.

drivers/quota/projectquota.go
rhatdan: I have modified the code to simplify.

rhatdan:
@TomSweeneyRedHat @vrothberg PTAL

drivers/quota/projectquota.go
mtrmac: Oops, `minProjectID` is not available to this function, that’s why it’s called in the first place. The current version looks reasonable.

docs/containers-storage.conf.5.md
```diff
@@ -263,13 +263,45 @@ The semanage command above tells SELinux to setup the default labeling of `NEWST
 
 Now all new content created in these directories will automatically be created with the correct label.
 
-## SEE ALSO
-`semanage(8)`, `restorecon(8)`, `mount(8)`, `fuse-overlayfs(1)`
+## QUOTAS
+
+Container storage impliments `XFS project quota controls for overlay storage
```
mtrmac, line 268:
```suggestion
Container storage implements `XFS project quota controls for overlay storage
```

mtrmac (commented):
Thanks for the documentation update.

Ideally I’d prefer to just extend the existing code, with the single allocation range, across both image store and volumes associated with a single `c/storage.Store`, but if that is not wanted, this implementation looks nice.

drivers/quota/projectquota.go
```diff
@@ -75,22 +80,44 @@ type Control struct {
 	quotas            map[string]uint32
 }
 
+// Attempt to generate a unigue projectid.  Multiple directories
+// per file system can have quota and they need a group of unigue
+// ids. This function attempts to allocate at least projectIDsAllocatedPerQuotaHome(10000)
+// unigue projectids, based on the inode of the basepath.
+func generateUnigueProjectID(path string) (uint32, error) {
```
mtrmac, line 87:
```suggestion
func generateUniqueProjectID(path string) (uint32, error) {
```
also elsewhere.

docs/containers-storage.conf.5.md
```diff
@@ -263,13 +263,45 @@ The semanage command above tells SELinux to setup the default labeling of `NEWST
...
+Container storage implements `XFS project quota controls` for overlay storage
+containers and volumes. The directory used to store the containers must be an
+`XFS` file system and be mounted with the `pquota` option.
+
+Example /etc/fstab entry:
+```
+/dev/podman/podman-var /var xfs defaults,x-systemd.device-timeout=0,pquota 1 2
+```
+
+Container storage generates project ids for each container and builtin volume, but these project ids need to be unigue for the XFS file system.
```
TomSweeneyRedHat, line 277:
```suggestion
Container storage generates project ids for each container and builtin volume, but these project ids need to be unique for the XFS file system.
```

docs/containers-storage.conf.5.md
```diff
@@ -263,13 +263,45 @@ The semanage command above tells SELinux to setup the default labeling of `NEWST
...
+
+xfs_quota tool can be used to assign a project id to the storage driver directory, e.g.:
```
TomSweeneyRedHat, line 279:
```suggestion
The xfs_quota tool can be used to assign a project id to the storage driver directory, e.g.:
```

docs/containers-storage.conf.5.md
```diff
@@ -263,13 +263,45 @@ The semanage command above tells SELinux to setup the default labeling of `NEWST
...
+echo 100000:/var/lib/containers/storage/storage >> /etc/projects
+echo 200000:/var/lib/containers/storage/volumes >> /etc/projects
+echo storage:100000 >> /etc/projid
+echo volumes:200000 >> /etc/projid
+xfs_quota -x -c 'project -s storage volumes' /<xfs mount point>
+```	     
+
+In that case, the storage directory project id will be used as a "start offset"
+and all containers will be assigned larger project ids (e.g. >= 100000).
+the volumes directory project id will be used as a "start offset"
```
TomSweeneyRedHat, line 291:
```suggestion
Then the volumes directory project id will be used as a "start offset"
```

docs/containers-storage.conf.5.md
```diff
@@ -263,13 +263,45 @@ The semanage command above tells SELinux to setup the default labeling of `NEWST
...
+
+```
+echo 100000:/var/lib/containers/storage/storage >> /etc/projects
+echo 200000:/var/lib/containers/storage/volumes >> /etc/projects
+echo storage:100000 >> /etc/projid
+echo volumes:200000 >> /etc/projid
+xfs_quota -x -c 'project -s storage volumes' /<xfs mount point>
+```	     
+
+In that case, the storage directory project id will be used as a "start offset"
```
TomSweeneyRedHat, line 289:
```suggestion
In the example above, the storage directory project id will be used as a "start offset"
```

drivers/quota/projectquota.go
```diff
@@ -75,22 +80,44 @@ type Control struct {
 	quotas            map[string]uint32
 }
 
+// Attempt to generate a unigue projectid.  Multiple directories
+// per file system can have quota and they need a group of unigue
```
TomSweeneyRedHat, line 84:
```suggestion
// per file system can have quota and they need a group of unique
```

drivers/quota/projectquota.go
```diff
@@ -75,22 +80,44 @@ type Control struct {
...
+// ids. This function attempts to allocate at least projectIDsAllocatedPerQuotaHome(10000)
+// unigue projectids, based on the inode of the basepath.
```
TomSweeneyRedHat, line 86:
```suggestion
// unique projectids, based on the inode of the basepath.
```

drivers/quota/projectquota.go
```diff
@@ -75,22 +80,44 @@ type Control struct {
...
+//    echo storage:100000 >> /etc/projid
+//    echo volumes:200000 >> /etc/projid
+//    xfs_quota -x -c 'project -s storage volumes' /<xfs mount point>
 //
-// In that case, the home directory project id will be used as a "start offset"
-// and all containers will be assigned larger project ids (e.g. >= 1000).
+// In that case, the storage directory project id will be used as a "start offset"
+// and all containers will be assigned larger project ids (e.g. >= 100000).
+// the volumes directory project id will be used as a "start offset"
+// and all volumes will be assigned larger project ids (e.g. >= 200000).
```
TomSweeneyRedHat, line 120: if you take the comments above, please use here too

drivers/quota/projectquota.go
```diff
@@ -105,8 +132,15 @@ func NewControl(basePath string) (*Control, error) {
 	if err != nil {
 		return nil, err
 	}
-	minProjectID++
+	if minProjectID == 0 {
+		// Indicates the storage was never initialized
+		// Generate a unigue range of Projectids for this basepath
```
TomSweeneyRedHat, line 137:
```suggestion
		// Generate a unique range of Projectids for this basepath
```

docs/containers-storage.conf.5.md
```diff
@@ -263,13 +263,45 @@ The semanage command above tells SELinux to setup the default labeling of `NEWST
...
+```
+/dev/podman/podman-var /var xfs defaults,x-systemd.device-timeout=0,pquota 1 2
+```
+
+Container storage generates project ids for each container and builtin volume, but these project ids need to be unique for the XFS file system.
+
+The xfs_quota tool can be used to assign a project id to the storage driver directory, e.g.:
+
+```
+echo 100000:/var/lib/containers/storage/storage >> /etc/projects
```
nalind, line 282:
```suggestion
echo 100000:/var/lib/containers/storage/overlay >> /etc/projects
```

rhatdan:
@mtrmac @TomSweeneyRedHat @nalind @saschagrunert @giuseppe @vrothberg @haircommander @mrunalp PTAL

Need to get this in, so that CRI-O can  use it.

mtrmac (commented):
Code LGTM, given the design; I’m not 100% happy about potentially taking over all of the uint32 project ID space but if it is opt-in like https://github.com/containers/podman/pull/10973, it’s acceptable (and good enough for RHCOS).

TomSweeneyRedHat:
LGTM

PR #976, opened 2021-07-28
overlay: check for aufs-style whiteout at startup

nalind:
At startup, check if we see any aufs-style whiteout in any of our layers.  This can be expensive, so if we don't find any at startup, note that and bypass the check later until the system reboots.  If we do find some, leave a note that we needed to use a mount_program so that we don't have to bother with even this in the future.

### DIFF
--- a/drivers/overlay/check_116.go
+++ b/drivers/overlay/check_116.go
@@ -1,0 +1,42 @@
+// +build go1.16
+
+package overlay
+
+import (
+	"io/fs"
+	"path/filepath"
+	"strings"
+
+	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/system"
+)
+
+func scanForMountProgramIndicators(home string) (detected bool, err error) {
+	err = filepath.WalkDir(home, func(path string, d fs.DirEntry, err error) error {
+		if detected {
+			return fs.SkipDir
+		}
+		if err != nil {
+			return err
+		}
+		basename := filepath.Base(path)
+		if strings.HasPrefix(basename, archive.WhiteoutPrefix) {
+			detected = true
+			return fs.SkipDir
+		}
+		if d.IsDir() {
+			xattrs, err := system.Llistxattr(path)
+			if err != nil {
+				return err
+			}
+			for _, xattr := range xattrs {
+				if strings.HasPrefix(xattr, "user.fuseoverlayfs.") || strings.HasPrefix(xattr, "user.containers.") {
+					detected = true
+					return fs.SkipDir
+				}
+			}
+		}
+		return nil
+	})
+	return detected, err
+}
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -266,9 +266,8 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 	}
 
 	if opts.mountProgram != "" {
-		f, err := os.Create(getMountProgramFlagFile(home))
-		if err == nil {
-			f.Close()
+		if err := ioutil.WriteFile(getMountProgramFlagFile(home), []byte("true"), 0600); err != nil {
+			return nil, err
 		}
 	} else {
 		// check if they are running over btrfs, aufs, zfs, overlay, or ecryptfs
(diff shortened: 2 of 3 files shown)


drivers/overlay/check_116.go
```diff
@@ -0,0 +1,26 @@
+// +build go1.16
```
rhatdan, line 1: What happens when go.1.17 happens?

rhatdan (approved):
This looks good, but can we assume that after we checked and found no aufs-style that we never need to check again.  IE I don't want to pay the penalty to check forever after every reboot.

WDYT @giuseppe

rhatdan:
@cevich Could you take a look at the devicemapper flakes,  These are happening regularly.

giuseppe:
> This looks good, but can we assume that after we checked and found no aufs-style that we never need to check again. IE I don't want to pay the penalty to check forever after every reboot.
> 
> WDYT @giuseppe

I think that would be better.  Could we store this information in the home directory, perhaps using the `getMountProgramFlagFile(home)` file, instead of the feature checks that are done after each reboot?

rhatdan:
Well  I would want the checks to happen after each reboot, until we are sure that all aufs style flags are gone.

@giuseppe Does newer fuse-overlay use the mknod(0,0) if the kernel supports it?

giuseppe:
> Well I would want the checks to happen after each reboot, until we are sure that all aufs style flags are gone.

I am afraid that will be expensive on a cold cache and depending on the storage size, it can add several seconds to the first access to the storage.

What could change between reboots to invalidate the storage having aufs style whiteout files?  I am fine if we do it until there are
aufs whiteout files and warn the user about them but IMO the check should not happen once all the aufs whiteout files are gone and the storage fully switched to native overlay.

> @giuseppe Does newer fuse-overlay use the mknod(0,0) if the kernel supports it?

yes, fuse-overlayfs uses it and then fallback to aufs style whiteout files

drivers/overlay/check_116.go
nalind: Each version of Go continues to define the build tags for versions that preceded it.

nalind:
> What could change between reboots to invalidate the storage having aufs style whiteout files?

People run versions of the software that predate any of the new state-of-things files that we're laying down.  We still prevent someone from using kernel overlay if they've ever used fuse-overlayfs, even after they remove every layer where it makes a difference, so it's not a complete fix, either.

rhatdan:
They could reuse older kernels, which do not allow mknod(0,0), is there a way we could remove the indicator file if we determine that mknod(0,0) is not supported.

Check if mknod(0.0) can happen.

I think we can come with a startup that would check if indicator files have been

rhatdan:
If I closed this, I certainly did not mean to.

cevich:
> @cevich Could you take a look at the devicemapper flakes, These are happening regularly.

Sorry, still playing catch up after only two weekdays away, whew!

I see all green check-marks, is there someplace specific I can look (assuming this is still relevant)?

cevich:
Perhaps this will help fix or show me the problem: https://github.com/containers/storage/pull/982

drivers/overlay/check_116.go
```diff
@@ -0,0 +1,39 @@
...
+	"io/fs"
+	"path/filepath"
+	"strings"
+
+	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/system"
+)
+
+func scanForMountProgramIndicators(home string) (detected bool, err error) {
+	err = filepath.WalkDir(home, func(path string, d fs.DirEntry, err error) error {
```
giuseppe, line 15: should we handle the `err != nil` case?

nalind: Evidently, we need to.  Pushing a tweak for that.

drivers/overlay/overlay.go
```diff
@@ -542,9 +541,29 @@ func SupportsNativeOverlay(graphroot, rundir string) (bool, error) {
...
+	var contents string
+	flagContent, err := ioutil.ReadFile(getMountProgramFlagFile(home))
+	if err == nil {
+		contents = strings.TrimSpace(string(flagContent))
+	}
+	switch contents {
+	case "true":
 		logrus.Debugf("overlay storage already configured with a mount-program")
 		return false, nil
+	case "":
```
rhatdan, line 553: Should this be the default case.  In cases with the file does not have "true" or "false" then scan.

nalind: Replacing `case ""` with `default`, which is what I think you're asking here.

nalind:
Rebased.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM, but out of date again.
@nalind could you spin up a WIP Podman PR to make sure nothing breaks over there when this is vendored in please?

giuseppe (approved):
LGTM

nalind:
Rebased, and opened https://github.com/containers/podman/pull/11168.

Issue #985, opened 2021-08-09
Run go get github.com/containers/storage/cmd/containers-storage@v1.33.2 failed


zhujian7:
Run `go get github.com/containers/storage/cmd/containers-storage@v1.33.2` failed with information:
```
root@6ec030b59786:/go# go get github.com/containers/storage/cmd/containers-storage@v1.33.2
go: finding github.com v1.33.2
go: finding github.com/containers/storage/cmd v1.33.2
go: finding github.com/containers v1.33.2
go: finding github.com/containers/storage/cmd/containers-storage v1.33.2
go: downloading github.com/opencontainers/go-digest v1.0.0
go: downloading github.com/ulikunitz/xz v0.5.10
go: downloading github.com/pkg/errors v0.9.1
go: downloading github.com/klauspost/compress v1.13.3
go: downloading github.com/json-iterator/go v1.1.11
go: downloading github.com/opencontainers/runc v1.0.1
go: downloading github.com/opencontainers/selinux v1.8.3
go: downloading github.com/docker/go-units v0.4.0
go: extracting github.com/opencontainers/go-digest v1.0.0
go: extracting github.com/pkg/errors v0.9.1
go: downloading golang.org/x/sys v0.0.0-20210426230700-d19ff857e887
go: downloading github.com/klauspost/pgzip v1.2.5
go: extracting github.com/json-iterator/go v1.1.11
go: extracting github.com/docker/go-units v0.4.0
go: extracting github.com/opencontainers/selinux v1.8.3
go: downloading github.com/google/go-intervals v0.0.2
go: downloading github.com/hashicorp/go-multierror v1.1.1
go: downloading github.com/vbatts/tar-split v0.11.1
go: downloading github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
go: downloading github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635
go: extracting github.com/ulikunitz/xz v0.5.10
go: extracting github.com/klauspost/pgzip v1.2.5
go: extracting github.com/hashicorp/go-multierror v1.1.1
go: downloading github.com/hashicorp/errwrap v1.0.0
go: extracting github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635
go: downloading github.com/moby/sys/mountinfo v0.4.1
go: extracting golang.org/x/sys v0.0.0-20210426230700-d19ff857e887
go: extracting github.com/google/go-intervals v0.0.2
[...]

PR #986, opened 2021-08-09
Invert libsubid tag

saschagrunert:
We now switch to a `libsubid` tag to increase the usability of the
library on systems not having the dependency in place.

Fixes https://github.com/containers/storage/issues/985

### DIFF
--- a/hack/libsubid_tag.sh
+++ b/hack/libsubid_tag.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 if test $(${GO:-go} env GOOS) != "linux" ; then
-	echo no_libsubid
 	exit 0
 fi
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

nalind:
LGTM

saschagrunert:
@rhatdan can we tag a new patch release to make the upgrade easier in CRI-O? :)

rhatdan:
I created a new release v1.34.0

mtrmac:
Is the default likely to flip again? I.e. should the build systems that were recently updated to add `no_libsubid` remove that as useless noise, or keep it?

rhatdan:
Yes they should remove.  I will open PRs for skopeo to fix.  We need an updted hack/libsubuid.sh as well.

PR #989, opened 2021-08-11
ApplyDiff: compress saved headers without concurrency

nalind:
When we're applying a diff, we compress the headers and stash them elsewhere so that we can use them to correctly reconstruct the layer if we need to extract its contents later.

By default, the compression uses a 1MB block, and up to GOMAXPROCS threads, which results in allocating GOMAXPROCS megabytes of memory up front.  That can be much more than we need, especially if the system has many, many cores.  Drop it down to 1 megabyte.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -27,6 +27,7 @@ import (
 	digest "github.com/opencontainers/go-digest"
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/vbatts/tar-split/archive/tar"
 	"github.com/vbatts/tar-split/tar/asm"
 	"github.com/vbatts/tar-split/tar/storage"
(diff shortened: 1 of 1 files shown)


layers.go
```diff
@@ -1529,6 +1530,9 @@ func (r *layerStore) ApplyDiff(to string, diff io.Reader) (size int64, err error
 	if err != nil {
 		compressor = pgzip.NewWriter(&tsdata)
 	}
+	if err := compressor.SetConcurrency(1024*1024, 1); err != nil { // 1024*1024 is the hard-coded default; we're not changing that
+		logrus.Debugf("error setting compression concurrency threads to 1: %v; ignoring", err)
```
rhatdan, line 1534: Should this be Info?

rhatdan (approved):
LGTM one question.

layers.go
nalind: Changed it.

TomSweeneyRedHat:
LGTM

mrunalp:
@saschagrunert Can we pick this into CRI-O branches? @nalind Thanks! :)

mrunalp:
Turns out this didn't help as we hoped. The pgzip library is using sync.Pool and I still see the 1MB buffer being created a lot of times during a pool. Next, looking into a fixed pool approach which doesn't get GC'ed.

PR #993, opened 2021-08-12
[1.24] ApplyDiff: compress saved headers without concurrency

saschagrunert:
Cherry-pick of https://github.com/containers/storage/pull/989

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -28,6 +28,7 @@ import (
 	digest "github.com/opencontainers/go-digest"
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


nalind:
LGTM

cevich:
> I believe we can ignore the ubuntu failures.

Confirming: Yes, this is safe in this case there was an actual problem in the kubik repositories.  That should be sorted, but I'm not very comfortable that we're doing live package installs/updates on release-branch CI.  This is highly prone for this kind of breakage.  It might help if we disabled the Ubuntu tests...since I know for sure those legacy repos. are not nearly as stable as for Fedora.  However this will negatively impact Confidence in Ubuntu for any needed backports.  I'm not qualified to determine if that's worry-worthy or not.

Let me know and I'll open a PR to disable them if desired.

PR #995, opened 2021-08-16
chunked: add new pull options use_hard_links and enable_partial_images

giuseppe:
if the option use_hard_links is set:
    
[storage.options]
pull_options = {use_hard_links = "true"}

then attempt to deduplicate files using hard links first.

the partial images feature is disabled by default, to enable it we need now:

pull_options = {enable_partial_images = "true"}

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -54,7 +54,7 @@ func timeToTimespec(time time.Time) (ts unix.Timespec) {
 	return unix.NsecToTimespec(time.UnixNano())
 }
 
-func copyFileContent(srcFd int, destFile, root string, dirfd int, missingDirsMode, mode os.FileMode, useHardLinks bool) (*os.File, int64, error) {
+func copyFileContent(srcFd int, destFile, root string, dirfd int, mode os.FileMode, useHardLinks bool) (*os.File, int64, error) {
 	src := fmt.Sprintf("/proc/self/fd/%d", srcFd)
 	st, err := os.Stat(src)
 	if err != nil {
@@ -173,7 +173,7 @@ func makeZstdChunkedDiffer(ctx context.Context, store storage.Store, blobSize in
 	}, nil
 }
 
-func copyFileFromOtherLayer(file internal.ZstdFileMetadata, source string, otherFile *internal.ZstdFileMetadata, root string, dirfd int, missingDirsMode os.FileMode, useHardLinks bool) (bool, *os.File, int64, error) {
+func copyFileFromOtherLayer(file internal.ZstdFileMetadata, source string, otherFile *internal.ZstdFileMetadata, root string, dirfd int, useHardLinks bool) (bool, *os.File, int64, error) {
 	srcDirfd, err := unix.Open(source, unix.O_RDONLY, 0)
 	if err != nil {
 		return false, nil, 0, err
@@ -186,17 +186,14 @@ func copyFileFromOtherLayer(file internal.ZstdFileMetadata, source string, other
 	}
 	defer srcFile.Close()
 
-	srcPath := fmt.Sprintf("/proc/self/fd/%d", srcFile.Fd())
-
-	dstFile, written, err := copyFileContent(srcPath, file.Name, root, dirfd, missingDirsMode, 0, useHardLinks)
+	dstFile, written, err := copyFileContent(int(srcFile.Fd()), file.Name, root, dirfd, 0, useHardLinks)
 	if err != nil {
 		return false, nil, 0, err
 	}
-
-	return true, dstFile, written, nil
+	return true, dstFile, written, err
 }
 
-func findFileInOtherLayers(file internal.ZstdFileMetadata, root string, dirfd int, layersMetadata map[string]map[string]*internal.ZstdFileMetadata, layersTarget map[string]string, missingDirsMode os.FileMode, useHardLinks bool) (bool, *os.File, int64, error) {
+func findFileInOtherLayers(file internal.ZstdFileMetadata, root string, dirfd int, layersMetadata map[string]map[string]*internal.ZstdFileMetadata, layersTarget map[string]string, useHardLinks bool) (bool, *os.File, int64, error) {
 	// this is ugly, needs to be indexed
 	for layerID, checksums := range layersMetadata {
 		m, found := checksums[file.Digest]
@@ -209,7 +206,7 @@ func findFileInOtherLayers(file internal.ZstdFileMetadata, root string, dirfd in
 			continue
 		}
 
-		found, dstFile, written, err := copyFileFromOtherLayer(file, source, m, root, dirfd, missingDirsMode, useHardLinks)
+		found, dstFile, written, err := copyFileFromOtherLayer(file, source, m, root, dirfd, useHardLinks)
 		if found && err == nil {
 			return found, dstFile, written, err
 		}
@@ -228,7 +225,7 @@ func getFileDigest(f *os.File) (digest.Digest, error) {
 // findFileOnTheHost checks whether the requested file already exist on the host and copies the file content from there if possible.
 // It is currently implemented to look only at the file with the same path.  Ideally it can detect the same content also at different
 // paths.
-func findFileOnTheHost(file internal.ZstdFileMetadata, root string, dirfd int, missingDirsMode os.FileMode, useHardLinks bool) (bool, *os.File, int64, error) {
+func findFileOnTheHost(file internal.ZstdFileMetadata, root string, dirfd int, useHardLinks bool) (bool, *os.File, int64, error) {
 	sourceFile := filepath.Clean(filepath.Join("/", file.Name))
 	if !strings.HasPrefix(sourceFile, "/usr/") {
 		// limit host deduplication to files under /usr.
@@ -266,7 +263,7 @@ func findFileOnTheHost(file internal.ZstdFileMetadata, root string, dirfd int, m
 		return false, nil, 0, nil
 	}
 
-	dstFile, written, err := copyFileContent(fmt.Sprintf("/proc/self/fd/%d", fd), file.Name, root, dirfd, missingDirsMode, 0, useHardLinks)
+	dstFile, written, err := copyFileContent(fd, file.Name, root, dirfd, 0, useHardLinks)
 	if err != nil {
 		return false, nil, 0, nil
 	}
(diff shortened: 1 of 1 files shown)


mtrmac:
> if the option use_hard_links is set:
> 
> [storage.options]
> pull_options = {use_hard_links = "true"}
> 
> then attempt to deduplicate files using hard links first.

(I know nothing about the problem space.) Why is this an user-facing option? Is there some kind of trade-off? (And is it safe to hard-link files across container layers? I have no idea).

Or is this something we _could_ decide automatically but just don’t have the code yet?

---

> the partial images feature is disabled by default, to enable it we need now:

So, for the record, the _primary_ purpose of this PR is to disable partial image use by default? Please write those kinds of things down so that we can, 5 years later, find out _why_ things happened.

giuseppe:
> Why is this an user-facing option? Is there some kind of trade-off? (And is it safe to hard-link files across container layers? I have no idea).
> 
> Or is this something we _could_ decide automatically but just don’t have the code yet?

it is for a very specific use case (deduplication and share inodes) that could be useful for automotive.  It should not be enabled by default as it changes how the inode looks like when you stat it and in general it is less safe.

rhatdan:
No documentation in man page?  Do you not want to document it, so we can experiment with it?

giuseppe:
I'd still like to keep it hidden for now, so we are more freedom to break the format and the configuration options.

rhatdan:
LGTM

rhatdan (approved):

pkg/chunked/storage_linux.go
```diff
@@ -148,7 +164,26 @@ func makeZstdChunkedDiffer(ctx context.Context, store storage.Store, blobSize in
 	}, nil
 }
 
-func findFileInOtherLayers(file internal.ZstdFileMetadata, root string, dirfd int, layersMetadata map[string]map[string]*internal.ZstdFileMetadata, layersTarget map[string]string, missingDirsMode os.FileMode) (*os.File, int64, error) {
+func copyFileFromOtherLayer(file internal.ZstdFileMetadata, source string, otherFile *internal.ZstdFileMetadata, root string, dirfd int, missingDirsMode os.FileMode, useHardLinks bool) (bool, *os.File, int64, error) {
```
mtrmac, line 167: Documenting the parameters and return values would be nice.

---

(`missingDirMode` seems unused, except passing it to functions that don’t use it.)

pkg/chunked/storage_linux.go
```diff
@@ -148,7 +164,26 @@ func makeZstdChunkedDiffer(ctx context.Context, store storage.Store, blobSize in
...
+	}
+	defer srcFile.Close()
+
+	srcPath := fmt.Sprintf("/proc/self/fd/%d", srcFile.Fd())
+
+	dstFile, written, err := copyFileContent(srcPath, file.Name, root, dirfd, missingDirsMode, 0, useHardLinks)
+	return true, dstFile, written, err
+}
+
+func findFileInOtherLayers(file internal.ZstdFileMetadata, root string, dirfd int, layersMetadata map[string]map[string]*internal.ZstdFileMetadata, layersTarget map[string]string, missingDirsMode os.FileMode, useHardLinks bool) (bool, *os.File, int64, error) {
```
mtrmac, line 186: Documenting the parameters and return values would be nice.

---

(`missingDirMode` seems unused, except passing it to functions that don’t use it.)

pkg/chunked/storage_linux.go
```diff
@@ -195,61 +215,61 @@ func getFileDigest(f *os.File) (digest.Digest, error) {
 // findFileOnTheHost checks whether the requested file already exist on the host and copies the file content from there if possible.
 // It is currently implemented to look only at the file with the same path.  Ideally it can detect the same content also at different
 // paths.
-func findFileOnTheHost(file internal.ZstdFileMetadata, root string, dirfd int, missingDirsMode os.FileMode) (*os.File, int64, error) {
+func findFileOnTheHost(file internal.ZstdFileMetadata, root string, dirfd int, missingDirsMode os.FileMode, useHardLinks bool) (bool, *os.File, int64, error) {
```
mtrmac, line 218: Documenting the parameters and return values would be nice.

---

(`missingDirMode` seems unused, except passing it to functions that don’t use it. Actually I can’t find a single real user in the whole file.)

pkg/chunked/storage_linux.go
```diff
@@ -195,61 +215,61 @@ func getFileDigest(f *os.File) (digest.Digest, error) {
...
+	dstFile, written, err := copyFileContent(fmt.Sprintf("/proc/self/fd/%d", fd), file.Name, root, dirfd, missingDirsMode, 0, useHardLinks)
 	if err != nil {
-		return nil, 0, nil
+		return false, nil, 0, nil
 	}
 
 	// calculate the checksum again to make sure the file wasn't modified while it was copied
 	if _, err := f.Seek(0, 0); err != nil {
-		return nil, 0, err
+		return false, nil, 0, err
```
mtrmac, line 263: Pre-existing: Shouldn’t all error paths from here on close `dstFile` if it is not `nil`?

pkg/chunked/storage_linux.go
```diff
@@ -654,10 +674,19 @@ func (d *chunkedZstdDiffer) ApplyDiff(dest string, options *archive.TarOptions)
...
+
+	if value := storeOpts.PullOptions["enable_partial_images"]; strings.ToLower(value) != trueVal {
+		return output, errors.New("enable_partial_images not configured")
+	}
+
+	enableHostDedup, useHardLinks := false, false
+	if value := storeOpts.PullOptions["enable_host_deduplication"]; strings.ToLower(value) == trueVal {
 		enableHostDedup = true
 	}
+	if value := storeOpts.PullOptions["use_hard_links"]; strings.ToLower(value) == trueVal {
```
mtrmac, line 687: Three copies seem enough for me to extract this into a helper function, `parseBooleanPullOption(storeOpts, option_name, default)` or something like that.

pkg/chunked/storage_linux.go
```diff
@@ -195,61 +215,61 @@ func getFileDigest(f *os.File) (digest.Digest, error) {
...
+		return false, nil, 0, err
 	}
 	checksum, err = getFileDigest(f)
 	if err != nil {
-		return nil, 0, err
+		return false, nil, 0, err
 	}
 	if checksum != manifestChecksum {
-		return nil, 0, nil
+		return false, nil, 0, nil
```
mtrmac, line 270: This leaves the non-matching file around. Isn’t that a problem? I’d expect `Linkat` to fail with `EEXIST` on any future attempts. I didn’t look into the non-hardlink case, so it might be pre-existing.

---

(Vague and non-blocking: Overall this second check feels laborious, looking also at the caller that must _conditionally_ do `setFileAttrs` only afterwards, and duplicates that code. It seems to me that there should be some way to structure this differently, maybe pass the attributes to set all the way down to `copyFileContent`, somehow.)

pkg/chunked/storage_linux.go
```diff
@@ -819,11 +848,13 @@ func (d *chunkedZstdDiffer) ApplyDiff(dest string, options *archive.TarOptions)
 				return output, err
 			}
 			dstFile.Close()
```
mtrmac, line 850: So the hard-link code ignores the new-container file attributes entirely? That feels like something that _must_ be documented while we still remember.

mtrmac (commented):
It’s tempting to suggest that if we are risk-averse and backporting this, adding _only_ a PR for the new `enable_partial_images` option now would be better; then we can add the new feature after the release.

---

Mostly non-blocking stylistic options, but see at least the `err == nil && found` and `/proc/self/fd` comments.

pkg/chunked/storage_linux.go
```diff
@@ -54,14 +55,29 @@ func timeToTimespec(time time.Time) (ts unix.Timespec) {
...
+func copyFileContent(src, destFile, root string, dirfd int, missingDirsMode, mode os.FileMode, useHardLinks bool) (*os.File, int64, error) {
 	st, err := os.Stat(src)
 	if err != nil {
 		return nil, -1, err
 	}
 
 	copyWithFileRange, copyWithFileClone := true, true
 
+	if useHardLinks {
+		fd, err := strconv.Atoi(strings.Replace(src, "/proc/self/fd/", "", 1))
```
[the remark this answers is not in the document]
mtrmac: It would even be more efficient, e.g. an `os.Fstat` instead of `os.Stat` just above.

giuseppe:
@mtrmac I've moved the patch to disable partial pulls here: https://github.com/containers/storage/pull/998

I'll re-work the other ones to address your comments

rhatdan:
@mtrmac PTAL

pkg/chunked/storage_linux.go
```diff
@@ -54,16 +54,41 @@ func timeToTimespec(time time.Time) (ts unix.Timespec) {
 	return unix.NsecToTimespec(time.UnixNano())
 }
 
-func copyFileContent(src, destFile, root string, dirfd int, missingDirsMode, mode os.FileMode) (*os.File, int64, error) {
+func copyFileContent(srcFd int, destFile, root string, dirfd int, mode os.FileMode, useHardLinks bool) (*os.File, int64, error) {
```
mtrmac, line 57: Isn’t `root` now unused? And that might propagate up through the call stack.

pkg/chunked/storage_linux.go
```diff
@@ -161,27 +196,12 @@ func findFileInOtherLayers(file internal.ZstdFileMetadata, root string, dirfd in
...
-		}
-		defer srcFile.Close()
-
-		srcPath := fmt.Sprintf("/proc/self/fd/%d", srcFile.Fd())
-
-		dstFile, written, err := copyFileContent(srcPath, file.Name, root, dirfd, missingDirsMode, 0)
-		if err != nil {
-			continue
+		found, dstFile, written, err := copyFileFromOtherLayer(file, source, m, root, dirfd, missingDirsMode, useHardLinks)
+		if found {
```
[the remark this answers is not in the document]
mtrmac: Still outstanding AFAICS. Am I incorrect about how this works?

pkg/chunked/storage_linux.go
```diff
@@ -640,6 +694,13 @@ type hardLinkToCreate struct {
 	metadata *internal.ZstdFileMetadata
 }
 
+func parseBooleanPullOption(storeOpts *storage.StoreOptions, name string, def bool) bool {
+	if value := storeOpts.PullOptions[name]; strings.ToLower(value) == "true" {
+		return true
```
mtrmac, line 699: Shouldn’t this be `if value, ok := … { return strings.ToLower(value) == "true" }` so that users can explicitly set options to `false`?

(Ultimately, ideally this should probably fail on typos. Or parse the options at store creation time, and fail then? But that’s expanding the scope quite a bit.)

pkg/chunked/storage_linux.go
mtrmac: Still outstanding I think.

giuseppe:
thanks for your review and to spot all these issues.  I've pushed a new version, but I'll go again through the patchset tomorrow

mtrmac (commented):
ACK — looks good to me purely reading the local changes. Thanks!

I don’t understand the intricacies of c/storage (or even what they are), so a review by an expert in that would probably still be useful.

PR #998, opened 2021-08-17
chunked: allow to disable partial images feature

giuseppe:
enable partial pulls only when it is explicitely configured in the
storage.conf file:

[storage.options]

pull_options = {enable_partial_images = "true"}

This is to prevent the experimental feature to leak into CRI-O.

The default value will change in future once the feature is stable.

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -654,8 +654,14 @@ func (d *chunkedZstdDiffer) ApplyDiff(dest string, options *archive.TarOptions) 
 		return output, err
 	}
 
+	const trueVal = "true"
+
+	if value := storeOpts.PullOptions["enable_partial_images"]; strings.ToLower(value) != trueVal {
+		return output, errors.New("enable_partial_images not configured")
+	}
+
 	enableHostDedup := false
-	if value := storeOpts.PullOptions["enable_host_deduplication"]; strings.ToLower(value) == "true" {
+	if value := storeOpts.PullOptions["enable_host_deduplication"]; strings.ToLower(value) == trueVal {
 		enableHostDedup = true
 	}
 


vrothberg (commented):
LGTM

rhatdan:
LGTM

Romain-Geissler-1A:
Hi @giuseppe,

You merged this more or less 6 months ago. Is partial pull still something considered experimental for now ?

Cheers,
Romain

giuseppe:
Hi Romain, yes this is still considered experimental

Romain-Geissler-1A:
Do you have already in mind a roadmap on when you will consider this fully supported ? Will this happen in 2022, or you don't know yet ?

I wish to use this on all our development machines in Amadeus, yet, before doing this, I need Docker to support zstd, which is not the case right now. FYI, I have just proposed a backport of your Docker commit in Docker 20.10 in https://github.com/moby/moby/pull/43331. I have no idea if it will be accepted, but trying costs nothing ;)

giuseppe:
"fully supported" on RHEL will take some time and it depends on how well it works upstream first.  We are currently blocked on quay.io to support HTTP range requests, so we cannot even test it at the moment.

As far as the open source project is concerned: this feature is already present so please play with it, and report how it performs for you.  It is far easier to change things now that it is still experimental and we can change the format than it would be later on when it will hopefully land in RHEL.

Romain-Geissler-1A:
Well, unfortunately I can't start using this until Docker supports it. The latest comment on Docker side doesn't make me feel confident this will ever be backported to 20.10, and on the other side, Docker 21.x (or shall this be named 22.x in 2022 ?) isn't released yet, so... I am blocked (we still heavily rely on Docker on Amadeus side).

rhatdan:
You could always just use Podman...

Romain-Geissler-1A:
I definitely wish podman becomes one day the de-facto standard for container hacking rather than docker, but we both know that's not the state of the industry right now ;) But I see podman being mentioned/supported in more and more places (both in the open source projects and internally in our own proprietary git repos), so hopefully it's just a matter of time till podman takes over docker as the de-facto standard.

Note: I am in the very low layers of what Amadeus produces. If I decide to use zstd:chunked in the images I produces, a big majority of the Amadeus teams will have to stop using docker from one day to another. I guess you don't ask everyday to all your Red Hat clients to stop using RHEL 6 and RHEL 7 so that they would use exclusively RHEL 8, while I am sure you would definitely like if that was possible (and to be honest, I would be glad if Red Hat asked this to Amadeus, that would push us to be more up to date !) ;)

PR #1001, opened 2021-08-20
chunked: estargz support

giuseppe:
in addition to zstd:chunked, add support for the estargz format.

estargz is maintained at github.com/containerd/stargz-snapshotter

Images using estargz can be used on old clients and registries that have no support for the zstd compression algorithm.


rhatdan:
@ktock PTAL

ktock (commented):
Thank you for this PR. Overall LGTM and works with docker.io.

off-topic: chunked pull doesn't seems work on ghcr.io with the following error? (can be addressed in the following-up PRs)

```
DEBU[0004] Failed to retrieve partial blob: invalid status code returned when fetching blob 200 (OK) 
```

giuseppe:
> `DEBU[0004] Failed to retrieve partial blob: invalid status code returned when fetching blob 200 (OK) `

yes, I saw that failure as well.

I need to handle the 200 return code differently (it seems that ghcr.io turns the partial request into a 200 when requesting too much of the original file)

giuseppe:
> > ``
> 
> yes, I saw that failure as well.
> 
> I need to handle the 200 return code differently (it seems that ghcr.io turns the partial request into a 200 when requesting too much of the original file)

I cannot see this issue anymore, as ghcr.io returns the correct 206 status code.  Could you please give it another try and see if something changed for you as well?

pkg/chunked/compressor/compressor.go
```diff
@@ -218,3 +222,89 @@ func ZstdCompressor(r io.Writer, metadata map[string]string, level *int) (io.Wri
...
+		return err
+	}
+	sr := io.NewSectionReader(c.file, 0, st.Size())
+
+	compressionLevel := pgzip.BestCompression
+	if c.compressionLevel != nil {
+		compressionLevel = *c.compressionLevel
+	}
+
+	blob, err := estargz.Build(sr, estargz.WithCompressionLevel(compressionLevel))
```
ktock, line 250: estargz changes DiffID(uncompressed digest) of the layer because it adds TOC entry. Could we propagate the converted digest to the callers?

ktock:
> I cannot see this issue anymore, as ghcr.io returns the correct 206 status code. Could you please give it another try and see if something changed for you as well?

Thank you for fixing this. Chunked pull works with ghcr.io now.

giuseppe:
> > 
> 
> Thank you for fixing this. Chunked pull works with ghcr.io now.

I've not done anything :) I could still see the behavior though for some layers, I am working on a fix for containers/image to handle a 200 status.

giuseppe:
do you think it could be helpful to add another annotation (like zstd:chunked does) to store the ToC offset so to avoid a round trip to read it?

ktock:
> do you think it could be helpful to add another annotation (like zstd:chunked does) to store the ToC offset so to avoid a round trip to read it?

Yes I do :)

pkg/chunked/compressor/compressor.go
mtrmac: It’s an extensive assumption of the current c/image/copy structure that compression/decompression/encryption/… doesn’t change DiffID values. Propagating this isn’t _impossible_ but it will touch a large part of the public API, probably triggering discussions about hiding some of the API, and so on…

mtrmac: In particular we definitely don’t want to add `c/image/types.BlobInfo.DiffID`; that historical abuse of the transport API for internal needs of the copy implementation needs to stop at some point, so let’s stop now.

giuseppe:
ready for review

ktock:
@giuseppe 

It seems to create wrong `diff_ids`? (maybe related to https://github.com/containers/storage/pull/1001#discussion_r693923431)

```console
$ skopeo copy -f oci --dest-compress-format gzip:estargz  docker://ghcr.io/stargz-containers/alpine:3.10.2-org docker://ghcr.io/ktock/myalpine:3.10.2-esgz
$ skopeo inspect --raw docker://ghcr.io/ktock/myalpine:3.10.2-esgz | jq
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "digest": "sha256:1f97f0559cbddbff6c872039e93f18c1abdc279cbe82e0eb40258c28f4c30bfd",
    "size": 585
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:09f11a52de5e1e41b8b7f226170df3a9b32881f13e1a330a3dcc344b576911c0",
      "size": 2803990,
      "annotations": {
        "containerd.io/snapshot/stargz/toc.digest": "sha256:2a5b40d9dac53b0ba42c6995648c5a3bb9d71d376c66123facc25d5a7b31957f",
        "io.containers.estargz.uncompressed-size": "5935104"
      }
    }
  ]
}
$ skopeo inspect --config --raw docker://ghcr.io/ktock/myalpine:3.10.2-esgz | jq '.rootfs.diff_ids'
[
  "sha256:03901b4a2ea88eeaad62dbe59b072b28b6efa00491962b8741081c5df50c65e0"
]
$ crane blob ghcr.io/ktock/myalpine:3.10.2-esgz@sha256:09f11a52de5e1e41b8b7f226170df3a9b32881f13e1a330a3dcc344b576911c0 | gunzip | sha256sum
b880ce78f96e260a9e030fe94a8c9070199bb7a20d5e86c221ceeea2c999e00b  -
```

`b880ce78f96e260a9e030fe94a8c9070199bb7a20d5e86c221ceeea2c999e00b` (sha256) doesn't match to any in `.rootfs.diff_ids`.

giuseppe:
@rhatdan @vrothberg @mtrmac any suggestions on how can we solve the issue above?

Should we drop the estargz build part for now?

mtrmac:
> @rhatdan @vrothberg @mtrmac any suggestions on how can we solve the issue above?

Well, build that capability? In addition to the API overhaul ~implied above, personally I’d say it’s mostly blocked on building tests for c/image/copy, the codebase is growing in size and especially special cases so much that it’s becoming untenable without tests; I’m working on them as time allows but that hasn’t been very fast recently.

> Should we drop the estargz build part for now?

Splitting the feature into two is an option, sure.

giuseppe:
I've dropped the compressor part until we figure out how it must be supported.

pkg/chunked/compression.go
```diff
@@ -119,47 +227,47 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, ann
...
 	if err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 	defer decoder.Close()
 
 	b := make([]byte, 0, lengthUncompressed)
 	if decoded, err := decoder.DecodeAll(manifest, b); err == nil {
-		return decoded, nil
+		return decoded, 0, nil
```
ktock, line 267: `return decoded, int64(offset), nil` ?

rhatdan:
@ktock @mtrmac PTAL

pkg/chunked/storage_linux.go
```diff
@@ -413,18 +449,46 @@ func createFileFromZstdStream(dest string, dirfd int, reader io.Reader, mode os.
...
+			}
+		}
+		defer c.gzipReader.Close()
+
+		if _, err := io.Copy(to, io.LimitReader(c.gzipReader, metadata.Size)); err != nil {
+			return err
+		}
+		if _, err := io.Copy(ioutil.Discard, c.gzipReader); err != nil {
+			return err
+		}
```
ktock, line 487: Could we do `io.Copy(ioutil.Discard, reader)` instead of `c.gzipReader` so that we can avoid any parse errors that occurred from this discarding part?

giuseppe:
@ktock thanks for the review.  Addressed the comments

pkg/chunked/compression.go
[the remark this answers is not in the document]
giuseppe: sorry, I've messed up with the git commit.  Should be fine now

ktock (approved):
Thank you! LGTM

PR #1002, opened 2021-08-20
Have NewReadCloserWrapper pass through io.WriterTo

mtrmac:
If the underlying `io.Reader` implements `WriteTo`, make the wrapped `ReadCloser` implement it as well.

This causes the values returned by `pkg/archive.DecompressStream` to implement `WriteTo` (for gzip and zstd, not bzip2 and xz), and that in turn eliminates a 32kB memory allocation when decompressing layers during pull.

### DIFF
--- a/pkg/ioutils/readers.go
+++ b/pkg/ioutils/readers.go
@@ -17,8 +17,25 @@ func (r *readCloserWrapper) Close() error {
 	return r.closer()
 }
 
+type readWriteToCloserWrapper struct {
+	io.Reader
+	io.WriterTo
+	closer func() error
+}
+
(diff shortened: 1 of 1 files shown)


mrunalp:
@nalind ptal

nalind:
LGTM

TomSweeneyRedHat:
Changes LGTM, but overlay test is failing.  I'm not sure if it is or is not related to the change.

mtrmac:
It doesn’t immediately look related, I have just restarted the test for now.

vrothberg (commented):
Late LGTM, nice catch!

PR #1003, opened 2021-08-20
Add LayerOptions.OriginalDigest and LayerOptions.UncompressedDigest

mtrmac:
This allows callers of `Store.PutLayer` to provide the values if they have already computed them, so that `ApplyDiff` does not need to compute them again.

Intended to allow fixing the second half of https://github.com/containers/image/issues/1160 .

This could quite significantly reduce CPU usage.

(Note that it remains the case that during pulls, both `c/image/storage` and `ApplyDiff` decompress the stream; c/image/storage stores the compressed, not the decompressed, version in a temporary file. Nothing changes about that here - it's not obvious that changing it is worth it, and anyway it's a different concept for a different PR/discussion.)

<s>WARNING: Absolutely untested at this point.</s>

### DIFF
--- a/store.go
+++ b/store.go
@@ -547,6 +547,15 @@ type LayerOptions struct {
 	// initialize this layer.  If set, it should be a child of the layer
 	// which we want to use as the parent of the new layer.
 	TemplateLayer string
+	// OriginalDigest specifies a digest of the tarstream (diff), if one is
+	// provided along with these LayerOptions, and reliably known by the caller.
+	// Use the default "" if this fields is not applicable or the value is not known.
+	OriginalDigest digest.Digest
+	// UncompressedDigest specifies a digest of the uncompressed version (“DiffID”)
+	// of the tarstream (diff), if one is provided along with these LayerOptions,
+	// and reliably known by the caller.
+	// Use the default "" if this fields is not applicable or the value is not known.
+	UncompressedDigest digest.Digest
 }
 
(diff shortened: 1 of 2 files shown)


store.go
```diff
@@ -1031,20 +1040,21 @@ func (s *store) PutLayer(id, parent string, names []string, mountLabel string, w
 			gidMap = s.gidMap
 		}
 	}
-	var layerOptions *LayerOptions
+	layerOptions := LayerOptions{
```
mtrmac, line 1043: It’s not immediately obvious to me whether not passing original `options` (possibly including `TemplateLayer` down to `rlstore.Put` was intentional or not; this preserves the current behavior.

mtrmac:
```
[+0331s]         	Error Trace:	testutil_unix.go:41
[+0331s]         	            				graphtest_unix.go:175
[+0331s]         	            				devmapper_test.go:90
[+0331s]         	Error:      	Received unexpected error:
[+0331s]         	            	devmapper: device Base already exists. Deleted=true
[+0331s]         	Test:       	TestDevmapperCreateFromTemplate
```
again.

mtrmac:
Somewhat tested now, and see https://github.com/containers/image/pull/1352 for measurements; ready for review.

rhatdan:
LGTM
@nalind @giuseppe PTAL

giuseppe (approved):
LGTM

mrunalp:
This is great. Thanks! Can we get this into a new release?

PR #1009, opened 2021-08-28
Add Inodes to OverlayOptionsConfig

QiWang19:
Fix error "Failed to decode the keys inodes" when using inodes in storage.conf
as documented https://github.com/containers/storage/blob/main/storage.conf#L73
https://github.com/containers/storage/blob/c9f1d016b9e7692e70b1338045af61c89fd63a0b/types/options.go#L280

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -93,6 +93,8 @@ type OverlayOptionsConfig struct {
 	MountProgram string `toml:"mount_program"`
 	// Size
 	Size string `toml:"size"`
+	// Inodes is used to set a maximum inodes of the container image.
+	Inodes string `toml:"inodes"`
 	// Do not create a bind mount on the storage home
 	SkipMountHome string `toml:"skip_mount_home"`
 	// ForceMask indicates the permissions mask (e.g. "0755") to use for new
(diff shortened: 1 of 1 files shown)


rhatdan (approved):
LGTM

rhatdan:
@saschagrunert @vrothberg @giuseppe @mrunalp @nalind PTAL

pkg/config/config.go
```diff
@@ -93,6 +93,8 @@ type OverlayOptionsConfig struct {
 	MountProgram string `toml:"mount_program"`
 	// Size
 	Size string `toml:"size"`
+	// Inodes is used to set a maximum inodes of the container image.
+	Inodes string `toml:"inodes"`
```
saschagrunert, line 97: Do we wanna indicate that it's a "maximum" by renaming to `MaxInodes` or `InodesMax`? OTOH right now it matches better to the overlay option `inodes=`

saschagrunert (approved):
LGTM

giuseppe (approved):
LGTM

I am not merging since there is still an open question in the review

pkg/config/config.go
QiWang19: @rhatdan WDYT about rename it with max mentioned?

rhatdan: I don't think people will be onfused, and we don't change size to MaxSize.

PR #1012, opened 2021-09-03
chunked: restrict dedup with hard links

giuseppe:
before deduplicating with hard links make sure the two files share the same UID, GID, file mode and extended attributes.

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -112,14 +112,14 @@ func copyFileContent(srcFd int, destFile string, dirfd int, mode os.FileMode, us
 	return dstFile, st.Size(), err
 }
 
-func prepareOtherLayersCache(layersMetadata map[string][]internal.FileMetadata) map[string]map[string]*internal.FileMetadata {
-	maps := make(map[string]map[string]*internal.FileMetadata)
+func prepareOtherLayersCache(layersMetadata map[string][]internal.FileMetadata) map[string]map[string][]*internal.FileMetadata {
+	maps := make(map[string]map[string][]*internal.FileMetadata)
 
 	for layerID, v := range layersMetadata {
-		r := make(map[string]*internal.FileMetadata)
+		r := make(map[string][]*internal.FileMetadata)
 		for i := range v {
 			if v[i].Digest != "" {
-				r[v[i].Digest] = &v[i]
+				r[v[i].Digest] = append(r[v[i].Digest], &v[i])
 			}
 		}
 		maps[layerID] = r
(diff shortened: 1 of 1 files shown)


rhatdan (approved):
LGTM

rhatdan:
@mtrmac @nalind PTAL

rhatdan:
@ktock @saschagrunert @vrothberg PTAL

pkg/chunked/storage_linux.go
```diff
@@ -235,30 +242,98 @@ func copyFileFromOtherLayer(file internal.FileMetadata, source string, otherFile
...
+}
+
 // findFileInOtherLayers finds the specified file in other layers.
 // file is the file to look for.
 // dirfd is an open file descriptor to the checkout root directory.
 // layersMetadata contains the metadata for each layer in the storage.
 // layersTarget maps each layer to its checkout on disk.
 // useHardLinks defines whether the deduplication can be performed using hard links.
-func findFileInOtherLayers(file internal.FileMetadata, dirfd int, layersMetadata map[string]map[string]*internal.FileMetadata, layersTarget map[string]string, useHardLinks bool) (bool, *os.File, int64, error) {
+func findFileInOtherLayers(file *internal.FileMetadata, dirfd int, layersMetadata map[string]map[string][]*internal.FileMetadata, layersTarget map[string]string, useHardLinks bool) (bool, *os.File, int64, error) {
```
ktock, line 310: No error seems returned from this func?

ktock (commented):
overall LGTM

vrothberg (commented):
LGTM

pkg/chunked/storage_linux.go
```diff
@@ -235,30 +242,98 @@ func copyFileFromOtherLayer(file internal.FileMetadata, source string, otherFile
...
+
+	path := fmt.Sprintf("/proc/self/fd/%d", fd)
+
+	listXattrs, err := system.Llistxattr(path)
+	if err != nil {
+		return false
+	}
+
+	xattrsToIgnore := map[string]interface{}{
+		"security.selinux": true,
```
mtrmac, line 279: Why is this {necessary, safe}?

rhatdan: I think @giuseppe is thinking that containers use a volume mount and do not rely on the label. But as @mtrmac this might actually break things, since a container might not be able to use the SELinux label in the lower.  The problem is going to be figuring out which labels can and can not be used.  If the label came from container storage it should be fine, and the check would not be necessary.  If the link came from the host system like in /usr, it probably would work.  If the link came from /var or other locations on the host system, it probably will not be readable within the confined container.

mtrmac: IIRC host reuse only uses `/usr`, so this might be fine — but having the full thinking documented in the code seems like something pretty likely to be useful 3–5 years from now, both to find out what the original motivation was, and so that future programmers and PRs that touch the same code notice that there’s something to think about.

This PR comment thread (hopefully we settle on a full definite story) can help recording what we were thinking, but a comment directly in code would help draw attention to the issue.

rhatdan: The question would be is can a user turn this off if it does cause problems?

mtrmac: “Problems?”

This seems to be to be the kind of thing where expert programmers should either determine that it works and that it is safe, or not have that feature; I find it unintuitive to treat this like performance tuning knob where there are many legitimate options and the best one is based on circumstances.

mtrmac: … and if we have no idea whether this is safe and works, why not just match the SELinux label exactly, and reuse files only in the case where we do have a good reason to think it is safe and works?

---

@giuseppe can you explain what you saw and how this works, please?

rhatdan: Well the content will never match if you are searching through /usr, versus /var/lib/containers/storage.

mtrmac: That’s a fair point. So what’s the property that we want (even if this _were_ a manual user option, what information would the user need to make a decision?), and the right way to check? A run-time read of the policy to see whether a specific operation is allowed? A build-time provided list of SELinux types that should be accepted, and compiling that list into the binary?

Looking at an old `/usr/bin` or  `/usr/libexec`, not just a `bin_t` or `lib_t` or `usr_t`; there are quite a few other types. What property are we looking for, and do they have that property?

rhatdan: $ sesearch -A -s container_t -t usr_t -c file -p read
allow domain base_ro_file_type:file { getattr ioctl lock open read };
allow svirt_sandbox_domain exec_type:file { entrypoint execute execute_no_trans getattr ioctl lock map open read };

$ seinfo -abase_ro_file_type -x

Type Attributes: 1
   attribute base_ro_file_type;
	bin_t
	boot_t
	etc_runtime_t
	etc_t
	lib_t
	shell_exec_t
	src_t
	system_conf_t
	system_db_t
	textrel_shlib_t
	usr_t

rhatdan: We could just hard code this list into opencontainers/selinux and then have storage use it.

rhatdan: Any other types we ignore.

mtrmac: Yup, that looks like a consistent story to (ignorant) me. Is the list of `base_ro_file_type` types relatively static? Or would we want to extract it at RPM build time? (OTOH that could potentially unexpectedly change the list on simple rebuilds.)

rhatdan: It has been static for years.

PR #1014, opened 2021-09-09
(*Store)Layer(): fix race when loading layers

vrothberg:
Make sure that the layer store is locked when loading the layers.

Reported-in: github.com/containers/podman/issues/11487

@nalind PTAL

### DIFF
--- a/store.go
+++ b/store.go
@@ -3054,10 +3054,15 @@ func (s *store) Layers() ([]Layer, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := lstore.LoadLocked(); err != nil {
-		return nil, err
-	}
-	layers, err := lstore.Layers()
+
+	layers, err := func() ([]Layer, error) {
+		lstore.Lock()
+		defer lstore.Unlock()
+		if err := lstore.Load(); err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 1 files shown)


store.go
```diff
@@ -3054,13 +3054,15 @@ func (s *store) Layers() ([]Layer, error) {
...
-	}
+
+	layers, err := func() ([]Layer, error) {
+		lstore.RLock()
+		defer lstore.Unlock()
+		if err := lstore.Load(); err != nil {
+			return nil, err
+		}
+		return lstore.Layers()
+	}()
```
nalind, line 3065: Looks like the check for `err` got lost here, otherwise LGTM.

vrothberg: Done. Also had to use a writer lock because of Load.

nalind: That was the only caller for LoadLocked().  We may want to try removing that at some point.

vrothberg: want me to drop a `TODO`?

nalind: Not a blocker, but if you need to revise this for any reason, that's probably a good idea.  Thanks!

vrothberg:
@rhatdan @giuseppe PTAL

rhatdan:
LGTM

PR #1018, opened 2021-09-10
store: don't ignore empty mappings for top layer

giuseppe:
do not pick the first available image if it was requested to not use
any mapping.

It causes a problem where an image is already present in the local
store with a mapping, then a container that doesn't specify any
mapping would end up to reuse that:

$ podman system reset -y
$ podman run --rm --uidmap 0:10000:1000 centos ls -ld /home
drwxr-xr-x. 2 root root 6 Nov  3  2020 /home
    
$ podman run --rm centos ls -ld /home
drwxr-xr-x. 2 10000 10000 6 Nov  3  2020 /home

### DIFF
--- a/cmd/containers-storage/container.go
+++ b/cmd/containers-storage/container.go
@@ -211,12 +211,8 @@ func containerParentOwners(flags *mflag.FlagSet, action string, m storage.Store,
 			json.NewEncoder(os.Stdout).Encode(mappings)
 		} else {
 			fmt.Printf("ID: %s\n", container.ID)
-			if len(uids) > 0 {
-				fmt.Printf("UIDs: %v\n", uids)
-			}
-			if len(gids) > 0 {
-				fmt.Printf("GIDs: %v\n", gids)
-			}
+			fmt.Printf("UIDs: %v\n", uids)
+			fmt.Printf("GIDs: %v\n", gids)
 		}
 	}
 	if len(matched) != len(args) {
--- a/store.go
+++ b/store.go
@@ -1131,10 +1131,6 @@ func (s *store) imageTopLayerForMapping(image *Image, ristore ROImageStore, crea
 		if options.HostGIDMapping && len(layer.GIDMap) != 0 {
 			return false
 		}
-		// If we don't care about the mapping, it's fine.
-		if len(options.UIDMap) == 0 && len(options.GIDMap) == 0 {
-			return true
-		}
 		// Compare the maps.
 		return reflect.DeepEqual(layer.UIDMap, options.UIDMap) && reflect.DeepEqual(layer.GIDMap, options.GIDMap)
 	}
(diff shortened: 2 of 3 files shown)


giuseppe:
@nalind PTAL.

This is a breaking change, but I could not understand why it was implemented this way.

nalind:
> do not pick the first available image if it was requested to not use
> any mapping.
> 
> It causes a problem where an image is already present in the local
> store with a mapping, then a container that doesn't specify any
> mapping would end up to reuse that:

Please elaborate on that.  Are callers not picking up the mappings that we set in the returned Container structure?  Are we not populating it correctly?

giuseppe:
I had messed up the commit message that includes a reproducer, fixed now:

```
$ podman system reset -y
$ podman run --rm --uidmap 0:10000:1000 centos ls -ld /home
drwxr-xr-x. 2 root root 6 Nov  3  2020 /home
    
$ podman run --rm centos ls -ld /home
drwxr-xr-x. 2 10000 10000 6 Nov  3  2020 /home
```

giuseppe:
> > 
> 
> Please elaborate on that. Are callers not picking up the mappings that we set in the returned Container structure? Are we not populating it correctly?

The issue is that if a mapped image already exists in the store, and then a new container is created that doesn't specify any mapping is created, then the existing mapped image is used.

nalind:
> The issue is that if a mapped image already exists in the store, and then a new container is created that doesn't specify any mapping is created, then the existing mapped image is used.

Right, because creating a mapped layer was costly both in terms of time and disk space used, and best avoided if the caller didn't express a preference about ID mappings.  The reproducer looks like it's showing podman ignoring the ID mappings that the storage library tells it the new container is set up to use, which I thought was required by the auto userns feature.

giuseppe:
>  The reproducer looks like it's showing podman ignoring the ID mappings that the storage library tells it the new container is set up to use, which I thought was required by the auto userns feature.





> Right, because creating a mapped layer was costly both in terms of time and disk space used, and best avoided if the caller didn't express a preference about ID mappings. The reproducer looks like it's showing podman ignoring the ID mappings that the storage library tells it the new container is set up to use, which I thought was required by the auto userns feature.

I think the expectation when there is no mapping specified is to run in the current user namespace, that is the same behavior we have when there is not already an image with a different mapping in the storage.  `--userns=auto` must be explicitly set

rhatdan:
I would assume the same, that if you don't specify a user namespace mapping, that the current host mapping would be used rather then the previously pulled imagees user namespace.

giuseppe:
FTR, @alexlarsson reported this issue to me

giuseppe:
@nalind @rhatdan so are you fine with the proposed fix?

nalind:
It'll do, but I'm still not cool with callers ignoring values we pass back to them as a practice.

rhatdan (approved):
LGTM

PR #1023, opened 2021-09-22
chunked: implement OSTree deduplication

giuseppe:
if the option ostree_repos is set:

    [storage.options]
    pull_options = {enable_partial_images = "true", ostree_repos = "/foo:/bar"}

then attempt to deduplicate from the specified list of OSTree repositories.

In order to be usable, an OSTree repository must be configured to track
the checksum for its files payload (payload link), that is disabled by
default:

ostree config --repo=/path/to/repo set core.payload-link-threshold N

Where N is the minimum size for files to be tracked by their payload
and must be a nonzero value.

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -345,6 +345,56 @@ func getFileDigest(f *os.File) (digest.Digest, error) {
 	return digester.Digest(), nil
 }
 
+// findFileInOSTreeRepos checks whether the requested file already exist in one of the OSTree repo and copies the file content from there if possible.
+// file is the file to look for.
+// ostreeRepos is a list of OSTree repos.
+// dirfd is an open fd to the destination checkout.
+// useHardLinks defines whether the deduplication can be performed using hard links.
+func findFileInOSTreeRepos(file *internal.FileMetadata, ostreeRepos []string, dirfd int, useHardLinks bool) (bool, *os.File, int64, error) {
+	digest, err := digest.Parse(file.Digest)
+	if err != nil {
+		return false, nil, 0, nil
+	}
+	payloadLink := digest.Encoded() + ".payload-link"
+	if len(payloadLink) < 2 {
+		return false, nil, 0, nil
+	}
+
+	for _, repo := range ostreeRepos {
+		sourceFile := filepath.Join(repo, "objects", payloadLink[:2], payloadLink[2:])
+		st, err := os.Stat(sourceFile)
+		if err != nil || !st.Mode().IsRegular() {
+			continue
+		}
+		if st.Size() != file.Size {
+			continue
+		}
+		fd, err := unix.Open(sourceFile, unix.O_RDONLY|unix.O_NONBLOCK, 0)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/chunked/storage_linux.go
```diff
@@ -1024,6 +1074,21 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions) (gra
 			continue
 		}
 
+		found, dstFile, _, err = findFileInOSTreeRepos(&r, ostreeRepos, dirfd, useHardLinks)
```
rhatdan, line 1077: Can found be false and dstFIle != nil?  Or found be true and dstFile be nil?
If not then the function should only return dstFile.

giuseppe: yes, it can return `found == true`, and `dstNil == nil` when doing hard links deduplication so the destination file is not overwritten.

rhatdan: How about found ==false dstNil != nil?

giuseppe: no that is not possible.  If the file is not found then the destination is not opened

rhatdan: It would make this clearer if it said
```
if found {
	if dstFile != nil {
			if err := setFileAttrs(dstFile, mode, &r, options); err != nil {
				dstFile.Close()
				return output, err
			}
			dstFile.Close()
		}
		
		continue
}
```

giuseppe: I've simplified it further and pushed a new version

rhatdan (approved):
LGTM

rhatdan:
@vrothberg @nalind @saschagrunert @flouthoc @umohnani8 @mtrmac PTAL

pkg/chunked/storage_linux.go
```diff
@@ -873,6 +923,9 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions) (gra
 	// modifies the source file as well.
 	useHardLinks := parseBooleanPullOption(&storeOpts, "use_hard_links", false)
 
+	// List of OSTree repositories to use for deduplication
+	ostreeRepos := strings.Split(storeOpts.PullOptions["ostree_repos"], ":")
```
flouthoc, line 927: @giuseppe I might be wrong should the convention be `,` instead of `:` we are using `:` at various places for extra mount args

flouthoc: Please ignore if multiple repos are generally seperated with `:` in `storage.conf`

giuseppe: `:` is often used as a path separator on unix, e.g. for `PATH`, and golang defines `os.PathListSeparator` as well as `:`.

I am fine to change it though if it is what we use in other places

flouthoc: @giuseppe thanks for clarifying i think `:` is better its a generally accepted convention

flouthoc:
LGTM

PR #1024, opened 2021-09-23
overlay: support project quota with partial pulls

giuseppe:
When quotas are enabled, partial pulls currently fail with an error like:

Error: committing the finished image: rename /var/lib/containers/storage/overlay/staging/012887766 /var/lib/containers/storage/overlay/b9b6a85242afa733a717aa45d7c0d80a444e224c96fb46677453aeef44caa0d6/diff: invalid cross-device link

Assign a project ID to the staging directory and copy it back to the target directory if the EXDEV error happens.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1631,6 +1631,13 @@ func (d *Driver) ApplyDiffWithDiffer(id, parent string, options *graphdriver.App
 		}
 	}
 
+	if d.quotaCtl != nil {
+		quota := quota.Quota{}
+		if err := d.quotaCtl.SetQuota(applyDir, quota); err != nil {
+			return graphdriver.DriverWithDifferOutput{}, err
+		}
+	}
+
 	logrus.Debugf("Applying differ in %s", applyDir)
 
 	out, err := differ.ApplyDiff(applyDir, &archive.TarOptions{
@@ -1657,7 +1664,16 @@ func (d *Driver) ApplyDiffFromStagingDirectory(id, parent, stagingDirectory stri
 	if err := os.RemoveAll(diff); err != nil && !os.IsNotExist(err) {
 		return err
 	}
-	return os.Rename(stagingDirectory, diff)
+
+	err = os.Rename(stagingDirectory, diff)
+	if err != nil && errors.Is(err, unix.EXDEV) {
+		// If we got EXDEV it might mean the two directories have a different quota projectid.  Copy
+		// the project ID from the staging directory to the destination and try again.
+		if errQuota := quota.CopyProjectID(stagingDirectory, filepath.Dir(diff)); errQuota == nil {
+			err = os.Rename(stagingDirectory, diff)
+		}
+	}
+	return err
 }
 
 // DifferTarget gets the location where files are stored for the layer.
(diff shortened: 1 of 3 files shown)


rhatdan (approved):
Nice catch
LGTM

rhatdan:
@nalind @TomSweeneyRedHat @flouthoc @vrothberg @mtrmac PTAL

drivers/overlay/overlay.go
```diff
@@ -1657,7 +1664,15 @@ func (d *Driver) ApplyDiffFromStagingDirectory(id, parent, stagingDirectory stri
 	if err := os.RemoveAll(diff); err != nil && !os.IsNotExist(err) {
 		return err
 	}
-	return os.Rename(stagingDirectory, diff)
+
+	err = os.Rename(stagingDirectory, diff)
+	if err != nil && errors.Is(err, unix.EXDEV) {
+		// If we got EXDEV, copy the project quota from the staging directory to the target path and try again.
```
mtrmac, line 1670: I think a comment like “EXDEV might mean that we failed because $reason, so $fix_that_reason” would be much more useful than just a comment trivially re-stating what the code below does.

drivers/overlay/overlay.go
```diff
@@ -1631,6 +1631,13 @@ func (d *Driver) ApplyDiffWithDiffer(id, parent string, options *graphdriver.App
 		}
 	}
 
+	if d.quotaCtl != nil {
+		quota := quota.Quota{}
```
mtrmac, line 1635: Why is it OK to use empty settings here?

mtrmac (commented):
How does quota _enforcement_ actually work with this PR?

I’d expect the staging directory creation to already be subject to quota; this doesn’t seem to do that.

mtrmac:
BTW it would be nice to have quota tests; if not in this repo, to extend the tests in Podman.

giuseppe:
> I’d expect the staging directory creation to already be subject to quota; this doesn’t seem to do that.

AFAICS we are not currently using any quota setting when creating images since `options` https://github.com/containers/storage/blob/main/store.go#L1057 is hardcoded to nil.

@rhatdan is setting a project id for the image layers a desired behavior or do we want to track/set limits only for containers?

mtrmac:
> > I’d expect the staging directory creation to already be subject to quota; this doesn’t seem to do that.
> 
> AFAICS we are not currently using any quota setting when creating images since `options` https://github.com/containers/storage/blob/main/store.go#L1057 is hardcoded to nil.

Oh, I didn’t know that. Still, I’d expect that hard-coding to exist at _one_ level of the call stack, ideally in one location. If the drivers are expected to enforce quota, they should enforce it consistently; if they are expected to ignore it, they should ignore it consistently.

giuseppe:
another issue there is with quota for images is that it is not possible to do any deduplication since each image has a different projid.

rhatdan:
I don't believe OpenShift controls based on the images pulled, but @mrunalp would know better then me.
Using Quota to get the size quickly might be useful though.

rhatdan:
I think we should avoid it for images.

giuseppe:
> I think we should avoid it for images.

@mrunalp @haircommander are you fine if we drop quota for images or does it break anything?

giuseppe:
checked with @mrunalp and @haircommander that this is not used by CRI-O at the moment.

Alternative fix: https://github.com/containers/storage/pull/1026

PR #1025, opened 2021-09-23
Standardize on capatalized logrus messages, and remove stutters

rhatdan:

### DIFF
--- a/drivers/fsdiff.go
+++ b/drivers/fsdiff.go
@@ -180,7 +180,7 @@ func (gdw *NaiveDiffDriver) ApplyDiff(id, parent string, options ApplyDiffOpts) 
 	start := time.Now().UTC()
 	logrus.Debug("Start untar layer")
 	if size, err = ApplyUncompressedLayer(layerFs, options.Diff, tarOptions); err != nil {
-		logrus.Errorf("Error while applying layer: %s", err)
+		logrus.Errorf("While applying layer: %s", err)
 		return
 	}
 	logrus.Debugf("Untar time: %vs", time.Now().UTC().Sub(start).Seconds())
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -174,21 +174,21 @@ func checkSupportVolatile(home, runhome string) (bool, error) {
 	var usingVolatile bool
 	if err == nil {
 		if volatileCacheResult {
-			logrus.Debugf("cached value indicated that volatile is being used")
+			logrus.Debugf("Cached value indicated that volatile is being used")
 		} else {
-			logrus.Debugf("cached value indicated that volatile is not being used")
+			logrus.Debugf("Cached value indicated that volatile is not being used")
 		}
 		usingVolatile = volatileCacheResult
 	} else {
 		usingVolatile, err = doesVolatile(home)
 		if err == nil {
 			if usingVolatile {
-				logrus.Debugf("overlay test mount indicated that volatile is being used")
+				logrus.Debugf("overlay: test mount indicated that volatile is being used")
 			} else {
-				logrus.Debugf("overlay test mount indicated that volatile is not being used")
+				logrus.Debugf("overlay: test mount indicated that volatile is not being used")
 			}
 			if err = cachedFeatureRecord(runhome, feature, usingVolatile, ""); err != nil {
-				return false, errors.Wrap(err, "error recording volatile-being-used status")
+				return false, errors.Wrap(err, "recording volatile-being-used status")
 			}
 		}
 	}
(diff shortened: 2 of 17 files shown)


rhatdan:
@vrothberg @nalind @TomSweeneyRedHat @giuseppe @flouthoc PTAL

drivers/fsdiff.go
```diff
@@ -180,7 +180,7 @@ func (gdw *NaiveDiffDriver) ApplyDiff(id, parent string, options ApplyDiffOpts)
 	start := time.Now().UTC()
 	logrus.Debug("Start untar layer")
 	if size, err = ApplyUncompressedLayer(layerFs, options.Diff, tarOptions); err != nil {
-		logrus.Errorf("Error while applying layer: %s", err)
+		logrus.Errorf("While applying layer: %s", err)
```
vrothberg, line 183:
```suggestion
		logrus.Errorf("Applying layer: %s", err)
```

drivers/overlay/overlay.go
```diff
@@ -1038,7 +1038,7 @@ func (d *Driver) getLowerDirs(id string) ([]string, error) {
 				if os.IsNotExist(err) {
 					logrus.Warnf("Can't read link %q because it does not exist. A storage corruption might have occurred, attempting to recreate the missing symlinks. It might be best wipe the storage to avoid further errors due to storage corruption.", lower)
 					if err := d.recreateSymlinks(); err != nil {
-						return nil, fmt.Errorf("error recreating the missing symlinks: %v", err)
+						return nil, fmt.Errorf("recreating the missing symlinks: %v", err)
```
vrothberg, line 1041:
```suggestion
						return nil, fmt.Errorf("recreating the missing symlinks: %w", err)
```

`%w` will preserve the error type.

drivers/overlay/overlay.go
```diff
@@ -1121,7 +1121,7 @@ func (d *Driver) recreateSymlinks() error {
 	// List all the directories under the home directory
 	dirs, err := ioutil.ReadDir(d.home)
 	if err != nil {
-		return fmt.Errorf("error reading driver home directory %q: %v", d.home, err)
+		return fmt.Errorf("reading driver home directory %q: %v", d.home, err)
```
vrothberg, line 1124:
```suggestion
		return fmt.Errorf("reading driver home directory %q: %w", d.home, err)
```

`%w` will preserve the error type.

vrothberg: There are many other cases below.  Maybe mgrating over to the new go-native error handling should be a different PR?

vrothberg (commented):
Just mini nits, but LGTM.

Shall we also migrate over to native go-native error handling (fmt.Error() + errors.Is()) in another PR? Not sure if that would impact the API or not. @mtrmac and I were chatting about that for c/image not so long ago.

drivers/overlay/overlay.go
rhatdan: Lets do that in a different PR.

TomSweeneyRedHat:
@rhatdan lint is shaking its fist at you on this one.

pkg/fileutils/fileutils_unix.go
```diff
@@ -14,7 +14,7 @@ import (
 // reading it via /proc filesystem.
 func GetTotalUsedFds() int {
 	if fds, err := ioutil.ReadDir(fmt.Sprintf("/proc/%d/fd", os.Getpid())); err != nil {
-		logrus.Errorf("Error opening /proc/%d/fd: %s", os.Getpid(), err)
```
flouthoc, line 17: This should be
```diff
+ logrus.Errorf("Error opening /proc/%d/fd: %v", os.Getpid(), err)
```

vrothberg: That would (re)introduce stutter since the path is already part of `err`.

pkg/fileutils/fileutils_unix.go
```diff
@@ -14,7 +14,7 @@ import (
...
+		logrus.Errorf(err)
```
TomSweeneyRedHat, line 17:
```suggestion
		logrus.Errorf("%v", err)
```
I think will make lint happy

pkg/fileutils/fileutils_unix.go
flouthoc: @vrothberg ah my bad if comment is confusing PR does a `%s` for err just wanted to point we must use `%v` or `%w`.

mtrmac:
> Just mini nits, but LGTM.
> 
> Shall we also migrate over to native go-native error handling (fmt.Error() + errors.Is()) in another PR? Not sure if that would impact the API or not. @mtrmac and I were chatting about that for c/image not so long ago.

`errors.Is` can handle `pkg/errors.Wrapf` values, but `pkg/errors.Cause` can’t handle `fmt.Errorf("%w")` values. So every migration can silently break callers; I’d caution against mass-migrating without reviewing every single caller.

For c/image we did agree on using `Errorf("%w")` in _new_ code, but the necessary careful migration of the old code just doesn’t _have_ to happen (the migration typically doesn’t have a user-visible benefit — `Errorf("%w")` is IIRC about 3x faster, but who is ever going to notice on a failure path?), so it’s way down on the priority list.

rhatdan:
@vrothberg @giuseppe @TomSweeneyRedHat @flouthoc @nalind PTAL

flouthoc:
LGTM , if  `%v` -> `%w` conversion is out of scope for this PR

vrothberg (commented):
LGTM

PR #1026, opened 2021-09-23
overlay: disable quota for images

giuseppe:
When quotas are enabled, partial pulls currently fail with an error like:

Error: committing the finished image: rename /var/lib/containers/storage/overlay/staging/012887766 /var/lib/containers/storage/overlay/b9b6a85242afa733a717aa45d7c0d80a444e224c96fb46677453aeef44caa0d6/diff: invalid cross-device link

Disable quotas when creating a read-only layer, which is also the
expectation from the existing code in Create() and from the
documentation:

**size**=""
  Maximum size of a read/write layer.   This flag can be used to set quota on the size of a read/write layer of a container. (format: <number>[<unit>], where unit = b (bytes), k (kilobytes), m (megabytes), or g (gigabytes))

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -821,7 +821,7 @@ func (d *Driver) CreateReadWrite(id, parent string, opts *graphdriver.CreateOpts
 		opts.StorageOpt["inodes"] = strconv.FormatUint(d.options.quota.Inodes, 10)
 	}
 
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
ACK, this looks consistent. (See #1024 for an earlier discussion.)

rhatdan (approved):
LGTM

PR #1028, opened 2021-09-29
overlay: use xattrs perms with network FSs

giuseppe:
when running in a user namespace on a network file system, automatically switch on the permission xattrs feature so that users
won't have to tweak their storage.conf file with difficult to guess options.
    
The list of remote file systems was taken from coreutils stat:

https://git.savannah.gnu.org/gitweb/?p=coreutils.git;a=blob;f=src/stat.c;h=0c34501e34410b9096edf377654b9ff27fe9fa24;hb=HEAD#l260

Also fix an issue with "silly renames" on NFS, more details in the commit log.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -248,6 +248,23 @@ func (d *Driver) getSupportsVolatile() (bool, error) {
 	return supportsVolatile, nil
 }
 
+// isNetworkFileSystem checks if the specified file system is supported by native overlay
+// as backing store when running in a user namespace.
+func isNetworkFileSystem(fsMagic graphdriver.FsMagic) bool {
+	switch fsMagic {
+	// a bunch of network file systems...
+	case graphdriver.FsMagicNfsFs, graphdriver.FsMagicSmbFs, graphdriver.FsMagicAcfs,
+		graphdriver.FsMagicAfs, graphdriver.FsMagicCephFs, graphdriver.FsMagicCIFS,
+		graphdriver.FsMagicFHGFSFs, graphdriver.FsMagicGPFS, graphdriver.FsMagicIBRIX,
+		graphdriver.FsMagicKAFS, graphdriver.FsMagicLUSTRE, graphdriver.FsMagicNCP,
+		graphdriver.FsMagicNFSD, graphdriver.FsMagicOCFS2, graphdriver.FsMagicPANFS,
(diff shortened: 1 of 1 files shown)


store.go
```diff
@@ -2498,7 +2499,15 @@ func (s *store) DeleteContainer(id string) error {
 			gcpath := filepath.Join(s.GraphRoot(), middleDir, container.ID)
 			wg.Add(1)
 			go func() {
-				errChan <- os.RemoveAll(gcpath)
+				var err error
+				for attempts := 0; attempts < 50; attempts++ {
+					err = os.RemoveAll(gcpath)
+					if err == nil || !errors.Is(err, unix.EBUSY) {
```
flouthoc, line 2505: I think we need move `erros.Is(err, unix.EBUSY)` with a wrapper to unix specific package with `// +build linux freebsd darwin` preferably https://github.com/containers/storage/blob/main/pkg/system/syscall_unix.go and return no-op for non-unix.

drivers/overlay/overlay.go
```diff
@@ -266,18 +283,27 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
 	}
 
 	if opts.mountProgram != "" {
+		if unshare.IsRootless() && isNetworkFileSystem(fsMagic) && opts.forceMask == nil {
+			m := os.FileMode(0700)
+			opts.forceMask = &m
+			logrus.Debugf("Network file system detected as backing store.  Enforcing overlay option `force_mask=\"%o\"`.", m)
```
rhatdan, line 289: This should be at the Info level, or should it be Warn?  Would it be better for users to configure this themselves within the storage.conf, and warn them that we did it automatically.

giuseppe: I had it at Warn level first, but seemed annoying.  I can switch it back to Warnf

giuseppe: reverted to Warning

store.go
giuseppe: thanks, fixed now

rhatdan:
We can see if the Warnf is too agravating, or is easy enough to fix.

rhatdan (approved):
LGTM

rhatdan:
@flouthoc @saschagrunert @nalind @TomSweeneyRedHat @ashley-cui @umohnani8 PTAL

flouthoc (approved):

flouthoc:
LGTM, Restarted failing test but not sure if it was a flake.

PR #1030, opened 2021-09-29
Create readonly locks if the lockfile does not exists

rhatdan:
Users are setting up readonly shares and sometimes they forget, or do
not even know that they need to create the lockfile.

Even in the quay.io/podman/stable images Dockerfile, I have to manually
create these files.

Fixes: https://github.com/containers/storage/issues/1029

### DIFF
--- a/pkg/lockfile/lockfile_unix.go
+++ b/pkg/lockfile/lockfile_unix.go
@@ -36,7 +36,7 @@ type lockfile struct {
 // necessary.
 func openLock(path string, ro bool) (fd int, err error) {
 	if ro {
-		fd, err = unix.Open(path, os.O_RDONLY|unix.O_CLOEXEC, 0)
+		fd, err = unix.Open(path, os.O_RDONLY|unix.O_CLOEXEC|os.O_CREATE, 0)
 	} else {
 		fd, err = unix.Open(path,
 			os.O_RDWR|unix.O_CLOEXEC|os.O_CREATE,


rhatdan:
@vrothberg @giuseppe @nalind @TomSweeneyRedHat @flouthoc @umohnani8 @saschagrunert @ashley-cui PTAL

vrothberg (approved):
LGTM but want a head nod from @nalind

nalind:
If the underlying filesystem is mounted read-only, we're not going to be able to create lock files on it, but otherwise, sure.

rhatdan:
Well in that case we will return a more clear error.

flouthoc:
LGTM ,   I might be completely wrong but since we are trying to create file in a `readonly` fs so i doubt if this solves the reported issue since this would just report different error `unable to create: permission denied` to the user however user wants if imagestore is empty and readonly podman should not try to acquire lock and simply bypass.

flouthoc:
I am assuming `ro` here stands for readonly https://github.com/containers/storage/blob/main/pkg/lockfile/lockfile_unix.go#L38

flouthoc:
LGTM, nvm that should be only when underlying FS is readonly.

rhatdan:
Right additional stores are only treated as read/only from container/storage point of view.  There is nothing that requires them to be actually on read/only file systems.

The create attempt when the file actually does not exist.

apollo13:
Looking through the change, does this actually fix the issue? After all the `overlay-images` won't exist either in an empty store…

rhatdan:
Maybe not, you would want it to create the entire path?

rhatdan:
Ok not fail on additonal stores that return errors.

apollo13:
Yeah it would be great if it would simply skip the store if there is nothing in there. Think of it like that: I can mount an empty store in and fill it at a later time for instance… Currently it requires it to have at least one image there (or at least run any other command that fills up that directory & lock structure).

rhatdan:
This patch in Podman worked for me.

I added "/var/lib/shared", to /etc/containers/storage.conf
```
# mkdir /var/lib/shared 
# ./bin/podman info | grep shared
    overlay.imagestore: /var/lib/shared
#  find /var/lib/shared/
/var/lib/shared/
/var/lib/shared/overlay-images
/var/lib/shared/overlay-images/images.lock
```
Seems like containers storage creates the /var/lib/shared/overlay-images now, and with my fix it now creates the images.lock file.

apollo13:
Ah well it probably did not create overlay-images for me since I mounted the volume `ro` (I mostly followed your article https://www.redhat.com/sysadmin/image-stores-podman and thought that it seemed like a nice idea to provide a read only cache for commonly used images that clients cannot modify).

PR #1033, opened 2021-10-03
fsdiff: [Race] fs lock before removing common parent layer from host

flouthoc:
Following PR ensures that we acquire a soft fs lock before removing
parent of any layer from host. Since parent layers could be re-used we
must lock if we are cleaning parent layer to prevent any race scenarios
for parallel invocation of process using storage (typically builds).

TLDR:
prevent race of missing parent layer when multiple processes are invoked

Already experienced scenarios
Missing parent lower when build is invoked in a parallel manner.

Closes: https://github.com/containers/fuse-overlayfs/issues/322


drivers/fsdiff.go
```diff
@@ -44,6 +44,11 @@ func NewNaiveDiffDriver(driver ProtoDriver, updater LayerIDMapUpdater) Driver {
 func (gdw *NaiveDiffDriver) Diff(id string, idMappings *idtools.IDMappings, parent string, parentMappings *idtools.IDMappings, mountLabel string) (arch io.ReadCloser, err error) {
 	startTime := time.Now()
 	driver := gdw.ProtoDriver
+	// since we are going to remove layers
+	// lock so other parallel instances can
+	// wait while parent layer can be restored.
+	driver.FileLock()
+	defer driver.FileUnlock()
```
giuseppe, line 51: Is it something we could fix directly in the overlay driver?

And it seems to be locked only here. Is this enough to protect from race conditions?

giuseppe (commented):
Good finding!

Some questions inline

drivers/fsdiff.go
flouthoc: Sure we could also lock here https://github.com/containers/storage/blob/main/drivers/overlay/overlay.go#L1775 my first intention was to make lock granular so we don't end up blocking for operations where lock is not needed.

During my experiments i saw race with fsdiff only so just to make things granular i decided to have it restricted only to fsdiff but i agree we could move this to overlay as well to be extra safe.

giuseppe: does the issue happen if there is a fsdiff and at the same time the container/image is being removed?

flouthoc: yes this happens whenever flow goes into https://github.com/containers/storage/blob/main/drivers/overlay/overlay.go#L1777 from more than one process and `parent` is same across multiple processes.

So typically one process might be trying to read things here https://github.com/containers/storage/blob/main/drivers/fsdiff.go#L92 while another process removes parent here https://github.com/containers/storage/blob/main/drivers/fsdiff.go#L90

TomSweeneyRedHat:
@flouthoc nice work!  If you have not already done so, could you put together a WIP PR in Podman and vendor this change in there to see if the tests there are happy with this change?

drivers/fsdiff.go
giuseppe: why doesn't the locking we have in place work here?

In overlay.go Put() has already a ref counter and it should not do any unmount when there is another user

flouthoc: I believe that is because `Put` has a mutex lock which only works for goroutines and thread in same process. However we do have a `fs` lock in `unmount` but at one place we are removing directory outside `fs` lock or `unmount` i.e https://github.com/containers/storage/blob/main/drivers/overlay/overlay.go#L1564 which is causing issues.

flouthoc: I think this is there in the first place because most of driver code is fetched from **moby** but they dont have inter-process race situations since everything works from a single daemon process.

PR #1035, opened 2021-10-04
Allow setting quota for individual containers

rhatdan:
Drivers have the ability to support size and inodes quota, but
we don't allow these options to be passed down to the driver.

With this fix, Podman should be able to support --storage-opt size=
per container.

Helps Fix: https://github.com/containers/podman/issues/11016


flouthoc:
LGTM, put already accepts storage opts

rhatdan:
@vrothberg @giuseppe @nalind @TomSweeneyRedHat @ashley-cui @umohnani8 @saschagrunert PTAL

umohnani8:
LGTM

PR #1047, opened 2021-10-19
Fix: relative path will result infinite loop

chenk008:
https://github.com/containers/storage/blob/73df15cb57ffd432da5f0e26ef6538441a00021a/pkg/idtools/idtools_unix.go#L49 will result an infinite loop when the path is relative.

Add validation to avoid it.

Signed-off-by: chenk008 <kongchen28@gmail.com>

### DIFF
--- a/pkg/idtools/idtools_unix.go
+++ b/pkg/idtools/idtools_unix.go
@@ -46,6 +46,9 @@ func mkdirAs(path string, mode os.FileMode, ownerUID, ownerGID int, mkAll, chown
 		// walk back to "/" looking for directories which do not exist
 		// and add them to the paths array for chown after creation
 		dirPath := path
+		if !filepath.IsAbs(dirPath) {
+			return fmt.Errorf("path: %s should be absolute", dirPath)
+		}
 		for {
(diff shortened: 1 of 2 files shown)


pkg/idtools/idtools_unix.go
```diff
@@ -46,6 +46,9 @@ func mkdirAs(path string, mode os.FileMode, ownerUID, ownerGID int, mkAll, chown
 		// walk back to "/" looking for directories which do not exist
 		// and add them to the paths array for chown after creation
 		dirPath := path
+		if !strings.HasPrefix(dirPath, "/") {
+			return fmt.Errorf("path: %s should start with /", dirPath)
```
flouthoc, line 50: Shouldnt we do `filepath.IsAbs` ?

chenk008: Thanks for your suggestion, I replaced with `filepath.IsAbs`

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
I've restarted the failed test, it looked to be a network flake.

PR #1052, opened 2021-11-03
chunked: fix another occurrence of linkat

giuseppe:
Using unix.AT_EMPTY_PATH requires CAP_DAC_READ_SEARCH.  Use an
equivalent variant that uses /proc/self/fd that can be used with
rootless.


flouthoc:
LGTM

rhatdan (approved):
LGTM

PR #1058, opened 2021-11-10
reexec.Command(): panic if reexec.Init() wasn't called

nalind:
If we start preparing to reexec ourselves, but didn't call `Init()` earlier, `panic()`.  The alternative is that the child process we start
won't do the work we want it to do as a dedicated subprocess, failing in a way which could be difficult to troubleshoot.

Fixes https://github.com/containers/image/issues/1401 by making it easier to spot the bug.


mtrmac (commented):
LGTM

rhatdan:
LGTM

PR #1062, opened 2021-11-17
store.DeleteImage: correctly account for mapped layers

nalind:
When walking the list of layers in an image to decide which of them to remove as part of removing that image, we check that each candidate layer is a) not being used as the top layer for any other images, and b) that the only layer that lists a candidate layer as its parent layer is either the top layer of the image we're removing, or the previous layer that we decided should be removed.

When the image has multiple top layers because some of them are alternate top layers which use different ID mappings, that second check didn't account for those mapped layers as being layers that shouldn't prevent us from removing a candidate layer.

Fixes https://github.com/containers/podman/issues/12316.

### DIFF
--- a/store.go
+++ b/store.go
@@ -2371,22 +2371,16 @@ func (s *store) DeleteImage(id string, commit bool) (layers []string, err error)
 		if err != nil {
 			return nil, err
 		}
-		childrenByParent := make(map[string]*[]string)
+		childrenByParent := make(map[string][]string)
 		for _, layer := range layers {
-			parent := layer.Parent
-			if list, ok := childrenByParent[parent]; ok {
-				newList := append(*list, layer.ID)
-				childrenByParent[parent] = &newList
-			} else {
-				childrenByParent[parent] = &([]string{layer.ID})
-			}
+			childrenByParent[layer.Parent] = append(childrenByParent[layer.Parent], layer.ID)
 		}
-		otherImagesByTopLayer := make(map[string]string)
+		otherImagesTopLayers := make(map[string]struct{})
 		for _, img := range images {
 			if img.ID != id {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@giuseppe @flouthoc PTAL

cmd/containers-storage/main.go
```diff
@@ -32,6 +33,7 @@ func main() {
 	if reexec.Init() {
 		return
 	}
+	unshare.MaybeReexecUsingUserNamespace(false)
```
flouthoc, line 36: I think this should be a separate commit. I am unable to figure out if this is related to logic change.

flouthoc:
@nalind Plz correct me if i am wrong. 

Actual logic change is in  only `hasOtherRefs` which was before accounting for layers as a positive reference although they were already marked as to be removed reason was we were only keeping recently used layer for check i.e in `lastRemoved`.

But now its changed to map `layersToRemoveMap`  keeping and validating against all the layers which we removed so `hasOtherRef` doesn't returns with a `false` positive. 

Change LGTM

flouthoc:
Not related to this PR: This needs a bit of re-phrasing/re-wording comment here is a bit confusing to read at first https://github.com/containers/storage/blob/main/store.go#L242 but could me only me.

flouthoc (approved):

giuseppe:
LGTM, as @flouthoc pointed out, could we move the change in `cmd/containers-storage/main.go` to a separate commit?

nalind:
Dropped the bit from `main.go`.

cmd/containers-storage/main.go
nalind: Dropped it.

rhatdan:
Going to force merge, to ignore devicemapper flake.

PR #1065, opened 2021-11-19
Make random value generation and use more reliable

mtrmac:
- Maintain a private RNG for `pkg/stringid` so that it is unaffected by anyone else thinking that `math/rand.Seed()` guarantees anything in a large codebase.
- When generating container/image/layer IDs, make sure not to generate duplicates, to eliminate the slightest risk of this randomly failing.

Compare the plight at https://github.com/containers/podman/issues/12155 .

⚠️ Please review the `getMaxSizeFromImage` part carefully; it’s quite possible that the use of the container ID there is intentional, I wouldn’t know.

---

It might make sense to merge this only after various other users of `math/rand` are modified/fixed, in case they unknowingly depend on this seeding the global RNG with unpredictable values (vs. the Go standard library of `Seed(1)`).


mtrmac:
> When generating container/image/layer IDs, make sure not to generate duplicates, to eliminate the slightest risk of this randomly failing.

There are many more instances of this in the tests; this PR doesn’t fix that.

mtrmac:
Another place where random values are assumed never to conflict is the `lockfile.lw` value.

So, this PR is now trying to make it consistently unique. That code complexity is bordering on not worth it — feel free to tell me to drop this part. The one possibly redeeming quality of it is that we now record a timestamp and PID of the lock owner into the lock state, which might help with later debugging.

rhatdan:
I think we need a PR opened against Podman and maybe buildah to make sure there are no regressions.

TomSweeneyRedHat:
@rhatdan tests aren't happy and you might need a rebase.

rhatdan:
@TomSweeneyRedHat You meant @mtrmac or commented in the wrong PR.

rhatdan:
@mtrmac Any idea why this is blowing up?

mtrmac:
Not yet, looking…

rhatdan:
LGTM

mtrmac:
I think so; but I’m not comfortable with merging this without someone much more familiar with c/storage reviewing it and making that call.

rhatdan:
@giuseppe @nalind PTAL

rhatdan:
LGTM

nalind:
Yes, this all looks fine.
Using strong entropy for 44 bytes of the ID might be overkill, but I can't guarantee that people aren't trying to share lock files across PID namespaces, so I can't rule out the need for it definitively.
I don't see any particular reason that we'd need to force the layer ID in `getMaxSizeFromImage()`.  In retrospect, though, not retrofitting it to use `store.MountImage()` when we added that method was a missed opportunity.

mtrmac:
> Using strong entropy for 44 bytes of the ID might be overkill

True. The previous version used crypto/rand for all of the 64 bytes, so this is at least a reduction. I can make the random part shorter (padding with zeroes), but then we get to discuss how short exactly :) Probably _at least_ 64 bits, to be commensurate with the PID namespace, and to account for the birthday paradox. 128 bits? 256?

rhatdan:
@mtrmac @nalind Is this ready to merge?

nalind:
LGTM

kolyshkin:
OTOH moby/moby removed the non-crypto functions from the package: https://github.com/moby/moby/pull/39336

rhatdan:
Want to open a PR to make the same change here?

kolyshkin:
> Want to open a PR to make the same change here?

Sure: https://github.com/containers/storage/pull/1337

PR #1066, opened 2021-11-19
Use math/rand.Rand.Read to generate slices of random data

mtrmac:
… instead of open-coding loops, primarily to simplify.

In one case this is a speedup, in the other it’s a noticeable slowdown, a correctness improvement in theory at least.

### DIFF
--- a/drivers/graphtest/graphtest_unix.go
+++ b/drivers/graphtest/graphtest_unix.go
@@ -1,3 +1,4 @@
+//go:build linux || freebsd || solaris
 // +build linux freebsd solaris
 
 package graphtest
@@ -10,9 +11,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"reflect"
 	"testing"
-	"unsafe"
 
 	graphdriver "github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/archive"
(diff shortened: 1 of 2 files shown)


mtrmac:
(If you prefer the code as is, feel completely free to close this PR.)

vrothberg (commented):
LGTM
@giuseppe @nalind @rhatdan PTAL

mtrmac:
FWIW
```
% go test -v -bench . -benchmem ./bench_test.go
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkStandard
BenchmarkStandard-12       	      28	  38833565 ns/op	52434254 B/op	       2 allocs/op
BenchmarkSourceOnly
BenchmarkSourceOnly-12     	      22	  51455058 ns/op	52428800 B/op	       1 allocs/op
BenchmarkSourceOnly8
BenchmarkSourceOnly8-12    	      22	  50682468 ns/op	52428903 B/op	       1 allocs/op
BenchmarkUnsafeSlice
BenchmarkUnsafeSlice-12    	      56	  20757283 ns/op	52428803 B/op	       1 allocs/op
```
on 50 MiB output, with
- Standard: `rand.New(rand.NewSource(seed)).Read`
- SourceOnly: Two nested loops turning `Int63` into 7 bytes, like one of the previous implementations
- SourceOnly8: Two nested loops turning `Int63` into 8 bytes (it turns out the compiler doesn’t unroll that any better)
- UnsafeSlice: A single loop storing `Int63` into an `int64`, like the other previous implementation

giuseppe (approved):
LGTM

PR #1072, opened 2021-11-23
chunked: various improvements

giuseppe:
some improvements and fixes for issues I've found while playing with docker.io/gscrivano/zstd-chunked:fcos-rawhide and docker.io/gscrivano/zstd-chunked:fcos-latest.

More details in each commit.

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -89,7 +89,7 @@ func copyFileContent(srcFd int, destFile string, dirfd int, mode os.FileMode, us
 	src := fmt.Sprintf("/proc/self/fd/%d", srcFd)
 	st, err := os.Stat(src)
 	if err != nil {
-		return nil, -1, err
+		return nil, -1, fmt.Errorf("copy file content for %q: %v", destFile, err)
 	}
 
 	copyWithFileRange, copyWithFileClone := true, true
@@ -111,15 +111,15 @@ func copyFileContent(srcFd int, destFile string, dirfd int, mode os.FileMode, us
 	// If the destination file already exists, we shouldn't blow it away
 	dstFile, err := openFileUnderRoot(destFile, dirfd, newFileFlags, mode)
 	if err != nil {
-		return nil, -1, err
+		return nil, -1, fmt.Errorf("open file %q under rootfs for copy: %v", destFile, err)
 	}
 
 	err = driversCopy.CopyRegularToFile(src, dstFile, st, &copyWithFileRange, &copyWithFileClone)
 	if err != nil {
 		dstFile.Close()
-		return nil, -1, err
+		return nil, -1, fmt.Errorf("copy to file %q under rootfs: %v", destFile, err)
 	}
-	return dstFile, st.Size(), err
+	return dstFile, st.Size(), nil
 }
 
 func prepareOtherLayersCache(layersMetadata map[string][]internal.FileMetadata) map[string]map[string][]*internal.FileMetadata {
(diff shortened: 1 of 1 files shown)


pkg/chunked/storage_linux.go
```diff
@@ -230,21 +230,21 @@ func makeEstargzChunkedDiffer(ctx context.Context, store storage.Store, blobSize
 func copyFileFromOtherLayer(file *internal.FileMetadata, source string, otherFile *internal.FileMetadata, dirfd int, useHardLinks bool) (bool, *os.File, int64, error) {
 	srcDirfd, err := unix.Open(source, unix.O_RDONLY, 0)
 	if err != nil {
-		return false, nil, 0, err
+		return false, nil, 0, errors.Wrapf(err, "open source file %q", source)
```
rhatdan, line 233: This looks like a stutter.

pkg/chunked/storage_linux.go
```diff
@@ -230,21 +230,21 @@ func makeEstargzChunkedDiffer(ctx context.Context, store storage.Store, blobSize
...
 	}
 	defer unix.Close(srcDirfd)
 
 	srcFile, err := openFileUnderRoot(otherFile.Name, srcDirfd, unix.O_RDONLY, 0)
 	if err != nil {
-		return false, nil, 0, err
+		return false, nil, 0, errors.Wrapf(err, "open source file %q under target rootfs", otherFile.Name)
```
rhatdan, line 239: I am worried about stutter here as well.  Doesn't this end up calling an unix.Open()? Which will report an error with the file that the open failed on?

pkg/chunked/storage_linux.go
```diff
@@ -526,27 +526,27 @@ func setFileAttrs(file *os.File, mode os.FileMode, metadata *internal.FileMetada
 
 	if err := unix.Fchown(fd, metadata.UID, metadata.GID); err != nil {
 		if !options.IgnoreChownErrors {
-			return err
+			return errors.Wrapf(err, "chown %q to %d:%d", metadata.Name, metadata.UID, metadata.GID)
```
rhatdan, line 572: Stutter.

pkg/chunked/storage_linux.go
```diff
@@ -828,7 +828,10 @@ func safeSymlink(dirfd int, mode os.FileMode, metadata *internal.FileMetadata, o
 		destDirFd = int(f.Fd())
 	}
 
-	return unix.Symlinkat(metadata.Linkname, destDirFd, destBase)
+	if err := unix.Symlinkat(metadata.Linkname, destDirFd, destBase); err != nil {
+		return errors.Wrapf(err, "create symlink %q pointing to %q", metadata.Name, metadata.Linkname)
```
rhatdan, line 944: stutter.

rhatdan:
LGTM other then concerns about stutter.

pkg/chunked/storage_linux.go
giuseppe: without the extra information, we only get the UNIX error without any pointer to what symlink we were creating

pkg/chunked/storage_linux.go
giuseppe: no, it won't use `unix.Open` but call directly the openat2 wrapper, that only produces the UNIX error without any additional information.

pkg/chunked/storage_linux.go
giuseppe: the `unix.Open` call doesn't include any extra information except the errno

rhatdan (approved):
Ok I fine then.
LGTM

rhatdan:
@flouthoc @nalind @vrothberg @saschagrunert @umohnani8 PTAL

pkg/chunked/storage_linux.go
```diff
@@ -89,7 +89,7 @@ func copyFileContent(srcFd int, destFile string, dirfd int, mode os.FileMode, us
 	src := fmt.Sprintf("/proc/self/fd/%d", srcFd)
 	st, err := os.Stat(src)
 	if err != nil {
-		return nil, -1, err
+		return nil, -1, errors.Wrapf(err, "copy file content for %q", destFile)
```
vrothberg, line 92: Nit: using `fmt.Errorf` would be more modern

vrothberg (commented):
Other than the nit LGTM

giuseppe:
> Other than the nit LGTM

thanks.  Changed to use `fmt.Errorf` instead of `errors.Wrapf`

PR #1075, opened 2021-11-25
chunked: add fallback for openat2

giuseppe:
add a fallback mechanism when openat2 is not supported by the underlying kernel.
    
If a call to openat2 fails with ENOSYS, then fallback to a user space lookup.  Generally the user space lookup is not safe, since symlinks lookups are vulnerable to TOCTOU attacks, but in this case where the rootfs is being created, there are no other processes modifying it thus such lookups can be considered safe.


pkg/chunked/storage_linux.go
```diff
@@ -580,36 +582,121 @@ func setFileAttrs(dirfd int, file *os.File, mode os.FileMode, metadata *internal
...
 		Flags:   flags,
 		Mode:    uint64(mode & 07777),
 		Resolve: unix.RESOLVE_IN_ROOT,
 	}
+	return unix.Openat2(dirfd, name, &how)
+}
+
+// skipOpenat2 is set when openat2 is not supported by the underlying kernel and avoid
+// using it again.
+var skipOpenat2 int32
```
flouthoc, line 672: I might be missing something :\  but got confused about where are we setting `skipOpenat2`. It seems to be uninitialized.

giuseppe: we are setting it here: https://github.com/containers/storage/pull/1075/files/94c46e4cc6f38b518a4f54dce82e0a09260432f4#diff-7badf29db7e16771a6012a87e00fc3aa64c61025c049584ee4737cadb09de517R686

flouthoc: got it thanks i missed it earlier

flouthoc (approved):
LGTM

rhatdan:
LGTM

PR #1077, opened 2021-11-26
archive: ignore the security.selinux xattr

giuseppe:
ignore the security.selinux xattr if it is present in the tarball header, since invalid labels cannot be set by unprivileged users and the `lsetxattr` syscall fails with EINVAL.

Closes: https://github.com/containers/storage/issues/1076

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -77,6 +77,10 @@ const (
 	containersOverrideXattr = "user.containers.override_stat"
 )
 
+var xattrsToIgnore = map[string]interface{}{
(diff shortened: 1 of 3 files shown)


giuseppe:
@flouthoc PTAL

flouthoc (approved):
LGTM. 

I remember facing same issue with `redroid` image while debugging some issue. I left comment of `selinux label` here: https://github.com/containers/podman/issues/12189#issuecomment-963921427 . I wonder what is the intention behind such image.

rhatdan:
LGTM

PR #1080, opened 2021-11-29
Reset defaultConfigFile if defaultOverrideConfigFile exits.

rhatdan:
Currently DefaultConfigFile does not report back the storage.conf file
that is actually used if /etc/containers/storage.conf exists it should
be reported back to the user. This issue was seen with Podman.

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -42,6 +42,10 @@ func init() {
 	defaultStoreOptions.GraphDriverName = ""
 
 	if _, err := os.Stat(defaultOverrideConfigFile); err == nil {
+		// The DefaultConfigFile(rootless) function returns the path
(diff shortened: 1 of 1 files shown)


types/options.go
```diff
@@ -42,6 +42,7 @@ func init() {
 	defaultStoreOptions.GraphDriverName = ""
 
 	if _, err := os.Stat(defaultOverrideConfigFile); err == nil {
+		defaultConfigFile = defaultOverrideConfigFile
```
vrothberg, line 45: Could you add a comment to this line?  I had to jump back two times to the commit message to understand why we need that change and am afraid of my garbage collector to forget so quickly.

vrothberg (commented):
LGTM, thank you

flouthoc (approved):

PR #1083, opened 2021-12-02
options: fall back to default `graph` and `run` root for empty case

flouthoc:
While reloading from config files `graph` and `run` root could be set to
empty. We should fall back to default if they are empty.

Closes: https://github.com/containers/podman/issues/12467

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -27,6 +27,13 @@ type tomlConfig struct {
 	} `toml:"storage"`
 }
 
+const (
+	// these are default path for run and graph root for rootful users
+	// for rootless path is constructed via getRootlessStorageOpts
+	defaultRunRoot   string = "/run/containers/storage"
+	defaultGraphRoot string = "/var/lib/containers/storage"
+)
+
 // defaultConfigFile path to the system wide storage.conf file
 var (
 	defaultConfigFile         = "/usr/share/containers/storage.conf"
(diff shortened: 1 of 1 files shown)


flouthoc:
Still need to add tests for this.

types/options.go
```diff
@@ -53,6 +53,13 @@ func init() {
 		}
 		ReloadConfigurationFileIfNeeded(defaultConfigFile, &defaultStoreOptions)
 	}
+	// reload could set values to empty for run and graph root if config does not contains anything
+	if defaultStoreOptions.RunRoot == "" {
+		defaultStoreOptions.RunRoot = "/run/containers/storage"
```
rhatdan, line 58: Will this work in rootless mode?

rhatdan: At least make these a constant.

flouthoc: @rhatdan I think issue only happens for rootful case. But i need to try it myself ill confirm it and make changes.

And sure I agree ill make these constant.

rhatdan:
LGTM
@giuseppe @nalind @vrothberg @TomSweeneyRedHat @umohnani8 @saschagrunert PTAL

umohnani8:
LGTM, are we waiting for tests to be added?

vrothberg:
> LGTM, are we waiting for tests to be added?

I am also in the "adding tests camp".  There's no guarantee we won't regress another time.

flouthoc:
> Still need to add tests for this.

@umohnani8 @vrothberg Yes i was trying to figure out the best way to test this in `c/storage` but unfortunately i was unable to test buggy behavior in `c/storage` would it be fine if i test this in `podman` ?

PR #1084, opened 2021-12-06
Make homedir functions work on Windows

rhatdan:
Needed for https://github.com/containers/podman/pull/12503


rhatdan:
@n1hility  PTAL

rhatdan:
@giuseppe @nalind @vrothberg @ashley-cui @umohnani8 @flouthoc @saschagrunert PTAL

vrothberg (commented):
LGTM, thanks @rhatdan

flouthoc (approved):

n1hility:
@rhatdan LGTM *but* as a disclaimer, technically Windows apps often use %LocalAppData%\[name of app] -> which is [user's home]\AppData\Local\[e.g. Podman])  In my draft win podman machine PR I just kept the Linux podman conventions and planned to revisit since I wasn't sure if it should be Local\Containers\Podman or Local\Podman\Containers. I personally think it's fine to leave it like this so, but up to you guys if we should be stricter adherents to the OS convention over an app convention.

rhatdan:
I am fine with fixing this to match Microsoft, perhaps we just do a if runtime.GOOS="windows" or make the functions in separate windows source code.

PR #1087, opened 2021-12-09
layerStore.Diff(): don't add an unnecessary buffer

nalind:
Don't `ReadAll()` from a `Reader` to create a buffer and then create another `Reader` to read from that buffer.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -23,6 +23,7 @@ import (
 	"github.com/containers/storage/pkg/system"
 	"github.com/containers/storage/pkg/tarlog"
 	"github.com/containers/storage/pkg/truncindex"
+	multierror "github.com/hashicorp/go-multierror"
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM, if this works.  Is this something we should test against buildah before we merge?

TomSweeneyRedHat:
Changes LGTM, tests do not.  A test with Buildah would be good.

nalind:
We were potentially closing the underlying file and the pgzip decompression filter before the JSON decoder had finished reading from the decompression filter, and that appears to have created a hang when we stopped buffering up all of the file's contents at the first step.  The JSON decoder and pgzip decompressor are both advertised as potentially reading more information than they strictly require if there's some available, so I'm guessing that a partial read left one or the other was waiting for more information that would never come.

rhatdan:
LGTM

PR #1093, opened 2021-12-22
GetDataHome accept uid

cdoern:
GetDataHome looks for the home directory on the user executing the function, if you want to get
the rootless storage config while running as root, this function fails to return the rootless home dir.
Adding a uid and checking if that uid matches ours, allows us to see if we need to find a different user's home dir.

This is necessary for a new podman function I am implementing `podman setup`. the setup function initializes all directories an files (DB included) for podman to be run successfully. Acquiring the home directory of the desired user is important for this task.

Signed-off-by: cdoern <cdoern@redhat.com>


pkg/homedir/homedir.go
```diff
@@ -21,11 +25,33 @@ func GetConfigHome() (string, error) {
 	return filepath.Join(home, ".config"), nil
 }
 
-// GetDataHome returns XDG_DATA_HOME.
+// GetDataHome returns XDG_DATA_HOME for the given UID
 // GetDataHome returns $HOME/.local/share and nil error if XDG_DATA_HOME is not set.
 //
 // See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
-func GetDataHome() (string, error) {
+func GetDataHome(uid int) (string, error) {
```
rhatdan, line 32: Add a new function
GetDataHomeByUID(uid int) ...

And have GetDataHome call it.

pkg/homedir/homedir.go
```diff
@@ -21,11 +25,14 @@ func GetConfigHome() (string, error) {
 	return filepath.Join(home, ".config"), nil
 }
 
-// GetDataHome returns XDG_DATA_HOME.
+// GetDataHome returns XDG_DATA_HOME for the given UID
```
rhatdan, line 28: This change is no longer valid.

pkg/homedir/homedir.go
```diff
@@ -21,11 +25,14 @@ func GetConfigHome() (string, error) {
...
 // GetDataHome returns $HOME/.local/share and nil error if XDG_DATA_HOME is not set.
 //
 // See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
-func GetDataHome() (string, error) {
+func GetDataHome(uid int) (string, error) {
```
rhatdan, line 32: This should no longer change.

pkg/homedir/homedir.go
```diff
@@ -21,11 +25,14 @@ func GetConfigHome() (string, error) {
...
+	if uid != os.Getuid() && os.Getuid() == 0 { // this means getRootlessDirInfo was called as root
```
rhatdan, line 33: I don't see why this code is here. It should just leave this call alone.

pkg/homedir/homedir.go
cdoern: but shouldn't `GetDataHome` call `GetDataHomeUID`? or do you want `GetDataHomeUID` to be called with `os.Getuid` in this case?

rhatdan: No, because we don't need to exec sudo for this. The GetDataHomeUID, is the weird case.

cdoern: ok so these should be separate functions that don't call one another? I thought having one call the other would make sense in some cases but splitting them completely makes sense to me as well.

pkg/homedir/homedir.go
```diff
@@ -34,6 +38,31 @@ func GetDataHome() (string, error) {
...
+		return "", err
+	}
+	cmd := exec.Command("sudo", "-Hiu", u.Username, "env")
+	out, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	env := strings.Split(string(out), "\n")
+	for _, val := range env {
+		if strings.Contains(val, "XDG_DATA_HOME") || strings.Contains(val, "HOME") {
```
giuseppe, line 58: for what functionality do we need this function?

giuseppe: I think it is risky to add such a function to c/storage as we are stuck with it forever.

This function cannot be used from a user namespace since `sudo` doesn't work there.

Maybe we could add it to Podman for now (or ignore `XDG_DATA_HOME` to detect rootless storage directories)?

I suggest to ignore `XDG_DATA HOME` since anyway it is a heuristic as the user can override it at any time, and I don't think we can detect it.  If the store is not present under `$HOME/.local/share/containers/storage` then handle it as not detected

cdoern: I am making a `podman system setup` feature that has the ability to take a user and setup podman storage from scratch. When run as root, `types.DefaultStorOptions` returns the root store options even if `rootless = true` is passed.

cdoern: the issue here is I have no way of getting the user's home directory from podman if we are running as root trying to find another user's default storage options.

rhatdan: Why not use machinectl login ... And execute podman in the user environment, falling back to sudo? I think this is the safest way forward. One issue with doing this via Code is you might miss things like storage.conf and containers.conf files sitting in the users homedir. Seems to me is what you want is to execute Podman in the same way a user would.

cdoern: @rhatdan would it make more sense to use `machinectl` within podman to avoid issues @giuseppe brings up about long-term maintenance w/in storage? I could basically create this function with machinectl and sudo in podman to get the home directory, and then modify the `DefaultStoreOption`'s run/graph root accordingly once they are returned?

rhatdan: I believe this function should be moved to Podman, and we should default to machinectl and fall back to sudo.

PR #1096, opened 2022-01-05
 GetStore modifications, Create GetLocalStorageConfig and SaveLocalStorageConfig

cdoern:
If passing either just a runroot or both a runroot and a graphroot
GetStore() was returning when finding a match for JUST the graph root,
overriding the runroot and using a different location than specified.

if either one or the other is passed, default the other option but if both are passed and not found together while searching the store, create a new store.

Also create two new helper functions which add the ability to retrieve and overwrite
the toml config in the user's storage.conf file

Signed-off-by: cdoern <cdoern@redhat.com>


cdoern:
@rhatdan PTAL, I'm adding this specifically for the podman reset flags I'm working on but either way the logic here seemed kind of broken

store.go
```diff
@@ -647,17 +647,21 @@ func GetStore(options types.StoreOptions) (Store, error) {
 	storesLock.Lock()
 	defer storesLock.Unlock()
 
+	// return if BOTH run and graph root are matched, otherwise our run-root can be overriden if the graph is found first
 	for _, s := range stores {
-		if s.graphRoot == options.GraphRoot && (options.GraphDriverName == "" || s.graphDriverName == options.GraphDriverName) {
+		if (s.graphRoot == options.GraphRoot) && (s.runRoot == options.RunRoot) && (options.GraphDriverName == "" || s.graphDriverName == options.GraphDriverName) {
```
rhatdan, line 652: What happens if s.runRoot != "" and options.RunRoot=""?

store.go
```diff
@@ -647,17 +647,21 @@ func GetStore(options types.StoreOptions) (Store, error) {
...
-	}
-	if options.RunRoot == "" {
-		return nil, errors.Wrap(ErrIncompleteOptions, "no storage runroot specified")
+	// if passed a run-root or graph-root alone, the other should be defaulted only error if we have neither.
+	switch {
+	case options.GraphRoot == "" && options.GraphRoot != "":
+		options.GraphRoot = types.Options().GraphRoot
+	case options.RunRoot == "" && options.GraphRoot != "":
+		options.RunRoot = types.Options().RunRoot
+	case options.RunRoot == "" && options.GraphRoot == "":
```
rhatdan, line 663: This should be default:

rhatdan: I think this could be done better with
```
 	case options.GraphRoot != "" && options.GraphRoot != "":
                    break;
       	case options.GraphRoot == "" && options.GraphRoot != "":
		options.GraphRoot = types.Options().GraphRoot
	case options.RunRoot == "" && options.GraphRoot != "":
		options.RunRoot = types.Options().RunRoot
	default:
...
```

store.go
cdoern: this loop just checks if a store already exists w/ the matching info we are giving right? so if the `s.runRoot != options.RunRoot` for all current stores, we want to create our own right?

cdoern:
@rhatdan let me know if this is what you meant for editing the toml. I am not sure if there is a cleaner way to override the file but if there is I will do that instead.

types/options.go
```diff
@@ -424,3 +425,40 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
...
+	}
+	f, err := os.Open(configFile)
+	if err != nil {
+		return err
+	}
+
+	return toml.NewEncoder(f).Encode(conf)
+}
+
+// GetLocalStorageConfig is used to retreive the storage.conf toml in order to overwrite it
```
rhatdan, line 449: Call this just StorageConfig()

types/options.go
```diff
@@ -424,3 +425,40 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
 func Options() StoreOptions {
 	return defaultStoreOptions
 }
+
+// SaveLocalStorageConfig overwrites the tomlConfig in storage.conf with the given conf
+func SaveLocalStorageConfig(conf TomlConfig) error {
```
rhatdan, line 430: Can you make this
func (* TomlConfig) Save() error {

types/options.go
```diff
@@ -424,3 +425,40 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) {
...
+	configFile, err := DefaultConfigFile(unshare.IsRootless())
+	if err != nil {
+		return err
+	}
+	if _, err := os.Stat(configFile); err == nil {
+		err = os.Remove(configFile)
```
rhatdan, line 436: One line 
if err = os.Remove(configFile); err {
...

store.go
```diff
@@ -647,17 +647,23 @@ func GetStore(options types.StoreOptions) (Store, error) {
...
-	if options.GraphRoot == "" {
-		return nil, errors.Wrap(ErrIncompleteOptions, "no storage root specified")
-	}
-	if options.RunRoot == "" {
-		return nil, errors.Wrap(ErrIncompleteOptions, "no storage runroot specified")
+	// if passed a run-root or graph-root alone, the other should be defaulted only error if we have neither.
+	switch {
+	case options.RunRoot != "" && options.GraphRoot != "":
+		break
+	case options.GraphRoot == "" && options.RunRoot != "":
```
rhatdan, line 661: No need for && I think this is simpler to read as 
```
case options.Graphroot == "":
                options.GraphRoot = types.Options().GraphRoot
case options.RunRoot == "":
		options.RunRoot = types.Options().RunRoot
```

types/options.go
cdoern: I think this will work, right?

```
if err = os.Remove(configFile); os.IsNotExist(err) {
	return err
}
```

store.go
cdoern: ok, I'm going to put the `case options.RunRoot == "" && options.GraphRoot == ""` at the top then since I think we want that to return an error, we need either a graph or a run dir to be passed or set by this point.

types/options.go
rhatdan: Yup

rhatdan:
LGTM
@giuseppe @nalind @vrothberg @saschagrunert @TomSweeneyRedHat @umohnani8 PTAL

rhatdan (approved):

store.go
```diff
@@ -647,17 +647,22 @@ func GetStore(options types.StoreOptions) (Store, error) {
...
-		return nil, errors.Wrap(ErrIncompleteOptions, "no storage runroot specified")
+	// if passed a run-root or graph-root alone, the other should be defaulted only error if we have neither.
+	switch {
+	case options.RunRoot == "" && options.GraphRoot == "":
+		return nil, errors.Wrap(ErrIncompleteOptions, "no storage runroot or graphroot specified")
+	case options.GraphRoot == "":
+		options.GraphRoot = types.Options().GraphRoot
+	case options.RunRoot == "":
+		options.RunRoot = types.Options().RunRoot
+
```
nalind, line 665: Errant blank line.

types/options.go
```diff
@@ -13,11 +13,12 @@ import (
 	"github.com/containers/storage/drivers/overlay"
 	cfg "github.com/containers/storage/pkg/config"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/unshare"
```
nalind, line 16: This is a large dependency to pull in for a package named "types".

cdoern: should I just pass a rootless bool to the function? @rhatdan WDYT?

rhatdan: Yes

umohnani8:
LGTM

rhatdan:
flake not related to PR. Merging.

PR #1097, opened 2022-01-07
Backport store: lock graphLock during to 1.37

TomSweeneyRedHat:
Backport the fix for https://github.com/containers/storage/pull/1037 to
the release-1.37 branch so that it can then be backported to Podman's
v3.4.2-rhel branch for RHEL 8.5.0.2

Signed-off-by: tomsweeneyredhat <tsweeney@redhat.com>

PR #1098, opened 2022-01-08
Backport store: lock graphLock during to 1.36

TomSweeneyRedHat:
Backport the fix for #1037 to
the release-1.36 branch so that it can then be backported to Podman's
v3.4.2-rhel branch for RHEL 8.5.0.2

Note: #1097 was completed to the 1.37 branch in error.  Podman
v3.4.2-rhel is still on c/storage v1.36

Signed-off-by: tomsweeneyredhat <tsweeney@redhat.com>

### DIFF
--- a/store.go
+++ b/store.go
@@ -2825,10 +2825,33 @@ func (s *store) Diff(from, to string, options *DiffOptions) (io.ReadCloser, erro
 	if err != nil {
 		return nil, err
 	}
+
+	// NaiveDiff could cause mounts to happen without a lock, so be safe
+	// and treat the .Diff operation as a Mount.
+	s.graphLock.Lock()
+	defer s.graphLock.Unlock()
+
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
@rhatdan PTAL
I created the release-1.36 branch.  As I went to vendor 1.37 into Podman, I realized it was at 1.36.  Ugh!  I think we probably should kill the release-1.37 branch to stop any confusion.

If you could verify this patch and the new release-1.36 branch, I'd appreciate it. 

t(rouble)

TomSweeneyRedHat:
Ugh, it looks like the two Ubuntu issues are due to problems with the Ubuntu repos, not this change.  Will try rerunning them.

TomSweeneyRedHat:
#@)(*$ just realized I pushed this to main instead of release-1.36.  Closing will resubmit to the right branch.

PR #1101, opened 2022-01-12
pkg: idtools: export RawTo{Container,Host}

cyphar:
While the IDMapping methods are preferable for most users, sometimes it
is necessary to map a single ID using a given mapping. In particular
this is needed for psgo to be able to map the user and group entries in
/proc/$pid/status using the user namespace of the target process.

See https://github.com/containers/psgo/pull/92.
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/pkg/idtools/idtools.go
+++ b/pkg/idtools/idtools.go
@@ -82,7 +82,7 @@ func GetRootUIDGID(uidMap, gidMap []IDMap) (int, int, error) {
 	if len(uidMap) == 1 && uidMap[0].Size == 1 {
 		uid = uidMap[0].HostID
 	} else {
-		uid, err = toHost(0, uidMap)
+		uid, err = RawToHost(0, uidMap)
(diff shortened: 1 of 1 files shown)


cyphar:
Is `IDMap` ever JSON-encoded? I ask because it would be really nice to switch to `rspec.LinuxIDMapping` so that different codebases don't need to convert between different implementations of the same three-entry struct.

giuseppe (approved):
LGTM

flouthoc (approved):
/lgtm
/approve

giuseppe:
> Is `IDMap` ever JSON-encoded? Is that JSON encoding ever parsed from disk or something similar?

it is at least used to store the mapping used by containers in the `containers.json` file

cyphar:
> it is at least used to store the mapping used by containers in the containers.json file

Yeah in that case we can't change it.

Issue #1091, opened 2021-12-20
Support of sparse files in container images


alicefr:
Today, it isn't possible to preserve the sparse files if they are copied inside a container image. This is reported as a limitation in the [OCI spec](https://github.com/opencontainers/image-spec/blob/main/layer.md#file-attributes). A [similar issue](https://github.com/moby/buildkit/issues/1749) has been reported in buildkit. Do we have any chances in the future to support this?

My particular use case affects [KubeVirt](https://github.com/kubevirt/kubevirt) and [container disks](https://kubevirt.io/user-guide/virtual_machines/disks_and_volumes/#containerdisk). I reported the issue in https://github.com/kubevirt/kubevirt/issues/6976 . In KubeVirt , we use container images to ship VM images and the spare files allow us to reduce the final container image size. I understand that this is a very specific use case but the feature might be useful for other situations.

giuseppe:
It sounds like a useful feature, and we can probably extend our format to support them (I've not looked though into it, so I am not sure how difficult it would be with the existing Go libraries for handling tarballs).

Should the holes to be automatically detected (like `cp --sparse=` can do) or use what the file system tells us (via `lseek(SEEK_HOLE)`?

alicefr:
It seems an old missing feature in go: https://github.com/golang/go/issues/13548

giuseppe:
could we circumvent the limitation and inject the missing metadata in the tar header?

alicefr:
I haven't dug too much into the technical alternatives. However, I think the main issue might be that we need to have the ability to write the missing metadata at build time and read it when we pull and extract the layer. So, this depends on which tool we used to build the image.

chainmail:
Hi,
I'm excited to see this coming back and with a positive chance of implementation. Our application stack uses sparse files heavily and means our test framework containers are really very big in some cases (10's of Gb). If you need any one to do some testing, please give me a shout.

giuseppe:
I think we could add it as part of the zstd:chunked format based on top of: https://github.com/containers/storage/pull/1092

I don't think it is worth trying to add this information to the tar stream itself, as anyway the compression deals with sparse files quite well.

giuseppe:
I've opened a PR for sparse files support in zstd:chunked: https://github.com/containers/storage/pull/1102

I think this is the best we can achieve with the current container images.

[...]

PR #1102, opened 2022-01-12
pkg/chunked: add support for sparse files

giuseppe:
automatically detect holes in sparse files (the threshold is hardcoded
at 1kb for now) and add this information to the manifest file.

The receiver will create a hole (using unix.Seek and unix.Ftruncate)
instead of writing the actual zeros.

Closes: https://github.com/containers/storage/issues/1091


rhatdan:
Should the size 1kb be configurable within storage.conf?

pkg/chunked/compressor/compressor.go
```diff
@@ -17,21 +17,137 @@ import (
...
+		// reading the file stream
+		case holesFinderStateRead:
+			if f.zeros > 0 {
+				f.zeros--
+				return 0, 0, nil
+			}
+			b, err := f.reader.ReadByte()
+			if err != nil {
+				return 0, b, err
+			} else if b == 0 {
```
rhatdan, line 51: No need for else.

pkg/chunked/compressor/compressor.go
```diff
@@ -17,21 +17,137 @@ import (
...
+		// accumulating zeros, but still didn't reach the threshold
+		case holesFinderStateAccumulate:
+			b, err := f.reader.ReadByte()
+			if err != nil {
+				if err == io.EOF {
+					f.state = holesFinderStateEOF
+					continue
+				}
+				return 0, b, err
+			} else if b == 0 {
```
rhatdan, line 70: Remove else

pkg/chunked/compressor/compressor.go
```diff
@@ -17,21 +17,137 @@ import (
...
+		case holesFinderStateFound:
+			b, err := f.reader.ReadByte()
+			if err != nil {
+				if err == io.EOF {
+					f.state = holesFinderStateEOF
+				}
+				holeLen := f.zeros
+				f.zeros = 0
+				return holeLen, 0, nil
+			} else if b == 0 {
```
rhatdan, line 91: Remove else

rhatdan (approved):

rhatdan:
@nalind @flouthoc @vrothberg @mtrmac @saschagrunert PTAL

pkg/chunked/compressor/compressor.go
giuseppe: fixed

pkg/chunked/compressor/compressor.go
giuseppe: fixed

rhatdan:
LGTM

pkg/chunked/compressor/compressor.go
```diff
@@ -17,21 +17,141 @@ import (
...
+}
+
+const (
+	holesFinderStateRead = iota
+	holesFinderStateAccumulate
+	holesFinderStateFound
+	holesFinderStateEOF
+)
+
+func (f *holesFinder) ReadByte() (int64, byte, error) {
```
nalind, line 39: Document this.

pkg/chunked/compressor/compressor.go
```diff
@@ -17,21 +17,141 @@ import (
...
+	closed      bool
+	rollsum     *RollSum
+	pendingHole int64
 
-	WrittenOut int64
+	WrittenOut       int64
+	IsLastChunkZeros bool
 }
 
 func (rc *rollingChecksumReader) Read(b []byte) (bool, int, error) {
```
nalind, line 128: Add a comment describing what these returned values are.

pkg/chunked/compressor/compressor.go
giuseppe: fixed

pkg/chunked/compressor/compressor.go
giuseppe: done

pkg/chunked/cache_linux_test.go
```diff
@@ -0,0 +1,255 @@
...
+      "offset": 99939,
+      "chunkSize": 86252,
+      "chunkOffset": 17615,
+      "chunkDigest": "sha256:2a9d3f1b6b37abc8bb35eb8fa98b893a2a2447bcb01184c3bafc8c6b40da099d"
+    }
+}
+`
+
+/*
+func TestHole(t *testing.T) {
```
flouthoc, line 58: Should this be uncommented.

giuseppe: no sorry, this should not be added.  Dropped now

pkg/chunked/compressor/compressor.go
```diff
@@ -119,8 +258,13 @@ func writeZstdChunkedStream(destFile io.Writer, outMetadata map[string]string, r
...
 		chunks := []chunk{}
 
+		hf := &holesFinder{
+			threshold: holesThreshold,
+			reader:    bufio.NewReader(tr),
+		}
+
 		rcReader := &rollingChecksumReader{
-			reader:  bufio.NewReader(tr),
+			reader:  hf,
```
mtrmac, line 267: (Embedding `holesFinder` directly into `rollingChecksumReader`, without the pointer indirection, might help the compiler a bit. Or `ReadByte` could just be a private method of `rollingChecksumReader` — OTOH that might make testing it a bit harder.)

pkg/chunked/compressor/compressor.go
```diff
@@ -17,21 +17,152 @@ import (
...
+		return rc.pendingHole == 0, int(toCopy), nil
+	}
+
 	if rc.closed {
 		return false, 0, io.EOF
 	}
+
 	for i := 0; i < len(b); i++ {
-		n, err := rc.reader.ReadByte()
+		holeLen, n, err := rc.reader.ReadByte()
```
mtrmac, line 165: Does `writeZstdChunkedStream` show up on a CPU profile, by any chance?

I have measured nothing, but all of this fairly deep call stack and state management for every single byte (especially across a public module boundary to `bufio.{ReadByte,UnreadByte}` seems like it it could be _quite_ costly, compared to filling a multi-kilobyte buffer and just doing a linear scan for a non-zero byte in a trivial (and easily-vectorizable) loop.

Of course, let’s not worry about it if it doesn’t matter.

pkg/chunked/storage_linux.go
```diff
@@ -42,9 +43,10 @@ const (
 	containersOverrideXattr = "user.containers.override_stat"
 	bigDataKey              = "zstd-chunked-manifest"
 
-	fileTypeZstdChunked   = iota
-	fileTypeEstargz       = iota
-	fileTypeNoCompression = iota
+	fileTypeZstdChunked = iota
+	fileTypeEstargz
+	fileTypeNoCompression
+	fileTypeHole
```
mtrmac, line 49: This name doesn’t make sense.

At a first glance, without actually trading down the full set of states/transitions in detail, the code should separate two concepts:
- How to decompress compressed data in this file (a function pointer, or an interface for “start decompressing and return a `Reader`” + “clean up”). Notably the `gzipReader`/`zstdReader` members fairly strongly suggest `chunkedDiffer` is inlining ~independent object types`.
- What’s the type of the current chunk (compressed, locally-available, a hole)

mtrmac (commented):
Note at least the `holeFinderStateFound` error path.

PR #1106, opened 2022-01-14
store: use system.EnsureRemoveAll to rm container

giuseppe:
if the container deletion fails with the simple os.RemoveAll(path),
then attempt again using system.EnsureRemoveAll(path).

The difference is that system.EnsureRemoveAll(path) tries to umount
eventual mounts that are keeping the directory busy.

It might help with: https://github.com/containers/podman/issues/11594

### DIFF
--- a/store.go
+++ b/store.go
@@ -2501,23 +2501,29 @@ func (s *store) DeleteContainer(id string) error {
 			gcpath := filepath.Join(s.GraphRoot(), middleDir, container.ID)
 			wg.Add(1)
 			go func() {
-				var err error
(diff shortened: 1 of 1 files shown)


umohnani8:
LGTM

flouthoc:
LGTM , EnsureRemoveAll will do the replaced retries I guess.

flouthoc (approved):

TomSweeneyRedHat:
LGTM
but it looks like a rebase is neeeded

giuseppe:
rebased and pushed

rhatdan:
Ignoring devicemapper flake.

PR #1118, opened 2022-01-25
archive: skip lchown if force_mask is set

giuseppe:
when the option ForceMask is in place and the file UID, GID and mode
are written to a xattr skip the lchown syscall.

[NO TESTS NEEDED] as it requires a setup with NFS.


rhatdan:
LGTM

ashley-cui:
LGTM

PR #1121, opened 2022-01-26
Naive diff should mount all layers read-only

leahneukirchen:
We don't need to write for a diff, and if the layers are on read-only
additionalimagestores, trying to mount them fails when they are used
as upperdir.

Similar to e96cd8656f24f56.

This is pretty much shot-in-the-dark, but it seems to fix the issue for me.
We should also investigate whether Diff and DiffSize need the same treatment.

The code also loses the default mount options here due to passing its own,
not sure if this is problematic.

Removing the /merged dir fails, but not sure why that is done in first place,
as overlay directories usually have empty /merged.

Podman issue: [#12926](https://github.com/containers/podman/issues/12926)

### DIFF
--- a/drivers/fsdiff.go
+++ b/drivers/fsdiff.go
@@ -127,6 +127,7 @@ func (gdw *NaiveDiffDriver) Changes(id string, idMappings *idtools.IDMappings, p
 	options := MountOpts{
 		MountLabel: mountLabel,
 	}
+	options.Options = append(options.Options, "ro")
 	layerFs, err := driver.Get(id, options)
 	if err != nil {
 		return nil, err
@@ -139,6 +140,7 @@ func (gdw *NaiveDiffDriver) Changes(id string, idMappings *idtools.IDMappings, p
 		options := MountOpts{
 			MountLabel: mountLabel,
 		}
+		options.Options = append(options.Options, "ro")
 		parentFs, err = driver.Get(parent, options)
 		if err != nil {
 			return nil, err


rhatdan:
@giuseppe @nalind PTAL

giuseppe:
The idea makes sense to me, but I have not tested it

giuseppe (approved):
tested locally.

LGTM

flouthoc:
@leahneukirchen Could you please sign the commit.

leahneukirchen:
Of course.

Can anyone look at the questions I stated above?

rhatdan:
LGTM

rhatdan:
@nalind PTAL
@leahneukirchen Do you need this in Podman soon?

leahneukirchen:
Yes, it's blocking me shipping a feature in our software. :)

nalind:
Mounting layers read-only could cause problems in cases where we suddenly needed them to be writable, for example if we needed to start a container that used such a layer as its rootfs while that rootfs was mounted read-only.  The naive differ currently mounts the parent layer read-only, and since we intend to never modify layers that are parents of other layers, it's fine if we always do that.

If we extended that to layers that don't have parents, though, committing containers would cause their root filesystems to be mounted read-only (even though they can't be in additional stores, so they're always on writable storage), and that's a potential problem that I would much prefer to avoid.

I'm guessing that having the lower-level driver (which in this case means overlay, since it and vfs are the only two that have the feature, and vfs doesn't actually mount anything) forcibly enable the read-only flag when it knows that a layer that it's being asked to mount is in an additional store would also solve the problem at hand.

leahneukirchen:
That sounds even better, but I wasn't sure how to get that information at that point.

nalind:
I've taken a stab at it using the two commits on the tip of https://github.com/nalind/storage/tree/ro-for-diff.  Does it also solve the problem you're seeing?

leahneukirchen:
That also solves the problem for me! I still think that `Diff` and `DiffSize` also should default the parent to `ro`, then it's good to go I think.

nalind:
@rhatdan managed to sneak that in for `Diff` back in #404, and `DiffSize` lets `Changes` mount the parent, so I think we're set there.  Opened #1123.

rhatdan:
@1123 is merged, do we still need this one?

leahneukirchen:
Nope, thanks!

PR #1128, opened 2022-02-07
containers: detect overlapping mappings

giuseppe:
when the container specifies some mappings to be applied, verify that
they are not overlapping and give a clearer error message.

Closes: https://github.com/containers/storage/issues/1127


flouthoc (approved):
LGTM

containers.go
```diff
@@ -324,6 +324,9 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
 				fmt.Sprintf("the container name \"%s\" is already in use by \"%s\". You have to remove that container to be able to reuse that name.", name, r.byname[name].ID))
 		}
 	}
+	if hasOverlappingRanges(options.UIDMap) || hasOverlappingRanges(options.GIDMap) {
+		return nil, errors.Wrapf(ErrInvalidMappings, "the specified mappings are overlapping")
```
rhatdan, line 328: Could you change this error to

the specified UID and/or GID mappings are overlapping.

nalind:
I could stand to see more detail about where the overlap is.  The more ranges there are in a map, the more work we're forcing someone who's troubleshooting this error to do, when we've already done all of that work.

giuseppe:
> I could stand to see more detail about where the overlap is. The more ranges there are in a map, the more work we're forcing someone who's troubleshooting this error to do, when we've already done all of that work.

good idea, changed the PR to report what mappings cause the conflict

rhatdan:
LGTM

PR #1130, opened 2022-02-10
fix types.Save()

cdoern:
switched os.Open to os.Create because os.Open() creates a file in read only mode

Signed-off-by: cdoern <cbdoer23@g.holycross.edu>


rhatdan:
Why did you remove the remove?

cdoern:
@rhatdan originally `os.Remove` was returning causing no writing to happen because I was not checking if err != nil alongside `os.IsNotExist` should work now

rhatdan:
LGTM

PR #1132, opened 2022-02-11
Ensure overlay.Driver.recreateSymlinks does not loop endlessly

mtrmac:
- Add missing error handling so that if one `Remove` keeps failing, we don’t loop forever
- As a failsafe (if the filesystem were somehow completely broken), terminate the loop after 10 iterations. (There probably shouldn’t even _need_ to be a loop, but that would be a more invasive and risky change.)

WARNING: Completely untested.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package overlay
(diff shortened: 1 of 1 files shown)


mtrmac:
(Overall I find the purpose of `recreateSymlinks` rather unclear. Why do we expect _these particular two files_ (`overlay/l/…` and `overlay/…/link`) to be corrupt, but not any other critical files that make up the layer storage?!)

giuseppe (approved):
LGTM

drivers/overlay/overlay.go
```diff
@@ -1233,7 +1238,10 @@ func (d *Driver) recreateSymlinks() error {
 			if len(targetComponents) != 3 || targetComponents[0] != ".." || targetComponents[2] != "diff" {
 				errs = multierror.Append(errs, errors.Errorf("link target of %q looks weird: %q", link, target))
 				// force the link to be recreated on the next pass
-				os.Remove(filepath.Join(linksDir, link.Name()))
+				if err := os.Remove(filepath.Join(linksDir, link.Name())); err != nil {
```
rhatdan, line 1241: Should this error if the error is ENOEXIST?

mtrmac: Good point — that shouldn’t be reported as an error, but it also shouldn’t be marked as having made progress.

rhatdan:
LGTM

PR #1133, opened 2022-02-11
RFC: Mark layers for deletion before actually doing so

mtrmac:
RFC. This is absolutely untested, and I don’t really know what I’m doing WRT any pre-existing design about crash-resilience or the like.

Currently, we just tell the graph driver to start deleting data, and only after that succeeds, delete the layer metadata. If the graph driver’s deletion process is interrupted for any reason, that could result in a visible / “usable” layer that will nevertheless fail somewhere in the graph driver if on any actual use.

We have an `incompleteFlag` mechanism for layers that have been created but don’t have the content populated on pull; just reuse it for any other deletes.

(In `layerStore.load`, `deleteInternal` is called on layers with that flag already set; in that case we don’t update it again and don’t need to save the layer again, until cleanup finishes.)

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -1166,6 +1166,18 @@ func (r *layerStore) deleteInternal(id string) error {
 	if !ok {
 		return ErrLayerUnknown
 	}
+	// Ensure that if we are interrupted, the layer will be cleaned up.
+	if !layerHasIncompleteFlag(layer) {
+		if layer.Flags == nil {
+			layer.Flags = make(map[string]interface{})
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@nalind @vrothberg PTAL

vrothberg (commented):
LGTM, let's wait for more pairs of eyes

giuseppe (approved):
LGTM

nalind:
LGTM

PR #1134, opened 2022-02-14
chunked: fix read layer cache

giuseppe:
if the layer cache doesn't already exist, automatically create it from
the layer TOC.

commit 10697a05a27fa82db0552d83a4bee5a3703f974f introduced this
regression.

### DIFF
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -108,35 +108,32 @@ func (c *layersCache) load() error {
 		}
 
 		bigData, err := c.store.LayerBigData(r.ID, cacheKey)
-		if err != nil {
-			if errors.Cause(err) == os.ErrNotExist {
+		// if the cache areadly exists, read and use it
+		if err == nil {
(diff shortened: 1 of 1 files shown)


pkg/chunked/cache_linux.go
```diff
@@ -108,35 +108,33 @@ func (c *layersCache) load() error {
...
-
-		metadata, err := readMetadataFromCache(bigData)
-		if err != nil {
 			logrus.Warningf("Error reading cache file for layer %q: %v", r.ID, err)
 		}
-
-		if metadata != nil {
-			c.addLayer(r.ID, metadata)
-			continue
+		if errors.Cause(err) != os.ErrNotExist {
```
vrothberg, line 121: else if?

giuseppe: sure, that is clearer to read.  Fixed

vrothberg (commented):
LGTM
Restarted the flaked job

Issue #1139, opened 2022-02-17
RFC: Create an “incomplete” layer record prior to invoking the graph driver


mtrmac:
Currently, `layerStore.Put` first invokes the graph driver to create a layer, and only afterwards it creates a `layers.json` entry (possibly an “incomplete” one if the layer is created from a tarball).

So, crashing/aborting `layerStore.Put` before a `layers.json` entry is committed to disk can result in a (possibly incomplete) graph driver layer object that we don’t have any record of, and we might well tell the graph driver to create a layer with the same ID (which can then result in unexpected situations).

With Kubernetes-colored glasses on, it seems elegant to me to have a “single source of truth”, so that if `layers.json` and the graph driver state get out of sync, we only ever do corrections in one direction. Considering the existing `incompleteFlag` logic that treats `layers.json` as authoritative (and also the fact that `layers.json` writing is taking more care to be atomic than the overlay graph driver, at least), it seems natural to do the same in `layerStore.Put`. so:

**Proposal:** In `layerStore.Put`, always create a `Layer` object with `incompleteFlag`, save it to `layers.json`, and only afterwards invoke the graph driver to create the layer.

This would be costless for layers created from a tarball (we would only change the timing of the save of the incomplete data), and add one more `layerStore.Save` (and a `fdatasync`) for read/write layers.

If creating the layer is aborted for any reason, the `incompleteFlag` set would cause the layer to be cleaned up in the future (but note #1136 ).

`Driver.Remove` implementations would still need to be fairly resilient against incomplete/corrupt layers, but that’s, to an extent, already the case today.

---

Alternatively, we could require `Driver.Create*` to be idempotent / resilient against pre-existing/partial layers with the same ID.

mtrmac:
… and #1148 implements the suggestion from this issue.

I tend to think that that #1140 is a short-term backportable workaround, and #1148 is closer to a long-term fix, but I really don’t know what I don’t know, so I’d love the experts to review and decide.

mtrmac:
#1148 was merged.

PR #1140, opened 2022-02-17
(Alternative to 1148): Don't blindly reuse state from a previous layer when re-creating it

mtrmac:
- <s>Use the `Driver.locker` lock when creating a layer in the overlay driver. I don’t know why that lock is necessary, but if it is necessary at all, it seems that it should be held on that code path. Alternatively, maybe that lock should be removed entirely?</s>
- Don't just add files to the layer’s directory; if we find anything pre-existing, remove that, including the symbolic link in `linkDir`, to ensure the layer is created in a known state.

Compare also **EDIT <s>#1133</s> #1139**.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -939,6 +939,16 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, disable
 		rootUID = int(st.UID())
 		rootGID = int(st.GID())
 	}
+
+	if _, err := system.Lstat(dir); err == nil {
+		logrus.Warnf("Trying to create a layer %#v while directory %q already exists; removing it first", id, dir)
+		// Don’t just os.RemoveAll(dir) here; d.Remove also removes the link in linkDir,
+		// so that we can’t end up with two symlinks in linkDir pointing to the same layer.
+		if err := d.Remove(id); err != nil {
+			return errors.Wrapf(err, "removing a pre-existing layer directory %q", dir)
+		}
+	}
+
 	if err := idtools.MkdirAllAndChownNew(dir, 0700, idPair); err != nil {
 		return err
 	}


mtrmac:
Note that removing a layer like this is also potentially risky — we can’t necessarily trust the contents of the `link` or `additionallayer` files in the layer’s directory.

mtrmac:
OK, I have tested this now and verified that it works as expected.

OTOH #1139 might be a better fix — I’d very much love for someone who understands the design intent to weigh in.

mtrmac:
Podman tests: https://github.com/containers/podman/pull/13315

mtrmac:
Marking as ready to review — please consider this _only_ in tandem with #1148.

mtrmac:
@nalind @giuseppe PTAL. This might be very wrong, I’d just like to make sure _something_ is done about these failure cases.

nalind:
The in-driver lock appears to be a carryover from docker — everything at this level is supposed to be protected by a LayerStore lock that happened before we called into any driver-specific logic.  The premise looks sound.

nalind:
This looks a decent remediation for cases where a process was killed sometime earlier, while it was attempting to create a layer from a template, which is often done for users of fedora toolbox, so I think this is worth pursuing.

drivers/overlay/overlay.go
```diff
@@ -942,6 +942,20 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, disable
 		rootUID = int(st.UID())
 		rootGID = int(st.GID())
 	}
+
+	// --- Only access "dir" below.
+	d.locker.Lock(id)
+	defer d.locker.Unlock(id)
+
+	if _, err := system.Lstat(dir); err == nil {
+		logrus.Warnf("Trying to create a layer %#v while directory %q already exists; removing it first", id, dir)
```
rhatdan, line 951: We default to showing warnings.  Is this something we want normal users to see by default?  Or should we drop this down to info?

mtrmac: I was thinking that, this indicates an unexpected abort of a c/storage operation, and it is definitely something we want to have captured _in any logs at the default level_ for postmortems if this cleanup is insufficient and/or incorrect and harmful.

I’m not so sure that it needs to be printed by default on a TTY during interactive use; at that point, there are unlikely to be any logs recorded, and the thing either works or it doesn’t. OTOH, it’s also a situation we shouldn’t be regularly getting into, so I don’t think that a warning should _hurt_.

Alternatively, #1148 should make this case unreachable assuming there are no bugs — and in that case we either don’t need this PR, or we can happily leave it on a warning level because it is clearly very unexpected. If I’m reading @nalind’s comments right, it seems preferable to have both this PR and #1148.

rhatdan: Ok leave it at Warning level.

rhatdan:
LGTM

mtrmac:
> The in-driver lock appears to be a carryover from docker — everything at this level is supposed to be protected by a LayerStore lock that happened before we called into any driver-specific logic.

OK, let’s discuss possibly removing that in #1214 . (That would break this PR, I will rebase either one as necessary.)

rhatdan:
After meeting with @nalind and @mtrmac we have decided to merge both PRs.

Issue #1143, opened 2022-02-20
podman run with `--uidmap` options randomly causes `Host ID ${HOSTID} cannot be mapped to a container ID` error


nia1048596:
Description:

If we specify `--uidmap` options to create container, the storage driver will `chown` files in container image filesystem to the specified uid/gid range.

https://github.com/containers/storage/blob/3db688341be34aa1e7c65b2944ac7aa2c81ad216/drivers/chown_unix.go#L15-L31

There is a race condition here during `chown`ing. The above function `platformLChown` gets called in `pwalk.walk`, which spawns many go routines to `chown` in parallel.

https://github.com/containers/storage/blob/3db688341be34aa1e7c65b2944ac7aa2c81ad216/drivers/chown.go#L53-L62

If there are hard linked files, they might be `chown`ed many times. After the first `chown`, the second `chown` will fail to map its uid/gid, which causes the `Host ID ${HOSTID} cannot be mapped to a container ID` error.

Steps to reproduce:

We can use `busybox` as example image, because there are 400 hard links in `/bin/*`, all files linking to `/bin/busybox`.

1. remove `busybox:latest` image

    ```
    podman rmi busybox:latest
    ```

2. setup podman subuid/subgid range

    ```
     $ grep podman /etc/{subuid,subgid}
    /etc/subuid:podman:2100000:2000000
    /etc/subgid:podman:2100000:2000000
    ```

3. launch new busybox container with `--uidmap=2200000`

   ```
    $ sudo podman run --rm -it --uidmap=0:2200000:65536 --gidmap=0:2200000:65536 docker.io/library/busybox true
   Error: error creating container storage: error creating an ID-mapped copy of layer "d31505fd5050f6b96ca3268d1db58fc91ae561ddf14eaabc41d63ea2ef8c1c6d": exit status 1: error during chown: error mapping host ID pair idtools.IDPair{UID:2200000, GID:2200000} for "bin/adduser" to container: Host ID 2200000 cannot be mapped to a container ID
   ```

`/bin/adduser`'s uid was formerly `chown`ed by other `/bin/*` files to `2200000`, but the expected uid is `2100000` which is the default subuid range for `podman` user in step 2.

giuseppe:
opened a PR: https://github.com/containers/storage/pull/1144

PR #1144, opened 2022-02-21
drivers, chown: support chown of hard links

giuseppe:
make sure the same inode is not chowned twice.  Track all the inodes that are chowned and skip the same inode if it is encountered multiple times.

Closes: https://github.com/containers/storage/issues/1143


rhatdan (approved):
LGTM

rhatdan:
@saschagrunert @flouthoc @vrothberg @nalind PTAL

nalind:
LGTM

Issue #1147, opened 2022-02-22
Failures creating a container/image/layer leaves a record in memory


mtrmac:
`layerStore.Put`, on various error paths, calls `layerStore.Driver.Remove`, but not `layerStore.Delete`; therefore, the failed layer is still kept around in memory and can be found and used within that process (and could even, potentially, be written to disk on any future `Save`, if that succeeds).

That wouldn’t be _that_ much of an issue of a failure to create a layer always caused a process termination, but that’s not the case for long-lived processes like CRI-O, and even Podman seems to default to 3 retries via `c/common/pkg.retry` (admittedly in _fairly_, but not _extremely_, limited situations).

I’m not sure what all the implications of this are. My head hurts just at the _thought of trying_ to enumerate that.

The “obvious” fix of calling `layerStore.Delete` on most error paths would test the limits of resiliency of the generic layer the deletion code, and graph drivers’ layer removal code, in previously-possibly-unanticipated ways.

mtrmac:
The same issue exists for images and containers.

mtrmac:
This was actually fixed in #1148 , AFAICS.

PR #1148, opened 2022-02-22
(Alternative to 1140): Record layers as incomplete before trying to create them

mtrmac:
This implements the suggestion in #1139, along with **EDIT partially** fixing #1147 ; as an alternative to #1140 .

Conceptually I like this much better than #1140, but it’s quite a bit more invasive, and potentially risky by invoking the layer removal code in previously-unreachable situations.

Marking as draft because it’s completely untested so far.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -841,19 +841,17 @@ func (r *layerStore) Put(id string, parentLayer *Layer, names []string, mountLab
 	for flag, value := range flags {
 		layer.Flags[flag] = value
 	}
-	savedIncompleteLayer := false
-	if diff != nil {
-		layer.Flags[incompleteFlag] = true
-		err = r.Save()
-		if err != nil {
-			// We don't have a record of this layer, but at least
-			// try to clean it up underneath us.
-			if err2 := r.driver.Remove(id); err2 != nil {
-				logrus.Errorf("While recovering from a failure saving incomplete layer metadata, error deleting layer %#v: %v", id, err2)
-			}
-			return nil, -1, err
+	layer.Flags[incompleteFlag] = true
+	err = r.Save()
+	if err != nil {
+		// We don't have a record of this layer, but at least
+		// try to clean it up underneath us.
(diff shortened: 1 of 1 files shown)


rhatdan:
Should we open a PR against this in Podman to get its test suite to run against it?

layers.go
```diff
@@ -732,105 +731,102 @@ func (r *layerStore) Put(id string, parentLayer *Layer, names []string, mountLab
...
+		r.byname[name] = layer
+	}
+	for flag, value := range flags {
+		layer.Flags[flag] = value
+	}
+	layer.Flags[incompleteFlag] = true
+
+	succeeded := false
+	cleanupFailureContext := ""
+	defer func() {
```
mtrmac, line 761: Note that using `defer` means that we would try to delete the WIP layer even on a panic. That’s probably correct for some banal `nil` dereference, but might make things worse if the in-memory state were significantly corrupt.

mtrmac:
> Marking as draft because it’s completely untested so far.

Tested now, at least for success and observing the intermediate states. Note that any ambition to being crash-resilient requires a solution to #1136 .

mtrmac:
> Should we open a PR against this in Podman to get its test suite to run against it?

Good idea, https://github.com/containers/podman/pull/13316 . (And similarly for the other PR.)

mtrmac:
Marking as ready to review — please consider this _only_ in tandem with #1140.

mtrmac:
@nalind @giuseppe PTAL. This might be very wrong, I’d just like to make sure _something_ is done about these failure cases.

nalind:
This looks like it has a decent chance of avoiding the types of cases that #1140 would have to clean up on behalf of binaries using a version of the storage library that doesn't include this change, so it looks like it's worth pursuing.

rhatdan:
Merging both PRs with agreement from @nalind and @mtrmac

PR #1152, opened 2022-02-23
image, layers, container: append new names to the ones which were written before and `dedupe`

flouthoc:

override if needed.

Following patch fixes concurrent builds via podman or buildah.

See: https://github.com/containers/buildah/pull/3794 for more details.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -378,11 +378,11 @@ func (r *containerStore) removeName(container *Container, name string) {
 }
 
 func (r *containerStore) SetNames(id string, names []string) error {
-	names = dedupeNames(names)
 	if container, ok := r.lookup(id); ok {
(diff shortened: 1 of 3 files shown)


giuseppe:
I think we need a new function to expose the functionality, otherwise it is a breaking change.

Maybe `AppendNames`?

flouthoc:
This is breaking change, this PR must be replaced with `SetNamesWithOptions` and all the `podman,buildah` etc and dependent code must use that, where deletion is not needed e.g `concurrent builds, tagging` etc.

flouthoc:
@giuseppe I'm closing this PR since i have few tests on this branch and created and final one which is here: https://github.com/containers/storage/pull/1153

Issue #595, opened 2020-04-13
Being-saved images create errors up the stack


nalind:
Currently, to commit an image in the image library's `containers-storage` transport, we create and populate layers, create the image record, and then save the non-layers under the directory that matches that image record's ID.  The manifest and config blobs are among those non-layers.

This creates a small window where there's an image record without a manifest and/or config blob, and if another thread tries to read information about the image during that window, it will get unexpected errors, and those errors percolating up the stack is a problem.

Option 1: add an `incomplete` flag that we use for screening out values that we'd otherwise return in an `Images()` list, and clean them up when we get a write lock, as we do for incomplete layers.  Won't work because we don't keep the lock during the commit operation, so things can get cleaned up from under us during the window.
Option 2: screen out values from the `Images()` list if they don't include a manifest.  The commit logic doesn't need to be changed because it gets the ID when it creates the image record, and uses that ID to save the manifest for the image.  Downsides: if we crash, the phantom image record could stay there forever, since it would be filtered for everyone who didn't know its ID.  The screening would also need to check for a config blob, which requires knowledge that lives in containers/image (schema 1 images don't have them, and logic for determining what kind of manifest we have is in containers/image), which we can't use as a dependency.
Option 3: if we fail to read an image's manifest, spin and try again.  Downside: hacky, and we can only guess how long we should spin before giving up and reporting an error.
Option 4: expose a way to manipulate `Flags` in the image record, add a flag that containers/image will set on works in progress, and teach everyone to skip over images that have the flag set.  Downside: kind of sucks for everyone.
Option 5: ???

rhatdan:
@vrothberg @saschagrunert @mtrmac @giuseppe @TomSweeneyRedHat @baude @mheon  @mrunalp WDYT?

TomSweeneyRedHat:
For option 1, could we add a lock to the commit operation to avoid the issue you're concerned about?

mrunalp:
I like option 2. Can we run some kind of gc from daemons like CRI-O periodically to remove the phantom records?

giuseppe:
could we write these files before the image is recorded?  We'd still need to have some kind of GC if these files are written but the image wasn't recorded.

Another option would be to use a staging area, similarly to OSTree, so we can write these files to a temporary (but guessable) location, e.g.: `/var/lib/containers/storage/overlay-images/staging/IMAGE-ID/manifest`

When we lookup for the manifest, we'd need to check like:

```
[...]

PR #1153, opened 2022-02-23
race, storage:  add independent `AddNames` and `RemoveNames` for images,layers and containers

flouthoc:
Adds `SetNameWithOptions` so operations which are invoke in parallel

For instance

PR replaces: https://github.com/containers/storage/pull/1152

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -84,7 +84,16 @@ type ContainerStore interface {
 
 	// SetNames updates the list of names associated with the container
 	// with the specified ID.
+	// Deprecated: Prone to race conditions, suggested alternatives are `AddNames` and `RemoveNames`.
 	SetNames(id string, names []string) error
+
+	// AddNames adds the supplied values to the list of names associated with the container with
+	// the specified id.
+	AddNames(id string, names []string) error
+
+	// RemoveNames removes the supplied values from the list of names associated with the container with
+	// the specified id.
+	RemoveNames(id string, names []string) error
 
 	// Get retrieves information about a container given an ID or name.
 	Get(id string) (*Container, error)
@@ -377,22 +386,40 @@ func (r *containerStore) removeName(container *Container, name string) {
 	container.Names = stringSliceWithoutValue(container.Names, name)
 }
 
+// Deprecated: Prone to race conditions, suggested alternatives are `AddNames` and `RemoveNames`.
 func (r *containerStore) SetNames(id string, names []string) error {
-	names = dedupeNames(names)
-	if container, ok := r.lookup(id); ok {
-		for _, name := range container.Names {
-			delete(r.byname, name)
-		}
-		for _, name := range names {
-			if otherContainer, ok := r.byname[name]; ok {
-				r.removeName(otherContainer, name)
-			}
-			r.byname[name] = container
-		}
-		container.Names = names
-		return r.Save()
+	return r.updateNames(id, names, setNames)
+}
+
+func (r *containerStore) AddNames(id string, names []string) error {
+	return r.updateNames(id, names, addNames)
+}
+
+func (r *containerStore) RemoveNames(id string, names []string) error {
+	return r.updateNames(id, names, removeNames)
+}
+
+func (r *containerStore) updateNames(id string, names []string, op updateNameOperation) error {
+	container, ok := r.lookup(id)
+	if !ok {
+		return ErrContainerUnknown
 	}
-	return ErrContainerUnknown
+	oldNames := container.Names
+	names, err := applyNameOperation(oldNames, names, op)
+	if err != nil {
+		return err
+	}
+	for _, name := range oldNames {
+		delete(r.byname, name)
+	}
+	for _, name := range names {
+		if otherContainer, ok := r.byname[name]; ok {
+			r.removeName(otherContainer, name)
+		}
+		r.byname[name] = container
+	}
+	container.Names = names
+	return r.Save()
 }
 
 func (r *containerStore) Delete(id string) error {
--- a/images.go
+++ b/images.go
@@ -136,7 +136,18 @@ type ImageStore interface {
 	// SetNames replaces the list of names associated with an image with the
 	// supplied values.  The values are expected to be valid normalized
 	// named image references.
+	// Deprecated: Prone to race conditions, suggested alternatives are `AddNames` and `RemoveNames`.
 	SetNames(id string, names []string) error
+
+	// AddNames adds the supplied values to the list of names associated with the image with
+	// the specified id. The values are expected to be valid normalized
+	// named image references.
+	AddNames(id string, names []string) error
+
+	// RemoveNames removes the supplied values from the list of names associated with the image with
+	// the specified id.  The values are expected to be valid normalized
+	// named image references.
+	RemoveNames(id string, names []string) error
 
 	// Delete removes the record of the image.
 	Delete(id string) error
@@ -505,26 +516,44 @@ func (i *Image) addNameToHistory(name string) {
 	i.NamesHistory = dedupeNames(append([]string{name}, i.NamesHistory...))
 }
 
+// Deprecated: Prone to race conditions, suggested alternatives are `AddNames` and `RemoveNames`.
 func (r *imageStore) SetNames(id string, names []string) error {
+	return r.updateNames(id, names, setNames)
+}
+
+func (r *imageStore) AddNames(id string, names []string) error {
+	return r.updateNames(id, names, addNames)
+}
+
+func (r *imageStore) RemoveNames(id string, names []string) error {
+	return r.updateNames(id, names, removeNames)
+}
+
+func (r *imageStore) updateNames(id string, names []string, op updateNameOperation) error {
 	if !r.IsReadWrite() {
 		return errors.Wrapf(ErrStoreIsReadOnly, "not allowed to change image name assignments at %q", r.imagespath())
 	}
-	names = dedupeNames(names)
-	if image, ok := r.lookup(id); ok {
-		for _, name := range image.Names {
-			delete(r.byname, name)
-		}
-		for _, name := range names {
-			if otherImage, ok := r.byname[name]; ok {
-				r.removeName(otherImage, name)
-			}
-			r.byname[name] = image
-			image.addNameToHistory(name)
-		}
-		image.Names = names
-		return r.Save()
+	image, ok := r.lookup(id)
+	if !ok {
+		return errors.Wrapf(ErrImageUnknown, "error locating image with ID %q", id)
 	}
-	return errors.Wrapf(ErrImageUnknown, "error locating image with ID %q", id)
+	oldNames := image.Names
+	names, err := applyNameOperation(oldNames, names, op)
+	if err != nil {
+		return err
+	}
+	for _, name := range oldNames {
+		delete(r.byname, name)
+	}
+	for _, name := range names {
+		if otherImage, ok := r.byname[name]; ok {
+			r.removeName(otherImage, name)
+		}
+		r.byname[name] = image
+		image.addNameToHistory(name)
+	}
+	image.Names = names
+	return r.Save()
 }
 
 func (r *imageStore) Delete(id string) error {
--- a/images_test.go
+++ b/images_test.go
@@ -91,4 +91,16 @@ func TestHistoryNames(t *testing.T) {
 	require.Len(t, secondImage.NamesHistory, 2)
 	require.Equal(t, secondImage.NamesHistory[0], "3")
 	require.Equal(t, secondImage.NamesHistory[1], "2")
+
+	// test independent add and remove operations
+	require.Nil(t, store.AddNames(firstImageID, []string{"5"}))
+	firstImage, err = store.Get(firstImageID)
+	require.Nil(t, err)
+	require.Equal(t, firstImage.NamesHistory, []string{"4", "3", "2", "1", "5"})
+
+	// history should still contain old values
+	require.Nil(t, store.RemoveNames(firstImageID, []string{"5"}))
+	firstImage, err = store.Get(firstImageID)
+	require.Nil(t, err)
+	require.Equal(t, firstImage.NamesHistory, []string{"4", "3", "2", "1", "5"})
 }
--- a/store.go
+++ b/store.go
@@ -31,6 +31,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+type updateNameOperation int
+
+const (
+	setNames updateNameOperation = iota
+	addNames
+	removeNames
+)
+
 var (
 	stores     []*store
 	storesLock sync.Mutex
@@ -368,7 +376,16 @@ type Store interface {
 
 	// SetNames changes the list of names for a layer, image, or container.
 	// Duplicate names are removed from the list automatically.
+	// Deprecated: Prone to race conditions, suggested alternatives are `AddNames` and `RemoveNames`.
 	SetNames(id string, names []string) error
+
+	// AddNames adds the list of names for a layer, image, or container.
+	// Duplicate names are removed from the list automatically.
+	AddNames(id string, names []string) error
+
+	// RemoveNames removes the list of names for a layer, image, or container.
+	// Duplicate names are removed from the list automatically.
+	RemoveNames(id string, names []string) error
 
 	// ListImageBigData retrieves a list of the (possibly large) chunks of
 	// named data associated with an image.
@@ -2050,7 +2067,20 @@ func dedupeNames(names []string) []string {
 	return deduped
 }
 
+// Deprecated: Prone to race conditions, suggested alternatives are `AddNames` and `RemoveNames`.
 func (s *store) SetNames(id string, names []string) error {
+	return s.updateNames(id, names, setNames)
+}
+
+func (s *store) AddNames(id string, names []string) error {
+	return s.updateNames(id, names, addNames)
+}
+
+func (s *store) RemoveNames(id string, names []string) error {
+	return s.updateNames(id, names, removeNames)
+}
+
+func (s *store) updateNames(id string, names []string, op updateNameOperation) error {
 	deduped := dedupeNames(names)
 
 	rlstore, err := s.LayerStore()
@@ -2063,7 +2093,16 @@ func (s *store) SetNames(id string, names []string) error {
 		return err
 	}
 	if rlstore.Exists(id) {
-		return rlstore.SetNames(id, deduped)
+		switch op {
+		case setNames:
+			return rlstore.SetNames(id, deduped)
+		case removeNames:
+			return rlstore.RemoveNames(id, deduped)
+		case addNames:
+			return rlstore.AddNames(id, deduped)
+		default:
+			return errInvalidUpdateNameOperation
+		}
 	}
 
 	ristore, err := s.ImageStore()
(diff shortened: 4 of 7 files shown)


containers.go
```diff
@@ -84,7 +84,7 @@ type ContainerStore interface {
 
 	// SetNames updates the list of names associated with the container
 	// with the specified ID.
-	SetNames(id string, names []string) error
+	SetNames(id string, names []string, opts SetNameOptions) error
```
vrothberg, line 87: Still a breaking change.

vrothberg (commented):
Would a `{Add,Delete}Name` API address the races?

containers.go
```diff
@@ -377,11 +377,18 @@ func (r *containerStore) removeName(container *Container, name string) {
...
 		for _, name := range container.Names {
-			delete(r.byname, name)
+			if opts.AppendOnly {
+				names = append(names, name)
+			} else {
+				delete(r.byname, name)
+			}
+		}
+		if opts.AppendOnly {
+			names = dedupeNames(names)
```
giuseppe, line 391: I think you can drop the first dedupe at line 381 and just do it here not guarded by `opts.AppendOnly`.

containers.go
flouthoc: @vrothberg Yeah i was not aware that container store is exposed dropped it.

containers.go
flouthoc: yes. that would be better.

flouthoc:
> Would a `{Add,Delete}Name` API address the races?

@vrothberg Bunch of code in `common or podman or buildah` needs to be instrumented slowly. So this seems most intuitive from what we discussed here   https://github.com/containers/storage/pull/1152 . 

There are some other things as well where we might have race but maybe that is part of a different PR.

giuseppe (approved):
code changes LGTM

giuseppe:
> Would a {Add,Delete}Name API address the races?

Wouldn't that be more difficult to handle for the caller?  If it is adding multiple names and the operation fails middle way, should it then try to delete previously added ones?

containers.go
```diff
@@ -378,11 +382,22 @@ func (r *containerStore) removeName(container *Container, name string) {
...
-			delete(r.byname, name)
+			if opts.AppendOnly {
+				names = append(names, name)
+			} else {
+				delete(r.byname, name)
+			}
 		}
+		names = dedupeNames(names)
 		for _, name := range names {
 			if otherContainer, ok := r.byname[name]; ok {
```
mtrmac, line 402: This looks quite broken: Not deleting `container` from `r.byname` will set `otherContainer` to `container` AFAICS.

I think that’s ultimately hidden by the `container.Names` assignment below, but it’s just not a clean and consistent way to implement this.

---

Same in the other two implementations.

---

One way to do this would be not to change the basic structure, only to change what names are set. I.e. keep the original `delete(r.byname)` loop as is, _then_ compute the final `names` value, and then proceed with the `otherContainer/byname/…` updates using the existing logic.

That would have the advantage that `updatedNames := applyEdits(container.Names, inputNames, opts)` could be shared across the three implementations.

Another way is to indeed deeply embed the add/remove cases into the implementation, so that names that don’t change are neither removed nor added in `r.byname`. That’s _technically_ more efficient but IMHO not worth the complexity.

mtrmac (commented):
If we need atomic name additions, we _certainly_ also need atomic name removals. So I’d expect something like  `UpdateNames(id string, updates NameUpdates)` with `type NameUpdates{ struct addNames, removeNames []string }`

There was also an issue I can’t quickly find about `store.CreateImage` doing name updates and other changes atomically with other operations, that’s why I suggest a `type NameUpdates` that could be reused in that other feature.

mtrmac:
> @vrothberg Bunch of code in `common or podman or buildah` needs to be instrumented slowly.

What difference does that make? Every line calling `SetNames` should to be edited either way, whatever form we design for the new API.

mtrmac (commented):
.

mtrmac:
Also, if we know the old APIs are racy, they should be documented as deprecated.

vrothberg:
> > @vrothberg Bunch of code in `common or podman or buildah` needs to be instrumented slowly.
> 
> What difference does that make? Every line calling `SetNames` should be edited either way, whatever form we design for the new API.

Yes, `SetNames` should not be used anymore. IIUIC, the race is in reading all data, followed by writing all data. An add(), delete() could just take care of synchronization and callers don't have to worry.

mtrmac:
> There was also an issue I can’t quickly find about `store.CreateImage` doing name updates and other changes atomically with other operations

https://github.com/containers/storage/issues/595

containers.go
mtrmac: … actually, with a clean `addNames` / `removeNames` (were we do need to handle no-op names but they are only no-op additions / no-op removals) the “deeply embed” approach also might be simple enough.

store.go
```diff
@@ -2051,6 +2060,13 @@ func dedupeNames(names []string) []string {
 }
 
 func (s *store) SetNames(id string, names []string) error {
+	opts := SetNameOptions{
+		AppendOnly: false,
+	}
```
nalind, line 2065: This is not a valid assumption, and could break untagging unless callers switch to `SetNamesWithOptions`.  An `AddNames/RemoveNames` pair instead of `SetNamesWithOptions()` would avoid this.

containers.go
flouthoc: Yeah this part is not needed. Let me alter that, lets discuss new in original thread.

flouthoc:
If we have a consensus on new API then i guess its cool. We can create a new API.

containers.go
```diff
@@ -395,6 +403,50 @@ func (r *containerStore) SetNames(id string, names []string) error {
 	return ErrContainerUnknown
 }
 
+func (r *containerStore) AddNames(id string, names []string) error {
+	if container, ok := r.lookup(id); ok {
```
vrothberg, line 407: I'd much prefer to error out early:
```go
container, found := r.lookup(id)
if !found {
    return ErrContainer...
}
// logic
```

containers.go
```diff
@@ -86,6 +86,14 @@ type ContainerStore interface {
 	// with the specified ID.
 	SetNames(id string, names []string) error
 
+	// AddNames adds the list of names associated with a container to storage with the
+	// supplied values.
```
vrothberg, line 90: I am not a native speaker but something similar to the comment below easier to brain parse:
```suggestion
	// AddNames adds the supplied values to the list of names associated with the container with the specified id.
```
Same at other locations.

containers.go
```diff
@@ -395,6 +403,50 @@ func (r *containerStore) SetNames(id string, names []string) error {
...
+		}
+		names = dedupeNames(names)
+		for _, name := range names {
+			if otherContainer, ok := r.byname[name]; ok {
+				r.removeName(otherContainer, name)
+			}
+			r.byname[name] = container
+		}
+		container.Names = names
+		return r.Save()
```
vrothberg, line 420: I am not entirely familiar with the code but I am surprised we're deleting something. I'd imagine something similar to:
```golang
container.Names = dedupeNames(append(containers.Names, names...))
```

vrothberg (commented):
Not at all a thorough review

containers.go
vrothberg: Scratch that: having read more code, we need to delete the previous references to other containers.

mtrmac (commented):
I’d prefer the `struct` approach from https://github.com/containers/storage/pull/1153#pullrequestreview-891044945 because it’s much easier to embed into a larger operation, as discussed above.

But I’ll completely defer to c/storage maintainers.

containers.go
```diff
@@ -395,6 +403,50 @@ func (r *containerStore) SetNames(id string, names []string) error {
...
+
+func (r *containerStore) RemoveNames(id string, names []string) error {
+	namesFiltered := make([]string, 0)
+	if container, ok := r.lookup(id); ok {
+		for _, name := range container.Names {
+			for _, n := range names {
+				if n == name {
+					continue
+				}
+			}
```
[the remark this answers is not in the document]
flouthoc: Oh my bad , there was a flag. I forgot to add it.

containers.go
```diff
@@ -395,6 +403,50 @@ func (r *containerStore) SetNames(id string, names []string) error {
 	return ErrContainerUnknown
 }
 
+func (r *containerStore) AddNames(id string, names []string) error {
```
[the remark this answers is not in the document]
flouthoc: Yes this can be done. Just like we do in store.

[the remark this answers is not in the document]
mtrmac: BTW there’s also the option of the _external_ API having separate `Add`/`Remove`/`Set`, but the three internal stores having an `Edit` with a struct that describes all three.

Up to c/storage maintainers.

images.go
```diff
@@ -527,6 +541,61 @@ func (r *imageStore) SetNames(id string, names []string) error {
...
+				if n == name {
+					continue
+				}
+			}
+			delete(r.byname, name)
+			namesFiltered = append(namesFiltered, name)
+		}
+		names = dedupeNames(namesFiltered)
+		// clear older entries from history since we are removing images
+		image.removeHistory()
```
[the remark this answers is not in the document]
mtrmac: AFAICS it _only_ adds the newly-set names; not the names that are being removed by this operation, nor the names that have been in the history in the distant past (and not even in the pre-removal `.Names`).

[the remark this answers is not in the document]
flouthoc: We pull names from current image and then re-write the ones which are only needed. I also added a test to verify this behavior.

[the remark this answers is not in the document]
flouthoc: @mtrmac got it let me rectify the test itself and remove the cleanup of history. I'll verify and re-open PR from draft.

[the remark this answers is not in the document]
flouthoc: fixed.

containers.go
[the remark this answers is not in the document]
flouthoc: Did it with `enum` just like i did it for store.

containers.go
flouthoc: @vrothberg Is there anything for me to do here ? or could you mark resolved.

store.go
flouthoc: implemented in below commit.

store.go
```diff
@@ -31,6 +31,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+type UpdateNameOperation int
+
+const (
+	SetNames UpdateNameOperation = iota
+	AddNames
+	RemoveNames
+)
```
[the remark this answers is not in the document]
flouthoc: It is being used by other stores as well.

store.go
```diff
@@ -2051,6 +2067,18 @@ func dedupeNames(names []string) []string {
 }
 
 func (s *store) SetNames(id string, names []string) error {
+	return s.updateNames(id, names, setNames)
+}
+
+func (s *store) AddNames(id string, names []string) error {
```
mtrmac, line 2073: I’m not sure whether to go there — others, feel free to say this is not the time.

If we are adding a new top-level API, do we _have_ to follow the generic `SetNames` model, where the code needs to lock all three stores and guess which one the user meant? I’d expect the user to _always_ know whether the goal is to set a name of a container/layer/image. So we could have `Add{Container,Image,Layer}Names`, that would actually be more efficient than the API provided now, and the tedious `switch` statements in `store.updateNames` would not be necessary.

mtrmac (commented):
The `containers.go` comments apply to the other two equally.

containers.go
```diff
@@ -378,21 +386,57 @@ func (r *containerStore) removeName(container *Container, name string) {
 }
 
 func (r *containerStore) SetNames(id string, names []string) error {
```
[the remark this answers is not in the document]
flouthoc: fixed

store.go
```diff
@@ -2051,6 +2067,18 @@ func dedupeNames(names []string) []string {
 }
 
 func (s *store) SetNames(id string, names []string) error {
```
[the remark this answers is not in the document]
flouthoc: fixed

containers.go
```diff
@@ -378,21 +386,57 @@ func (r *containerStore) removeName(container *Container, name string) {
...
+}
+
+func (r *containerStore) updateNames(id string, names []string, op UpdateNameOperation) error {
+	namesFiltered := make([]string, 0)
+	container, ok := r.lookup(id)
+	if !ok {
+		return ErrContainerUnknown
+	}
+	for _, name := range container.Names {
+		if op == setNames {
```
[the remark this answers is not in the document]
flouthoc: fixed.

containers.go
```diff
@@ -378,21 +386,57 @@ func (r *containerStore) removeName(container *Container, name string) {
...
+		if op == removeNames {
+			found := false
+			for _, n := range names {
+				if n == name {
+					found = true
+				}
+			}
+			if !found {
+				delete(r.byname, name)
+				namesFiltered = append(namesFiltered, name)
```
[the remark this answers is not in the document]
flouthoc: Nice idea, I have implemented similar in new commit.

containers.go
```diff
@@ -84,8 +84,18 @@ type ContainerStore interface {
 
 	// SetNames updates the list of names associated with the container
 	// with the specified ID.
+	// SetNames is deprecated and prone to race conditions.
```
mtrmac, line 87: Please follow https://github.com/golang/go/wiki/Deprecated  in these notices.

utils.go
```diff
@@ -40,3 +41,27 @@ func validateMountOptions(mountOptions []string) error {
...
+func applyNameOperation(oldNames []string, names []string, op updateNameOperation) ([]string, error) {
+	result := make([]string, 0)
+	switch op {
+	case setNames:
+		// ignore all old names and just return new names
+		return names, nil
+	case removeNames:
+		// remove given names from old names
+		for _, name := range oldNames {
+			if !stringutils.InSlice(names, name) {
```
mtrmac, line 54: I don’t think the case-folding behavior of `stringutils.InSlice` is expected here.

images_test.go
```diff
@@ -91,4 +91,18 @@ func TestHistoryNames(t *testing.T) {
 	require.Len(t, secondImage.NamesHistory, 2)
 	require.Equal(t, secondImage.NamesHistory[0], "3")
 	require.Equal(t, secondImage.NamesHistory[1], "2")
+
+	// test independent add and remove operations
+	require.Nil(t, store.AddNames(firstImageID, []string{"5"}))
+	firstImage, err = store.Get(firstImageID)
+	require.Nil(t, err)
+	require.Len(t, firstImage.NamesHistory, 5)
```
mtrmac, line 99: (Even the `.Len` checks can be dropped, the deep equality comparison also implies the same length, AFAIK.)

mtrmac (commented):
Implementation mostly LGTM. I’ll defer to c/storage experts on the (public and internal) API.

giuseppe:
do you also have a patch for buildah that uses these new APIs?

It would be helpful to validate the change and see if it fixes the race condition before we merge

flouthoc:
> do you also have a patch for buildah that uses these new APIs?
> 
> It would be helpful to validate the change and see if it fixes the race condition before we merge

@giuseppe I think new API has to go in `c/common/libimage. Tag` no patch in buildah is needed AFAICS.

utils.go
flouthoc: using custom loop here, stringutils seems to be case insensitive, maybe later we can add a new function in stringutils itself.

utils.go
```diff
@@ -40,3 +40,35 @@ func validateMountOptions(mountOptions []string) error {
...
+	switch op {
+	case setNames:
+		// ignore all old names and just return new names
+		return names, nil
+	case removeNames:
+		// remove given names from old names
+		for _, name := range oldNames {
+			// only keep names in final result which do not intersect with input names
+			// basically `result = oldNames - names`
+			nameExistsInOldNames := false
```
mtrmac, line 55: `nameExistsInOldNames` is guaranteed true by the loop containing this line.

Isn’t this variable actually `nameShouldBeRemoved`, or `nameExistsInNames`?[1]


([1] an illustration for the rationale of naming the input value `opParameters`, to make it distinct from all the “names” variables around.)

flouthoc: I also think `nameShouldBeRemoved` is better, and switched `names` -> `opParameters` for better readability.

flouthoc:
@containers/podman-maintainers @containers/storage-maintainers Following PR and use-case is verified here: https://github.com/containers/podman/pull/13339

giuseppe (approved):
LGTM

vrothberg:
`podman/buildah untag $IMAGE` without a tag will remove all tags/names from the specified image.  To prevent race conditions there, we'd need means to remove all names from an image.

Could we do this in case of `RemoveNames(id, nil)` (i.e., when the argument is `nil`)?

flouthoc:
@vrothberg They can still use `SetNames` for that, cause deleting all names will anyways produce race for other builds.

vrothberg:
> @vrothberg They can still use `SetNames` for that, cause deleting all names will anyways produce race for other builds.

`SetNames` is deprecated, so I don't think any caller should continue using it.

flouthoc:
@vrothberg How did we used to solve this with `SetNames` did we used to pass empty array to `SetNames` ?

vrothberg:
> @vrothberg How did we used to solve this with `SetNames` did we used to pass empty array to `SetNames` ?

Currently, we're first reading all names, and then untag them one-by-one.  But I think we can continue doing that what we're doing and leave this PR as is.  

LGTM

giuseppe:
@nalind are you fine to merge the last version?

mtrmac:
> Could we do this in case of `RemoveNames(id, nil)` (i.e., when the argument is `nil`)?

(I understand we are not planning to do this, so just a side note:) That would be a rather irregular behavior, and potentially a trap to a caller that builds the list of names from components.

We are not going to run out of strings, so adding a special `RemoveAllImageNames` (should that even include digested references?) is always an option.

mtrmac (commented):
Code LGTM, I’ll defer to others on the API, e.g. (https://github.com/containers/storage/pull/1153#discussion_r813256132 https://github.com/containers/storage/pull/1153#discussion_r813273185 ). From my POV both of these ideas are not a reason to block this PR.

containers.go
```diff
@@ -86,6 +86,14 @@ type ContainerStore interface {
 	// with the specified ID.
 	SetNames(id string, names []string) error
 
+	// AddNames adds the supplied values to the list of names associated with the container with
+	// the specified id.
+	AddNames(id string, names []string) error
+
+	// RemoveNames removes the supplied values from the list of names associated with the container with
+	// the specified id.
+	RemoveNames(id string, names []string) error
```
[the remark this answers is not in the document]
nalind: I'm not attached to the new methods being public.  The early guess was that we'd find ourselves needing to call into LayerStore/ImageStore/ContainerStore methods directly far more often than we do now, but I don't think that's turned out to be the case in practice.

utils.go
```diff
@@ -40,3 +40,35 @@ func validateMountOptions(mountOptions []string) error {
 	}
 	return nil
 }
+
+func applyNameOperation(oldNames []string, opParameters []string, op updateNameOperation) ([]string, error) {
```
nalind, line 44: If all of the name manipulation logic ends up running through here, it's a good place to do the deduplication.

giuseppe (approved):
LGTM

PR #1155, opened 2022-02-23
insufficient IDs errors: mention podman-system-migrate

vrothberg:
Based on a conversation on the Podman mailing list:
Mentioning podman-system-migrate in the error message may help users
resolve their issues faster.

Signed-off-by: Valentin Rothberg <vrothberg@redhat.com>

@giuseppe PTAL
It would be displayed for non-Podman clients as well but I think that's a fair trade off.


giuseppe (approved):
LGTM

flouthoc (approved):
LGTM

rhatdan:
It would be nice if this returned a constant error and then Podman could add the text, but this is good for now.

PR #1156, opened 2022-02-28
Preserve extra overlayfs mount options when loading many-layered images

kousu:
When overlayfs loads a container image that has many layers, each becomes
an extra argument to the 'lowerdir' list, and if it has many many many
of them it will try to squeeze more in by rewriting them as relative paths.

Unfortunately, this case was forgetting 'userxattr' and 'volatile' and
any other user-supplied mount options were being forgotten, which meant
that in rootless mode (which needs userxattr to behave properly) you'd
get an EXDEV error anytime you tried to rename() a directory.

This swaps a couple steps around to make it easier to make both cases the same.

This should fix https://github.com/containers/podman/issues/13123, but I haven't actually tested it yet because I haven't figured out how. But I traced enough out to spot the bug and wanted to get some code up before I forgot it.

Cheers!


drivers/overlay/overlay.go
```diff
@@ -1531,6 +1532,9 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		} else {
 			opts = fmt.Sprintf("lowerdir=%s", strings.Join(absLowers, ":"))
 		}
+		if len(optsList) > 0 {
+			opts = fmt.Sprintf("%s,%s", opts, strings.Join(optsList, ","))
+		}
```
kousu, line 1537: This part is the fix. The rest of the patch is just supporting changes for this line.

flouthoc (approved):
Code looks good to me. Just need to try it once.

@containers/storage-maintainers PTAL

flouthoc:
or a dummy PR against `podman` would be a good verifier.

rhatdan:
@giuseppe PTAL

giuseppe (approved):
LGTM

rhatdan:
Thanks @kousu 
LGTM

PR #1157, opened 2022-02-28
overlay: mountFrom read-only mounts using relative paths

nalind:
When we need to shell out to a helper binary to be able to pass relative locations for directories listed in the mount options, whether or not the mount is meant to be writable isn't a factor.


rhatdan:
LGTM

saschagrunert (approved):
LGTM

vrothberg (approved):

PR #1162, opened 2022-03-08
ExecRunnable: prevent logging on stderr when its not expected

flouthoc:
ExecRunnable is invoked by buildah and error from c/storage is logged to
stdout. This creates a drift between the between the behavior which is
generated by rootless and rootful buildah.

This corrects the behavior for rootless buildah so it has parity with
buildah invoked for root user.

This helps in issue being discussed here: https://github.com/containers/buildah/pull/3804#discussion_r820904895


flouthoc:
@containers/storage-maintainers @giuseppe @nalind @rhatdan PTAL

giuseppe (approved):
LGTM

vrothberg (approved):
LGTM

PR #1165, opened 2022-03-16
Move the mount_program autodetection into the graphdriver Init()

dcermak:
This prevents issues inside podman where the `mount_program` is detected and subsequently erased, e.g. via the usage of the `--root` option.

This fixes https://github.com/containers/podman/issues/13459
This fixes https://github.com/containers/podman/issues/13458

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -292,6 +292,31 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 		backingFs = fsName
 	}
 
+	runhome := filepath.Join(options.RunRoot, filepath.Base(home))
+	rootUID, rootGID, err := idtools.GetRootUIDGID(options.UIDMaps, options.GIDMaps)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create the driver home dir
+	if err := idtools.MkdirAllAs(path.Join(home, linkDir), 0700, rootUID, rootGID); err != nil {
+		return nil, err
+	}
+
+	if err := idtools.MkdirAllAs(runhome, 0700, rootUID, rootGID); err != nil {
+		return nil, err
+	}
+
+	if opts.mountProgram == "" {
+		if supported, err := SupportsNativeOverlay(home, runhome); err != nil {
+			return nil, err
+		} else if !supported {
+			if path, err := exec.LookPath("fuse-overlayfs"); err == nil {
+				opts.mountProgram = path
+			}
+		}
+	}
+
 	if opts.mountProgram != "" {
 		if unshare.IsRootless() && isNetworkFileSystem(fsMagic) && opts.forceMask == nil {
 			m := os.FileMode(0700)
@@ -314,20 +339,6 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 		if unshare.IsRootless() && isNetworkFileSystem(fsMagic) {
 			return nil, errors.Wrapf(graphdriver.ErrIncompatibleFS, "A network file system with user namespaces is not supported.  Please use a mount_program")
 		}
-	}
-
-	rootUID, rootGID, err := idtools.GetRootUIDGID(options.UIDMaps, options.GIDMaps)
-	if err != nil {
-		return nil, err
-	}
-
-	// Create the driver home dir
-	if err := idtools.MkdirAllAs(path.Join(home, linkDir), 0700, rootUID, rootGID); err != nil {
-		return nil, err
-	}
-	runhome := filepath.Join(options.RunRoot, filepath.Base(home))
-	if err := idtools.MkdirAllAs(runhome, 0700, rootUID, rootGID); err != nil {
-		return nil, err
 	}
 
 	var usingMetacopy bool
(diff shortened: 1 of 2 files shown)


dcermak:
cc @giuseppe @vogtinator

drivers/overlay/overlay.go
```diff
@@ -330,6 +330,14 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
 		return nil, err
 	}
 
+	supported, err := SupportsNativeOverlay(home, runhome)
+
+	if !supported {
+		if path, err := exec.LookPath("fuse-overlayfs"); err == nil && opts.mountProgram == "" {
+			opts.mountProgram = path
```
Vogtinator, line 337: Shouldn't this happen before the `if opts.mountProgram != "" {` block above?

dcermak: Right, it should, otherwise it could still cause issues in certain situations.

TomSweeneyRedHat:
LGTM, but definitely want a head nod from @giuseppe .  @dcermak TY!  Did you by chance vendor this commit into a WIP Podman PR to validate that the tests work there with these changes?  If not, could you please?

drivers/overlay/overlay.go
```diff
@@ -292,6 +292,15 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
 		backingFs = fsName
 	}
 
+	runhome := filepath.Join(options.RunRoot, filepath.Base(home))
+	supported, err := SupportsNativeOverlay(home, runhome)
```
giuseppe, line 296: should we check for `err`?

drivers/overlay/overlay.go
```diff
@@ -292,6 +292,15 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
 		backingFs = fsName
 	}
 
+	runhome := filepath.Join(options.RunRoot, filepath.Base(home))
```
giuseppe, line 295: `runhome` might not exist at this point.  You'd need to move the `idtools.MkdirAllAs` call before using it

drivers/overlay/overlay.go
```diff
@@ -292,6 +292,15 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
...
+	supported, err := SupportsNativeOverlay(home, runhome)
+
+	if !supported {
+		if path, err := exec.LookPath("fuse-overlayfs"); err == nil && opts.mountProgram == "" {
+			opts.mountProgram = path
```
giuseppe, line 300: if `mountProgram` is already configured, we don't have to override it.

drivers/overlay/overlay.go
dcermak: Fixed, thanks!

drivers/overlay/overlay.go
dcermak: Indeed, the user might have specified a different value and we don't want to remove that one.

drivers/overlay/overlay.go
dcermak: I think we should. All the errors that `SupportsNativeOverlay` reports are rather low level issues that could indicate something is up with your file system or storage, so I think these shouldn't be ignored.

dcermak:
> @dcermak TY! Did you by chance vendor this commit into a WIP Podman PR to validate that the tests work there with these changes? If not, could you please?

I have tried in https://github.com/containers/podman/pull/13536, but I have not managed to get the correct vendoring in place.

dcermak:
@TomSweeneyRedHat https://github.com/containers/podman/pull/13536 is green

drivers/overlay/overlay.go
```diff
@@ -292,6 +292,31 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
...
+	// Create the driver home dir
+	if err := idtools.MkdirAllAs(path.Join(home, linkDir), 0700, rootUID, rootGID); err != nil {
+		return nil, err
+	}
+
+	if err := idtools.MkdirAllAs(runhome, 0700, rootUID, rootGID); err != nil {
+		return nil, err
+	}
+
+	if opts.mountProgram != "" {
```
giuseppe, line 310: should this be `if opts.mountProgram == ""`?  Otherwise if it is already set it should just be honored

dcermak: Of course… Apologies, I must have been asleep when coding that…

drivers/overlay/overlay.go
```diff
@@ -292,6 +292,31 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
...
+
+	if err := idtools.MkdirAllAs(runhome, 0700, rootUID, rootGID); err != nil {
+		return nil, err
+	}
+
+	if opts.mountProgram == "" {
+		if supported, err := SupportsNativeOverlay(home, runhome); err != nil {
+			return nil, err
+		} else if !supported {
+			if path, err := exec.LookPath("fuse-overlayfs"); err == nil && opts.mountProgram == "" {
```
rhatdan, line 314: This is not necessary `&& opts.mountProgram == ""` Since this was checked above.

dcermak: Whoops, you're of course correct. I overlooked that one in the last rebase.

giuseppe (approved):
LGTM

PR #1170, opened 2022-03-21
MaybeReexecUsingUserNamespace: handle SIGHUP/SIGINT/SIGTERM

nalind:
Pass along SIGHUP, SIGINT, and SIGTERM if the parent receives them while the child process is running.  Set SIGKILL as the child's parent-death signal so that, if we exit for some reason before the child does, it will be stopped.  This addresses part of https://github.com/containers/buildah/issues/3544.

### DIFF
--- a/pkg/unshare/unshare_linux.go
+++ b/pkg/unshare/unshare_linux.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"os/signal"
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
LGTM
@giuseppe PTAL

flouthoc (approved):

giuseppe (approved):
LGTM

PR #1172, opened 2022-03-23
[v1.36 Backport] race, storage: add independent `AddNames` and `RemoveNames` for images,layers and containers

flouthoc:
Backport of: https://github.com/containers/storage/pull/1153

Adds SetNameWithOptions so operations which are invoke in parallel

For instance


flouthoc:
PR on wrong branch. My bad.

PR #1173, opened 2022-03-23
[v1.36 Backport] race, storage: add independent AddNames and RemoveNames for images,layers and containers

flouthoc:
Backport of: https://github.com/containers/storage/pull/1153

Adds AddNames and RemoveNames so operations which are invoked in parallel

For instance


flouthoc:
@containers/storage-maintainers @giuseppe @nalind @rhatdan PTAL. I'm not sure if failing ubuntu test are related to the change in this PR.

vrothberg (commented):
Could add two more commits:
1) Bump the version for a new release
2) Bump back to a -dev version

vrothberg:
Restarted the Ubuntu jobs.  If they continue failing, we may need to call @cevich's hotline.

flouthoc:
@vrothberg Done, although i see `c/storage` is using convention of `+dev` instead of `-dev`,  so I kept it `+dev`.

flouthoc:
@TomSweeneyRedHat FYI

cevich:
FWIW: The pattern we follow in every other repo. running tests on VMs is (especially for Ubuntu):

* Never do a blanket package update at test runtime, removals are okay.
* Make sure any runtime package installs ([example](https://github.com/containers/buildah/blob/bc5080ccdd742f7e24535ba2fc3fbf2acd3ba6b9/contrib/cirrus/setup.sh#L31-L34)) happen from pre-downloaded cache ([example](https://github.com/containers/automation_images/blob/752b774bb3e29fb11e43890f07fd978d8d1eac1e/cache_images/ubuntu_packaging.sh#L157-L168))
* or simply add the necessary packages into the list ([example](https://github.com/containers/automation_images/blob/752b774bb3e29fb11e43890f07fd978d8d1eac1e/cache_images/ubuntu_packaging.sh#L50)) to be installed (assuming they don't cause any interference in any other CI)

Legacy Fedora VMs are slightly more forgiving since their package repositories tend to remain available for quite a long time.  But ultimately they will be archived, so much of the same advice applies.

TomSweeneyRedHat:
I'm going to close this PR.  If we need to backport, which I doubt, we can resurrect if necessary.

flouthoc:
@containers/storage-maintainers @TomSweeneyRedHat It seems we need this patch in `v3.4.2-rhel` again. Could we review this PR and cut a new release once this is merged. 

Backport is similar to this https://github.com/containers/storage/pull/1174 so once this is merged and a new version I think `v1.36.2` is released then `v1.36.2` can be used by other dependencies.

rhatdan:
LGTM

TomSweeneyRedHat:
@flouthoc which BZ does this backport address?

flouthoc:
@TomSweeneyRedHat This was supposed to go into `v3.4.2-rhel` but I'm not sure if that is under consideration anymore or not. So nothing as of now.

TomSweeneyRedHat:
Just for historical searches, it's good to put the BZ number in the PR, preferably in the Description.  If there's not one that's OK, but would like to have one added if there is a particular BZ this is aimed at.

PR #1174, opened 2022-03-23
[v1.38, backport] store: add independent `AddNames` and `RemoveNames` for images,layers,containers

flouthoc:
Backport of https://github.com/containers/storage/pull/1153

Adds AddNames and RemoveNames so operations which are invoked in parallel

For instance


rhatdan:
LGTM

flouthoc:
nvm its already here: https://github.com/containers/storage/tree/v1.38.3

TomSweeneyRedHat:
Partially addresses: https://bugzilla.redhat.com/show_bug.cgi?id=2066519 and https://bugzilla.redhat.com/show_bug.cgi?id=2055487

TomSweeneyRedHat:
Exceptions approved:  RHEL 9.0 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2066519  RHEL 8.6 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2055487

PR #1176, opened 2022-03-26
Switch most calls to filepath.Walk to filepath.WalkDir

rhatdan:
It is faster then Walk, when you don't need to stat every
file and directory.


rhatdan:
@nalind @cevich This looks like the images does not have a newer golang?

cevich:
> This looks like the images does not have a newer golang?

Newer golang is F36 only and in https://github.com/containers/podman/pull/13376

drivers/overlay/check_115.go
```diff
@@ -1,8 +1,10 @@
+//go:build !go1.16
 // +build !go1.16
 
 package overlay
 
 import (
+	"io/fs"
```
nalind, line 7: This source file is a fallback for the sake of compilers earlier than 1.16, which don't have `io/fs` in their standard library.  You want to remove this file and merge `check_116.go` into `check.go` instead.

rhatdan:
@cevich could you update the images for storage.  Podman's images do not want.  But I need newer versions of golang on f34.

cevich:
> could you update the images for storage

Yep.

cevich:
Let's see what https://github.com/containers/storage/pull/1182 do.

TomSweeneyRedHat:
All kinds of test unhappiness @rhatdan

kolyshkin (approved):
LGTM

rhatdan:
Yup I am at a loss to why this works on Podman but not in storage.

TomSweeneyRedHat:
Changes LGTM
lint is complaining about not being able to find a package, plus you need to rebase, apparently.

nalind:
The `lint` task in the Cirrus config specifies that it runs in `golang:1.15` on line 120 of `.cirrus.yml`, and `io/fs` was new in 1.16.

cevich:
> The lint task in the Cirrus config specifies that it runs in golang:1.15 on line 120 of .cirrus.yml, and io/fs was new in 1.16.

IIUC, it's totally fine to just bump that up in the YAML, shouldn't affect anything other than that one lint task.

PR #1180, opened 2022-03-29
overlay: use idmapped lower layers where supported

giuseppe:
use idmapped mounts for the overlay lower layers when the kernel
supports them.

For each lower directory with ID=0...N-1, it creates a idmapped mount
at $GRAPHROOT/overlay/$LAYER/mapped/$ID.  The final overlay mount will
use these idmapped mounts instead of the original source directory.

The upperdir is not idmapped, so files are created with the same
IDs used by the user namespace.

### DIFF
--- a/drivers/overlay/check.go
+++ b/drivers/overlay/check.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package overlay
@@ -11,6 +12,7 @@ import (
 	"syscall"
 
 	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/mount"
 	"github.com/containers/storage/pkg/system"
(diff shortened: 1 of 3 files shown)


giuseppe:
tested with these kernel patches: https://lore.kernel.org/linux-unionfs/20220329103526.1207086-1-brauner@kernel.org/T/#m84a0bbc1476af67073a48f5800c858a361848413

drivers/overlay/check.go
```diff
@@ -218,3 +220,54 @@ func doesVolatile(d string) (bool, error) {
...
+	}
+
+	mergedDir := filepath.Join(layerDir, "merged")
+	lowerDir := filepath.Join(layerDir, "lower")
+	lowerMappedDir := filepath.Join(layerDir, "lower-mapped")
+	upperDir := filepath.Join(layerDir, "upper")
+	workDir := filepath.Join(layerDir, "work")
+
+	defer func() {
+		_ = unix.Unmount(mergedDir, unix.MNT_DETACH)
```
rhatdan, line 239: Should this defer be lower down at least for the Umount.

drivers/overlay/overlay.go
```diff
@@ -205,6 +207,29 @@ func checkSupportVolatile(home, runhome string) (bool, error) {
...
+func checkAndRecordIDMappedSupport(home, runhome string) (bool, error) {
+	feature := "idmapped-lower-dir"
+	overlayCacheResult, overlayCacheText, err := cachedFeatureCheck(runhome, feature)
+	if err == nil {
+		if overlayCacheResult {
+			logrus.Debugf("Cached value indicated that overlay is supported")
+		} else {
+			logrus.Debugf("Cached value indicated that overlay is not supported")
+		}
+		if !overlayCacheResult {
```
rhatdan, line 221: No need for this check if, just move return false below after Debugf line above.

rhatdan (approved):

giuseppe:
thanks for the review.

I've addressed the comments and pushed a new version

drivers/overlay/idmapped_utils.go
```diff
@@ -0,0 +1,181 @@
...
+	// MOUNT_ATTR_NOSUID - Ignore suid and sgid bits
+	MOUNT_ATTR_NOSUID = 0x00000002 //nolint:golint
+	// MOUNT_ATTR_NODEV - Disallow access to device special files
+	MOUNT_ATTR_NODEV = 0x00000004 //nolint:golint
+	// MOUNT_ATTR_NOEXEC - Disallow program execution
+	MOUNT_ATTR_NOEXEC = 0x00000008 //nolint:golint
+	// MOUNT_ATTR__ATIME - Setting on how atime should be updated
+	MOUNT_ATTR__ATIME = 0x00000070 //nolint:golint
+	// MOUNT_ATTR_RELATIME - Update atime relative to mtime/ctime
+	MOUNT_ATTR_RELATIME = 0x00000000 //nolint:golint
```
rhatdan, line 37: Quick question is this one correct?  No bits set?

giuseppe: yes, it is correct:

```
$ grep MOUNT_ATTR_RELATIME /usr/include/linux/mount.h
#define MOUNT_ATTR_RELATIME	0x00000000 /* - Update atime relative to mtime/ctime. */
```

I am not sure if we should add all of these (added for completeness) and if they should be public.

rhatdan: Can we use them from here?
 grep -r MOUNT_ATTR_NODEV .
./vendor/golang.org/x/sys/unix/zerrors_linux.go:	MOUNT_ATTR_NODEV                            = 0x4

giuseppe: I've dropped the ones not used by the module and made the other 3 private

rhatdan:
LGTM

PR #1057, opened 2021-11-09
idtools: fix ToHost conversion

giuseppe:
drop a special case for RootPair that prevented the conversion of the
UID/GID when it was matching the root user in the container.

Closes: https://github.com/containers/storage/issues/1056

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #1185, opened 2022-03-30
idtools: don't say ToHost() makes exceptions in its godoc when it doesn't

nalind:
The godoc for pkg/idtools.ToHost() mentions that it makes an exception when either of the IDs to be mapped is 0, but that special case was removed in #1057.


rhatdan (approved):
LGTM

saschagrunert (approved):
LGTM

flouthoc (approved):
LGTM

PR #1188, opened 2022-04-04
Check filecaps on newuidmap and newgidmap on failure

rhatdan:
A fairly common failure is newuidmap and newgidmap not being setuid and
setgid. If these commands fail, check the permissions on the binaries in
order to give the user better information on why they failed.

### DIFF
--- a/pkg/unshare/unshare_linux.go
+++ b/pkg/unshare/unshare_linux.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package unshare
@@ -74,6 +75,26 @@ func getRootlessGID() int {
 		return u
 	}
 	return os.Getegid()
+}
+
+func isSetID(path string, modeid os.FileMode, capid capability.Cap) (bool, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return false, err
+	}
+
+	mode := info.Mode()
+	if mode&modeid == modeid {
+		return true, nil
+	}
+	cap, err := capability.NewFile2(path)
+	if err != nil {
+		return false, err
+	}
+	if err := cap.Load(); err != nil {
+		return false, err
+	}
+	return cap.Get(capability.EFFECTIVE, capid), nil
 }
 
 func (c *Cmd) Start() error {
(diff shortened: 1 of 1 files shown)


rhatdan:
Fixes: https://github.com/containers/podman/issues/13751

pkg/unshare/unshare_linux.go
```diff
@@ -76,6 +77,42 @@ func getRootlessGID() int {
 	return os.Getegid()
 }
 
+func isSetUid(path string) (bool, error) {
+	info, err := os.Stat(os.Args[1])
```
Luap99, line 81:
```suggestion
	info, err := os.Stat(path)
```

pkg/unshare/unshare_linux.go
```diff
@@ -76,6 +77,42 @@ func getRootlessGID() int {
...
+	}
+
+	cap, err := capability.NewFile(path)
+	if err != nil {
+		return false, err
+	}
+	return cap.Get(capability.EFFECTIVE, capability.CAP_SETUID), nil
+}
+
+func isSetGid(path string) (bool, error) {
```
Luap99, line 98: Can you combine isSetUid() and isSetGid() into one function, right now the logic is duplicated, i.e.
```
modeSetid := os.ModeSetuid
capSetid := capability.CAP_SETGID
if gid {
    modeSetid := os.ModeSetuid
    capSetid := capability.CAP_SETGID
}
```

Luap99 (commented):
Podman doesn't use this code AFAIK.
You need to change https://github.com/containers/podman/blob/839744af56ee6a3e54c63eb8da495974666dfdeb/pkg/rootless/rootless_linux.go#L115 for podman.

pkg/unshare/unshare_linux.go
```diff
@@ -76,6 +77,36 @@ func getRootlessGID() int {
...
 }
 
+type idtype int64
+
+const (
+	uid idtype = 0
+	gid        = 1
+)
+
+func isSetID(id idtype, path string) (bool, error) {
```
nalind, line 87: This would be a bit shorter if the callers passed in the mode bit and the capability to look for instead of an enumerated value.

rhatdan: Fixed.

pkg/unshare/unshare_linux.go
```diff
@@ -76,6 +78,23 @@ func getRootlessGID() int {
...
+	info, err := os.Stat(path)
+	if err != nil {
+		return false, err
+	}
+
+	mode := info.Mode()
+	if mode&modeid == modeid {
+		return true, nil
+	}
+	cap, err := capability.NewFile(path)
```
nalind, line 91: The docs for `github.com/syndtr/gocapability/capability` say `NewFile()` is deprecated, and to use `NewFile2()` and then `cap.Load()` instead.

rhatdan: I had that on a previous version.

pkg/unshare/unshare_linux.go
```diff
@@ -76,6 +78,26 @@ func getRootlessGID() int {
 	return os.Getegid()
 }
 
+func isSetID(path string, modeid fs.FileMode, capid capability.Cap) (bool, error) {
```
nalind, line 81: It's okay to call `modeid` a `os.FileMode`.

rhatdan: Now you just want to aggravate me....

nalind: Trying to get you past [the linter](https://cirrus-ci.com/task/6722549160607744?logs=test#L16), which seems to not understand that package name.

nalind:
LGTM

giuseppe (approved):
LGTM

PR #1191, opened 2022-04-06
[backport v1.24] Lock during store diff and add race-free `AddNames` and `RemoveNames` api.

flouthoc:
Backport of https://github.com/containers/storage/pull/1153 and https://github.com/containers/storage/pull/1037

* Introduces race-free `AddNames` and `RemoveNames` API.
* Fixes race by locking graphLock during diff

Needed for: BZ 2066364

PR #1193, opened 2022-04-07
Revert "[backport v1.24] Lock during store diff and add race-free `AddNames` and `RemoveNames` api."

flouthoc:
Reverts containers/storage#1191

Backport was re-evaluated only for `release-1.23` buildah branch and backport for `release-1.19` buildah is no longer under consideration. So we must revert this.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -85,16 +85,7 @@ type ContainerStore interface {
 
 	// SetNames updates the list of names associated with the container
 	// with the specified ID.
-	// Deprecated: Prone to race conditions, suggested alternatives are `AddNames` and `RemoveNames`.
 	SetNames(id string, names []string) error
-
-	// AddNames adds the supplied values to the list of names associated with the container with
-	// the specified id.
-	AddNames(id string, names []string) error
-
-	// RemoveNames removes the supplied values from the list of names associated with the container with
-	// the specified id.
-	RemoveNames(id string, names []string) error
 
 	// Get retrieves information about a container given an ID or name.
 	Get(id string) (*Container, error)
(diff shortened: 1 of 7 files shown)


flouthoc:
@nalind @giuseppe @rhatdan @vrothberg @TomSweeneyRedHat Could we revert this please since this branch is not being used in backports anymore. Ubuntu tests are unrelated and were failing in previous PR as well.

giuseppe:
LGTM but I am not sure we should also revert the VERSION number

flouthoc:
@giuseppe I can drop the version change, but the version bumped in https://github.com/containers/storage/pull/1191 was never released.

vrothberg:
> @nalind @giuseppe @rhatdan @vrothberg @TomSweeneyRedHat Could we revert this please since this branch is not being used in backports anymore. Ubuntu tests are unrelated and were failing in previous PR as well.

If the branch isn't used anymore, why bother?

flouthoc:
@vrothberg Just in-case if we ever revisit this branch anytime in future it contains commits which never got released and they were never intended to be here. But I don't have a strict opinion on this cause merged commits seems harmless but we discussed here  https://github.com/containers/storage/pull/1191 that we should revert so just created a PR.

rhatdan:
Fix the version and then we can merge.

flouthoc:
@rhatdan Dropped the version revert.

vrothberg:
I just noticed that this PR was opened with a branch on c/storage.  Please be careful; we should not push to c/storage but use our forks to open PRs.

flouthoc:
@vrothberg Yeah my bad actually I used Github GUI's revert button so branch was created on `c/storage` and not on my fork. Will not use GUI from next time its much safe to do it by hand.

flouthoc:
Deleted PR branch.

vrothberg:
Thanks!  No worries at all.

PR #1197, opened 2022-04-12
Fully populate layer records for mapped image top layers

nalind:
When we create a copy of an image's top layer that's intended to be identical to the top layer, except for having some set of ID mappings already applied to it, copy over the template layer's compressed and uncompressed digest and size information, compression information, tar-split data, and lists of used UIDs and GIDs, if we have them.

The lack of sizing information was forcing ImageSize() to regenerate the diffs to determine the size of the mapped layers, which shouldn't have been necessary.

Teach the overlay DiffGetter to look for files in the diff directories of lower layers if we can't find them in the current layer, so that tar-split can retrieve content that we didn't have to pull up.

Fixes https://github.com/containers/podman/issues/13755.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -39,7 +39,6 @@ import (
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"github.com/vbatts/tar-split/tar/storage"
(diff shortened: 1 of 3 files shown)


vrothberg (commented):
LGTM
@giuseppe PTAL

rhatdan:
LGTM

debarshiray:
Thanks, @nalind (because of https://github.com/containers/toolbox/issues/1027) !

PR #1209, opened 2022-04-14
DeleteImage(): always delete mapped top layers

nalind:
When deleting an image, always delete an image's mapped top layers, even ~if~ when the unmapped top layer is shared with other images and must therefore not be deleted.


rhatdan:
Your desciption is a little contradictory?

nalind:
The main distinction is mapped vs unmapped?

rhatdan:
LGTM

baude:
LGTM

jwhonce (approved):
/lgtm

PR #1210, opened 2022-04-18
layerStore.Put(): update digest-based indexes when creating from templates

nalind:
When we're creating a layer using another layer as a template, add the new layer's uncompressed and compressed digest to the maps we use to index layers using those digests.

When we forgot to do that, searching for a layer by either would still turn up the original template, so this didn't really break anything.

We also had a case in layerStore.PutAdditionalLayer() where we mistakenly mixed up the uncompressed and compressed digests when populating the by-uncompressed-digest map; fix that while we're in here.


vrothberg (commented):
LGTM

Restarted the flaked job.

rhatdan (approved):
LGTM

PR #1219, opened 2022-04-27
FreeBSD fixes for pkg/mount

dfr:
This uses MNT_FORCE as an approximation of Linux MNT_DETACH and passes through key/value options such as devfs' ruleset=N option to the kernel. Also, I removed the code which truncates the zfs mount point to work around a path name limit in obsolete versions of FreeBSD which was fixed in FreeBSD 12.

### DIFF
--- a/drivers/zfs/zfs_freebsd.go
+++ b/drivers/zfs/zfs_freebsd.go
@@ -2,7 +2,6 @@ package zfs
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/containers/storage/drivers"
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

rhatdan:

flouthoc (approved):
LGTM

flouthoc:
@dfr Could you rebase please although its not needed but github UI blocks merge button and only option is force merge.

dfr:
Rebased

rhatdan:
Ignoring the flakes.

PR #1220, opened 2022-04-27
chown: use overflow id as fallback when chowning

giuseppe:
when chowning an image, fall back to the overflow ID when a UID or GID cannot be mapped to the target user namespace.
    
This ensures the chown driver works similar to what we do with idmapped mounts when it is supported for overlay.
    
It is needed for CRI-O to support user namespaces in Kubernetes since the Kubelet picks a static size for the user namespace and it might break some images using IDs outside the picked range.

### DIFF
--- a/drivers/chown_unix.go
+++ b/drivers/chown_unix.go
@@ -76,7 +76,7 @@ func (c *platformChowner) LChown(path string, info os.FileInfo, toHost, toContai
 			UID: uid,
 			GID: gid,
 		}
-		mappedPair, err := toHost.ToHost(pair)
+		mappedPair, err := toHost.ToHostOverflow(pair)
 		if err != nil {
 			return fmt.Errorf("error mapping container ID pair %#v for %q to host: %v", pair, path, err)
 		}
--- a/drivers/graphtest/graphtest_unix.go
+++ b/drivers/graphtest/graphtest_unix.go
@@ -125,11 +125,11 @@ func DriverTestCreateBase(t testing.TB, drivername string, driverOptions ...stri
 	driver := GetDriver(t, drivername, driverOptions...)
 	defer PutDriver(t)
 
-	createBase(t, driver, "Base")
+	createBase(t, driver, "Base1")
 	defer func() {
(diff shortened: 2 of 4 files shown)


rhatdan:
What happens if I have no UIDs in /etc/subuid, does this effectively do ignore_chown?  Or is this different?

rhatdan (approved):
LGTM

TomSweeneyRedHat:
LGTM

giuseppe:
> What happens if I have no UIDs in /etc/subuid, does this effectively do ignore_chown? Or is this different?

they happen in two different times.

ignore_chown is needed when a rootless user pulls an image in its initial user namespace (the `podman unshare` environment) and the image requires more IDs than are available for that user.  So it won't be possible to represent it correctly in the local storage.

The use case I am trying to address here instead is when an image already present in the storage must be chowned before it can be used in a user namespace.  In this case, both for rootful and rootless, the user specifies the user namespace mappings or size (through `auto:size=N`) and some IDs cannot be mapped to the target user namespace.

I think in this case we should not error out but do what the user asked to do.  We need to fit the image in the specified mappings and if something cannot be mapped then use the overflow IDs.

giuseppe:
> Should we at least logrus.Warnf?

I'd expect that to be too verbose when used from CRI-O.  Maybe Debugf?

rhatdan:
Ok add debugf.

giuseppe:
> Ok add debugf.

added

nalind:
What's the intent of the  `graphtest: use unique names for the file system layers` commit?

giuseppe:
> What's the intent of the `graphtest: use unique names for the file system layers` commit?

I've seen this test failing multiple times, and it wasn't clear where exactly it was failing, so the intent is to make it clearer.

If you don't like it, I can open a PR to revert the change

nalind:
The intent wasn't documented, AFAICT, anywhere, which means there's nothing preventing another maintainer from concluding that it was done on a whim, and then quietly reverting it if we do another rebase onto from moby's version.

PR #1221, opened 2022-05-02
Change default location for storage.conf to...

dfr:
... /usr/local/etc/containers on FreeBSD


dfr:
Looks like I broke the linux build - I'll update the PR shortly

types/options_freebsd.go
```diff
@@ -0,0 +1,17 @@
...
+const (
+	// these are default path for run and graph root for rootful users
+	// for rootless path is constructed via getRootlessStorageOpts
+	defaultRunRoot   string = "/var/run/containers/storage"
+	defaultGraphRoot string = "/var/db/containers/storage"
+)
+
+// defaultConfigFile path to the system wide storage.conf file
+var (
+	defaultConfigFile         = "/usr/local/share/containers/storage.conf"
```
TomSweeneyRedHat, line 12: should this be `/usr/local/containers/storage.conf` to match https://github.com/containers/storage/pull/1221/files#diff-3e70e441de9baa998963aaec77ca4cfe228f061b38d08eda62f28bdc53e3fdcbR8

dfr: I think the example conf file should change instead. Generally, this kind of thing tends to live either in /usr/local/share or /usr/local/etc with any file likely to be changed locally strongly preferred to be in /usr/local/etc.

dfr: I changed storage.conf-freebsd to match the compiled in default

rhatdan:
LGTM

rhatdan:
@giuseppe PTAL

giuseppe:
LGTM

flouthoc (approved):
LGTM

types/options_freebsd.go
TomSweeneyRedHat: THat works for me, I was mostly concerned about the difference between the two.  Thanks @dfr

PR #1229, opened 2022-05-05
style clean up in storage.conf code

vrothberg:
`ReloadConfigurationFile()` does not return an error but `fmt.Print`s them which is quite ugly but I didn't touch that.

@rhatdan @giuseppe PTAL


giuseppe (approved):
LGTM

rhatdan:
LGTM

rhatdan:
This seems to be breaking tests.

vrothberg:
Updated

Issue #1232, opened 2022-05-09
error creating an ID-mapped copy of layer when the layer is in a different store


giuseppe:
The reproducer for the error:

```
# podman --root /tmp/store pull alpine
# podman --root /tmp/store2 --storage-opt .imagestore=/tmp/store pull alpine
# podman --root /tmp/store2 --storage-opt .imagestore=/tmp/store run --rm --uidmap=0:100:10000 alpine true
Error: error creating container storage: error creating an ID-mapped copy of layer "4fc242d58285699eca05db3cc7c7122a2b8e014d9481f323bd9277baacfa0628": layer not known
```

We should probably extend `imageTopLayerForMapping` to return `nil` if the layers are in a different store.

giuseppe:
PR here: https://github.com/containers/storage/pull/1233

PR #1233, opened 2022-05-09
store: skip template copy with layer in different store

giuseppe:
do not attempt to create a layer using a TemplateLayer if the source
layer is in a different store, since the lookup for the source layer
would fail in the layers store.

The reproducer for the error:

Error: error creating container storage: error creating an ID-mapped copy of layer "4fc242d58285699eca05db3cc7c7122a2b8e014d9481f323bd9277baacfa0628": layer not known

Closes: https://github.com/containers/storage/issues/1232

### DIFF
--- a/store.go
+++ b/store.go
@@ -1195,6 +1195,11 @@ func (s *store) imageTopLayerForMapping(image *Image, ristore ROImageStore, crea
 				if layer == nil {
 					layer = cLayer
 					parentLayer = cParentLayer
+					if store != rlstore {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@nalind @saschagrunert @flouthoc @vrothberg PTAL

vrothberg (commented):
LGTM
Restarted the flaked job

giuseppe:
tests are green!

PR #1234, opened 2022-05-10
overlay: do not check for idmapped mounts if euid!=0

giuseppe:
do not store the result of the check when running with euid != 0.  The
check must happen and its result stored only once the process is
re-executed in its user namespace and runs with euid == 0.

It doesn't make any real difference at the moment since the  kernel
feature is going to be supported only for root in the initial user
namespace.


flouthoc (approved):
LGTM

PR #1239, opened 2022-05-12
Implement ReadMemInfo on FreeBSD

dfr:

### DIFF
--- a/pkg/system/meminfo_freebsd.go
+++ b/pkg/system/meminfo_freebsd.go
@@ -1,0 +1,84 @@
+//go:build freebsd && cgo
+// +build freebsd,cgo
+
+package system
+
+import (
+	"fmt"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+// #include <unistd.h>
+// #include <sys/vmmeter.h>
+// #include <sys/sysctl.h>
+// #include <vm/vm_param.h>
+import "C"
+
+func getMemInfo() (int64, int64, error) {
+	data, err := unix.SysctlRaw("vm.vmtotal")
+	if err != nil {
+		return -1, -1, fmt.Errorf("Can't get kernel info: %v", err)
+	}
+	if len(data) != C.sizeof_struct_vmtotal {
+		return -1, -1, fmt.Errorf("unexpected vmtotal size %d", len(data))
+	}
+
+	total := (*C.struct_vmtotal)(unsafe.Pointer(&data[0]))
+
+	pagesize := int64(C.sysconf(C._SC_PAGESIZE))
+	npages := int64(C.sysconf(C._SC_PHYS_PAGES))
+	return pagesize * npages, pagesize * int64(total.t_free), nil
+}
+
+func getSwapInfo() (int64, int64, error) {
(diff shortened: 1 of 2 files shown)


pkg/system/meminfo_freebsd.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	pagesize := int64(C.sysconf(C._SC_PAGESIZE))
+	return pagesize * total, pagesize * used, nil
+}
+
+// ReadMemInfo retrieves memory statistics of the host system and returns a
+//  MemInfo type.
+func ReadMemInfo() (*MemInfo, error) {
+	MemTotal, MemFree, err := getMemInfo()
+	SwapTotal, SwapFree, err := getSwapInfo()
```
flouthoc, line 60: This overrides err from `getMemInfo`. Probably this should also check if `err != nil` and return early in that case.

pkg/system/meminfo_freebsd.go
```diff
@@ -0,0 +1,75 @@
...
+	npages := int64(C.sysconf(C._SC_PHYS_PAGES))
+	return pagesize * npages, pagesize * int64(total.t_free), nil
+}
+
+func getSwapInfo() (int64, int64, error) {
+	var (
+		total int64 = 0
+		used  int64 = 0
+	)
+	for i := 0; ; i++ {
```
flouthoc, line 40: Just a question: Can you get number of swap devices using `nswapdev` instead infinite loop.

flouthoc:
Two questions above.

pkg/system/meminfo_freebsd.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, fmt.Errorf("error getting memory totals %v\n", err)
+	}
+	SwapTotal, SwapFree, err := getSwapInfo()
+	if err != nil {
+		return nil, fmt.Errorf("error getting swap totals %v\n", err)
+	}
+
+	if MemTotal < 0 || MemFree < 0 || SwapTotal < 0 || SwapFree < 0 {
+		return nil, fmt.Errorf("error getting system memory info %v\n", err)
+	}
```
dfr, line 70: I don't think any of these can be negative if the two calls above succeed. Perhaps this can change to an assert?

pkg/system/meminfo_freebsd.go
dfr: Good idea - I was kind of following the algorithm used in the base system's swapinfo command but this is better.

flouthoc: @dfr I am not aware about BSD a lot so i am not even sure about it. It was just a question or rather a discussion, you can feel free to discuss more about rather than changing code right away. Could you check again if it not breaks anything else.

dfr: I re-tested using 'podman info' and verified that it still reads the swap details. I also noticed that it mixed up used and free totals for swap so I fixed that.

flouthoc (approved):
LGTM @containers/storage-maintainers PTAL

rhatdan:
LGTM

PR #1240, opened 2022-05-12
Limited support for pkg/unshare on FreeBSD

dfr:
This does not really 'unshare' anything but does implement the setsid,
setpgrp and ctty options as well as the pipe-based signalling protocol
with the parent process which made it slightly easier to port buildah's
chroot isolation mechanism.

### DIFF
--- a/pkg/unshare/unshare.c
+++ b/pkg/unshare/unshare.c
@@ -1,4 +1,4 @@
-#ifndef UNSHARE_NO_CODE_AT_ALL
+#if !defined(UNSHARE_NO_CODE_AT_ALL) && defined(__linux__)
 
 #define _GNU_SOURCE
 #include <sys/types.h>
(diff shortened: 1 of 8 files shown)


giuseppe (approved):
LGTM

nalind:
`OOMScoreAdj` is in the non-Linux version of the `Cmd` struct, but it seems to be ignored.  Should it be there?

dfr:
> `OOMScoreAdj` is in the non-Linux version of the `Cmd` struct, but it seems to be ignored. Should it be there?

I will remove it - as you can see its not implemented here and I'm not sure if there is a direct equivalent

nalind:
The FreeBSD implementation could probably stand to drop the `_Containers-unshare` environment variable, which the Linux implementation uses for passing flags for `unshare()` down to the child process, but which doesn't appear to serve a purpose in the FreeBSD version.

Go ahead and move `bailOnError` from `unshare_linux.go` to `unshare.go` to avoid having to duplicate it in `unshare_freebsd.go`.

dfr:
I've updated the branch with the suggested changes, please take a look.

rhatdan:
LGTM

PR #1241, opened 2022-05-16
PullOptions should work for rootless users also

rhatdan:
Also add interfaces to allow callers to see the pull options.

While experimenting with pushing and pulling with zstd, I found
that storage pulloptions were not being used in rootless mode.

### DIFF
--- a/store.go
+++ b/store.go
@@ -173,6 +173,7 @@ type Store interface {
 	GraphRoot() string
 	GraphDriverName() string
 	GraphOptions() []string
+	PullOptions() map[string]string
 	UIDMap() []idtools.IDMap
 	GIDMap() []idtools.IDMap
 
@@ -607,6 +608,7 @@ type store struct {
 	graphRoot       string
 	graphDriverName string
 	graphOptions    []string
+	pullOptions     map[string]string
 	uidMap          []idtools.IDMap
 	gidMap          []idtools.IDMap
 	autoUsernsUser  string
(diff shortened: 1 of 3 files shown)


rhatdan:
@giuseppe @flouthoc @nalind @saschagrunert @vrothberg PTAL

store.go
```diff
@@ -712,20 +714,21 @@ func GetStore(options types.StoreOptions) (Store, error) {
...
+		runRoot:         options.RunRoot,
 		uidMap:          copyIDMap(options.UIDMap),
-		gidMap:          copyIDMap(options.GIDMap),
-		autoUsernsUser:  options.RootAutoNsUser,
-		autoNsMinSize:   autoNsMinSize,
-		autoNsMaxSize:   autoNsMaxSize,
-		additionalUIDs:  nil,
-		additionalGIDs:  nil,
 		usernsLock:      usernsLock,
-		disableVolatile: options.DisableVolatile,
```
nalind, line 728: This makes it harder to use `git annotate` to find which patch/PR introduced a particular option, which hopefully includes rationale for the changes.

store.go
```diff
@@ -776,6 +779,10 @@ func (s *store) GraphOptions() []string {
 	return s.graphOptions
 }
 
+func (s *store) PullOptions() map[string]string {
+	return s.pullOptions
```
nalind, line 783: This should be returning a copy so that callers can't mess with the configuratoin by modifying the map's contents.

vrothberg (commented):
LGTM

flouthoc (approved):
LGTM

PR #1249, opened 2022-06-01
storage.conf: respect CONTAINERS_STORAGE_CONF env

Luap99:
CONTAINERS_STORAGE_CONF env var should overwrite the default path even
for the first config load at init time.

Fixes containers/podman#14267

types: do not load storage.conf at init time

Using init in libs is bad. init() has to be run every time the
application which imports this package starts. This slows down startup
time even if the application will never use this code path.

Instead use a sync.Once to only the config when it is required.

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -31,7 +31,11 @@ const (
 	storageConfEnv = "CONTAINERS_STORAGE_CONF"
 )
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #1250, opened 2022-06-01
Don't try to use %w in testing package logging functions

nalind:
Go 1.18's testing package is adamant that it doesn't support the `%w` specifier for wrapping errors, which makes sense because it outputs text rather than producing an error object like `fmt.Errorf()` does.  Change attempts to use `%w` to `%v`.


TomSweeneyRedHat:
LGTM
fwiw

PR #1256, opened 2022-06-07
chunked: drop host dedup feature

giuseppe:
drop host deduplication by just looking at the file path.  It could be
useful in very specific use cases, but it is too expensive for generic
images.  If the need arises, we first need to create an index of the
files that we can deduplicate so there is no need to calculate the
checksum on the fly.


rhatdan:
You need to fix man pages.

rhatdan (approved):

rhatdan:
LGTM
@alexlarsson @saschagrunert @nalind @vrothberg @umohnani8 @flouthoc PTAL

nalind:
LGTM

flouthoc:
LGTM for the record.

PR #1258, opened 2022-06-13
chown: use Link instead of Chown for dup inodes

giuseppe:
If the inode was already encountered and chowned, use link(2) instead
of chown(2).

This is needed when the underlying storage (as it could be overlay
with index=off) breaks the hard link on copy up.

https://github.com/containers/storage/pull/1144 added the initial
check.

Closes: https://github.com/containers/storage/issues/1257

### DIFF
--- a/drivers/chown_unix.go
+++ b/drivers/chown_unix.go
@@ -21,12 +21,12 @@ type inode struct {
 
 type platformChowner struct {
 	mutex  sync.Mutex
-	inodes map[inode]bool
+	inodes map[inode]string
 }
 
 func newLChowner() *platformChowner {
 	return &platformChowner{
-		inodes: make(map[inode]bool),
+		inodes: make(map[inode]string),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


drivers/chown_unix.go
```diff
@@ -41,13 +41,21 @@ func (c *platformChowner) LChown(path string, info os.FileInfo, toHost, toContai
...
 	c.mutex.Lock()
-	_, found := c.inodes[i]
+	oldOwner, found := c.inodes[i]
 	if !found {
-		c.inodes[i] = true
+		owner := idtools.IDPair{
+			UID: int(st.Uid),
+			GID: int(st.Gid),
+		}
+		c.inodes[i] = owner
```
flouthoc, line 50: Non-blocking nit
```suggestion
		c.inodes[i] = idtools.IDPair{
			UID: int(st.Uid),
			GID: int(st.Gid),
		}
```

flouthoc (approved):
LGTM

drivers/chown_unix.go
```diff
@@ -41,13 +41,21 @@ func (c *platformChowner) LChown(path string, info os.FileInfo, toHost, toContai
...
 	}
 	c.mutex.Unlock()
 
-	if found {
+	// If the dev/inode was already chowned but the file has the original
+	// owner then it means the underlying hard link was broken (as it happens with
+	// overlay when index=off) so we need to chown again and not rely on the previous
+	// chown for the same inode.
```
nalind, line 57: This comment really confused me - it didn't describe the checks on the next line, and after suggesting that if the checks were true, we still needed to do something, it just `return`ed.

nalind:
Do we want to take the hit of checking inode numbers after the `system.Lchown()`, and re-linking things that we know were linked before we started walking the tree?

nalind:
I suppose if we only checked for "that multiply-linked thing got a new inode number" when `st.Nlinks > 1`, it might not hurt too bad.  If we kept a mapping from old inode numbers to new link targets, and relinked it instead of chowning the no-longer-linked lower item, would that work?

giuseppe:
> I suppose if we only checked for "that multiply-linked thing got a new inode number" when `st.Nlinks > 1`, it might not hurt too bad. If we kept a mapping from old inode numbers to new link targets, and relinked it instead of chowning the no-longer-linked lower item, would that work?

that is a much better solution.  I've played with it and it seems to work fine, I'll test it more and update the PR

rhatdan:
LGTM
@nalind PTAL

nalind:
Are there any tests we can add for this?

nalind:
LGTM

PR #1265, opened 2022-06-16
idtools: add lookup by UID with libsubid

giuseppe:
We lost this feature when we moved to using libsubid for looking up
user additional ranges.

If the lookup using the username fails then attempt again using the
UID, since /etc/subuid and /etc/subgid allow that.

Closes: https://github.com/containers/storage/issues/1264


giuseppe:
> LGTM
> 
> Any chance we can add a regression test?

that would require a system configuration change since we are going through libsubid

PR #1268, opened 2022-06-21
[release-1.24] drivers: enhance check for mount fs type

giuseppe:
when checking that a mount has a specific file system type, also
validate that it is on a different than its parent directory.

This helps when using virtiofs as the underlying file system since the
check used for fuse-overlayfs would fail since they both use FUSE.

(cherry picked from commit da09b934b6e019279b0489ace2e0194450f36a5b)

### DIFF
--- a/drivers/driver_linux.go
+++ b/drivers/driver_linux.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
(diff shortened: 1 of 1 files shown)


gkurz (approved):
FWIW. Thanks @giuseppe !

Issue #1270, opened 2022-06-27
rawhide(f37)  + --uidmap: Unmounting error: EINVAL


edsantiago:
On current (2022-06-27) f37:
```console
# podman run --uidmap 0:10000:10000 quay.io/libpod/testimage:20210610 true
ERRO[0000] Unmounting /var/lib/containers/storage/overlay/89bd3d0bfb07220c3b715562c8fcad9656350522d31348125a4b5548fff541b1/merged: invalid argument 
Error: error mounting storage for container 2f987a9a5038866172185259e8543d4bd0d443c8281d1a1c63fe194f99b14f41: creating overlay mount to /var/lib/containers/storage/overlay/89bd3d0bfb07220c3b715562c8fcad9656350522d31348125a4b5548fff541b1/merged, mount_data="lowerdir=/var/lib/containers/storage/overlay/89bd3d0bfb07220c3b715562c8fcad9656350522d31348125a4b5548fff541b1/mapped/0/l/NOGMYKSTI2JBFYP2GGCC6PHX2L,upperdir=/var/lib/containers/storage/overlay/89bd3d0bfb07220c3b715562c8fcad9656350522d31348125a4b5548fff541b1/diff,workdir=/var/lib/containers/storage/overlay/89bd3d0bfb07220c3b715562c8fcad9656350522d31348125a4b5548fff541b1/work,nodev,metacopy=on,context=\"system_u:object_r:container_file_t:s0:c96,c513\"": permission denied
```
It's not SELinux. Only error reported by journalctl is:
```
... podman[10845]: 2022-06-27 12:21:42.696031095 -0400 EDT m=+0.155750992 container create 2f987a9a5038866172185259e8543d4bd0d443c8281d1a1c63fe194f99b14f41 (image=quay.io/libpod/testimage:20210610, name=eloquent_noyce, created_at=2021-06-10T18:55:36Z, created_by=test/system/build-testimage, io.buildah.version=1.21.0)
... kernel: overlayfs: failed to resolve '/var/lib/containers/storage/overlay/89bd3d0bfb07220c3b715562c8fcad9656350522d31348125a4b5548fff541b1/mapped/0/l/NOGMYKSTI2JBFYP2GGCC6PHX2L': -13
```
[...]

PR #1271, opened 2022-06-28
overlay: create linkDir with mask 0755

giuseppe:
on newer kernels we use idmapped mounts for the overlayfs.

When running in a user namespace that doesn't include the root user
from the host, the linDir cannot be accessed since CAP_DAC_OVERRIDE
doesn't work with IDs that are not part of the mapping.

To avoid this problem, create the linkDir with permissions 0755.

Closes: https://github.com/containers/storage/issues/1270

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1214,7 +1214,6 @@ func (d *Driver) recreateSymlinks() error {
 	if err != nil {
 		return fmt.Errorf("reading driver home directory %q: %v", d.home, err)
 	}
-	linksDir := filepath.Join(d.home, "l")
 	// This makes the link directory if it doesn't exist
 	rootUID, rootGID, err := idtools.GetRootUIDGID(d.uidMaps, d.gidMaps)
 	if err != nil {
@@ -1258,9 +1257,12 @@ func (d *Driver) recreateSymlinks() error {
 				continue
 			}
 		}
+
+		// linkDirFullPath is the full path to the linkDir
+		linkDirFullPath := filepath.Join(d.home, "l")
 		// Now check if we somehow lost a "link" file, by making sure
 		// that each symlink we have corresponds to one.
-		links, err := ioutil.ReadDir(linksDir)
+		links, err := ioutil.ReadDir(linkDirFullPath)
 		if err != nil {
 			errs = multierror.Append(errs, err)
 			continue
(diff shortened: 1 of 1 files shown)


drivers/overlay/overlay.go
```diff
@@ -1216,11 +1216,7 @@ func (d *Driver) recreateSymlinks() error {
 	}
 	linksDir := filepath.Join(d.home, "l")
 	// This makes the link directory if it doesn't exist
-	rootUID, rootGID, err := idtools.GetRootUIDGID(d.uidMaps, d.gidMaps)
-	if err != nil {
-		return err
-	}
-	if err := idtools.MkdirAllAs(path.Join(d.home, linkDir), 0700, rootUID, rootGID); err != nil {
+	if err := idtools.MkdirAllAs(path.Join(d.home, linkDir), 0755, 0, 0); err != nil {
```
edsantiago, line 1219: I don't know nothing about nothing, but line 1217 made me sweat. Since `linksDir` (plural) isn't actually used until line 1259 (far, far below), and since a casual monkey reader like me could easily think that the singular `linkDir` right here could be a typo or misused variable... would you consider doing a little cleanup while you're here, moving the `linksDir` plural definition way way down, and adding a comment explaining how it differs from singular-`linkDir`?

giuseppe: I've not noticed it, thanks for the hint.  Pushed a new patch

rhatdan (approved):
LGTM

rhatdan:
@vrothberg @saschagrunert @nalind @flouthoc @umohnani8 PTAL

drivers/overlay/overlay.go
```diff
@@ -1214,13 +1214,8 @@ func (d *Driver) recreateSymlinks() error {
...
 		return fmt.Errorf("reading driver home directory %q: %v", d.home, err)
 	}
-	linksDir := filepath.Join(d.home, "l")
 	// This makes the link directory if it doesn't exist
-	rootUID, rootGID, err := idtools.GetRootUIDGID(d.uidMaps, d.gidMaps)
-	if err != nil {
-		return err
-	}
-	if err := idtools.MkdirAllAs(path.Join(d.home, linkDir), 0700, rootUID, rootGID); err != nil {
+	if err := idtools.MkdirAllAs(path.Join(d.home, linkDir), 0755, 0, 0); err != nil {
```
flouthoc, line 1218: Not blocking: Since commit is missing description for this change just a doubt previously this configured the remapped root uid/gid from provided uid and gid maps which is not same now i.e always `0/0` could this affect containers custom userns.

PR #1276, opened 2022-06-29
[release-1.40] drivers: enhance check for mount fs type

nalind:
When checking that a mountpoint is a specific file system type, also validate that it is on a different filesystem than its parent directory.

This helps when using virtiofs as the underlying file system for storage since the check used for fuse-overlayfs would previously fail since they both use FUSE.

Cherry-picked from https://github.com/containers/storage/pull/1261.

### DIFF
--- a/drivers/driver_linux.go
+++ b/drivers/driver_linux.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #1281, opened 2022-07-07
idtools: optimize SafeChown

giuseppe:
instead of first stat'ing the file to verify its ownership and in case
do a chown, attempt first the chown and in case it fails fallback to
verify that the ownership is already correct.

### DIFF
--- a/pkg/idtools/idtools.go
+++ b/pkg/idtools/idtools.go
@@ -387,12 +387,16 @@ func SafeChown(name string, uid, gid int) error {
 		uid = os.Getuid()
 		gid = os.Getgid()
 	}
+	chownErr := os.Chown(name, uid, gid)
+	if chownErr == nil {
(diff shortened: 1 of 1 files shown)


giuseppe:
@rhatdan this should address the point I've raised for using SafeChown with podman volumes

flouthoc (approved):
LGTM

rhatdan:
The problem with this would be that it changes the mtime of files as opposed to the original way which would not.  Also this will cause a copyup on overlay whereas before it would not.  Correct?

giuseppe:
> The problem with this would be that it changes the mtime of files as opposed to the original way which would not. Also this will cause a copyup on overlay whereas before it would not. Correct?

yes correct.  It is going to cause a copyup with overlay and it always update the time.  Given that, I am closing this PR

PR #1286, opened 2022-07-13
Fix wrong error return value

saschagrunert:
We now pre-check the error before returning it in wrapped state.

Refers to https://github.com/containers/storage/commit/3455d12729c882e5bbc37179e0f52e7819c9e251#r78367588


vrothberg (approved):
LGTM
@giuseppe PTAL

vrothberg:
I wonder why the linters didn't catch that

flouthoc (approved):
LGTM

Issue #1288, opened 2022-07-13
If layer is missing leading directory component(s), they get 0600 permission


graywolf-at-work:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I'm not really sure if this is a bug or not, but when you have a layer, that does not have a whole directory path defined, the missing components get 0600, which is not great.

It does not allow regular user to enter those directories nor do anything with them.

**Steps to reproduce the issue:**

0/ Combine the files since github sucks:

    cat xa* >x.tgz

[xaa.tgz](https://github.com/containers/podman/files/9102854/xaa.tgz)
[xab.tgz](https://github.com/containers/podman/files/9102855/xab.tgz)
[xac.tgz](https://github.com/containers/podman/files/9102856/xac.tgz)

1/ Load the attached file

    zcat <x.tgz | podman load

2/ Run it:

    podman run --rm localhost/bazel/packages/tarball:x ls -l /

3/ Notice that the `ruby-3.1.2` directory has only rw permission:

    drw-------   1 root   root    4734 Jul 13 13:49 ruby-3.1.2

**Describe the results you received:**

    drw-------

**Describe the results you expected:**

    drwxr-xr-x

**Additional information you deem important (e.g. issue happens only occasionally):**

N/A

**Output of `podman version`:**

```
Client:       Podman Engine
Version:      4.1.0
API Version:  4.1.0
Go Version:   go1.18.2
Git Commit:   73c021e0588393f18d2c54dec335ec9e960287a4
Built:        Thu May 12 02:00:55 2022
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.26.1
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  cgroupManager: cgroupfs
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.2-r0
    path: /usr/bin/conmon
    version: 'conmon version 2.1.2, commit: f25214cda836296fe8695c2e582ca79d2246053a'
  cpuUtilization:
    idlePercent: 80.83
    systemPercent: 5.22
    userPercent: 13.95
  cpus: 16
  distribution:
    distribution: alpine
    version: 3.16.0
  eventLogger: file
  hostname: foobar
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1001
      size: 1
    - container_id: 1
[...]

PR #1289, opened 2022-07-14
diff: create missing layers dirs with 0755

giuseppe:
if the parent directory is missing when creating a directory or
extracting a file from the tarball, then create the missing parent
directories with mode 0755 instead of 0600 so that a container can run
with uid != 0.

Closes: https://github.com/containers/storage/issues/1288


saschagrunert (approved):
LGTM

flouthoc (approved):
LGTM

PR #1300, opened 2022-07-25
Remove a couple of cases of possibly wrapping `nil` as errors

nalind:
There are a few places that look like we could end up passing a `nil` value to `fmt.Errorf()`'s "%w" specifier, and one where we were calling `Validate()` on a `Digest` a second time to wrap an error result instead of just reusing it.

### DIFF
--- a/images.go
+++ b/images.go
@@ -242,8 +242,8 @@ func (i *Image) recomputeDigests() error {
 		if !bigDataNameIsManifest(name) {
 			continue
 		}
-		if digest.Validate() != nil {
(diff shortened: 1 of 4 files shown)


rhatdan (approved):
LGTM

rhatdan:
Is this worth rushing into release?

saschagrunert (approved):

nalind:
> Is this worth rushing into release?

I don't see any calls to either version of `CopyFileWithTarAndChown()` when I scan the podman source tree, so I'm not sure it's used at all.  The other two sites this PR touches shouldn't be negatively affecting anything in their current form.

rhatdan:
Ok we will get this on the next vendor dance.

Issue #1304, opened 2022-08-04
gcc failed to link binary: multiple definition of 'init'


gustavosbarreto:
Take following code for example:
```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/containers/image/v5/copy"
	"github.com/containers/image/v5/signature"
	"github.com/containers/image/v5/transports/alltransports"
	_ "github.com/opencontainers/runc/libcontainer/nsenter"
)

func main() {
	policyContext, err := signature.NewPolicyContext(&signature.Policy{
		Default: []signature.PolicyRequirement{
			signature.NewPRInsecureAcceptAnything(),
		},
	})
	if err != nil {
		log.Print(err)
		os.Exit(1)
	}
	defer policyContext.Destroy()

	src, err := alltransports.ParseImageName("docker://registry-1.docker.io/...")
	if err != nil {
		log.Print(err)
		os.Exit(1)
	}

	dir, err := os.MkdirTemp("", "image")
	if err != nil {
		log.Print(err)
		os.Exit(1)
	}
	defer os.RemoveAll(dir)

	dst, err := alltransports.ParseImageName(fmt.Sprintf("dir:%s", dir))
	if err != nil {
		log.Print(err)
		os.Exit(1)
	}

	if _, err = copy.Image(context.Background(), policyContext, dst, src, nil); err != nil {
		log.Print(err)
		os.Exit(1)

	}

}

func runContainerUsingLibcontainer() {
	// libcontainer stuff
}
```

Error message when running `go build`:
```
/usr/lib/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: /tmp/go-link-1603053777/000034.o: in function `init':
/home/gustavo/go/pkg/mod/github.com/containers/storage@v1.42.0/pkg/unshare/unshare_cgo.go:9: multiple definition of `init'; /tmp/go-link-1603053777/000001.o:/home/gustavo/go/pkg/mod/github.com/opencontainers/runc@v1.1.3/libcontainer/nsenter/nsenter.go:10: first defined here
collect2: error: ld returned 1 exit status
```

rhatdan:
@nalind @giuseppe @saschagrunert PTAL

PR #1305, opened 2022-08-04
pkg/unshare: declare init() as static

gustavosbarreto:
This commit declare C init() function as static, since this is a constructor there is no need to export this symbol, avoiding errors like 'multiple definitions error' when linking the binary.

Fixes: #1304


saschagrunert (approved):

otavio (approved):

nalind:
Thanks for spotting this!  Can you also hit the version in `unshare_gccgo.go`?  LGTM with that.

nalind:
Thanks!  LGTM

PR #1307, opened 2022-08-16
replace syscall.Pause with <-time.After(math.MaxInt) for arm64 platform

derekhjray:
function `syscall.Pause` is not available for linux/arm64 platform, using `<-time.After(math.MaxInt)` instead

### DIFF
--- a/drivers/overlay/idmapped_utils.go
+++ b/drivers/overlay/idmapped_utils.go
@@ -6,8 +6,10 @@ package overlay
 import (
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"syscall"
+	"time"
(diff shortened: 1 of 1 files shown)


rhatdan:
Please sign your commits.

git commit -a --amend -s
git push --force

nalind:
When I give the compiler this:
```Go
package main

import (
	"fmt"
	"runtime"
	"syscall"
)

func main() {
	fmt.Printf("GOARCH: %s\n", runtime.GOARCH)
	syscall.Pause()
}
```
with GOARCH=arm64 in my environment, it seems to build correctly.  What are we missing?

rhatdan:
Since we never heard back, I am closing.

PR #1312, opened 2022-08-24
unshare: improve rootless detection

giuseppe:
in addition to check the UID of the user that launched the command,
also check whether the process has CAP_SYS_ADMIN (in the same way as
podman/pkg/rootless does) and also check that the current user
namespace has all the IDs available.

Closes: https://github.com/containers/storage/issues/1311

After this change, podman/pkg/rootless can use the function directly
instead of defining another version with similar functionalities.

### DIFF
--- a/pkg/unshare/unshare_linux.go
+++ b/pkg/unshare/unshare_linux.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -387,10 +388,39 @@ const (
 	UsernsEnvName = "_CONTAINERS_USERNS_CONFIGURED"
 )
 
+// hasFullUsersMappings checks whether the current user namespace has all the IDs mapped.
+func hasFullUsersMappings() (bool, error) {
+	content, err := ioutil.ReadFile("/proc/self/uid_map")
+	if err != nil {
+		return false, err
+	}
+	// if the uid_map contains 4294967295, the entire IDs space is available in the
+	// user namespace, so it is likely the initial user namespace.
+	return bytes.Contains(content, []byte("4294967295")), nil
+}
+
 // IsRootless tells us if we are running in rootless mode
 func IsRootless() bool {
 	isRootlessOnce.Do(func() {
 		isRootless = getRootlessUID() != 0 || getenv(UsernsEnvName) != ""
+		if !isRootless {
+			hasCapSysAdmin, err := HasCapSysAdmin()
+			if err != nil {
+				logrus.Warnf("Failed to read CAP_SYS_ADMIN presence for the current process")
+			}
+			if err == nil && !hasCapSysAdmin {
+				isRootless = true
+			}
+		}
+		if !isRootless {
+			hasMappings, err := hasFullUsersMappings()
+			if err != nil {
+				logrus.Warnf("Failed to read current user namespace mappings")
+			}
+			if err == nil && !hasMappings {
+				isRootless = true
+			}
+		}
 	})
 	return isRootless
 }


pkg/unshare/unshare_linux.go
```diff
@@ -387,10 +388,42 @@ const (
...
+			hasCapSysAdmin, err := HasCapSysAdmin()
+			if err != nil {
+				logrus.Warnf("Failed to read CAP_SYS_ADMIN presence for the current process")
+			}
+			if err == nil && !hasCapSysAdmin {
+				isRootless = true
+			}
+		}
+		if !isRootless {
+			hasMappings, err := hasFullUsersMappings()
```
rhatdan, line 419: Why would this indicate isRootless.

If I have a range of `0 0 1000000` won't the isRootless return true?  Aren't I running as root in this case?

giuseppe: But we are using `isRootless` to mean "not in the initial user namespace".  That is what the kernel does when checking what is allowed and what not.  For example, we need to use `user. ` xattrs for overlay whiteouts even if we are root in a user namespace

rhatdan:
LGTM
@flouthoc @nalind @vrothberg @saschagrunert @ashley-cui @umohnani8 PTAL

rhatdan (approved):

pkg/unshare/unshare_linux.go
```diff
@@ -387,10 +388,42 @@ const (
...
+func hasFullUsersMappings() (bool, error) {
+	content, err := ioutil.ReadFile("/proc/self/uid_map")
+	if err != nil {
+		return false, err
+	}
+	// Both runc and crun check whether the current process is in a user namespace
+	// by looking up 4294967295 in /proc/self/uid_map.  If the mappings would be
+	// copied as they are, the check in the OCI runtimes would fail.  So just split
+	// it in two different ranges.
+	return bytes.Contains(content, []byte("4294967295")), nil
```
nalind, line 402: This comment doesn't seem to describe what's happening before or after it, so I don't understand what's happening here.

giuseppe: I've fixed the comment and pushed a new version

flouthoc (approved):
LGTM

pkg/unshare/unshare_linux.go
mtrmac: I think by now, we **desperately need** documentation. “IsRootless tells us if we are running in rootless mode” says nothing.

Describe in which situations that function returns true (possibly how can one get into such situations), and what does that means for application behavior. What should, or shouldn’t, they be doing differently?

It might _well be the case_ that we need three or more different functions, `IsNotInTheInitialUserNamespace`, `IsInsideAPodmanCreatedContainerNamespace`,  and so on.

Just vaguely saying “rootless” is apparently causing misunderstandings and unanticipated breakage.

pkg/unshare/unshare_linux.go
```diff
@@ -387,10 +388,39 @@ const (
...
 
+// hasFullUsersMappings checks whether the current user namespace has all the IDs mapped.
+func hasFullUsersMappings() (bool, error) {
+	content, err := ioutil.ReadFile("/proc/self/uid_map")
+	if err != nil {
+		return false, err
+	}
+	// if the uid_map contains 4294967295, the entire IDs space is available in the
+	// user namespace, so it is likely the initial user namespace.
+	return bytes.Contains(content, []byte("4294967295")), nil
```
mtrmac, line 399: Can it happen that that that single ID is mapped but not the others?

nalind: The kernel rejects attempts to create mappings where either start point is 4294967295 with EINVAL.

nalind: Oh, right with you there.  I had to look it up.

PR #1314, opened 2022-08-29
overlay: use the right lowers for read-only mounts via subprocess

### LINKED ISSUE #640: Using mountopt metacopy=on results in build layers that include everything beneath
When running buildah on my system, new build layers included everything from layers beneath. When I do a `podman save` and extract the contents of the two layers, diff shows that most of the files in the two layers are identical. The new layer is slightly larger than the old layer.

@nalind helped me track it down to the `metacopy=on` `mountopt`. Removing it fixed the issue for me. Here's what he had to say on slack:

> Looks like something goes wrong when we try to use the `metacopy=on` option there.  it disables the native diff logic in the overlay driver, and the naive diff logic doesn't seem to react well to metacopy. 

**buildah info output:**
```json
{
    "host": {
        "CgroupVersion": "v1",
        "Distribution": {
            "distribution": "fedora",
            "version": "30"
        },
        "MemTotal": 66861326336,
        "MenFree": 4907630592,
        "OCIRuntime": "runc",
        "SwapFree": 0,
        "SwapTotal": 0,
        "arch": "amd64",
        "cpus": 8,
        "hostname": "xanadu.remote.redhat.com",
        "kernel": "5.5.16-100.fc30.x86_64",
        "os": "linux",
        "rootless": false,
        "uptime": "931h 10m 46.36s (Approximately 38.79 days)"
    },
    "store": {
        "ContainerStore": {
            "number": 19
        },
        "GraphDriverName": "overlay",
        "GraphOptions": [
            "overlay.mountopt=nodev,metacopy=on"
        ],
        "GraphRoot": "/var/lib/containers/storage",
        "GraphStatus": {
            "Backing Filesystem": "btrfs",
            "Native Overlay Diff": "false",
            "Supports d_type": "true",
            "Using metacopy": "true"
        },
        "ImageStore": {
            "number": 16
        },
        "RunRoot": "/var/run/containers/storage"
    }
}
```

I'm on Fedora 30 with podman-1.8.0-4.fc30.x86_64 and buildah-1.12.0-2.fc30.x86_64

nalind:
When we use a subprocess to mount a read-only layer, include the layer's own diff in the list of lowers, since it's not being used as an upper.  Fixes #640.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1608,7 +1608,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 			diffDir := path.Join(id, "diff")
 			opts = fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s", strings.Join(relLowers, ":"), diffDir, workdir)
 		} else {
-			opts = fmt.Sprintf("lowerdir=%s", strings.Join(relLowers, ":"))
+			opts = fmt.Sprintf("lowerdir=%s:%s", diffDir, strings.Join(relLowers, ":"))
 		}
 		if len(optsList) > 0 {
 			opts = fmt.Sprintf("%s,%s", opts, strings.Join(optsList, ","))


TomSweeneyRedHat:
LGTM
but my goodness, that get function is getting gi-normous.

rhatdan:
LGTM

PR #1318, opened 2022-08-31
Don't use subprocess output as a format specifier

### LINKED ISSUE #1316: Command output logged incorrectly
https://github.com/containers/storage/blob/134614cc082ff9dd804f1f349ddbb043ecd10eba/drivers/chown.go#L96 is using output as a format string.

nalind:
Don't pass the output of a subprocess to `fmt.Errorf()`, which will try to treat it as a format specifier.  Just use `errors.New()` to create an error using the text verbatim.

Fixes #1316.

### DIFF
--- a/drivers/chown.go
+++ b/drivers/chown.go
@@ -2,6 +2,7 @@ package graphdriver
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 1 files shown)


mtrmac (approved):
LGTM. Thanks!

flouthoc (approved):
LGTM

vrothberg (approved):
/lgtm

Issue #1321, opened 2022-09-05
Llistxattr fails on merged proc file system


blkt:
Hello folks, I'm reporting a problem we're facing with IBM VPC-based Kubernetes Clusters. I don't really know whether there's a way to reproduce it elsewhere, but this is basically what happens.

```
# ./skopeo --version
skopeo version 1.9.2 commit: ca1b0f34d1d2c3e2074c9c231249f1a0e4d3ceb7

# ./skopeo inspect containers-storage:10c83cb7b1e01cf593b568de30e88e3ca5f2e9655187c51050af0bb7980dce79
FATA[0013] Error parsing image name "containers-storage:10c83cb7b1e01cf593b568de30e88e3ca5f2e9655187c51050af0bb7980dce79": llistxattr /var/data/criorootstorage/overlay/5152d020458c568c1b8346ef74e2fa06ca7c20a643aa1914627a0b6fa5c42383/merged/proc/fs/nfsd: operation not supported
```

In this example, `skopeo` is shipping `containers/storage` version `v1.42.0`.

Am I correct saying that this routine [here](https://github.com/containers/storage/blob/main/pkg/system/xattrs_linux.go#L75-L77) should be modified to handle `EOPNOTSUPP` as well? I don't really see any reason for this use case not to be supported, but please let me know if there's some kind of spec that forbids this.

nalind:
Going by the other call sites in the storage library, we've been checking for `EOPNOTSUPP` at places that call `system.Llistxattr()` rather than trying to quietly ignore that error in the wrapper itself, so I think we'd more likely teach `drivers/overlay.scanForMountProgramIndicators()` to treat that error as a non-issue if `filepath.WalkDir()` returns it.

PR #1323, opened 2022-09-06
Don't fail on walking os of EOPNOTSUPP for Llistxatt

rhatdan:
Fixes: https://github.com/containers/storage/issues/1321

### DIFF
--- a/drivers/overlay/check_116.go
+++ b/drivers/overlay/check_116.go
@@ -1,14 +1,17 @@
-// +build go1.16
+//go:build linux
+// +build linux
 
 package overlay
 
 import (
+	"errors"
 	"io/fs"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind @giuseppe @vrothberg @saschagrunert @TomSweeneyRedHat @flouthoc PTAL

nalind:
LGTM.  I was thinking it would be faster to let us break out of `WalkDir()` on the first failure, but we probably do need to continue scanning at least the top-level directory.
Non-blocking: we should just merge `check_116.go` in with the regular `check.go`, since we dropped the 1.15-and-earlier implementation in #1176.

flouthoc (approved):
LGTM

saschagrunert (approved):

vrothberg (approved):
LGTM

PR #1327, opened 2022-09-08
chrootarchive: propagate env from parent

giuseppe:
propagate the environment variables from the current process since they are needed by pkg/unshare to read the UID for the user that created the user namespace (through the _CONTAINERS_ROOTLESS_UID env variable).


nalind:
This currently changes two of the places where we call `userns.RunningInUserNS()` without touching several others.  What's the rationale?

giuseppe:
> This currently changes two of the places where we call `userns.RunningInUserNS()` without touching several others. What's the rationale?

I've touched only the file that I was working on as part of the fix.  I agree it makes no sense to change only that, so I've adapted other files as well and pushed a new version

rhatdan:
LGTM

Luap99:
I applied this fix in https://github.com/containers/podman/pull/15695 and it seems to work (at least the compose and API tests which failed previously are now green).

flouthoc (approved):
LGTM

PR #1328, opened 2022-09-08
Remove hasMountProgram when running storage without mount program

rhatdan:
Fixes: https://github.com/containers/podman/issues/15698

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -307,6 +307,16 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 		return nil, err
 	}
 
+	if opts.mountProgram == "" {
+		if err := os.Remove(getMountProgramFlagFile(home)); err != nil && !errors.Is(err, os.ErrNotExist) {
+			return nil, err
+		}
+	} else {
+		if err := ioutil.WriteFile(getMountProgramFlagFile(home), []byte("true"), 0600); err != nil {
+			return nil, err
+		}
+	}
+
 	fsMagic, err := graphdriver.GetFSMagic(home)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


rhatdan:
@edsantiago @nalind @giuseppe PTAL

drivers/overlay/overlay.go
```diff
@@ -307,6 +307,16 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
 		return nil, err
 	}
 
+	if opts.mountProgram == "" {
+		if err := os.Remove(getMountProgramFlagFile(home)); err != nil && !errors.Is(err, os.ErrNotExist) {
+			return nil, err
```
giuseppe, line 312: I think this can be a problem if you run a podman command with custom storage opts.  You could force to not use the mount program for a single command, but I am not sure it should affect the storage because images stored there to be consumed by the mount program are still present and won't be usable with native overlay.

giuseppe:
alternative fix: https://github.com/containers/storage/pull/1329

drivers/overlay/overlay.go
rhatdan: The question I have is does this change effect any other operation.  Bottom line, with the current code, as soon as you specify a --storage-opt, the file is written and from then on the system assumes that it has a mount program.

For example, will the system stop doing Metacopyup?

giuseppe: it doesn't affect any operation, it affects only operations that create a new layer, like pulling an image; but we would need to add this logic and propagate it down to c/storage, otherwise we risk of writing images in a different format.

Is there any real use case where the user wants to specify a mount program just for one command?

The only reason for using fuse-overlayfs as root is to handle a storage that is on NFS, I don't see any other valid use case otherwise.

Changing the mount program with `--storage-opt` is more of a testing tool for trying out a new fuse-overlayfs than something we should worry about.

PR #1333, opened 2022-09-10
Fix various warnings

mtrmac:
… as reported by the default configuration of VS Code, just because they are distracting me.

Primarily this updates users of `io/ioutil`, but many other smaller things as well.

More to follow sometime…


TomSweeneyRedHat:
LGTM

PR #1334, opened 2022-09-11
ReadMemInfo: Add missing import of errors

dfr:
This wasn't picked up by the freebsd cross build because meminfo_freebsd.go relies on cgo.


rhatdan:
LGTM

PR #1336, opened 2022-09-12
pkg/lockfile: nits

kolyshkin:
Fix one issue and a couple of nits in pkg/lockfile. Please see individual commits for details.

### DIFF
--- a/pkg/lockfile/lockfile_unix.go
+++ b/pkg/lockfile/lockfile_unix.go
@@ -71,15 +71,11 @@ func newLastWriterID() []byte {
 // openLock will create the file and its parent directories,
 // if necessary.
 func openLock(path string, ro bool) (fd int, err error) {
-	if ro {
-		fd, err = unix.Open(path, os.O_RDONLY|unix.O_CLOEXEC|os.O_CREATE, 0)
-	} else {
-		fd, err = unix.Open(path,
-			os.O_RDWR|unix.O_CLOEXEC|os.O_CREATE,
-			unix.S_IRUSR|unix.S_IWUSR|unix.S_IRGRP|unix.S_IROTH,
-		)
+	flags := os.O_RDONLY | unix.O_CLOEXEC | os.O_CREATE
+	if !ro {
+		flags |= os.O_RDWR
 	}
-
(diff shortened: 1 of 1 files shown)


pkg/lockfile/lockfile_unix.go
```diff
@@ -71,15 +71,11 @@ func newLastWriterID() []byte {
...
-	if ro {
-		fd, err = unix.Open(path, os.O_RDONLY|unix.O_CLOEXEC|os.O_CREATE, 0)
-	} else {
-		fd, err = unix.Open(path,
-			os.O_RDWR|unix.O_CLOEXEC|os.O_CREATE,
-			unix.S_IRUSR|unix.S_IWUSR|unix.S_IRGRP|unix.S_IROTH,
-		)
+	flags := os.O_RDONLY | unix.O_CLOEXEC | os.O_CREATE
+	if !ro {
+		flags |= os.O_RDWR
```
kolyshkin, line 76: This relies on the fact that O_RDONLY value is 0.

rhatdan:
LGTM
@nalind @flouthoc @vrothberg @giuseppe @cdoern @TomSweeneyRedHat @umohnani8 PTAL

giuseppe (approved):
LGTM

pkg/lockfile/lockfile_unix.go
mtrmac: That’s
- Extremely non-obvious to careless users.
- Not guaranteed to be true, and in fact not true on at least one Go-supported platform right now (though, to be fair, not a platform _currently_ supported by Podman).

Please just do the obviously-correct thing.

mtrmac: https://github.com/containers/storage/pull/1343

kolyshkin: More generally, the assumption is O_RDONLY does not have any bits set that O_RDWR doesn't.

That is true even for zOS (the only Golang platform where O_RDONLY != 0). In zOS, O_RDONLY is 2 and O_RDWR is 3.

Having said that, your code is definitely better as it does not have any such undocumented assumptions.

PR #1337, opened 2022-09-13
pkg/stringid: deprecate GenerateNonCryptoID

kolyshkin:
This is essentially a port of https://github.com/moby/moby/pull/39336.

Instead of removing the function, mark it as deprecated (and hope that linter errors will force people to fix the code).

Followups:
 * https://github.com/containers/podman/pull/15788
 * https://github.com/containers/common/pull/1155

### DIFF
--- a/drivers/aufs/aufs_test.go
+++ b/drivers/aufs/aufs_test.go
@@ -732,7 +732,7 @@ func BenchmarkConcurrentAccess(b *testing.B) {
 	// create a bunch of ids
 	var ids []string
 	for i := 0; i < numConcurrent; i++ {
-		ids = append(ids, stringid.GenerateNonCryptoID())
+		ids = append(ids, stringid.GenerateRandomID())
 	}
 
 	if err := d.Create(ids[0], "", nil); err != nil {
--- a/pkg/stringid/stringid.go
+++ b/pkg/stringid/stringid.go
@@ -2,18 +2,12 @@
 package stringid
 
 import (
-	cryptorand "crypto/rand"
+	"crypto/rand"
 	"encoding/hex"
 	"fmt"
-	"io"
-	"math"
-	"math/big"
-	"math/rand"
 	"regexp"
 	"strconv"
 	"strings"
-	"sync"
-	"time"
 )
 
 const shortLen = 12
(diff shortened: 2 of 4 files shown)


kolyshkin:
@mtrmac @rhatdan @nalind PTAL

mtrmac (commented):
Neither this version nor Docker’s one contains any rationale…

If we want to be _pedantically_ reliable, we need to look for ID collisions and retry; a better RNG might make collisions less likely (especially pre-#1065) but they can still happen.

Given that, I’m neutral on this.

mtrmac:
Also, I think this adds a syscall per ID generation, as of Go 1.19.

But then again, most users of this subpackage call the fully-random function already, and there’s _so much_ going on when creating containers and the like already, so I guess it makes little difference either way (?).

---

Still neutral on the merits; maybe just being the same as moby/moby is valuable enough to do this.

kolyshkin:
@mtrmac thanks for your input!

> Neither this version nor Docker’s one contains any rationale…

Alas, I don't fully understand the reason behind this change either, thus better no rationale than a bogus one. Please consider this an RFC.

> Also, I think this adds a syscall per ID generation, as of Go 1.19.

AFAIK syscalls (per se) are pretty cheap nowadays (the primary reason why `readfile(2)` was not accepted). Having said that, I didn't do any benchmarks.

saschagrunert (approved):

rhatdan:
@nalind @mrunalp WDYT?

rhatdan:
@giuseppe WDYT?

mtrmac:
Meanwhile, https://github.com/containers/common/pull/1155 adds a new `math/rand` used ID generator.

Is there some overall goal or a reason to do all these changes?

kolyshkin:
> Meanwhile, [containers/common#1155](https://github.com/containers/common/pull/1155) adds a new `math/rand` used ID generator.
> 
> Is there some overall goal or a reason to do all these changes?

Great question. Moving this one to draft until I get an answer.

rhatdan:
@kolyshkin are you still working on this?  Interested in it?

TomSweeneyRedHat:
@kolyshkin any status on this or https://github.com/containers/common/pull/1155?  We just got a backport request for it in: https://bugzilla.redhat.com/show_bug.cgi?id=2157930

kolyshkin:
OK, I guess I don't have the guts to move forward with this change.

PR #1339, opened 2022-09-14
Use defined constants for flag names

nalind:
Use constants for the names of flags that we set in `Flags` maps that we store in layer/image/container records, to make it easier to avoid possible breakages due to typos in the future.


rhatdan:
LGTM
@giuseppe @vrothberg @flouthoc @ashley-cui @umohnani8 @saschagrunert PTAL

giuseppe (approved):
LGTM

PR #1340, opened 2022-09-14
overlay: when we warn about the kernel, mention the version

nalind:
If we're going to complain about the booted kernel, include information about which kernel it is.


rhatdan:
LGTM
@giuseppe @vrothberg @saschagrunert @TomSweeneyRedHat @umohnani8 @flouthoc PTAL

giuseppe (approved):
LGTM

PR #1341, opened 2022-09-14
Tweak a couple of error messages

nalind:
Add missing punctuation to a couple of error messages, and include which mount options we tried to use when we complain about `mount()` failing.


rhatdan:
LGTM
@giuseppe @vrothberg @saschagrunert @TomSweeneyRedHat @umohnani8 @flouthoc PTAL

flouthoc (approved):
LGTM

PR #1342, opened 2022-09-14
layers.Wipe(): try to remove layers before their parents

nalind:
When removing layers, try to remove layers before removing their parents, as the lower-level driver may enforce the dependency beyond what the higher-level logic does or knows.

Do this by sorting by creation time as an attempt at flattening the layer tree into an ordered list.  It won't be correct if a parent layer needed to be recreated, but it's more likely to be correct otherwise.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -1388,6 +1388,9 @@ func (r *layerStore) Wipe() error {
 	for id := range r.byid {
 		ids = append(ids, id)
 	}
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@giuseppe @vrothberg @flouthoc @ashley-cui @umohnani8 @saschagrunert PTAL

giuseppe (approved):
LGTM

PR #1343, opened 2022-09-14
Remove an undocumented assumption that O_RDONLY == 0

mtrmac:
There's at least one Go platform where that is not true.

Making that assumption without any warning is just a trap for future readers.

Instead, just do the obviously correct thing, per the API documentation.

On macOS, this compiles to _literally exactly the same binary code_.

So, let the compiler be clever so that we don't have to.


rhatdan:
LGTM

giuseppe (approved):
LGTM

PR #1344, opened 2022-09-14
Deprecate lockfile.Locker.RecursiveLock (alternative to #1376)

mtrmac:
It is not valid as currently implemented, and there is no known user.

Cc: @vrothberg : This was added in #347 , supposedly for the concurrent copy detection.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -617,6 +617,8 @@ func (r *containerStore) Lock() {
 	r.lockfile.Lock()
 }
 
+// Deprecated: This can block indefinitely if the current goroutine owns the lock, and another goroutine is trying to acquire a writer lock.
+// Do not use this.
 func (r *containerStore) RecursiveLock() {
 	r.lockfile.RecursiveLock()
 }
--- a/images.go
+++ b/images.go
@@ -798,6 +798,8 @@ func (r *imageStore) Lock() {
 	r.lockfile.Lock()
 }
 
+// Deprecated: This can block indefinitely if the current goroutine owns the lock, and another goroutine is trying to acquire a writer lock.
+// Do not use this.
 func (r *imageStore) RecursiveLock() {
 	r.lockfile.RecursiveLock()
 }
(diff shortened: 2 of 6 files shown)


vrothberg (commented):
Why not fix it?

I didn't think it entirely through but assume we still need it for the concurrent pull work, don't we?

mtrmac:
> Why not fix it?

Sure, in principle.

I don’t immediately know how to do that
- Go does not provide a recursive locking primitive
- It does not even provide a goroutine ID, which would be necessary for my first guess at how to build one ourselves
- We need to somehow integrate that goroutine with non-recursive uses of the same locks — or maybe we don’t; what does the currently-nonexistent concurrent-pull implementation truly need?

> … assume we still need it for the concurrent pull work, don't we?

I don’t know, I haven’t looked into that.

---

… and I can’t spend time on this for the next few weeks (months?) at least. In the meantime, I want the problem to be recorded and documented, to prevent accidents.

We can always un-deprecate this after the implementation is fixed.

mtrmac:
- and we need tests to actually exercise the recursiveness of the locks, contention in recursive situations, and interaction with non-recursive read/write lock owners; IIRC the current ones don’t do (at least some of?) that.

vrothberg:
> We can always un-deprecate this after the implementation is fixed.

Yes, that sounds like a good idea.

LGTM, @nalind are you on board?

nalind:
It's fine, though the prospect of keeping an API around with known breakage is troubling.  If it's only there for a WIP caller that hasn't been merged and isn't actively being worked on, I'm more inclined to just remove it outright and revisit it when it becomes relevant again.

mtrmac:
> I'm more inclined to just remove it outright

Sure: #1376 . I don’t have a strong preference.

---

(I have updated this one to deprecate also the callers in C/I/L stores; they seem to be public but can’t actually be directly accessed by external users anyway.)

rhatdan:
Lets remove it.

PR #1345, opened 2022-09-14
Make {container, layer, image} store APIs private

mtrmac:
c/storage currently exports `ContainerStore`,`ImageStore`,`LayerStore`, and several component interfaces, publicly.

Meanwhile there are no _public_ ways to obtain implementations of these interfaces (and no non-public callers in Podman or CRI-O that I can find); and the public interfaces at least _look_ like an impediment to changing the C/I/L store interfaces (e.g. to change locking).

To clarify that, make the used C/I/L store implementations private (and keep the public interfaces in `deprecated.go` purely for formal API compatibility).

Then, as a proof of concept, remove unused <s>`RecursiveLock` (compare #1344),</s> `LoadLocked`, and `Lookup` methods.

@nalind RFC. I might well be missing something about external users of these interfaces.
Cc: @vrothberg for the `RecursiveLock` method removal, along with #1344.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -66,12 +66,12 @@ type Container struct {
 	Flags map[string]interface{} `json:"flags,omitempty"`
 }
 
-// ContainerStore provides bookkeeping for information about Containers.
-type ContainerStore interface {
-	FileBasedStore
-	MetadataStore
-	ContainerBigDataStore
-	FlaggableStore
+// rwContainerStore provides bookkeeping for information about Containers.
+type rwContainerStore interface {
+	fileBasedStore
+	metadataStore
+	containerBigDataStore
+	flaggableStore
 
 	// Create creates a container that has a specified ID (or generates a
 	// random one if an empty value is supplied) and optional names,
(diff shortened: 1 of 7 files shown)


mtrmac:
(This could go further, making all the methods in the interfaces also package-private, but that would, right now, be a lost of noise that detracts from the core of this PR. That can happen over time, as methods are added/modified.)

mtrmac:
> This could go further, making all the methods in the interfaces also package-private,

Actually #1346 would end up doing that, if we accept the idea of lock ownership tokens.

rhatdan:
LGTM

rhatdan:
@nalind @saschagrunert @vrothberg @flouthoc @ashley-cui @umohnani8 PTAL

rhatdan:
@mtrmac needs a rebase.  Then I think we should get this in.

rhatdan (approved):

rhatdan:
@nalind @saschagrunert @vrothberg @flouthoc @ashley-cui @umohnani8 PTAL

flouthoc (approved):
LGTM @nalind @giuseppe @vrothberg PTAL

Issue #1332, opened 2022-09-09
layerStore.Load under layerStore.RLock is racy


mtrmac:
From #1322:
- `store.Mounted` (locks `layerStore` for reading) → `ReloadIfChanged` (locks `loadMut`, irrelevant because there is no other concurrent `ReloadIfChanged`) → `Load` (locks nothing, updates things like `r.byid`)
- vs. `store.Layer` (locks `layerStore` for reading → `Get` (locks nothing, reads things like `r.byid`)

The read lock is shared, right? So there is no exclusion of `Load` modifying the data structure vs. any other reader of the store AFAICT.

I just can’t see how calling `ReloadIfChanged` with a shared reader lock is supposed to work in this case. My initial guess is that it is not supported _at all_ to use the same store from multiple goroutines, and each goroutine should have its own store — but that’s clearly not correct, `store.GetStore` carefully returns a pre-existing matching store instance for concurrent callers.

mtrmac:
@nalind @haircommander I’ve been told that performance with in-process concurrency is a major concern in this area.

Is there any way I could quantify that? Ideally a benchmark I can run on-demand? Or specific code paths that matter? (I _assume_ that’s mostly CRI-O, which nowadays has its own in-memory caching for both images and sandboxes. Which parts of CRI-O?)

haircommander:
I don't think we have a good reproducer for storage deadlocks/lock contention/races as of now. We've only seen theses situations in production so far

mtrmac:
A tentative RFC:

Currently, all getters within a single process lock a `layerStore` for reading, via `layerStore.lockfile`. But they, also, on each call, serialize on `layerStore.loadMut` (inside `layerStore.ReloadIfChanged`).

I propose to, more or less, turn `loadMut` into a RWLock, which is locked for reading during any read accesses to `layerStore`, and locked for writing during any updates. That means:
- ~No changes to correct writers to store; they current lock for exclusive use cross-process, and also intra-process.
- Hopefully trivially small changes to concurrent readers _without a concurrent external modification_: They can all hold `loadMut` for reading concurrently. The difference is only the extra overhead to lock/unlock the intra-process lock. This should be the CRI-O case, where there _usually_ are no concurrent external writers.
[...]

PR #1346, opened 2022-09-15
In-process exclusion of readers and writers

mtrmac:
**EDIT: See the description in https://github.com/containers/storage/pull/1346#issuecomment-1314316901 instead.**

To see what fixing #1332 would be like. This is on top of, and includes, #1345.

The goal is to protect in-process readers against concurrent writes to memory by `Load`. Let’s discuss the theory of that, and the performance, in #1332 , this is an attempt to draft what that could look like.

(Most of this PR could be merged, and might be useful, without actually changing the locking scheme — but without changing the locking scheme I don’t think we have a reason to do and review all of those code changes).

Right now, locking is open-coded all over the >3000 lines of `storage.store`:
```go
func (s *store) ListLayerBigData(id string) ([]string, error) {
        lstore, err := s.LayerStore()
        if err != nil {
                return nil, err
        }
        lstores, err := s.ROLayerStores()
        if err != nil {
                return nil, err
        }
        foundLayer := false
        for _, s := range append([]ROLayerStore{lstore}, lstores...) {
                store := s
                store.RLock()
                defer store.Unlock()
                if err := store.ReloadIfChanged(); err != nil {
                        return nil, err
                }
                data, err := store.BigDataNames(id)
                if err == nil {
                        return data, nil
                }
                if store.Exists(id) {
                        foundLayer = true
                }
        }
        if foundLayer {
                return nil, fmt.Errorf("locating big data for layer with ID %q: %w", id, os.ErrNotExist)
        }
        return nil, fmt.Errorf("locating layer with ID %q: %w", id, ErrLayerUnknown)
}
```
and adding _extra_ code to each of these sites to deal with the in-process lock would be a tedious nightmare with too many opportunities to overlook something..

So, this PR proposes consolidating almost all of these repeated snippets into helpers that call closures: `layerReadAccess` and `layerWriteAccess` at the level of a `layerStore`, and even higher-level (especially for reads) `store.writeToLayerStore` and `readAllLayerStores[T](*store)`.

The same function can now be:
```go
func (s *store) ListLayerBigData(id string) ([]string, error) {
	foundLayer := false
	if data, err, done := readAllLayerStores(s, func(store roLayerStore, token layerReadToken) ([]string, error, bool) {
		data, err := store.bigDataNames(token, id)
		if err == nil {
			return data, nil, true
		}
		if store.exists(token, id) {
			foundLayer = true
		}
		return nil, nil, false
	}); done {
		return data, err
	}
	if foundLayer {
		return nil, fmt.Errorf("locating big data for layer with ID %q: %w", id, os.ErrNotExist)
	}
	return nil, fmt.Errorf("locating layer with ID %q: %w", id, ErrLayerUnknown)
}
```
which is < 2/3 of the original size.

At least for the read side, it seems to me that any variant without Go 1.18 generics would need 2 or 3 more tedious lines at every call site. That’s not a show-stopper, but let’s discuss whether we can do it this way. (A notable downside is that the Go 1.18 requirement might not allow us to backport the locking changes to any earlier version … but then with the size of the locking changes, more than almost all of `store.go`, we might not want to do that anyway?)

After having the closures, we can now quite cheaply add `layerReadToken` and `layerWriteToken` types that serve as markers that the layer store’s methods are called with the correct lock held. That’s not nearly as good as a safe language-level support (making the lock-protected fields outright unreachable without holding a lock), but it’s _something_.

---

This does not actually show handling of the harder cases, like `CreateContainer`, where we might need to lock several stores, and the like. I’m confident that can be done — we can always do locking manually instead of through the closures; but I didn’t want to spend time on this without having basic consensus on the approach at all.

Also, there seem to be _quite a few_ opportunities for small-scale optimization, typically to check for `ErrLayerUnknown` from a call instead of calling `Exists` before/after the main query. After a few false starts, this PR tries _very hard_ not to touch that logic, because that would ballon the scope of this work to a completely unmanageable and impossible to review size.

---

I’d _very much_ appreciate
- opinions on whether this style is reasonable and acceptable overall
- any suggestions to make the code cleaner. (I never thought I’d say this but I found myself thinking that this would probably be significantly easier to do, _and_ even easier to prove correct, in C++, due to tuples, variadic templates and `const`.)

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -141,16 +141,21 @@ type rwContainerStore interface {
 }
 
 type containerStore struct {
-	lockfile   *lockfile.LockFile
-	dir        string
-	jsonPath   [numContainerLocationIndex]string
+	// The following fields are only set when constructing containerStore, and must never be modified afterwards.
+	// They are safe to access without any other locking.
+	lockfile *lockfile.LockFile // Synchronizes readers vs. writers of the _filesystem data_, both cross-process and in-process.
+	dir      string
+	jsonPath [numContainerLocationIndex]string
+
+	inProcessLock sync.RWMutex // Can _only_ be obtained with lockfile held.
+	// The following fields can only be read/written with read/write ownership of inProcessLock, respectively.
+	// Almost all users should use startReading() or startWriting().
 	lastWrite  lockfile.LastWrite
 	containers []*Container
 	idindex    *truncindex.TruncIndex
 	byid       map[string]*Container
 	bylayer    map[string]*Container
 	byname     map[string]*Container
-	loadMut    sync.Mutex
 }
 
 func copyContainer(c *Container) *Container {
@@ -202,6 +207,7 @@ func (c *Container) MountOpts() []string {
 	}
 }
 
+// The caller must hold r.inProcessLock for reading.
 func containerLocation(c *Container) containerLocations {
 	if c.volatileStore {
 		return volatileContainerLocation
@@ -216,9 +222,11 @@ func containerLocation(c *Container) containerLocations {
 // should use startWriting() instead.
 func (r *containerStore) startWritingWithReload(canReload bool) error {
 	r.lockfile.Lock()
+	r.inProcessLock.Lock()
 	succeeded := false
 	defer func() {
 		if !succeeded {
+			r.inProcessLock.Unlock()
 			r.lockfile.Unlock()
 		}
 	}()
@@ -241,21 +249,34 @@ func (r *containerStore) startWriting() error {
 
 // stopWriting releases locks obtained by startWriting.
 func (r *containerStore) stopWriting() {
+	r.inProcessLock.Unlock()
 	r.lockfile.Unlock()
 }
 
 // startReading makes sure the store is fresh, and locks it for reading.
 // If this succeeds, the caller MUST call stopReading().
 func (r *containerStore) startReading() error {
+	// inProcessLocked calls the nested function with r.inProcessLock held for writing.
+	inProcessLocked := func(fn func() error) error {
+		r.inProcessLock.Lock()
+		defer r.inProcessLock.Unlock()
+		return fn()
+	}
+
 	r.lockfile.RLock()
-	unlockFn := r.lockfile.Unlock // A function to call to clean up, or nil
+	unlockFn := r.lockfile.Unlock // A function to call to clean up, or nil.
 	defer func() {
 		if unlockFn != nil {
 			unlockFn()
 		}
 	}()
 
-	if tryLockedForWriting, err := r.reloadIfChanged(false); err != nil {
+	var tryLockedForWriting bool
+	if err := inProcessLocked(func() error {
+		var err error
+		tryLockedForWriting, err = r.reloadIfChanged(false)
+		return err
+	}); err != nil {
 		if !tryLockedForWriting {
 			return err
 		}
@@ -264,7 +285,10 @@ func (r *containerStore) startReading() error {
 
 		r.lockfile.Lock()
 		unlockFn = r.lockfile.Unlock
-		if _, err := r.reloadIfChanged(true); err != nil {
+		if err := inProcessLocked(func() error {
+			_, err := r.reloadIfChanged(true)
+			return err
+		}); err != nil {
 			return err
 		}
 		unlockFn()
@@ -272,22 +296,34 @@ func (r *containerStore) startReading() error {
 
 		r.lockfile.RLock()
 		unlockFn = r.lockfile.Unlock
-		// We need to check for a reload reload once more because the on-disk state could have been modified
+		// We need to check for a reload once more because the on-disk state could have been modified
 		// after we released the lock.
 		// If that, _again_, finds inconsistent state, just give up.
 		// We could, plausibly, retry a few times, but that inconsistent state (duplicate container names)
 		// shouldn’t be saved (by correct implementations) in the first place.
-		if _, err := r.reloadIfChanged(false); err != nil {
+		if err := inProcessLocked(func() error {
+			_, err := r.reloadIfChanged(false)
+			return err
+		}); err != nil {
 			return fmt.Errorf("(even after successfully cleaning up once:) %w", err)
 		}
 	}
 
+	// NOTE that we hold neither a read nor write inProcessLock at this point. That’s fine in ordinary operation, because
+	// the on-filesystem r.lockfile should protect us against (cooperating) writers, and any use of r.inProcessLock
+	// protects us against in-process writers modifying data.
+	// In presence of non-cooperating writers, we just ensure that 1) the in-memory data is not clearly out-of-date
+	// and 2) access to the in-memory data is not racy;
+	// but we can’t protect against those out-of-process writers modifying _files_ while we are assuming they are in a consistent state.
+
 	unlockFn = nil
+	r.inProcessLock.RLock()
 	return nil
 }
 
 // stopReading releases locks obtained by startReading.
 func (r *containerStore) stopReading() {
+	r.inProcessLock.RUnlock()
 	r.lockfile.Unlock()
 }
 
@@ -296,16 +332,20 @@ func (r *containerStore) stopReading() {
 // The caller must hold r.lockfile for reading _or_ writing; lockedForWriting is true
 // if it is held for writing.
 //
+// The caller must hold r.inProcessLock for WRITING.
+//
 // If !lockedForWriting and this function fails, the return value indicates whether
 // reloadIfChanged() with lockedForWriting could succeed.
 func (r *containerStore) reloadIfChanged(lockedForWriting bool) (bool, error) {
-	r.loadMut.Lock()
-	defer r.loadMut.Unlock()
-
 	lastWrite, modified, err := r.lockfile.ModifiedSince(r.lastWrite)
 	if err != nil {
 		return false, err
 	}
+	// We require callers to always hold r.inProcessLock for WRITING, even if they might not end up calling r.load()
+	// and modify no fields, to ensure they see fresh data:
+	// r.lockfile.Modified() only returns true once per change.  Without an exclusive lock,
+	// one goroutine might see r.lockfile.Modified() == true and decide to load, and in the meanwhile another one could
+	// see r.lockfile.Modified() == false and proceed to use in-memory data without noticing it is stale.
 	if modified {
 		if tryLockedForWriting, err := r.load(lockedForWriting); err != nil {
 			return tryLockedForWriting, err // r.lastWrite is unchanged, so we will load the next time again.
@@ -373,6 +413,7 @@ func (r *containerStore) datapath(id, key string) string {
 //
 // The caller must hold r.lockfile for reading _or_ writing; lockedForWriting is true
 // if it is held for writing.
+// The caller must hold r.inProcessLock for WRITING.
 //
 // If !lockedForWriting and this function fails, the return value indicates whether
 // retrying with lockedForWriting could succeed.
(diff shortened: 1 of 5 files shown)


mtrmac:
⚠️ Do not merge as is: On top of the upgrade to Go 1.18, this outright disables all of `golangci-lint`, because it reports _so many_ warnings, many of them at least a bit valid, and fixing them is not the point of this PR.

mtrmac:
> After having the closures, we can now quite cheaply add `layerReadToken` and `layerWriteToken` types that serve as markers that the layer store’s methods are called with the correct lock held.

_External_ callers of `layerStore` could probably be protected easier by being provided a “open with lock” / “operations allowed when locked for reading” / ”operations allowed when locked for writing”, making it impossible to call the wrong methods without a cast.

But that would not protect _internal_ calls within `layerStore` methods, because regardless of which interface they implement, they would be methods on the underlying object.

Also, the “Token” objects, unlike interface pointers, could _in principle_ carry state to allow diagnosing their misuse (They don’t currently, they are zero-sized structs.). Interfaces pointers are ~stateless.

mtrmac:
For the record, the empty …Token structs are optimized out and cause no extra machine code to be generated, at least in toy examples.

mtrmac:
Per the #1332 discussion, to see whether this locking has acceptable performance, essentially _all_ of it has to be built first.

So, as much as I _like_ the tokens and the extra bit of safety, that’s a bit too much churn.

So, I’ll be trying:
- To reduce boilerplate in `store` as much as possible, at least for the simple cases — to decrease the size of the code and the number of lines that need to be touched by any other changes.
- To replace `{,R,Un}Lock` on the C/I/L stores, + a manual `ReloadIfChanged`, with a higher-level `{start,stop}{Reading,Writing}`, explicitly so that callers don’t explicitly manage the inter-process locks, and instead express intent. That will also allow us to do cleanup in `start*` for #1136 — even if the intended operation is a “read” the stores will have the freedom to lock for writing (or to loop obtaining and releasing a lock, if necessary). (The token approach is nice _when it matches_, but many of the important write code paths would require significant restructuring of the code or the locking scheme, and that’s an explicit non-goal of this effort.)
- Both of the above could be an improvement in every situation, at least decreasing code size. But we can then change what the locks _do_, just by modifying the very few `start*/stop*/Load` methods.

mtrmac:
Next steps are to merge #1345 , then I have a set of thematic branches lined up for PRs.

vrothberg:
This is on my todo list but it's a massive PR that requires some uninterrupted time to review which is rare at the moment.

rhatdan:
@mtrmac needs a rebase.

rhatdan:
@mtrmac is this still a WIP?

mtrmac:
Yes — it’s only on the layer store, not on the container/image ones (or maybe even the global `storage.Store`), there are FIXMEs…

#1407 is the next step — that one should be review-ready but I still want to actually inject failures manually, and test that code.

mtrmac:
OK, this is now ready for review.

- Fixes #1332:

  The original design https://github.com/containers/storage/pull/898#issuecomment-840365674 works for the container and image store, but not for the layer store, which can become “modified” and cause reload from disk during the lifetime of any reader.

  So, instead of a short-term `r.loadMut` along with `r.lockfile` _together_ protecting data access, we now do the traditional thing, and all fields of the stores are protected by an ordinary `r.inProcessLock`, completely independently from the cross-process state.

 This should make no performance difference for writers, which are already completely exclusive via `r.lockfile`; they now hold an additional exclusive lock (`r.inProcessLock`) over the fun lifetime of their exclusive access, where before they held an exclusive lock (`r.loadMut`) for a short time.

  For readers, in _most_ cases, the period of exclusively holding `r.loadMut` is directly replaced by a period of exclusively holding `r.inProcessLock`, but on top they now hold a shared `r.inProcessLock` for the duration of their operation. So the cost is just the code & cache impact of acquiring and releasing that lock; this lock is _almost never_ going to block — only in the targeted case of actually triggering a `layerStore` reload.

  `containerStore` and `imageStore`, AFAICS, don’t strictly need this redesign, but they were changed as well, to keep the design as similar as possible. (Also, reading those stores first might help with reviewing, because they are the simpler cases.)

- Fixes #1429, and one half of #1379 **except for** Btrfs and ZFS graph drivers.

  The original intent maintaining separate mount data, https://github.com/containers/storage/commit/0183a293dc01f22561b83448340c08e55f9b979b , seems to have been to allow mounting / unmounting without exclusive locks; that has not actually materialized, except for `Diff` (i.e. image push and maybe builds). But because the separate mount data is stored in the `Layer` objects, this does not actually allow any more parallelism, because any accesses to `Layer` need to be serialized against reloads of the mount data (at the very least for `MountPoint`, a string field, i.e. containing a pointer and size, which can’t be updated atomically).

  The `Diff` code path remains unresolved: for Btrfs and ZFS, there are concurrent readers and writers of data without any synchronization. I don’t have the time / patience / concentration to fix that right now — the choice seems to be between adding more locks and actually decreasing parallelism if `Diff` is called, or breaking the API to move the data out of `storage.Layer`.

  For other graph drivers, we can actually remove some of the locks without any _actual_ decrease in safety (mostly because the accesses to `MountPoint` and `MountCount` are fundamentally unsafe / out-of-sync anyway); OTOH that exposes that some uses of `MountCount` are not really synchronized, which was resolved by adding an “atomic conditional unmount” operation, and eliminating #1429 in the process.

  The concerns about `MountPoint` and `MountCount` being out-of-sync do affect Podman; some of the code seems to use that data to make decisions. I don’t know whether those decisions are actually safe due to other synchronization, not important because they are only heuristics to protect against accidents, or actual problems.

- Annotates the *`Store` fields, and methods, with the required lock access, with the intent to make that code style a requirement for any future additions. Let’s see if that holds…

mtrmac:
This is now ready to review.

Downstream CI runs:
- https://github.com/containers/buildah/pull/4416
- https://github.com/containers/podman/pull/16508
- https://github.com/cri-o/cri-o/pull/6376

@nalind PTAL, both in general, and at the mount locking code after https://github.com/containers/storage/commit/0183a293dc01f22561b83448340c08e55f9b979b .

@haircommander could you see if there are any performance concerns about this, as discussed around https://github.com/containers/storage/issues/1332#issuecomment-1263840289 , please? CI in GitHub could run based on the PR above; any other manual evaluation would also be appreciated.

haircommander:
> @haircommander could you see if there are any performance concerns about this, as discussed around https://github.com/containers/storage/issues/1332#issuecomment-1263840289 , please? CI in GitHub could run based on the PR above; any other manual evaluation would also be appreciated.

thanks for the heads up! I am working with @wgahnagl to see if we can detect any changes in performance

mtrmac:
This needs to be updated for #1433 , so that pushes can continue to be parallel. (That shouldn’t directly affect CRI-O… but the fix might make things better?)

rhatdan:
Is this still a draft?

mtrmac:
@rhatdan read the last comment

mtrmac:
This now depends on #1436 and #1438 , in order to fix #1433 so that `store.Diff` isn’t completely serialized on the new `inProcessLock` for checks whether to reload layer stores.

Based on #1436, the individual C/I/L stores can now safely check whether the on-disk state has changed with a read lock (instead of the current `loadMut`), so the fast path of `store.Diff` is completely parallel (well, apart from `store.graphLock` serializing the actual mounts, if any).

Similarly, for heavily multi-threaded within a single process (like, possibly, CRI-O), read-heavy workloads can now run completely in parallel, without any serialization on the old `loadMut` locks; only _true_ writes will exclude readers.

Also fixes #1421 , in order to validate that the read-only users are truly parallel without any reloads.

Note to self: Validate this in all consumers, including CRI-O, before making this ready for review.

mtrmac:
Note to self: This is now rebased and stand-alone fore review again, but Buildah unit tests are failing in a way that doesn’t seem _obviously_ related. So I need to debug that.

rhatdan:
@mtrmac needs a rebase.

mtrmac:
Rebased, downstream tests seem to pass now.Rebased, downstream tests seem to pass now.

rhatdan:
Ready for Review?

rhatdan (approved):

mtrmac:
Yes, I think this is ready for review (or possible CRI-O benchmarking).

rhatdan:
@mrunalp @haircommander PTAL

rhatdan:
LGTM
@alexlarsson @giuseppe @vrothberg @nalind @flouthoc @umohnani8 PTAL

containers.go
```diff
@@ -141,16 +141,21 @@ type rwContainerStore interface {
...
-	jsonPath   [numContainerLocationIndex]string
+	// The following fields are only set when constructing containerStore, and must never be modified afterwards.
+	// They are safe to access without any other locking.
+	lockfile *lockfile.LockFile // Synchronizes readers vs. writers of the _filesystem data_, both cross-process and in-process.
+	dir      string
+	jsonPath [numContainerLocationIndex]string
+
+	inProcessLock sync.RWMutex // Can _only_ be obtained with lockfile held.
+	// The following fields can only be read/written with read/write ownership of inProcessLock, respectively.
+	// Almost all users should use startReading() or startWriting().
```
vrothberg, line 152: Absolutely non-blocking and just an idea.  The `Image` struct in libimage has some fields that may be cached.  I think you (@mtrmac) suggested to create an embedded struct to highlight the caching nature of these fields (see https://github.com/containers/common/blob/main/libimage/image.go#L38). Cached fields are always accessed via `(*Image).cached.field` which IMO makes the code much more expressive and comprehensible.

Would it help to add the fields that require locking in a similar embedded struct? `(*containerStore).requiresLock.containers`?

layers.go
```diff
@@ -83,13 +83,26 @@ type Layer struct {
...
-	// Mount() calls, so the caller should consult this value after Mount()
-	// succeeds to find the location of the container's root filesystem.
+	// recently mounted.
+	//
+	// WARNING: This field is a snapshot in time: (except for users inside c/storage that
+	// hold the mount lock) the true value can change between subsequent
+	// calls to c/storage API.
+	//
+	// Users that need to handle concurrent mount/unmount attempts should not access this
+	// field at all, and should only user the path returned by .Mount() (and that’s only
```
vrothberg, line 93: s/only users/only use/

vrothberg (commented):
Impressive work. LGTM

giuseppe (approved):
LGTM

haircommander:
I have tasked @wgahnagl with running some performance tests to see how this works. I can follow up here when we have results

containers.go
mtrmac: Ooh, that did not occur to me at all, and it’s definitely worth doing.

wgahnagl:
> I have tasked @wgahnagl with running some performance tests to see how this works. I can follow up here when we have results
I've run the kubeBurner high-density stresstest against this PR, and it's maxing out at 500 pods, which is normal! :+1:

mtrmac:
@wgahnagl Thank you so much.

rhatdan:
@mtrmac Could you fix the comments and then we can merge and release a new containers/storage.

rhatdan:
@mtrmac Could you fix the comments and then we can merge and release a new containers/storage.

containers.go
mtrmac: I have prototyped this, and it does seem valuable — but it’s a lot of churn.

E.g. on `containers.go`, the current PR is +80-14 non-whitespace non-comment changes (including some changes with comments), and moving the fields around would be about +51-45 . Also, in addition to C/I/L stores, the top-level `store` would benefit from the same updates, but `store.go` is mostly out of scope of this PR.

Overall, I’d prefer to do it in a separate PR a bit later: it would help maintenance but it should not have immediately user-visible impact, which makes it harder for me to prioritize right now — and the work so far churn has already been disrupting Nalin’s work, so coordinating a bit better seems warranted.

I have added the suggestion to #1389; it might well be worth tracking as a higher-priority issue separate from the somewhat desperate #1389.

layers.go
mtrmac: Fixed, thanks!

containers.go
vrothberg: Yes, let's do that in a separate PR. Thank you!

vrothberg (commented):
LGTM

rhatdan:
LGTM

PR #1347, opened 2022-09-15
tests/tools/Makefile: we don't use ffjson any more

nalind:
We don't use ffjson any more, so don't try to build it.


rhatdan:
LGTM

PR #1348, opened 2022-09-15
tests/idmaps.bats: only test .imagestore on overlay and vfs

nalind:
Don't try to test with a .imagestore option on btrfs, devicemapper, or zfs.


rhatdan:
LGTM
@giuseppe PTAL

giuseppe (approved):
LGTM

PR #1349, opened 2022-09-15
Update to go-zfs v3.0.0

nalind:
go-zfs migrated to Go modules, so we update our import path for it.


rhatdan:
LGTM
@giuseppe @flouthoc @saschagrunert @vrothberg @umohnani8 PTAL

PR #1353, opened 2022-09-17
Fix LUtimesNano on FreeBSD

dfr:
The code was passing timespec arguments to lutimes which expects timeval arguments. This likely only ever 'worked' because the nanoseconds field is nearly always zero.

### DIFF
--- a/pkg/system/meminfo_unix_test.go
+++ b/pkg/system/meminfo_unix_test.go
@@ -1,4 +1,5 @@
-// +build linux freebsd
+//go:build linux
+// +build linux
 
 package system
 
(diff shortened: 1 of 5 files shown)


dfr:
Note: this makes the Linux and FreeBSD versions of LUtimesNano identical and perhaps it would be better to rename utimes_linux.go to utimes_common.go here?

dfr:
Alternatively, the version in docker/docker/pkg/system avoids the use of unix.Syscall6, replacing with unix.UtimesNanoAt which might be even better.

dfr:
I added a couple of extra commits which enable go unit tests for pkg/system.

rhatdan (approved):
LGTM

rhatdan:
@flouthoc @giuseppe @nalind @vrothberg @umohnani8 @ashley-cui PTAL

flouthoc (approved):
LGTM

PR #1355, opened 2022-09-21
pkg/system: Allow EnsureRemoveAll to delete trees with immutable files

### LINKED ISSUE #1354: Deleting images or containers with immutable files fails on FreeBSD
On normal FreeBSD installations, a few security related files (e.g. /usr/bin/su) are installed as immutable. This causes problems for the vfs storage driver when it deletes file trees containing these files. Running this script shows the problem:
```
#! /bin/sh
graphdir=$(mktemp -d)
l=$(sudo ./containers-storage -g=$graphdir -s=vfs create-layer)
i=$(sudo ./containers-storage -g=$graphdir -s=vfs create-image $l)
c=$(sudo ./containers-storage -g=$graphdir -s=vfs create-container $i)
m=$(sudo ./containers-storage -g=$graphdir -s=vfs mount $c)
sudo touch $m/foobar
sudo chflags schg $m/foobar
sudo ./containers-storage -g=$graphdir -s=vfs unmount $c
sudo ./containers-storage -g=$graphdir -s=vfs delete $c
sudo chflags -R 0 $graphdir
sudo rm -rf $graphdir
```
This generates an error something like this:
```
3c8c85cc2f58555d62ec435008d96c4e09aefd5f7cc1027151942024a0f711bb: unlinkat /tmp/tmp.9oimuB3U/vfs/dir/4cdf5c2b9b32a71b84ec003dfc1df4aaf2a121013524550766b842b6dafd2122/foobar: operation not permitted
```
The immutable flag can be removed by root unless the system is running with a raised securelevel so it is possible to delete these files by resetting the flags with chflags.

dfr:
Fixes #1354.

### DIFF
--- a/pkg/system/rm.go
+++ b/pkg/system/rm.go
@@ -35,6 +35,9 @@ func EnsureRemoveAll(dir string) error {
 	}
 
 	for {
+		if err := resetFileFlags(dir); err != nil {
+			return fmt.Errorf("resetting file flags: %w", err)
+		}
 		err := os.RemoveAll(dir)
 		if err == nil {
 			return nil
--- a/pkg/system/rm_common.go
+++ b/pkg/system/rm_common.go
@@ -1,0 +1,10 @@
+//go:build !freebsd
+// +build !freebsd
+
+package system
+
+// Reset file flags in a directory tree. This allows EnsureRemoveAll
+// to delete trees which have the immutable flag set.
+func resetFileFlags(dir string) error {
+	return nil
+}
--- a/pkg/system/rm_freebsd.go
+++ b/pkg/system/rm_freebsd.go
@@ -1,0 +1,32 @@
+package system
+
+import (
+	"io/fs"
+	"path/filepath"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+func lchflags(path string, flags int) (err error) {
+	p, err := unix.BytePtrFromString(path)
+	if err != nil {
+		return err
+	}
+	_, _, e1 := unix.Syscall(unix.SYS_LCHFLAGS, uintptr(unsafe.Pointer(p)), uintptr(flags), 0)
+	if e1 != 0 {
+		return e1
+	}
+	return nil
+}
+
+// Reset file flags in a directory tree. This allows EnsureRemoveAll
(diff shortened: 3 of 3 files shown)


rhatdan:
LGTM
@nalind @vrothberg @giuseppe @flouthoc @saschagrunert PTAL

nalind:
There's probably a corresponding function on Linux that uses the FS_IOC_SETFLAGS ioctl to clear immutable flags, but we don't tend to run into that in routine use, so we can leave that for another time.
LGTM.

rhatdan:
```
int attr;
fd = open("pathname", ...);

ioctl(fd, FS_IOC_GETFLAGS, &attr);  /* Read current flags */
attr |= FS_NOATIME_FL;              /* Add desired flag */
ioctl(fd, FS_IOC_SETFLAGS, &attr);  /* Set updated flags */

The interesting part is the flags themselves. There are some that help optimize file-operations, others that manipulate the file operational mode and yes, there are always some exotic flags that even after reading their description twice I still don’t know how it would actually affect the system.

But, inside this haystack I found two shiny needles that really caught my eye and got my heart rate up, and yeah, I know this sounds nerdy :)

    FS_APPEND_FL: Mark a file, so that it can be opened only with the O_APPEND flag.
    FS_IMMUTABLE_FL: The file is immutable: no changes are permitted to the file contents or metadata (permissions, timestamps, ownership, link count and so on).
```

rhatdan:
https://github.com/snapcore/snapd/blob/master/osutil/chattr.go

rhatdan:
One issue with this is that it will be slow, since it needs to walk the trees twice.

Would it be better to RemoveAll then if it fails attempt to  change the immutable and try to removeall again.

dfr:
I think immutable and append flags are generally useful. BSDs have used immutable, combined with securelevel, to protect suid utilities like login and su for a very long time. The point about efficiency is a good one - I used WalkDir which should avoid excessive stat calls but still, it would be better to check for an EPERM error from RemoveAll first.

I was working on a separate PR which will preserve the BSD flags in pkg/archive but got slightly distracted by fixing tests on FreeBSD. I will add a commit to that to avoid the double tree walk in simple cases.

PR #1357, opened 2022-09-22
If XDG_CONFIG_HOME env is set we should use it and ignore defaults

rhatdan:
HPC Customers noticed that storage was attempting to read files in /usr and /etc, even though they set XDG_CONFIG_HOME, they expect to only read config files in this directory.

Fixes: https://github.com/containers/podman/issues/15680

(Actually partial fixes), need to look at other config files.

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -48,6 +48,12 @@ func loadDefaultStoreOptions() {
 			loadDefaultStoreOptionsErr = err
 			return
 		}
+	} else if path, ok := os.LookupEnv("XDG_CONFIG_HOME"); ok {
+		defaultOverrideConfigFile = filepath.Join(path, "containers", "storage.conf")
+		if err := ReloadConfigurationFileIfNeeded(defaultOverrideConfigFile, &defaultStoreOptions); err != nil {
+			loadDefaultStoreOptionsErr = err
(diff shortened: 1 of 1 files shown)


rhatdan:
@vrothberg PTAL

rhatdan:
@giuseppe @nalind @vrothberg @saschagrunert @TomSweeneyRedHat @flouthoc PTAL

types/options.go
```diff
@@ -48,6 +48,12 @@ func loadDefaultStoreOptions() {
 			loadDefaultStoreOptionsErr = err
 			return
 		}
+	} else if path, ok := os.LookupEnv("XDG_CONFIG_HOME"); ok {
+		defaultOverrideConfigFile = filepath.Join(path, "containers/storage.conf")
```
nalind, line 52: Pass the "containers" and "storage.conf" path elements to `Join()` as two arguments, not one.

rhatdan: Done

giuseppe (approved):
LGTM

vrothberg:
Restarted the failed job.

vrothberg:
@rhatdan feel free to merge or rebase

PR #1358, opened 2022-09-23
tests/idmaps.bats: only run idmaps tests on Linux

dfr:
All the other integration tests pass on FreeBSD with the zfs driver. Some of them fail with the vfs driver but those also fail on Linux.


rhatdan:
LGTM
@vrothberg @giuseppe @nalind @TomSweeneyRedHat @flouthoc @saschagrunert PTAL

giuseppe (approved):
LGTM

PR #1360, opened 2022-09-25
Make the unit tests work for FreeBSD

dfr:
This exposed a couple of FreeBSD bugs in pkg/archive which are also fixed here.

### DIFF
--- a/pkg/archive/archive_unix.go
+++ b/pkg/archive/archive_unix.go
@@ -119,3 +119,13 @@ func handleLChmod(hdr *tar.Header, path string, hdrInfo os.FileInfo, forceMask *
 	}
 	return nil
 }
+
+// Hardlink without symlinks
+func handleLLink(targetPath, path string) error {
+	// Note: on Linux, the link syscall will not follow symlinks.
+	// This behavior is implementation-dependent since
+	// POSIX.1-2008 so to make it clear that we need non-symlink
+	// following here we use the linkat syscall which has a flags
+	// field to select symlink following or not.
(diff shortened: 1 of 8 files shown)


rhatdan:
LGTM

rhatdan (approved):

pkg/archive/archive_unix.go
```diff
@@ -118,3 +119,8 @@ func handleLChmod(hdr *tar.Header, path string, hdrInfo os.FileInfo, forceMask *
 	}
 	return nil
 }
+
+// Hardlink without symlinks
+func handleLLink(targetPath, path string) error {
+	return unix.Linkat(unix.AT_FDCWD, targetPath, unix.AT_FDCWD, path, 0)
```
flouthoc, line 125: Is there any reason for dropping `if err := os.Link(targetPath, path); err != nil {` for `unix` variant and directly using `Linkat(`.

dfr: I don't think this is strictly needed. The confusing thing here when working on the tests was that the code implicitly assumed that hard links to symlinks do not follow the symlink and changing to Linkat perhaps makes that slightly clearer. It could be improved with a comment.

Looking at archive_unix.go and archive_freebsd.go, they are quite similar and it would be worth trying to share the implementation and reduce archive_freebsd.go to a minimum. Probably in a separate PR.

dfr: I added a comment to clarify a little.

flouthoc (approved):
LGTM @giuseppe @nalind @rhatdan PTAL

giuseppe (approved):
LGTM

PR #1367, opened 2022-09-27
drivers/driver_linux: Show fstype in GetFSMagic() if debug logging is enabled

aarontomlin:
In the context of github.com/containers/storage/drivers.GetFSMagic() when log-level debug is enabled show the filesystem type/or ID that is returned by the statfs(2) system call in particular when the ID is not known to podman. This might suggest an unsupported configuration.

Signed-off-by: Aaron Tomlin <atomlin@redhat.com>

### DIFF
--- a/drivers/driver_linux.go
+++ b/drivers/driver_linux.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 
 	"github.com/containers/storage/pkg/mount"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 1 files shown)


drivers/driver_linux.go
```diff
@@ -130,6 +131,10 @@ func GetFSMagic(rootpath string) (FsMagic, error) {
 	if err := unix.Statfs(filepath.Dir(rootpath), &buf); err != nil {
 		return 0, err
 	}
+
+	if _, ok := FsNames[FsMagic(buf.Type)]; ok != true {
```
nalind, line 135: It's more idiomatic to use `!ok` here.  Please include the pathname of the directory we were looking at in the debug message.

aarontomlin: Understood and will do.

nalind:
LGTM

flouthoc (approved):
LGTM

PR #1369, opened 2022-09-27
Fix accessing containers and image locks

rhatdan:
Match containers and image lock to the new layer lock method. in https://github.com/containers/storage/pull/1351

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -255,8 +255,6 @@ func newContainerStore(dir string) (ContainerStore, error) {
 	if err != nil {
 		return nil, err
 	}
-	lockfile.Lock()
-	defer lockfile.Unlock()
 	cstore := containerStore{
 		lockfile:   lockfile,
 		dir:        dir,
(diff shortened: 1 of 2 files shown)


rhatdan:
@mtrmac PTAL

containers.go
```diff
@@ -265,6 +263,8 @@ func newContainerStore(dir string) (ContainerStore, error) {
 		bylayer:    make(map[string]*Container),
 		byname:     make(map[string]*Container),
 	}
+	cstore.Lock()
+	defer cstore.Unlock()
 	if err := cstore.Load(); err != nil {
```
vrothberg, line 268: It's somehow surprising that callers of a public API are responsible for locking.

rhatdan:
@vrothberg I think @mtrmac has a PR to make all of these functions Private to containers/storage.

nalind:
LGTM

containers.go
mtrmac: The `{container,image,layer}Store` objects implement public interfaces, but AFAICS there is no public way to access them (one would have to take a `storage.Store` and manually cast it to an interface that contains accessors to the three stores).

Compare #1345 .

I think the basic idea is that various `storage.Store` operations need to coordinate _multiple_ operations on a single `layerStore`, so they need a consistent view, and a lock held across the `layerStore` calls.

See #1346 which tries to make the locking requirements much more explicit to make sure there are no mistakes there.

mtrmac:
@rhatdan thanks!

PR #1371, opened 2022-09-29
pkg/archive: Use archive_unix.go on FreeBSD

dfr:
Almost all the code can be shared between Linux, FreeBSD and Darwin with the only different in the handling of lchmod. On FreeBSD and Darwin (and probably other BSDs) we can access lchmod functionality via unix.Fchmodat.


rhatdan:
LGTM
@nalind @flouthoc @saschagrunert @giuseppe @vrothberg PTAL

giuseppe (approved):
LGTM

PR #1374, opened 2022-09-29
drivers/overlay: Notify the user if the backing filesystem is not supported with OverlayFS

aarontomlin:
If the backing filesystem is not supported with OverlayFS as a container storage driver/or graph driver then do not continue. In this context, report the incompatibility by notifying the user of an unsupported configuration. The filesystem type/or ID that is returned by the statfs(2) system call is reported too, to assist further troubleshooting efforts. It can be used to deduce the actual filesystem used for the container storage graph directory.

Signed-off-by: Aaron Tomlin <atomlin@redhat.com>

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -309,9 +309,11 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 	if err != nil {
 		return nil, err
 	}
-	if fsName, ok := graphdriver.FsNames[fsMagic]; ok {
-		backingFs = fsName
+	fsName, ok := graphdriver.FsNames[fsMagic]
+	if !ok {
+		return nil, fmt.Errorf("filesystem type %#x reported for %s is not supported with 'overlay': %w", fsMagic, filepath.Dir(home), graphdriver.ErrIncompatibleFS)
 	}
+	backingFs = fsName
 
 	runhome := filepath.Join(options.RunRoot, filepath.Base(home))
 	rootUID, rootGID, err := idtools.GetRootUIDGID(options.UIDMaps, options.GIDMaps)


giuseppe (commented):
good idea, could you fix the issue reported by the linter?

aarontomlin:
> good idea, could you fix the issue reported by the linter?

Done.

drivers/overlay/overlay.go
```diff
@@ -309,7 +310,9 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
 	if err != nil {
 		return nil, err
 	}
-	if fsName, ok := graphdriver.FsNames[fsMagic]; ok {
+	if fsName, ok := graphdriver.FsNames[fsMagic]; !ok {
+		return nil, errors.Wrapf(graphdriver.ErrIncompatibleFS, "filesystem type %#x reported for %s is not supported with 'overlay'", fsMagic, filepath.Dir(home))
+	} else {
```
rhatdan, line 315: Just remove the else { } block

and only have
backingFs=fsName

Should fix the lint complaint.

drivers/overlay/overlay.go
```diff
@@ -309,7 +310,9 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
 	if err != nil {
 		return nil, err
 	}
-	if fsName, ok := graphdriver.FsNames[fsMagic]; ok {
+	if fsName, ok := graphdriver.FsNames[fsMagic]; !ok {
+		return nil, errors.Wrapf(graphdriver.ErrIncompatibleFS, "filesystem type %#x reported for %s is not supported with 'overlay'", fsMagic, filepath.Dir(home))
```
rhatdan, line 314: Could you also use
fmt.Erorrf("filesystem type %#x reported for %s is not supported with 'overlay': %w", fsMagic, filepath.Dir(home), graphdriver.ErrIncompatibleFS)

aarontomlin: Done.

rhatdan:
LGTM
Thanks @aarontomlin

drivers/overlay/overlay.go
```diff
@@ -309,9 +309,10 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
 	if err != nil {
 		return nil, err
 	}
-	if fsName, ok := graphdriver.FsNames[fsMagic]; ok {
-		backingFs = fsName
+	if fsName, ok := graphdriver.FsNames[fsMagic]; !ok {
+		return nil, fmt.Errorf("filesystem type %#x reported for %s is not supported with 'overlay': %w", fsMagic, filepath.Dir(home), graphdriver.ErrIncompatibleFS)
 	}
+	backingFs = fsName
```
flouthoc, line 315: Please move this to an `else` block, since `fsName` does not exists outside of the scope.

aarontomlin: I agree. That was a mistake.

giuseppe:
> good idea, could you fix the issue reported by the linter?

so I disagree with my assessment from some months ago.  I think it is better if we revert, or do not make it a failure as it is hiding valid use cases: https://github.com/containers/storage/issues/1546

tis-rpage:
> > good idea, could you fix the issue reported by the linter?
> 
> so I disagree with my assessment from some months ago. I think it is better if we revert, or do not make it a failure as it is hiding valid use cases: #1546

I believe the failure condition is reasonable, however I recommend adding a flag at a minimum to ignore the check and let users own the tool failing by using some `--i-acknowledge-overlayfs-support-is-good-but-i-dont-care`

giuseppe:
no there is no need to fail just because we don't know the file system.  At most, we can fail/raise warning if we know something won't work correctly

PR #1375, opened 2022-09-29
overlay: try harder with more layers

nalind:
If we have enough layers that even using relative paths for symlinks to the diff directories won't make the list of lower directories and options fit in a single page,
* ~create a temporary directory~
* ~mount a tmpfs there~
* ~bind mount from the diff directories to subdirectories of the temp directory with short names~
* ~mount using those bind-mounted directories as the lower directories~
* ~detach the temporary directory~
* ~remove the temporary directory~
* open the various lowers for reading
* use the descriptor names in the `lowerdir=` list

This is ~noticeably~ slower, so don't take these extra steps unless we know we absolutely have to.

Should help with https://github.com/cri-o/cri-o/issues/6261.

### DIFF
--- a/drivers/overlay/mount.go
+++ b/drivers/overlay/mount.go
@@ -8,14 +8,16 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/containers/storage/pkg/reexec"
 	"golang.org/x/sys/unix"
 )
 
 func init() {
-	reexec.Register("storage-mountfrom", mountFromMain)
+	reexec.Register("storage-mountfrom", mountOverlayFromMain)
 }
 
 func fatal(err error) {
@@ -31,7 +33,7 @@ type mountOptions struct {
 	Flag   uint32
 }
 
-func mountFrom(dir, device, target, mType string, flags uintptr, label string) error {
+func mountOverlayFrom(dir, device, target, mType string, flags uintptr, label string) error {
 	options := &mountOptions{
 		Device: device,
 		Target: target,
@@ -67,7 +69,7 @@ func mountFrom(dir, device, target, mType string, flags uintptr, label string) e
 }
 
 // mountfromMain is the entry-point for storage-mountfrom on re-exec.
-func mountFromMain() {
+func mountOverlayFromMain() {
 	runtime.LockOSThread()
 	flag.Parse()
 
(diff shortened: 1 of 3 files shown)


drivers/overlay/mount.go
```diff
@@ -8,21 +8,33 @@ import (
...
-	reexec.Register("storage-mountfrom", mountFromMain)
+	reexec.Register("storage-mountfrom", mountOverlayFromMain)
 }
 
 func fatal(err error) {
 	fmt.Fprint(os.Stderr, err)
 	os.Exit(1)
 }
 
+func bindMount(source, destination, data string) (string, error) {
```
flouthoc, line 28: `pkg/mount` already has generic mount function, could we reuse it ?

drivers/overlay/mount.go
```diff
@@ -77,13 +89,147 @@ func mountFromMain() {
...
 	}
 
-	if err := unix.Mount(options.Device, options.Target, options.Type, uintptr(options.Flag), options.Label); err != nil {
+	// Split out the various options, since we need to manipulate the
+	// paths, but we don't want to mess with other options.
+	lower := ""
+	upper := ""
+	work := ""
+	label := ""
+	other := ""
```
flouthoc, line 124: Nit: non-blocker
```suggestion
	// paths, but we don't want to mess with other options.
	var lower, upper, work, label, other string
```

drivers/overlay/mount.go
```diff
@@ -77,13 +89,147 @@ func mountFromMain() {
...
+				if err := unix.Unmount(tmpdir, unix.MNT_DETACH); err != nil {
+					fmt.Fprint(os.Stderr, err)
+				}
+				if err := os.Remove(tmpdir); err != nil {
+					fmt.Fprint(os.Stderr, err)
+				}
+				fatal(err)
+			}
+		}
+		lower = "lowerdir=" + strings.Join(lowers, ":")
```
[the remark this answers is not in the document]
nalind: I didn't expect it to work with fuse-overlayfs, but it seems to.  Switching over.

drivers/overlay/mount.go
nalind: Dropping the function.

drivers/overlay/mount.go
nalind: Picking this up.

nalind:
This is making me consider if we'd just be better off recording layer IDs in "${layer}/lower" in "${layer}/diff" form and dispensing with the symbolic link farm in "l" entirely.  We'd fork/exec for mounting a lot more often.

drivers/overlay/overlay.go
```diff
@@ -73,21 +73,23 @@ const (
...
 	// which can be used to create the unique link identifier
 	// for every layer. If this value is too long then the
 	// page size limit for the mount command may be exceeded.
 	// The idLength should be selected such that following equation
-	// is true (512 is a buffer for label metadata).
-	// ((idLength + len(linkDir) + 1) * maxDepth) <= (pageSize - 512)
+	// is true (512 is a buffer for label metadata, 128 is the
+	// number of lowers we want to be able to use without having
+	// to use bind mounts to get all the way to the kernel limit).
+	// ((idLength + len(linkDir) + 1) * maxFastDepth) <= (pageSize - 512)
```
mtrmac, line 92: `maxFastDepth` does not exist AFAICS.

drivers/overlay/mount.go
```diff
@@ -77,12 +79,111 @@ func mountFromMain() {
...
+
+	// Those arguments still took up too much space.  Open the diff
+	// directories and use their descriptor numbers as lowers, using
+	// /proc/self/fd as the current directory.
+
+	// Split out the various options, since we need to manipulate the
+	// paths, but we don't want to mess with other options.
+	var upper, work, lower, label, other string
+	for _, arg := range strings.Split(options.Label, ",") {
+		if strings.HasPrefix(arg, "upperdir=") {
```
mtrmac, line 106: Style only:

```go
nv := strings.SplitN(arg, "=")
switch nv[0] {
case "upperdir":
    upper = nv[1]
…
}
```
(and eventually `strings.Cut` with Go 1.18.)

---

Separately: The caller has all of this data. It feels a bit tedious to format it into a single string, and then parse it out again — but changing that would be a rather larger rewrite.

drivers/overlay/mount.go
```diff
@@ -77,12 +79,111 @@ func mountFromMain() {
...
+			upper = strings.TrimPrefix(arg, "upperdir=")
+		} else if strings.HasPrefix(arg, "workdir=") {
+			work = strings.TrimPrefix(arg, "workdir=")
+		} else if strings.HasPrefix(arg, "lowerdir=") {
+			lower = strings.TrimPrefix(arg, "lowerdir=")
+		} else if strings.HasPrefix(arg, "label=") {
+			label = arg
```
mtrmac, line 113: It’s a bit confusing that some variables contain just the data, and some the data + the option — and that for some, it _changes_ during the lifetime of this function.

Maybe … [A]

drivers/overlay/mount.go
```diff
@@ -77,12 +79,111 @@ func mountFromMain() {
...
+				fatal(err)
+			}
+			lowers[i] = fmt.Sprintf("%d", lowerFd)
+		}
+		lower = "lowerdir=" + strings.Join(lowers, ":")
+	}
+
+	// Reconstruct the Label field.
+	options.Label = ""
+	for _, arg := range []string{upper, work, lower, label, other} {
```
mtrmac, line 161: [A] …

something like
```go
for _, pair := range []struct{optionName, value string} {
   { "upperdir=", upper },
   …
   { "", other },
}
```
would allow most of this function to only deal with the individual values instead of the mount syntax.

---

(Ultimately, this “turn option values into a mount option syntax string” code might be useful to completely factor out…)

drivers/overlay/mount.go
```diff
@@ -77,12 +79,111 @@ func mountFromMain() {
...
+		}
+		err = unix.Mount(options.Device, options.Target, options.Type, uintptr(options.Flag), options.Label)
+	} else {
+		err = fmt.Errorf("cannot mount layer, mount data %q too large %d >= page size %d", options.Label, len(options.Label), pageSize)
+	}
+
+	// Clean up.
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "creating overlay mount to %s, mount_data=%q\n", options.Target, options.Label)
+		panic(err)
```
mtrmac, line 186: Shouldn’t this be `fatal(err)`?

mtrmac (commented):
The idea and overall implementation LGTM — mostly mild style suggestions.

(Parenthetically, I’m a bit unhappy with `overlay.Driver.get` growing even more, to 329 lines now. I suppose splitting that up would be a separate PR…)

drivers/overlay/mount.go
nalind: Yes, it should.  Fixing.

drivers/overlay/overlay.go
nalind: Oof, forgot to change that back to refer to `maxDepth` after some experimenting.  Fixing that.

rhatdan:
@mtrmac PTAL again.

giuseppe (approved):
LGTM

drivers/overlay/overlay.go
```diff
@@ -73,20 +73,22 @@ const (
...
 	// idLength represents the number of random characters
 	// which can be used to create the unique link identifier
 	// for every layer. If this value is too long then the
 	// page size limit for the mount command may be exceeded.
 	// The idLength should be selected such that following equation
-	// is true (512 is a buffer for label metadata).
+	// is true (512 is a buffer for label metadata, 128 is the
+	// number of lowers we want to be able to use without having
+	// to use bind mounts to get all the way to the kernel limit).
 	// ((idLength + len(linkDir) + 1) * maxDepth) <= (pageSize - 512)
```
mtrmac, line 92: Now the inequality does not hold. `s/maxDepth/128/` or maybe … `maxDirect = 128 is the … ` and `* maxDirect`?

drivers/overlay/mount.go
```diff
@@ -77,11 +79,96 @@ func mountFromMain() {
...
+				fatal(err)
+			}
+			lowers[i] = fmt.Sprintf("%d", lowerFd)
+		}
+		lowerv = strings.Join(lowers, ":")
+	}
+
+	// Reconstruct the Label field.
+	options.Label = upperk + upperv + "," + workk + workv + "," + lowerk + lowerv + "," + labelk + labelv + "," + others
+	options.Label = strings.ReplaceAll(options.Label, ",,", ",")
```
mtrmac, line 156: OK I didn’t expect that … but it’s quite elegant.

mtrmac (commented):
Code LGTM.

drivers/overlay/overlay.go
nalind: Opened https://github.com/containers/storage/pull/1392 to do the first one.

PR #1376, opened 2022-09-30
Remove lockfile.Locker.RecursiveLock (alternative to #1344)

mtrmac:
It is not valid as currently implemented, and there is no known user.

See #1344 for more discussion.


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (commented):
LGTM

PR #1364, opened 2022-09-26
Another bunch of warning cleanups

mtrmac:
… slowly grinding away on `golangci-lint` outputs with various settings.

_Mostly_ just cleanups, but includes a few real bug fixes — _and_ a change to the `golangci-lint` configuration.

Note the `TarWithOptions` and `ImagesByTopLayer` changes, at least.

Some recurring themes:
- More error checking, primarily in tests. (I have deferred the harder work of error checking in production)
- Silence warnings about unused code on macOS.

Eventually we should update to a recent Go, run `gofmt`, and re-enable all default linters of `golangci-lint`, some of which are currently disabled.

PR #1377, opened 2022-10-01
Pen-ultimate? bunch of warning cleanups

mtrmac:
Continuing to grind away on this…

The primary theme is error handling, in `defer`s and on error recovery paths — that may introduce extra user-visible noise via `logrus`.

Also includes various cleanups even from linters that probably don’t make sense to enable in enforcing mode, where it seemed reasonable.


rhatdan:
@nalind @giuseppe @saschagrunert @vrothberg @umohnani8 @ashley-cui PTAL

store.go
```diff
@@ -1444,7 +1445,13 @@ func (s *store) CreateContainer(id string, names []string, image, layer, metadat
 	}
 	container, err := rcstore.Create(id, names, imageID, layer, metadata, options)
 	if err != nil || container == nil {
-		rlstore.Delete(layer)
+		if err2 := rlstore.Delete(layer); err2 != nil {
+			if err == nil {
+				err = fmt.Errorf("deleting layer %#v: %w", layer, err2)
```
flouthoc, line 1450: In past changing error message or introducing new error have regressed `podman` or `buildah` tests. Should we open a dummy PR in podman/buildah to verify these ?

flouthoc (commented):
PR looks good, but a small concern about change of error messages or new error messages w.r.t to `podman` and `buildah`.

store.go
mtrmac: I wouldn’t expect Podman tests to trigger _these_ error recovery paths, but, sure, it’s easy enough to create the PRs, and this set of changes is tedious to review, so the extra confidence is definitely worth it.

https://github.com/containers/buildah/pull/4311
https://github.com/containers/podman/pull/16024

cover both this PR and #1364 .

mtrmac: Both succeeded.

rhatdan:
LGTM

PR #1378, opened 2022-10-01
Add support for exporting and importing file flags on FreeBSD

dfr:
The implementation could be shared with darwin and other BSDs if needed but initially is only enabled on FreeBSD. This uses the SCHILY.fflags PAX option to encode flags which is supported by libarchive (bsdtar) and star. The flag save/restore entry points are exported, allowing them to be used by buildah/copier.

To support applying diffs to trees with flags, I added logic to reset immutable flags during the UnpackLayer process - this is slightly complicated by the desire to support immutable directories which needs to defer setting fiags on directories to after all modifications to the directory contents. Fortunately, something similar is already in place for setting directory modify times.

### DIFF
--- a/pkg/archive/fflags_bsd.go
+++ b/pkg/archive/fflags_bsd.go
@@ -1,0 +1,167 @@
+//go:build freebsd
+// +build freebsd
+
+package archive
+
+import (
+	"archive/tar"
+	"fmt"
+	"math/bits"
+	"os"
+	"strings"
+	"syscall"
+
+	"github.com/containers/storage/pkg/system"
+)
+
+const (
+	paxSCHILYFflags = "SCHILY.fflags"
+)
+
+var (
+	flagNameToValue = map[string]uint32{
+		"sappnd":     system.SF_APPEND,
+		"sappend":    system.SF_APPEND,
+		"arch":       system.SF_ARCHIVED,
+		"archived":   system.SF_ARCHIVED,
+		"schg":       system.SF_IMMUTABLE,
+		"schange":    system.SF_IMMUTABLE,
+		"simmutable": system.SF_IMMUTABLE,
+		"sunlnk":     system.SF_NOUNLINK,
+		"sunlink":    system.SF_NOUNLINK,
+		"snapshot":   system.SF_SNAPSHOT,
+		"uappnd":     system.UF_APPEND,
+		"uappend":    system.UF_APPEND,
+		"uarch":      system.UF_ARCHIVE,
+		"uarchive":   system.UF_ARCHIVE,
+		"hidden":     system.UF_HIDDEN,
+		"uhidden":    system.UF_HIDDEN,
+		"uchg":       system.UF_IMMUTABLE,
+		"uchange":    system.UF_IMMUTABLE,
+		"uimmutable": system.UF_IMMUTABLE,
+		"uunlnk":     system.UF_NOUNLINK,
+		"uunlink":    system.UF_NOUNLINK,
+		"offline":    system.UF_OFFLINE,
+		"uoffline":   system.UF_OFFLINE,
+		"opaque":     system.UF_OPAQUE,
+		"rdonly":     system.UF_READONLY,
+		"urdonly":    system.UF_READONLY,
+		"readonly":   system.UF_READONLY,
+		"ureadonly":  system.UF_READONLY,
+		"reparse":    system.UF_REPARSE,
+		"ureparse":   system.UF_REPARSE,
+		"sparse":     system.UF_SPARSE,
+		"usparse":    system.UF_SPARSE,
+		"system":     system.UF_SYSTEM,
+		"usystem":    system.UF_SYSTEM,
+	}
+	// Only include the short names for the reverse map
+	flagValueToName = map[uint32]string{
+		system.SF_APPEND:    "sappnd",
+		system.SF_ARCHIVED:  "arch",
+		system.SF_IMMUTABLE: "schg",
+		system.SF_NOUNLINK:  "sunlnk",
+		system.SF_SNAPSHOT:  "snapshot",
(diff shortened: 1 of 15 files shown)


pkg/system/lchflags_bsd.go
```diff
@@ -0,0 +1,22 @@
...
+
+package system
+
+import (
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+func Lchflags(path string, flags uint32) error {
```
rhatdan, line 46: Does this need to be exported?

rhatdan (approved):
One question 
LGTM

rhatdan:
@giuseppe @vrothberg @nalind @flouthoc @umohnani8 PTAL

dfr:
Rebased

rhatdan:
@nalind @saschagrunert @giuseppe @vrothberg @umohnani8 PTAL

flouthoc:
There are some commits in `archive` on which I don't have lot of idea, but they will affect buildah and podman afaics on *nix as well. I'd defer review for this to @nalind @giuseppe.

flouthoc:
@dfr commit 8bd87e9f52ef4e5a81e2c1799ccde4f37d8661c4 adds changes to *nix counterpart as well right ? But commit only talks about `FreeBSD`. Could we split commit which affects `*nix` into subcommits ?

Please ignore my concern if i understood this wrong :)

dfr:
@flouthoc that commit could be split into functional parts, e.g.:

- Adding fflags_*.go with entry points to encode/decode flags
- Adding support for detecting flag changes in changes_unix.go and changes_bsd_test.go
- Adding calls to those encode/decode flags in archive.go and diff.go

Would that help to clarify?

flouthoc:
> @flouthoc that commit could be split into functional parts, e.g.:
> 
>     * Adding fflags_*.go with entry points to encode/decode flags
> 
>     * Adding support for detecting flag changes in changes_unix.go and changes_bsd_test.go
> 
>     * Adding calls to those encode/decode flags in archive.go and diff.go
> 
> 
> Would that help to clarify?

@dfr Yes I think that would be really helpful thanks for considering my comment :) .

dfr:
I split the pkg/archive commit along the lines I suggested - hope that helps.

dfr:
Rebased

pkg/system/rm.go
```diff
@@ -34,14 +35,24 @@ func EnsureRemoveAll(dir string) error {
...
-		}
 		err := os.RemoveAll(dir)
 		if err == nil {
 			return nil
 		}
 
+		// If the RemoveAll fails with a permission error, we
+		// may have immutable files so try to remove the
+		// immutable flag and redo the RemoveAll.
+		if errors.Is(err, syscall.EPERM) {
```
flouthoc, line 46: You missed importing `syscall` package.

dfr: Somehow this got lost in the rebase - should be fixed now

rhatdan:
@flouthoc PTAL

rhatdan:
@flouthoc @nalind @saschagrunert @giuseppe @vrothberg @umohnani8 PTAL

dfr:
Rebased

rhatdan:
@flouthoc @nalind @saschagrunert @giuseppe @vrothberg @umohnani8 PTAL

flouthoc:
> There are some commits in `archive` on which and I don't have lot of idea on `archive` part, but they will affect buildah and podman afaics on *nix as well. I'd defer review for this to @nalind @giuseppe.

PR LGTM but I'd like @nalind @giuseppe @mtrmac @vrothberg to take a look for above reason.

pkg/archive/fflags_bsd.go
```diff
@@ -0,0 +1,163 @@
...
+		system.UF_OFFLINE:   "offline",
+		system.UF_OPAQUE:    "opaque",
+		system.UF_READONLY:  "rdonly",
+		system.UF_REPARSE:   "reparse",
+		system.UF_SPARSE:    "sparse",
+		system.UF_SYSTEM:    "system",
+	}
+)
+
+func parseFileFlags(fflags string, set, clear *uint32) error {
```
mtrmac, line 75: Non-blocking:
```suggestion
func parseFileFlags(fflags string) (uint32, uint32, error) {
```
would be a bit more idiomatic in Go.

pkg/archive/fflags_bsd.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	fflags, err := formatFileFlags(st.Flags())
+	if err != nil {
+		return err
+	}
+	if fflags != "" {
+		if hdr.PAXRecords == nil {
+			hdr.PAXRecords = map[string]string{}
+		}
+		hdr.PAXRecords["SCHILY.fflags"] = fflags
```
mtrmac, line 125: Non-blocking: Maybe turn `SCHILY.fflags` into a top level `bsdFlagsPAXRecordsName` constant?

mtrmac (commented):
LGTM — I’m really not that familiar with `pkg/archive`, but at least the added tests add quite a bit of confidence.

Just a few optional stylistic nits.

rhatdan:
Could you fix the two issues @mtrmac found.

pkg/archive/fflags_bsd.go
dfr: In bsdtar this option is used to encode file flags on Linux as well so I think I will follow the convention used in the golang archive/tar package and call the constant paxSHILYFflags.

dfr:
> Could you fix the two issues @mtrmac found.

Done, thanks!

pkg/archive/fflags_bsd.go
mtrmac: My fault, I didn’t look for any existing constants.

PR #1383, opened 2022-10-13
Consolidate the common parts of simple writers

mtrmac:
Factor out the repetitive “find and lock C/I/L store” parts out of the top-level `storage.Store` API into helpers: `writeTo{Container,Image,Layer}Store` and `writeToAllStores`.

Overall, this is just a refactoring that should not change behavior. The goal is to decrease the number of individual call sites to the locking code — so that the code refactored here does not need to be touched again for locking changes.

The more complex writers that don’t fit the simple locking schemes of these helpers remain unchanged; the locking in them will be updated individually.

See individual commit messages for (a bit) more detail. It might be more convenient to review this in a view that ignores whitespace changes.

### DIFF
--- a/store.go
+++ b/store.go
@@ -1003,6 +1003,23 @@ func (s *store) readAllLayerStores(fn func(store roLayerStore) (bool, error)) (b
 	return false, nil
 }
 
+// writeToLayerStore is a helper for working with store.getLayerStore():
+// It locks the store for writing, checks for updates, and calls fn()
+// It returns the return value of fn, or its own error initializing the store.
+func (s *store) writeToLayerStore(fn func(store rwLayerStore) error) error {
+	store, err := s.getLayerStore()
+	if err != nil {
+		return err
+	}
+
+	store.Lock()
+	defer store.Unlock()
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@saschagrunert @nalind @vrothberg @giuseppe @umohnani8 @flouthoc PTAL

mtrmac:
#1384 is a counterpart for the more complex read paths.

store.go
```diff
@@ -1003,6 +1003,23 @@ func (s *store) readAllLayerStores(fn func(store roLayerStore) (bool, error)) (b
 	return false, nil
 }
 
+// writeToLayerStore is a helper for working with store.getLayerStore():
+// It locks the store for writing, checks for updates, and calls fn()
+// It returns the return value of fn, or its own error initializing the store.
+func (s *store) writeToLayerStore(fn func(store rwLayerStore) error) error {
```
vrothberg, line 1009: Very elegant, I like it!

vrothberg (approved):
LGTM

PR #1384, opened 2022-10-13
Factor out the infrastructure for reading primary + additional stores

mtrmac:
Factor out the repetitive “find primary + additional stores, and lock them in order” parts out of the top-level storage.Store API into helpers: `readAll{Layer,Image}Stores`.

Overall, this is just a refactoring that should not change behavior. The goal is to decrease the number of individual call sites to the locking code — so that the code refactored here does not need to be touched again for locking changes.

The more complex readers that don’t fit the simple locking schemes of these helpers remain unchanged; the locking in them will be updated individually.

This could be a bit better with Go 1.18, but that’s for the future.

See individual commit messages for (a bit) more detail.

#1383 is a counterpart for the write paths.

### DIFF
--- a/store.go
+++ b/store.go
@@ -1036,6 +1036,44 @@ func (s *store) allImageStores() ([]roImageStore, error) {
 	return append([]roImageStore{primary}, additional...), nil
 }
 
+// readAllImageStores processes allImageStores() in order:
+// It locks the store for reading, checks for updates, and calls
+//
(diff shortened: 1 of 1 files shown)


rhatdan:
@saschagrunert @nalind @vrothberg @giuseppe @umohnani8 @flouthoc PTAL

giuseppe (approved):
LGTM

PR #1385, opened 2022-10-13
Simplify the name update call stack

mtrmac:
Instead of going
-	3 store methods
-	store.updateNames+enum
-	3 sub-store methods
-	subStore.updateNames+enum
-	applyNameOperation+enum,

simplify to
-	3 store methods
-	store.updateNames+enum
-	subStore.updateNames+enum
-	applyNameOperation+enum,

Should not change behavior. Looking purely at `updateNameOperation`, invalid values would now be detected after doing more work, but there is no way for an external caller to trigger use of an invalid value anyway.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -81,18 +81,8 @@ type rwContainerStore interface {
 	// convenience of the caller, nothing more.
 	Create(id string, names []string, image, layer, metadata string, options *ContainerOptions) (*Container, error)
 
-	// SetNames updates the list of names associated with the container
-	// with the specified ID.
-	// Deprecated: Prone to race conditions, suggested alternatives are `AddNames` and `RemoveNames`.
-	SetNames(id string, names []string) error
-
-	// AddNames adds the supplied values to the list of names associated with the container with
-	// the specified id.
-	AddNames(id string, names []string) error
-
-	// RemoveNames removes the supplied values from the list of names associated with the container with
-	// the specified id.
-	RemoveNames(id string, names []string) error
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
@nalind @flouthoc @giuseppe @vrothberg @saschagrunert @umohnani8 PTAL

vrothberg (commented):
LGTM, needs a rebase @mtrmac

flouthoc (changes requested):
`RemoveNames` and `AddNames` are not deprecated instead they were added only to deprecate racy `SetNames`. Here: https://github.com/containers/storage/pull/1153

Downstream is still using `RemoveNames` or `AddNames` and this breaks public API without any deprecation notice IMHO.

Last switch in one of the downstreams happend here https://github.com/containers/buildah/pull/4190 but there are many more PR's.

nalind:
I'm fine with dropping the methods from rwLayerStore/rwImageStore/rwContainerStore - those are package-private types.  The exported interfaces don't change, so LGTM.

flouthoc:
> I'm fine with dropping the methods from rwLayerStore/rwImageStore/rwContainerStore - those are package-private types. The exported interfaces don't change, so LGTM.

Ah i see, SGTM as well since I see downstreams only using `Store` directly which is still public so that clears my doubt. LGTM

@mtrmac There is a conflict in the PR though, could you please rebase.

flouthoc (approved):

mtrmac:
Rebased, tests pass.

PR #1386, opened 2022-10-13
Lock layers only for reading

mtrmac:
Only lock the primary layer store for reading in `CreateImage` and `Layers`; I can’t see any reason for the exclusivity.

Cc: @nalind in case I’m missing something, primarily in `CreateImage`.


rhatdan (approved):
LGTM

nalind:
LGTM

PR #1387, opened 2022-10-14
Fix layer locking in CreateContainer

mtrmac:
- `imageTopLayerForMapping`, in the `cParentLayer` loop, was reading all additional layer stores, although only some of them might have been locked at that point.
- `getAutoUserNS` → `getMaxSizeFromImage` was assuming that all layer stores are locked, but in fact the additional layer stores were all unlocked at that point
- `getMaxSizeFromImage` was calling `store.getLayerStore()` and the like, which could potentially trigger a reload via `graphLock` and return a _different_ store than the one that was locked

So, lock _all_ layer stores in `CreateContainer` (at least on the path where images are involved), and pass the locked layer stores down the call stack.

Also, document a bit more explicitly what's going on.

### DIFF
--- a/store.go
+++ b/store.go
@@ -1249,6 +1249,11 @@ func (s *store) CreateImage(id string, names []string, layer, metadata string, o
 	return ristore.Create(id, names, layer, metadata, creationDate, options.Digest)
 }
 
+// imageTopLayerForMapping does ???
+// On entry:
+// - ristore must be locked EITHER for reading or writing
+// - rlstore must be locked for writing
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
Nice catch, LGTM

rhatdan (approved):
LGTM

mtrmac:
For the record, I would never just _find_ this at staring at the code. (I don’t have that kind of patience and focus, for starters.) This is an outcome of prototyping lock tokens per #1389 .

PR #1392, opened 2022-10-17
Fixup a comment in the overlay driver

nalind:
Since maxDepth is 500 and not 128, the comment doesn't make sense. Instead of defining a new named constant, just use the value, since the choice of 128 was already explained in the paragraph just above it.


flouthoc (approved):
LGTM, CI is stuck because of flake.

flouthoc:
restarted CI

mtrmac (commented):
LGTM. Thanks!

PR #1395, opened 2022-10-17
Move locking responsibility from store.go to the container/image/layer stores

mtrmac:
Final big part of the preparatory infrastructure work on locking.

Move the conceptual responsibility for locking/loading `containerStore`, `imageStore` and `layerStore` from the caller (`store`) into the three individual C/I/L objects. And in those objects, centralize it.

I.e. this goes from
```go
func (s *store) method(…) {
    …
    layerStore.Lock()
    defer layerStore.Unlock()
    if err := store.ReloadIfChanged(); err != nil { … }
```

to
```go
func (s *store) method(…) {
    …
    if err := layerStore.startWriting(); err != nil { … }
    defer layerStore.stopWriting()
```

that will, finally, allow us to change the design in a single place, `startReading` / `startWriting`. E.g. to:
- Do in-process locking correctly (#1332 )
- Clean up inconsistent state, and write it back to disk, even if the caller is a reader (#1136 )
- Avoid the need to call `Locked()` or `IsReadWrite` in the `Load` method, passing the values from the (now single) caller instead

---

This should not change observable behavior, apart from the “Remove a redundant rwImageStore.Save() call” commit.

(The need to change every lock caller in `store` by this PR is the primary motivation for preparing #1383 / #1384 , which decreased the number of places to update.)

This is organized by the individual store, changing each one separately; i.e. each conceptual change is repeated three times. If desired, I can instead reorganize this to do same conceptual change in a single commit, across all three stores.

This ends up removing `roFileBasedStore` / `rwFileBasedStore`, and declaring the `{start,stop}{Reading,Writing}` methods in each individual store. If preferred, I could change this PR to keep (or re-introduce) those interfaces again.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -235,6 +235,18 @@ func (r *containerStore) stopReading() {
 	r.lockfile.Unlock()
 }
 
+// ReloadIfChanged reloads the contents of the store from disk if it is changed.
+func (r *containerStore) ReloadIfChanged() error {
+	r.loadMut.Lock()
+	defer r.loadMut.Unlock()
+
+	modified, err := r.lockfile.Modified()
+	if err == nil && modified {
+		return r.Load()
+	}
+	return err
+}
+
 func (r *containerStore) Containers() ([]Container, error) {
 	containers := make([]Container, len(r.containers))
 	for i := range r.containers {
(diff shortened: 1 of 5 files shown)


rhatdan (approved):
LGTM
@nalind @flouthoc @giuseppe @vrothberg @saschagrunert PTAL

giuseppe (approved):
LGTM

vrothberg (commented):
LGTM

PR #1396, opened 2022-10-18
Fix lockfiles never detecting changes

nalind:
If we ever check the last-writer of a lock file and get back a zero value, don't only check if subsequent last-writer values match its first zero bytes, which is always true, as a test to see if the last-writer value has remained unchanged.

Related: https://github.com/containers/podman/pull/16211


mtrmac (commented):
LGTM.

Do we need a test for the expected behavior of `Modified` (both triggering once, per this PR, , and not triggering afterwards, per #1391) on a fresh lock?

rhatdan:
LGTM

PR #1397, opened 2022-10-18
Add TestLockfileMultiprocessModified()

nalind:
Test changes introduced in #1396.  Should fail if either #1396 or #1391 is reverted.

### DIFF
--- a/pkg/lockfile/lockfile_test.go
+++ b/pkg/lockfile/lockfile_test.go
@@ -626,7 +626,8 @@ func TestLockfileMultiprocessModified(t *testing.T) {
 	// Take a read lock somewhere, then see if we incorrectly detect changes.
 	cmd, wc, rc1, err := subLock(lock)
 	wc.Close()
-	cmd.Wait()
+	err = cmd.Wait()
+	require.NoError(t, err)
 	rc1.Close()
 
 	lock.Lock()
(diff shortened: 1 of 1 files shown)


nalind:
Tweaked to also verify calling `Modified()` again after calling it the first time, but before the last-writer is changed by anyone.

rhatdan:
LGTM

pkg/lockfile/lockfile_test.go
```diff
@@ -605,3 +605,58 @@ func TestLockfileMultiprocessMixed(t *testing.T) {
...
+	lock.Lock()
+	modified, err = lock.Modified()
+	lock.Unlock()
+	assert.NoError(t, err, "checking if lock was modified")
+	assert.False(t, modified, "expected Modified() to be false after we check the first time, before anyone Touched it")
+
+	// Take a read lock somewhere, then see if we incorrectly detect changes.
+	cmd, wc, rc1, err := subLock(lock)
+	wc.Close()
+	cmd.Wait()
```
mtrmac, line 629:
```suggestion
	err = cmd.Wait()
	require.NoError(t, err)
```
please, to avoid risk of undetected failure modes, and to reduce noise from linters. (Everywhere.)

mtrmac (commented):
Thanks for adding this!

Overall approach LGTM.

PR #1398, opened 2022-10-18
Don't create $root/{mounts,tmp}, which we don't use

nalind:
Don't create "mounts" or "tmp" subdirectories under the graph root at startup, since we ended up not using them.

Resolves https://github.com/containers/podman/discussions/16195.


rhatdan:
LGTM

mtrmac (approved):
LGTM.

PR #1399, opened 2022-10-18
API break: Remove Lockfile.Locked

### LINKED ISSUE #1388: lockfile.Locked() must not be used to make any decisions
Looking into how https://github.com/containers/storage/issues/1324 crashed:
```
github.com/containers/storage/pkg/lockfile.(*lockfile).Touch
	github.com/containers/storage@v1.41.1-0.20220616120034-7df64288ef35/pkg/lockfile/lockfile_unix.go:247 +0x198
```
failed because `layerStore.lockfile` was not held.

But we got there via
```
github.com/containers/storage.(*layerStore).saveLayers
	github.com/containers/storage@v1.41.1-0.20220616120034-7df64288ef35/layers.go:498 +0x1d4
github.com/containers/storage.(*layerStore).Load
	github.com/containers/storage@v1.41.1-0.20220616120034-7df64288ef35/layers.go:422 +0x534
```
which can only happen if `layerStore.lockfile.Locked()`!

My guess at what has happened (apart from the mystery of `graphLock`, https://github.com/containers/storage/issues/1324#issuecomment-1278400558 ):
- There was a `layerStore` instance for the path in question.
- Somehow, we got another `layerStore` instance. Most likely, `store.graphLock.TouchedSince(…)` triggered a re-load, which abandons the current `store.layerStore` and created a new instance, without writing for the old one to go quiescent. That’s _should_ be fine, the code is supposed to be reliable against out-of-process writers, so two independent in-process writers is easier.
- The two `layerStore` instances get `layerStore.lockfile` via `GetLockfile(path)`, and both receive the same `pkg/lockfile.lockfile` instance. That’s as expected, and fine.
- But now, `Locked()` is no longer useful! It shows that _one_ of the two owners of that lock object has locked it for writing, but it does not say that it’s the current one.
 
So, with A and B, both living in the same process, owning the same lock, this becomes possible:
- A locks the lock
- B checks `.Locked()` and thinks that B owns it
- B starts to make destructive changes
- At this point A and B are racing and creating invalid state, although out-of-process readers and writers are still excluded
- A can unlock the lock
- B continues to make destructive changes, racing against other processes.

AFAICS `Locked` is fundamentally not useful in Go (at least not without a concept of a goroutine ID, which is intentionally not available).

So, following the precedent of #1376 , I plan to break the API and remove that method.

mtrmac:
As detailed in #1388, `Lockfile.Locked()` does not return the right values if there are multiple users of the same lock within a single process.

So, replace the users by callers correctly tracking locking state; and add `LockFile.AssertLocked` and `Lockfile.AssertLockedForWriting` purely for diagnostic purposes. These functions can trigger a panic in clearly invalid situations, but can’t be used by callers to avoid tracking state correctly. That’s a bit risky, so I’ll at least trigger Buildah and Podman CI runs with this change

Following the precedent of #1376 , this removes the `Locked` method, breaking API of c/storage.

Fixes #1388 (and makes another small step towards fixing #1136; the relevant places are now annotated).

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -191,7 +191,7 @@ func (r *containerStore) startWritingWithReload(canReload bool) error {
 	}()
 
 	if canReload {
-		if err := r.ReloadIfChanged(); err != nil {
+		if err := r.reloadIfChanged(true); err != nil {
 			return err
 		}
 	}
@@ -222,7 +222,7 @@ func (r *containerStore) startReading() error {
 		}
 	}()
 
-	if err := r.ReloadIfChanged(); err != nil {
+	if err := r.reloadIfChanged(false); err != nil {
 		return err
 	}
 
@@ -235,14 +235,17 @@ func (r *containerStore) stopReading() {
 	r.lockfile.Unlock()
 }
 
-// ReloadIfChanged reloads the contents of the store from disk if it is changed.
-func (r *containerStore) ReloadIfChanged() error {
+// reloadIfChanged reloads the contents of the store from disk if it is changed.
+//
+// The caller must hold r.lockfile for reading _or_ writing; lockedForWriting is true
+// if it is held for writing.
+func (r *containerStore) reloadIfChanged(lockedForWriting bool) error {
 	r.loadMut.Lock()
 	defer r.loadMut.Unlock()
 
 	modified, err := r.lockfile.Modified()
 	if err == nil && modified {
-		return r.Load()
+		return r.load(lockedForWriting)
 	}
 	return err
 }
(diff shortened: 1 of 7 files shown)


mtrmac:
This should very slightly improve efficiency because we no longer take the lock’s state mutex to determine if we can write to the store.

containers.go
```diff
@@ -302,17 +307,19 @@ func (r *containerStore) Load() error {
 	r.bylayer = layers
 	r.byname = names
 	if needSave {
+		if !lockedForWriting {
+			// Eventually, the callers should be modified to retry with a write lock, instead.
+			return errors.New("container store is inconsistent and the current caller does not hold a write lock")
```
tyler92, line 312: How can callers know when they need to try again? I mean how to distinguish this error (which requires retry) from any other? I see ErrDuplicateImageNames in layerStore.Load()

mtrmac: They can’t, currently.

I plan to add that (“try again”, instead of callers hard-coding specific error values) both to `load` and the callers simultaneously.

mtrmac:
Please don’t merge until this is proven not to break Buildah and Podman tests:
- https://github.com/containers/buildah/pull/4350
- https://github.com/containers/podman/pull/16222

mtrmac:
Buildah tests succeeded; Podman failures seem unrelated.

rhatdan (approved):
LGTM

rhatdan:
@saschagrunert @nalind @vrothberg @flouthoc @umohnani8 PTAL

containers.go
mtrmac: The retries will be added in #1407 .

vrothberg (commented):
LGTM

PR #1400, opened 2022-10-19
Fix locking bugs

mtrmac:
Don't immediately unlock the layer store after locking it.

Luckily it only matters for additional stores.


mtrmac:
Also:

Fix `{layerStore,imageStore}.startWriting()`
    
This is a bad bug: we wouldn't reload state from disk at all in writers.

vrothberg (commented):
LGTM

vrothberg:
Feel free to merge.  I don't feel comfortable with the 'update branch' button.

PR #1401, opened 2022-10-19
fix break during unmount

vrothberg:
Move the break statement out of the switch to actually break out of the loop as intended.

Signed-off-by: Valentin Rothberg <vrothberg@redhat.com>


flouthoc (approved):
LGTM

rhatdan:
LGTM

PR #1406, opened 2022-10-21
Don’t silently ignore errors when decoding JSON or deleting incomplete layers

mtrmac:
See individual commit messages for details

Cc: @nalind to make sure I’m not missing some error resiliency trade-off.

I’ll run this through Buildah and Podman CI together with some follow-up PRs.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -281,26 +281,30 @@ func (r *containerStore) load(lockedForWriting bool) error {
 	if err != nil && !os.IsNotExist(err) {
 		return err
 	}
+
 	containers := []*Container{}
-	layers := make(map[string]*Container)
-	idlist := []string{}
-	ids := make(map[string]*Container)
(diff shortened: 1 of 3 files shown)


mtrmac:
Downstream tests (from #1406, a superset):
- https://github.com/containers/buildah/pull/4368
- https://github.com/containers/podman/pull/16253

vrothberg (commented):
LGTM

Let's wait for @nalind's head nod before merging.

mtrmac:
Rebased, and added one more commit to avoid more `err == nil` checks.

mtrmac:
I don’t personally have any preference in `%q` vs `%#v`, do you want me to change that to be consistent for layer IDs, one way or another?

nalind:
Nah, it's already in too many places to be worth the churn.

PR #1407, opened 2022-10-21
Clean up inconsistent stores also in readers

### LINKED ISSUE #1136: Crash resiliency: deleting incomplete layers doesn’t reliably happen

mtrmac:
When we reload the store in a reader, drop the lock, lock for writing, reload again to trigger cleanup, and then lock again for reading and proceed with the originally intended operation.

This should fix #1136 , although that’s not yet proven. (It only took 6 weeks of work to get this far…)

<s>Built on top of #1399 and #1406 .</s>

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -278,7 +278,6 @@ func (r *containerStore) datapath(id, key string) string {
 // The caller must hold r.lockfile for reading _or_ writing; lockedForWriting is true
 // if it is held for writing.
 func (r *containerStore) load(lockedForWriting bool) error {
-	needSave := false
 	rpath := r.containerspath()
 	data, err := os.ReadFile(rpath)
 	if err != nil && !os.IsNotExist(err) {
@@ -295,6 +294,7 @@ func (r *containerStore) load(lockedForWriting bool) error {
 	layers := make(map[string]*Container)
 	ids := make(map[string]*Container)
 	names := make(map[string]*Container)
+	var errorToResolveBySaving error // == nil
 	for n, container := range containers {
 		idlist = append(idlist, container.ID)
 		ids[container.ID] = containers[n]
@@ -302,7 +302,7 @@ func (r *containerStore) load(lockedForWriting bool) error {
 		for _, name := range container.Names {
 			if conflict, ok := names[name]; ok {
 				r.removeName(conflict, name)
-				needSave = true
+				errorToResolveBySaving = errors.New("container store is inconsistent and the current caller does not hold a write lock")
 			}
 			names[name] = containers[n]
 		}
@@ -313,10 +313,10 @@ func (r *containerStore) load(lockedForWriting bool) error {
 	r.byid = ids
 	r.bylayer = layers
 	r.byname = names
-	if needSave {
+	if errorToResolveBySaving != nil {
 		if !lockedForWriting {
 			// Eventually, the callers should be modified to retry with a write lock, instead.
-			return errors.New("container store is inconsistent and the current caller does not hold a write lock")
+			return errorToResolveBySaving
 		}
 		return r.Save()
 	}
--- a/containers.go
+++ b/containers.go
@@ -191,7 +191,7 @@ func (r *containerStore) startWritingWithReload(canReload bool) error {
 	}()
 
 	if canReload {
-		if err := r.reloadIfChanged(true); err != nil {
+		if _, err := r.reloadIfChanged(true); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


mtrmac:
Downstream tests:
- https://github.com/containers/buildah/pull/4368
- https://github.com/containers/podman/pull/16253

containers.go
```diff
@@ -215,18 +215,41 @@ func (r *containerStore) stopWriting() {
...
-	if err := r.reloadIfChanged(false); err != nil {
-		return err
+	if tryLockedForWriting, err := r.reloadIfChanged(false); err != nil {
+		if !tryLockedForWriting {
+			return err
+		}
+		unlockFn()
+		unlockFn = nil
+
+		r.lockfile.Lock()
```
tyler92, line 232: It looks like an upgradable lock (when the reader can become a writer but atomically). I didn't find a popular library for Go, but I have experience with [boost](https://www.boost.org/doc/libs/1_71_0/doc/html/interprocess/synchronization_mechanisms.html#interprocess.synchronization_mechanisms.sharable_upgradable_mutexes.sharable_upgradable_mutexes_operations.upgradable_mutexes_operations_upgradable) (just for example of upgradable mutex). Maybe it's good idea to use this approach here? I will allow avoiding double call of reloadIfChanged.

mtrmac: Yes, a lock type like that can be built.

A limitation here is that we need an _inter-process_ lock, a lock that works across user namespaces without users having to set up anything (and if at all possible one that is interoperable with a previous version of c/storage). That basically limits us to the two kernel-provided file locking mechanisms, and other concerns direct us to the `fcntl` mechanism.

Now, `fcntl` locks do, it seems, support lock type change; the formal specification seems to be a bit ambiguous about whether it is atomic or whether it internally releases the lock.

---

Purely in principle:

For container and image stores, this is AFAICT a “shouldn’t happen” code path; in that case, performance really isn’t a concern at all, and it definitely doesn’t warrant adding a dependency or using a non-trivial locking implementation.

The layer store, with its recovery of incomplete layers, which _can_ happen in production for unavoidable reasons, an upgradeable would be useful because it would mean that we don’t have to loop, and don’t have to decide how much to loop. But, in that case, all locking costs are dwarfed by costs of actual cleanup filesystem operations anyway…

---

So, overall, I rather favor the most simple approach possible. Not that the loops in `layerStore.startReadingWithReload` are _that_ simple, but still, better than custom lock implementations. I guess?

tyler92: Thank you for the explanation. I agree, and I did not take into account that upgradable requires an inter-process lock.

rhatdan (approved):
LGTM

mtrmac:
This is still a draft: I’d like to take the time to manually set up the inconsistent state and actually verify that this fixes (the current variant of) #1136 .

rhatdan:
@giuseppe @nalind @vrothberg @flouthoc @saschagrunert @umohnani8 PTAL

mtrmac:
So this was completely ineffective (calling `reloadIfChanged` the second time after a change does nothing). At least based on the test case in #1136 , it seems to work now.

Waiting for another round of downstream tests : https://github.com/containers/buildah/pull/4368 https://github.com/containers/podman/pull/16253 ; if they pass, I’m calling this ready for review.

rhatdan:
@vrothberg @giuseppe @nalind @flouthoc @saschagrunert @umohnani8 PTAL

vrothberg (commented):
Changes LGTM

Note that I am far from being familiar with the low-level details.

giuseppe (approved):
LGTM

flouthoc (approved):
LGTM

flouthoc:
@mtrmac https://github.com/containers/buildah/pull/4368 https://github.com/containers/podman/pull/16253 did not run test the most recent push on this PR

mtrmac:
@flouthoc There were no changes to this PR’s code, but, sure, let me rebase and re-test to be sure.

TomSweeneyRedHat:
LGTM
and happy green test buttons.

mtrmac:
@flouthoc Rebased tests passed again.

This PR now includes #1428 , alternatively it could be rebased on top with no changes, but the Buildah/Podman vendoring tests wouldn’t match commit IDs. I don’t care which way this PR is merged.

flouthoc (approved):
LGTM
/lgtm

PR #1411, opened 2022-10-25
overlay: drop relLowers workaround

giuseppe:
commit 7c5964df95c892cfbdbce594cf5a8e2973c70fd7 changed the way the mountFrom program works by always using the /proc/self/fd/%d path for the lower dirs.  Hence there is no gain to pass a path relative to the storage, but we can just pass the absolute path as the resulting file path length is the same and won't affect how many layers we can specify.

This fixes idmapped mounts when the total length of lower layers is > 4096 and it also solves the problem of using layers from additional stores.

Closes: https://github.com/containers/storage/issues/1410

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1378,28 +1378,9 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		return "", errors.New("max depth exceeded")
 	}
 
-	// absLowers is the list of lowers as absolute paths, which works well with additional stores.
+	// absLowers is the list of lowers as absolute paths.
 	absLowers := []string{}
-	// relLowers is the list of lowers as paths relative to the driver's home directory.  If
-	// additional stores are being used, some of these will still be absolute paths.
-	relLowers := []string{}
 
-	// Check if $link/../diff{1-*} exist.  If they do, add them, in order, as the front of the lowers
-	// lists that we're building.  "diff" itself is the upper, so it won't be in the lists.
-	link, err := os.ReadFile(path.Join(dir, "link"))
-	if err != nil {
-		if !os.IsNotExist(err) {
-			return "", err
-		}
-		logrus.Warnf("Can't read parent link %q because it does not exist. Going through storage to recreate the missing links.", path.Join(dir, "link"))
-		if err := d.recreateSymlinks(); err != nil {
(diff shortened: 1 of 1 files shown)


giuseppe:
@nalind @rhatdan PTAL

rhatdan:
LGTM
@nalind @saschagrunert @flouthoc @mtrmac @umohnani8 @vrothberg PTAL

giuseppe:
The issue happened with an image that has many layers. I think it is fine for now to simplify the code if the additional cost happens only with these images

mtrmac (commented):
- It’s completely unclear to me how this is fixing anything. Was the relative-path code buggy in some way WRT additional stores?
- If this code is code the primary reason for creating the `link` symlink forest, should we actually drop _all_ users of it? There are some left.
- … and show we stop creating the symlinks at all? (Would that break older-version readers, or would they workaround via `recreateSymlink`? And if it would break them, do we need to care, or is one-directional compatibility sufficient?)

giuseppe:
> * It’s completely unclear to me how this is fixing anything. Was the relative-path code buggy in some way WRT additional stores?

when idmapped mounts are used, we were just rewriting the absolute lower dirs, but not updating the `relLowers`.  This is probably not possible since the idmapped mounts use a subdirectory where the idmapped bind mount is created, thus the relative dirs in `link` won't work.



> * If this code is code the primary reason for creating the `link` symlink forest, should we actually drop _all_ users of it? There are some left.
> * … and show we stop creating the symlinks at all? (Would that break older-version readers, or would they workaround via `recreateSymlink`? And if it would break them, do we need to care, or is one-directional compatibility sufficient?)

I had a quick look at it, and it seems the `link` file is still used by `getLower` and to point to layers into additional stores so we would first need to fix these two other occurrences.

mtrmac:
> when idmapped mounts are used, we were just rewriting the absolute lower dirs, but not updating the `relLowers`. 

Ah. Thanks for writing that down.

giuseppe:
can we move this forward or is there anything blocking I should take care of?

mtrmac:
The _implementation_ LGTM.

As for the PR as a whole, do we want to drop all of the `link` data structure now, or not? I don’t feel qualified to have an opinion.

rhatdan:
If we need to `drop all of the link data structure now, or not`
we can do it in a separate PR.
LGTM

PR #1412, opened 2022-10-25
Fix some warnings again

mtrmac:
This is my penance for not actually finishing the warning cleanup, so that I could bump the `golangci-lint` version and make these things enforced on every pull request.

So, I get to run the recent version of `golangci-lint` locally, ignore the things I didn’t fix yet, and fix the newly introduced issues after the fact.


vrothberg (commented):
LGTM
@giuseppe PTAL

rhatdan:
LGTM

PR #1415, opened 2022-10-28
unshare: fix creating a userns when running as root

giuseppe:
this check would prevent the function to ever be used when running as root, since we won't check what capabilities are currently available to the process.

### DIFF
--- a/pkg/unshare/unshare_linux.go
+++ b/pkg/unshare/unshare_linux.go
@@ -465,7 +465,7 @@ func bailOnError(err error, format string, a ...interface{}) { // nolint: golint
 // MaybeReexecUsingUserNamespace re-exec the process in a new namespace
 func MaybeReexecUsingUserNamespace(evenForRoot bool) {
 	// If we've already been through this once, no need to try again.
-	if os.Geteuid() == 0 && IsRootless() {
+	if os.Geteuid() == 0 && GetRootlessUID() > 0 {
 		return
 	}
 


pkg/unshare/unshare_linux.go
```diff
@@ -464,11 +464,6 @@ func bailOnError(err error, format string, a ...interface{}) { // nolint: golint
 
 // MaybeReexecUsingUserNamespace re-exec the process in a new namespace
 func MaybeReexecUsingUserNamespace(evenForRoot bool) {
-	// If we've already been through this once, no need to try again.
```
rhatdan, line 467: Doesn't this change it from Maybe to just Reexec?

giuseppe: we don't always reexec unless `evenForRoot` is set.

If `evenForRoot` is not set, then we check if CAP_SYS_ADMIN is in the effective capabilities

rhatdan:
LGTM
@flouthoc @saschagrunert @nalind @mtrmac @vrothberg @flouthoc @umohnani8 PTAL

pkg/unshare/unshare_linux.go
```diff
@@ -464,11 +464,6 @@ func bailOnError(err error, format string, a ...interface{}) { // nolint: golint
 
 // MaybeReexecUsingUserNamespace re-exec the process in a new namespace
 func MaybeReexecUsingUserNamespace(evenForRoot bool) {
```
mtrmac, line 466: If this changes the semantics of the function in caller-visible ways, I’d expect that to show up _at least_ in the documentation, if not in the name.

IMHO doc strings should be enough for a caller to understand what the function does (from the caller’s perspective) without needing to read the implementation, and this one is way below that standard.

mtrmac (commented):
What is this actually fixing?

Some situation where we are in a user namespace already, one that wasn’t created by this subpackage (no environment variables), we do have CAP_ADMIN (per `IsRootless`), but don’t have some other capability?

Something else?

Does that mean that the semantics of `IsRootless` is also incorrect and needs updating? (The documentation of `IsRootless` says basically nothing, so I have absolutely no idea.)

---

I’m not saying that this change is incorrect, I’m saying that I have no way to tell, and this might be a good opportunity to actually clarify the semantics.

giuseppe:
> What is this actually fixing?

the issue I've observed is that if we run as root but without CAP_SYS_ADMIN, then the function exits immediately without creating the user namespace as it was asked to do.

I've changed the patch so we avoid the additional checks if we an unprivileged user (uid > 0) created the user namespace and we are not already running with euid == 0

flouthoc (approved):
LGTM Just confirming that this PR does not appears to be functionally changing anything and just avoiding checks for capabilities when it is guaranteed that efffective UID is 0 and this is rootless session.

giuseppe:
can we please move this forward?  It is needed for buildah inside a container

flouthoc (approved):
LGTM
/lgtm

slonopotamus:
> It is needed for buildah inside a container

This change _broke_ my usage of buildah inside container.

Before this change:

```
$ docker run --rm -it quay.io/buildah/stable:v1.28.0 buildah unshare echo hi
hi
```

After this change:

```
$ docker run --rm -it quay.io/buildah/upstream:latest buildah unshare echo hi
Error during unshare(CLONE_NEWUSER): Operation not permitted
ERRO[0000] parsing PID "": strconv.Atoi: parsing "": invalid syntax 
ERRO[0000] (Unable to determine exit status)
```

pkg/unshare/unshare_linux.go
slonopotamus: > we don't always reexec unless `evenForRoot` is set.

Uh? The only codepath in this function that doesn't reexec doesn't involve `evenForRoot` in any way.

rhatdan:
@giuseppe WDYT?

slonopotamus:
We've discussed that a bit in #4563, but since they're kinda separate issues, I've created https://github.com/containers/buildah/issues/4563 specifically to address regression.

PR #1417, opened 2022-11-02
fix inconsitent go build tags

Luap99:
Run `go fmt ./...` which automatically adds the new build tag syntax.
This change is backwards compatible.

CI: run make local-validate

Make sure all PRs are formated with `go fmt` correctly.
I added it to the existing lint job as this seems to make the most
sense to me.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -114,7 +114,7 @@ lint_task:
       echo "deb http://deb.debian.org/debian stretch-backports main" > /etc/apt/sources.list.d/backports.list
       apt-get update
       apt-get install -y libbtrfs-dev libdevmapper-dev
-    test_script: make lint
+    test_script: make local-validate && make lint
 
 
 # Update metadata on VM images referenced by this repository state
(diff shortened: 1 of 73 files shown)


rhatdan:
Lint is broken.

Luap99:
This should good now, the one failure is caused by a CI issue. I do not have permission to restart it here.

lsm5 (approved):
/lgtm
/hold

lsm5:
looks like it needs a rebase

lsm5:
@rhatdan i think would be nice to get this vendored into podman, buildah, skopeo soon as possible.

Luap99:
@lsm5 did you checked if it fixes your issue?
AFAIK go is still backwards compatible with the old tags so it should at least in theory work even without this PR.

lsm5:
Didn't check prior. Anyway, trying to vendor it now on f37 with golang v1.19.2 and I'm seeing unrelated vendoring issues
```
github.com/containers/podman/v4/cmd/podman/images imports
	github.com/containers/ocicrypt/helpers imports
	github.com/containers/ocicrypt/utils imports
	gopkg.in/square/go-jose.v2: unrecognized import path "gopkg.in/square/go-jose.v2": https fetch: Get "https://gopkg.in/square/go-jose.v2?go-get=1": dial tcp 185.125.188.236:443: i/o timeout
```

I'll spin up a c9s vm in a bit and look further

PR #1418, opened 2022-11-02
overlay: fix call to clone on s390x

giuseppe:
the two arguments to clone are swapped on s390x. This patch fixes the call when running on s390x.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2108887


rhatdan:
LGTM
@saschagrunert @vrothberg @nalind @flouthoc @mtrmac @umohnani8 PTAL

flouthoc (approved):
LGTM

PR #1423, opened 2022-11-04
[release-1.42] overlay: fix call to clone on s390x

nalind:
the two arguments to clone are swapped on s390x. This patch fixes the call when running on s390x.  Cherry-picked from #1418.


rhatdan:
LGTM

Issue #1389, opened 2022-10-14
Ideas wanted: Rely on language features to enforce locking correctness


mtrmac:
Looking at #1324, #1332, #1379 , it’s not absolutely required but boy it would be _nice_ if we could have the programming language enforce locking invariants
- That locks that are supposed to be held are held
- That reads to struct fields only happen with a read lock held
- That writes to struct fields only happen with a write lock held.

At the thousands of lines of c/storage code, and about half a dozen locks involved, it’s only natural that _some_ mistakes will happen and we could certainly use some help noticing them.

Sadly, Go is very weak for these purposes. So, this is an **open-ended call for ideas**.

mtrmac:
---

- I have nothing at all for protecting access to struct fields. AFAICS we could make the struct completely private and wrap everything in getters/setters … written _manually_. Not worth it.
- For enforcing API invariants:
  - Interfaces _sort of_ work:
    ```go
    type unlockedStore interface {
       startReading() readableStore
       startWriting() writableStore
    }
    
    type readableStore interface {
       // readers
       Close()
    }
    // and writableStore similarly
    ```
    but if the three interfaces are implemented by the same struct, calls _within_ that struct are no longer protected (e.g. a “reader” method can call a “writer” method no the same object)
  - “lock tokens” work:
    ```go
    type readToken struct {} // unique type, zero cost to pass around
    type writeToken struct {} // unique type, zero cost to pass around
    type store interface { // or just methods on a struct
       startReading() readToken
       stopReading(readToken)
       startWriting() writeToken
       stopWriting(writeToken)

       readSomething(readToken)
       writeSomething(writeToken, valueType)
    }
    ```
    but require all call sites to carry around an _extra_ token value. That has zero or close-to-zero run-time cost, but the code is larger.

Is there something else that would work? Anything that would be worth doing?

mtrmac:
> * I have nothing at all for protecting access to struct fields

@vrothberg came up with
```go
type stateful struct {
    lock sync.Mutex
    requiresLock struct {
        field …
    }
}
```

which is much better than nothing. A downside is that such a type can’t be created with a trivial struct initializer: either `.requiresLock.field` must be initialized after the struct is created, or `requiresLock` must become an out-of-line separately-defined named type — that’s probably still well worth the cost.

PR #1430, opened 2022-11-12
More small locking fixes

mtrmac:
- Don’t write to `BigDataDigests` and `BigDataSizes` in methods that only have a read lock — either avoid that, or use a write lock
- Stop using type checks to see if an image store is primary, or read-write; it just doesn’t work like that.

---

\*Grumble\* I’m getting increasingly desperate for #1389 . Staring at locking code trying to not make a mistake feels like writing C memory allocation code manually — however good a human can get at it, machines do it faster, and they never get tired or inattentive.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -629,6 +629,7 @@ func (r *containerStore) BigDataSize(id, key string) (int64, error) {
 	return -1, ErrSizeUnknown
 }
 
+// Requires startWriting. Yes, really, WRITING (see SetBigData).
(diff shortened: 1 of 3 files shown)


rhatdan (approved):
LGTM
@nalind @giuseppe @vrothberg @flouthoc @saschagrunert @umohnani8 PTAL

saschagrunert (approved):

vrothberg (approved):
LGTM

PR #1434, opened 2022-11-15
Don't cross mount points when building change lists

dfr:
This is needed on FreeBSD to avoid traversing into /dev or mounted volumes when compiling diff sets for running containers. I'm not sure this approach is the best one but I can't think of a scenario where crossing mount points is intended when comparing storage layers.

### DIFF
--- a/pkg/archive/changes_other.go
+++ b/pkg/archive/changes_other.go
@@ -43,7 +43,12 @@ func collectFileInfoForChanges(oldDir, newDir string, oldIDMap, newIDMap *idtool
 func collectFileInfo(sourceDir string, idMappings *idtools.IDMappings) (*FileInfo, error) {
 	root := newRootFileInfo(idMappings)
 
-	err := filepath.WalkDir(sourceDir, func(path string, d fs.DirEntry, err error) error {
+	sourceStat, err := system.Lstat(sourceDir)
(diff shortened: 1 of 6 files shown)


pkg/archive/changes_other.go
```diff
@@ -86,8 +91,12 @@ func collectFileInfo(sourceDir string, idMappings *idtools.IDMappings) (*FileInf
 		if err != nil {
 			return err
 		}
-		info.stat = s
 
+		if s.Dev() != sourceStat.Dev() {
```
flouthoc, line 95: Is this also applicable for `*nix` ? or does this additional check `if os == freebsd`

flouthoc (commented):
changes LGTM other than one question above

pkg/archive/changes_other.go
dfr: The st_dev field of struct stat is common across all *nix since at least v7. I added the field to StatT for *unix but so far I only initialised it on FreeBSD and Linux - I can go back and add it for darwin, openbsd and solaris but have no good way of testing.

The additional check is currently only needed for FreeBSD - on Linux, the use of mount namespaces seems to avoid the problem of traversing into sub-mounts.

rhatdan:
LGTM

Issue #1433, opened 2022-11-15
Consecutive call to Store.Diff() never returns


negrel:
Hello,
I've a problem using `storage` package, whenever I run a call to Store.Diff() followed by another call to Store.Diff(), the second call never returns. In the following code, the second print is never called.
I've also tried to debug the code using delve and it seems that store.Diff() is waiting indefinitely for the storage lockfile.

```go
package main

import (
	"fmt"

	"github.com/containers/storage"
	"github.com/containers/storage/pkg/reexec"
	"github.com/containers/storage/pkg/unshare"
	"github.com/containers/storage/types"
)

func main() {
	if reexec.Init() {
		return
	}
	unshare.MaybeReexecUsingUserNamespace(false)

	storeOptions, err := types.DefaultStoreOptionsAutoDetectUID()
	if err != nil {
		panic(err)
	}
	store, err := storage.GetStore(storeOptions)
	if err != nil {
		panic(err)
	}

	img, err := store.Image("docker.io/library/ubuntu:22.04")
	if err != nil {
		panic(err)
	}

	diff, err := store.Diff(img.TopLayer, img.TopLayer, nil)
	if err != nil {
		panic(err)
	}
	fmt.Println("first diff done", diff)

	diff, err = store.Diff(img.TopLayer, img.TopLayer, nil)
	if err != nil {
		panic(err)
	}
	fmt.Println("second diff done", diff)
}

```

Here is the `go.mod`:
```
module github.com/negrel/storage-issue

go 1.18

require github.com/containers/storage v1.44.0

require (
	github.com/BurntSushi/toml v1.2.1 // indirect
	github.com/Microsoft/go-winio v0.5.2 // indirect
	github.com/Microsoft/hcsshim v0.9.5 // indirect
	github.com/containerd/cgroups v1.0.1 // indirect
	github.com/docker/go-units v0.5.0 // indirect
	github.com/gogo/protobuf v1.3.2 // indirect
	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
	github.com/google/go-intervals v0.0.2 // indirect
	github.com/google/uuid v1.2.0 // indirect
	github.com/hashicorp/errwrap v1.0.0 // indirect
	github.com/hashicorp/go-multierror v1.1.1 // indirect
	github.com/json-iterator/go v1.1.12 // indirect
	github.com/klauspost/compress v1.15.12 // indirect
	github.com/klauspost/pgzip v1.2.5 // indirect
[...]

PR #1436, opened 2022-11-21
Replace Locker.Modified with a more correct change tracking

### LINKED ISSUE #1420: lockfile.Modified() is designed incorrectly
The lock object (intentionally) is shared across the whole process, and apparently intended to allow multiple in-process callers to work correctly, similarly to how it work with multiple processes using the same lock.

But `Modified()` is an “edge-triggered” signal, which returns `true` _once_ after observing the last-writer ID to be updated.

If there are multiple stores pointing at the same `lockfile` object, only one of them will see `Modified` to be set to `true`, and the others will miss the update.

AFAICS the `lockfile.lw` state should, instead, be maintained in callers of `lockfile.Modified`, not in the lock object itself.

---

Right now, this is _probably_ OK enough; we _can_ have multiple stores in a single process pointing at the same lock file simultaneously (via something like https://github.com/containers/storage/blob/74e37433a2a04ef9ba62d7ae15c9110933b9aca6/store.go#L867 , and there’s a recent-ish panic report suggesting this code path), but we only call `Modified` at the start of the operation, _and_ when creating a store object. So if the linked code path triggers abandoning an existing store object, that object has _probably_ (not certainly) already called `Modified`, and doesn’t care any more and won’t trigger the lockfile update any more; and the new replacement store object is loading data anyway.

mtrmac:
This, primarily,
- modifies `pkg/lockfile` to expose a struct instead of an interface, so that we can keep adding methods without formally breaking API all the time
- adds a new “last writer” state tracking API, where the state is tracked in the data consumer (one per consumer), instead of in the lock file object (one per process). This is necessary for basic correctness, and fixes #1420 (for callers that update).
  - This is also a tiny bit more efficient, because we don’t need to use an extra mutex to protect the per-process “last writer” data
- uses the explicitly managed state to:
  - improve reliability: we now record “last known state” only _after_ successfully reloading from disk; so it’s not possible for a caller to use old data just because we failed to reload. (OTOH that broken state will now result in permanent downtime until the cause is fixed.)
  - improve performance: we no longer reload state on first access _after_ creating the store and loading (fixes the reproducer of #1433)
- modifies the layer reload code not to reload all layers if just the mounts file is modified

See individual commit messages for more changes, and more details.


mtrmac:
The commit description is now updated.

Downstream tests:
- https://github.com/containers/buildah/pull/4430
- https://github.com/containers/podman/pull/16581

mtrmac:
The downstream tests have passed, this is ready for review.

rhatdan (approved):

layers.go
```diff
@@ -503,12 +491,19 @@ func (r *layerStore) reloadIfChanged(lockedForWriting bool) (bool, error) {
...
-	modified, err := r.Modified()
+	layersModified, err := r.layersModified()
 	if err != nil {
 		return false, err
 	}
-	if modified {
+	if layersModified {
+		// r.load also reloads mounts data
 		return r.load(lockedForWriting)
+	} else if r.lockfile.IsReadWrite() {
```
rhatdan, line 504: No need for else here.

rhatdan:
One minor NIT but LGTM
@nalind @vrothberg @giuseppe @saschagrunert @flouthoc PTAL

layers.go
mtrmac: Thanks, updated.

rhatdan:
@alexlarsson PTAL

rhatdan:
@nalind @vrothberg @giuseppe @saschagrunert @flouthoc PTAL

vrothberg (commented):
LGTM

PR #1438, opened 2022-11-22
Document and improve store locking

mtrmac:
Motivated by #1433 seeing unexpected layer store refreshes / existence of parallel layer stores for the same directory in a single process, this:
- Attempts to document the concurrency design and rules for accessing `store` fields and `store.graphLock`. I’m not at all sure it’s correct, at least I want to establish the precedent that the design is expected to be documented
- Consolidates the uses of `store.graphLock` to guard against `Shutdown` calls to `store.startUsingGraphDriver` / `store.stopUsingGraphDriver` helpers, instead of copy&pasted and code slowly deviating over time. This is similar to the CIL`Store.startWriting` approach.
- Fixes users that specifically rely on `store.graphLock` and can trigger a graph driver reload to actually use the reloaded graph driver for their `layerStore` operations, instead of using an obsolete `layerStore` (and graph driver) instance immediately after the reload.
- Introduces a larger set of helpers to obtain the `layerStore` instances for the primary and additional stores, so that every operation only locks `graphLock` once instead of up to three times.
- OTOH, removes helpers to obtain `imageStore` and `containerStore` instances; the users can just refer to a constant field, so they are modified to do that, removing many error checks and local variables
- Fixes various individual instances of missing locking in `store`.

<s>This is on top of unmerged #1436.</s> See individual commit messages for details.

@nalind @giuseppe PTAL, it’s quite likely I’m missing something about the current locking implementation

Cc: @alexlarsson : this might significantly reduce the number of locking operations on `storage.lock`. I have no idea if it is noticeably _faster_, but it’s probably a noticeable change to the `strace` at least, so it might be necessary to track this separately from other changes if you are quantifying other work.

### DIFF
--- a/store.go
+++ b/store.go
@@ -2941,9 +2941,20 @@ func (s *store) Layer(id string) (*Layer, error) {
 }
 
 func (s *store) LookupAdditionalLayer(d digest.Digest, imageref string) (AdditionalLayer, error) {
-	adriver, ok := s.graphDriver.(drivers.AdditionalLayerStoreDriver)
-	if !ok {
-		return nil, ErrLayerUnknown
+	var adriver drivers.AdditionalLayerStoreDriver
+	if err := func() error { // A scope for defer
+		if err := s.startUsingGraphDriver(); err != nil {
+			return err
+		}
+		defer s.stopUsingGraphDriver()
+		a, ok := s.graphDriver.(drivers.AdditionalLayerStoreDriver)
+		if !ok {
+			return ErrLayerUnknown
+		}
+		adriver = a
+		return nil
+	}(); err != nil {
+		return nil, err
 	}
 
 	al, err := adriver.LookupAdditionalLayer(d, imageref)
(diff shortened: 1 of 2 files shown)


mtrmac:
Now ready for review.

Downstream tests:
- https://github.com/containers/buildah/pull/4442
- https://github.com/containers/podman/pull/16705

mtrmac:
Added one more commit to document the locking hierarchy within `store`.

rhatdan:
@alexlarsson @nalind @vrothberg @giuseppe @flouthoc @umohnani8 PTAL

store.go
```diff
@@ -982,11 +982,8 @@ func (s *store) getLayerStore() (rwLayerStore, error) {
 
 // getROLayerStores obtains additional read/only layer store objects used by the
```
rhatdan, line 989: This comment is repeated I believe.

rhatdan (approved):
LGTM, One potential issue with duplicated comments.

store.go
mtrmac: Thanks, fixed.

rhatdan:
@alexlarsson @nalind @vrothberg @giuseppe @flouthoc @umohnani8 PTAL

vrothberg (approved):
LGTM
Nice work!

giuseppe (approved):
LGTM

PR #1439, opened 2022-11-24
Avoid unnecessary reloads of store data

alexlarsson:
We're currently reloading the layers json file every time we (ourselves) write it

This PR fixes this.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -809,15 +809,14 @@ func (r *layerStore) saveLayers(saveLocations layerLocations) error {
 		if err != nil {
 			return err
 		}
-		var opts *ioutils.AtomicFileWriterOptions
+		opts := ioutils.AtomicFileWriterOptions{}
 		if location == volatileLayerLocation {
-			opts = &ioutils.AtomicFileWriterOptions{
-				NoSync: true,
-			}
+			opts.NoSync = true
 		}
-		if err := ioutils.AtomicWriteFileWithOpts(rpath, jldata, 0600, opts); err != nil {
+		if err := ioutils.AtomicWriteFileWithOpts(rpath, jldata, 0600, &opts); err != nil {
 			return err
 		}
+		r.layerspathsModified[locationIndex] = opts.ModTime
 	}
 	lw, err := r.lockfile.RecordWrite()
 	if err != nil {
--- a/pkg/ioutils/fswriters.go
+++ b/pkg/ioutils/fswriters.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 )
 
 // AtomicFileWriterOptions specifies options for creating the atomic file writer.
@@ -13,6 +14,9 @@ type AtomicFileWriterOptions struct {
 	// storage after it has been written and before it is moved to
 	// the specified path.
 	NoSync bool
+	// On successful return from Close() this is set to the mtime of the
+	// newly written file.
+	ModTime time.Time
 }
 
 var defaultWriterOptions = AtomicFileWriterOptions{}
@@ -74,6 +78,11 @@ func AtomicWriteFileWithOpts(filename string, data []byte, perm os.FileMode, opt
 	if err1 := f.Close(); err == nil {
 		err = err1
 	}
+
+	if opts != nil {
+		opts.ModTime = f.modTime
+	}
+
 	return err
 }
 
@@ -87,6 +96,7 @@ type atomicFileWriter struct {
 	writeErr error
 	perm     os.FileMode
 	noSync   bool
+	modTime  time.Time
 }
 
 func (w *atomicFileWriter) Write(dt []byte) (int, error) {
@@ -109,9 +119,25 @@ func (w *atomicFileWriter) Close() (retErr error) {
 			return err
 		}
 	}
+
+	// fstat before closing the fd
+	info, statErr := w.f.Stat()
+	if statErr == nil {
+		w.modTime = info.ModTime()
+	}
+	// We delay error reporting until after the real call to close()
+	// to match the traditional linux close() behaviour that an fd
+	// is invalid (closed) even if close returns failure. While
+	// weird, this allows a well defined way to not leak open fds.
+
 	if err := w.f.Close(); err != nil {
 		return err
 	}
+
+	if statErr != nil {
+		return statErr
+	}
+
 	if err := os.Chmod(w.f.Name(), w.perm); err != nil {
 		return err
 	}


alexlarsson:
Hmm, the tests seem to be built on the idea that the first Modified *should* return true. I think this is sort of weird, but I guess we can work with that by calling Modified() and extra time in the store.

rhatdan (approved):
LGTM
@mtrmac PTAL

pkg/ioutils/fswriters.go
```diff
@@ -109,6 +117,12 @@ func (w *atomicFileWriter) Close() (retErr error) {
 			return err
 		}
 	}
+
+	info, statErr := w.f.Stat()
+	if statErr == nil {
+		w.modTime = info.ModTime()
```
mtrmac, line 123: AFAIK it’s not documented that `w.f.Close()` won’t update the timestamp (e.g. it could be buffering writes for some reason).

OTOH, sure, `Stat()` after closing the file may lose updates. But then all of this protects against ~rogue external writers who don’t lock correctly…

mtrmac (commented):
No, `Modified` needs to be dropped entirely, not made more prevalent: https://github.com/containers/storage/issues/1420 .

I think #1436 + #1346 is a superset of this.

rhatdan:
@alexlarsson Needs a rebase/perhaps a rewrite.

pkg/ioutils/fswriters.go
alexlarsson: I dunno how *documented* this is per se, but that is now how delayed writes work. The mtime information is part of the struct inode, and it will be modified when you do actual vfs operations, not when dirty buffers are flushed to disk.

alexlarsson:
This is not needed anymore on master.

pkg/ioutils/fswriters.go
mtrmac: I was thinking of the Go `os.File` possibly buffering data, the way `<stdio.h>` traditionally does. I agree it would be surprising if it happened, but nothing explicitly promises that it won’t.

>  The mtime information is part of the struct inode, and it will be modified when you do actual vfs operations, not when dirty buffers are flushed to disk.

(BTW, https://pubs.opengroup.org/onlinepubs/9699919799/ “file times update” says that it might be delayed after the actual write even inside the kernel — in a way that wouldn’t matter here.)

alexlarsson:
Ah, sorry, we still need the second half of this

pkg/ioutils/fswriters.go
alexlarsson: I mean, it better have pushed all os.File data to the kernel, or the fdatasync() call we just did was completely wasted, having persisted to disk a partial json file...

Anyway, I'm pretty sure that in practice (i.e. on linux) this will work. Even if Xenix 0.0.1 might do something weird and still be according to spec. And anwyay, as you say, if we get the wrong mtime all that happens is that we unnecessarily load the file again.

rhatdan:
LGTM
@mtrmac PTAL

pkg/ioutils/fswriters.go
mtrmac: > I mean, it better have pushed all os.File data to the kernel, or the fdatasync() call we just did was completely wasted,

Oh, that’s a strong argument.

pkg/ioutils/fswriters.go
```diff
@@ -109,6 +117,12 @@ func (w *atomicFileWriter) Close() (retErr error) {
 			return err
 		}
 	}
+
+	info, statErr := w.f.Stat()
+	if statErr == nil {
```
mtrmac, line 122: Why wouldn’t this be a fatal error?

pkg/ioutils/fswriters.go
```diff
@@ -12,7 +13,8 @@ type AtomicFileWriterOptions struct {
 	// If NoSync is not specified, the file is synced to the
 	// storage after it has been written and before it is moved to
 	// the specified path.
-	NoSync bool
+	NoSync  bool
+	ModTime time.Time
```
mtrmac, line 17: Returning a value in “options” seems very surprising to me. At the very least, it should be clearly documented.

Going by the “`…Options` is input to the operation” logic, a `WriteModTimeOnSuccess *time.Time` field would make at bit more sense … and, OTOH, be more cumbersome to use.

Ultimately it would be cleanest to add a new function that returns that value (or a struct containing that value, for future expansion?)

layers.go
```diff
@@ -809,15 +809,16 @@ func (r *layerStore) saveLayers(saveLocations layerLocations) error {
...
-			opts = &ioutils.AtomicFileWriterOptions{
-				NoSync: true,
-			}
+			opts.NoSync = true
 		}
-		if err := ioutils.AtomicWriteFileWithOpts(rpath, jldata, 0600, opts); err != nil {
+		if err := ioutils.AtomicWriteFileWithOpts(rpath, jldata, 0600, &opts); err != nil {
 			return err
 		}
+		if !opts.ModTime.IsZero() {
```
mtrmac, line 819: This special case is not documented … and if it represents a `fstat` failure, I don’t think it’s really necessary to handle that this way.

mtrmac (commented):
Logic WRT locking LGTM (and with the “pending writes” debate decisively settled, this is a clearly better approach than the one in #1346).

As for the API design, I’ll fully defer to actual c/storage maintainers.

pkg/ioutils/fswriters.go
alexlarsson: That is a tricky question, 99.9% of all callers don't need that Mod time so might not care about this failure. On the other hand, if this actually fails then something pretty weird is going an and maybe you still want to know.

pkg/ioutils/fswriters.go
alexlarsson: I agree that this is somewhat weird, but the primary driver of this particular API was to avoid changing the API and also not needlessly complicating the full API. It is weird, I agree, even ugly. But it is pretty effective, and unlikely to bother anyone. The alternative would be to create yet another "even more options" alternative of the function and have everything chain to that. That is clearly aesthetically better, but increases the amount of api, code and indirect calls at runtime.

pkg/ioutils/fswriters.go
alexlarsson: Is fatal now

alexlarsson:
Minor change, made the fstat() failure fatal and added some docs.

vrothberg (commented):
LGTM

pkg/ioutils/fswriters.go
mtrmac: AFAICS this error is still ignored.

mtrmac: I’m sorry, never mind, my mistake. It’s even very clearly documented.

PR #1447, opened 2022-12-12
userns: stop printing size as escaped string in error and instead use `%v`

flouthoc:
Using `%q` as format specifier for size prints non-meaningful error messages instead use `%v` which will print in default format.

Example Bad error message
```console
Error: identifier is not a container: preparing container for next step:
creating build container: creating container: the container needs a user
namespace with size '𘛋' that is bigger than the maximum value allowed
with userns=auto '𐀀'
```

See error msg here: https://github.com/containers/podman/issues/16795


giuseppe (approved):
LGTM

rhatdan:
LGTM

PR #1450, opened 2022-12-13
Use unix.Open instead of os.OpenFile for /dev/zfs

dfr:
The code is testing that its possible to open the /dev/zfs device. Using os.OpenFile does other things to the open file, including trying to set non-blocking i/o which triggers a cryptic error message on the console on FreeBSD.

I plan to fix that upstream in openzfs (https://github.com/openzfs/zfs/issues/14286) but in the meantime, using unix.Open instead of os.OpenFile avoids the problem.


rhatdan:
LGTM
@nalind @mtrmac @saschagrunert @flouthoc @umohnani8 @giuseppe @vrothberg

giuseppe (approved):
LGTM

PR #1458, opened 2023-01-02
docs: fix fcn name ref for `AssertLockedForWriting`

chenrui333:
relates to #1399


rhatdan:
LGTM
Thanks @chenrui333

Issue #1457, opened 2023-01-02
Proposal: gracefully fallback to other drivers when loading the provided one fails


dcermak:
c/storage will pick a new driver by name in https://github.com/containers/storage/blob/a747b27fe4caee6c37d4d41c087d31b0990da80a/drivers/driver.go#L332

In certain cases the situation could arise that a given driver is requested, but it is not a valid choice, e.g. the system dictates to use the zfs driver, but $HOME is on anything but zfs.

It would be simple in principle to allow for such a case and gracefully fall back to a supported driver, but that could be quite unexpected to API consumers.

Would this be something that could be added? If yes, what would you think for such a change to be the best place?

vrothberg:
> but that could be quite unexpected to API consumers.

Yes, I agree.  If the user explicitly requested a specific driver, I think that we must error out.

If the user doesn't request a specific driver, c/storage already picks a working and reasonable default.

dcermak:
So if podman would want to gracefully fall back or pick a new driver, it would be up to podman to implement such behavior then?

vrothberg:
> So if podman would want to gracefully fall back or pick a new driver, it would be up to podman to implement such behavior then?

No, Podman should error out as well.  Maybe I am misreading the proposal but if the user does a `podman --storage-driver=foobar` and podman would (silently) pick btrfs because foobar doesn't work, it would be a kind of magic that may be OK for some users but would be a break for others.

dcermak:
No, this is more for the case when e.g. the system wide driver is set to btrfs/zfs via /etc/containers/storage.conf, but your home partition is on xfs. It would be nice if podman would then just display a warning and gracefully fallback to the best working driver for rootless containers (would be overlay I guess?).

vrothberg:
> If you explicitly ask for a driver, then repicking shouldn't happen imho.

If `/etc/containers/storage.conf` has a specific driver set, then the admin or user have changed it which to me is the same as `--storage-driver` on the CLI.

@giuseppe WDYT?

dcermak:
> > If you explicitly ask for a driver, then repicking shouldn't happen imho.
> 
> If `/etc/containers/storage.conf` has a specific driver set, then the admin or user have changed it which to me is the same as `--storage-driver` on the CLI.

`/etc/containers/storage.conf` could also have a specific driver set by the distribution vendor, which is the case that I am trying to address here.

giuseppe:
[...]

PR #1460, opened 2023-01-03
Make the priority for picking the storage driver configurable

dcermak:
This fixes https://github.com/containers/storage/issues/1457

I have tested this locally by vendoring the specific PR into podman and then setting the following `~/.config/containers/storage.conf`:
```toml
[storage]

driver_priority = [
                "zfs",
                "btrfs",
                "overlay"
]
```
podman then tried zfs (but failed as I don't have any zfs partitions) and picked btrfs as the next viable candidate.

Removing the setting causes podman to correctly pick up overlay as the driver.

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -59,6 +59,11 @@ A common use case for this field is to provide a local storage directory when us
   container storage run dir (default: "/run/containers/storage")
 Default directory to store all temporary writable content created by container storage programs. The rootless runroot path supports environment variable substitutions (ie. `$HOME/containers/storage`)
 
+**driver_priority**=[]
+  Priority list for the storage drivers that will be tested one after the other to pick the storage driver if it is not defined. The first storage driver in this list that can be used, will be picked as the new one and all subsequent ones will not be tried. If all drivers in this list are not viable, then **all** known drivers will be tried and the first working one will be picked.
+By default, the storage driver is set via the `driver` option. If it is not defined, then the best driver will be picked according to the current platform. This option allows you to override this internal priority list with a custom one to prefer certain drivers.
+Setting this option only has an effect if the local storage has not been initialized yet and the driver name is not set.
+
 ### STORAGE OPTIONS TABLE
 
 The `storage.options` table supports the following options:
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -322,6 +322,7 @@ func getBuiltinDriver(name, home string, options Options) (Driver, error) {
 type Options struct {
 	Root                string
 	RunRoot             string
+	DriverPriority      []string
 	DriverOptions       []string
 	UIDMaps             []idtools.IDMap
 	GIDMaps             []idtools.IDMap
@@ -337,9 +338,18 @@ func New(name string, config Options) (Driver, error) {
 
 	// Guess for prior driver
 	driversMap := scanPriorDrivers(config.Root)
-	for _, name := range priority {
-		if name == "vfs" {
-			// don't use vfs even if there is state present.
+
+	// use the supplied priority list unless it is empty
+	prioList := config.DriverPriority
+	if len(prioList) == 0 {
+		prioList = priority
+	}
+
+	for _, name := range prioList {
+		if name == "vfs" && len(config.DriverPriority) == 0 {
+			// don't use vfs even if there is state present and vfs
+			// has not been explicitly added to the override driver
+			// priority list
 			continue
 		}
 		if _, prior := driversMap[name]; prior {
@@ -372,7 +382,7 @@ func New(name string, config Options) (Driver, error) {
 	}
 
 	// Check for priority drivers first
-	for _, name := range priority {
+	for _, name := range prioList {
 		driver, err := getBuiltinDriver(name, config.Root, config)
 		if err != nil {
 			if isDriverNotSupported(err) {
--- a/types/options.go
+++ b/types/options.go
@@ -19,6 +19,7 @@ import (
 type TomlConfig struct {
 	Storage struct {
 		Driver              string            `toml:"driver,omitempty"`
+		DriverPriority      []string          `toml:"driver_priority,omitempty"`
 		RunRoot             string            `toml:"runroot,omitempty"`
 		GraphRoot           string            `toml:"graphroot,omitempty"`
 		RootlessStoragePath string            `toml:"rootless_storage_path,omitempty"`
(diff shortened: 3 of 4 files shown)


rhatdan:
Please update the man page and make it clear the priority list is only used if graphdriver name is NOT specified.

docs/containers-storage.conf.5.md
```diff
@@ -57,7 +57,12 @@ A common use case for this field is to provide a local storage directory when us
 
 **runroot**=""
   container storage run dir (default: "/run/containers/storage")
-Default directory to store all temporary writable content created by container storage programs. The rootless runroot path supports environment variable substitutions (ie. `$HOME/containers/storage`)
+Default directory to store all temporary writable content created by container storage programs. The rootless runroot path supports environment variable substitutions (ie. `$HOME/containers/storage**)
```
rhatdan, line 60: This change looks like a mistake.

rhatdan:
One issue in the man page, once this is fixed, LGTM
@vrothberg PTAL

docs/containers-storage.conf.5.md
dcermak: Indeed, thanks for the catch!

dcermak:
One question about this change though: when invoking podman now, I **always** get the following error:
```
ERRO[0000] The storage 'driver' option must be set in /home/dan/.config/containers/storage.conf to guarantee proper operation
```
which is also suboptimal from a UX perspective. Does it really need to be an error? The driver is after all set in the store itself.

vrothberg:
> One question about this change though: when invoking podman now, I **always** get the following error:
> 
> ```
> ERRO[0000] The storage 'driver' option must be set in /home/dan/.config/containers/storage.conf to guarantee proper operation
> ```
> 
> which is also suboptimal from a UX perspective. Does it really need to be an error? The driver is after all set in the store itself.

@giuseppe @rhatdan WDYT?  I am surprised as well.  Shouldn't Podman / storage be smart enough to pick a default driver if none is set?

drivers/driver.go
```diff
@@ -329,15 +329,22 @@ type Options struct {
 }
 
 // New creates the driver and initializes it at the specified root.
-func New(name string, config Options) (Driver, error) {
+func New(name string, driverPriority []string, config Options) (Driver, error) {
```
vrothberg, line 332: This is a breaking chance.  Please move it to the `Options` struct.

types/options.go
```diff
@@ -213,10 +214,16 @@ type StoreOptions struct {
 	// RootlessStoragePath is the storage path for rootless users
 	// default $HOME/.local/share/containers/storage
 	RootlessStoragePath string `toml:"rootless_storage_path"`
-	// GraphDriverName is the underlying storage driver that we'll be
-	// using.  It only needs to be specified the first time a Store is
-	// initialized for a given RunRoot and GraphRoot.
+	// If the driver is not specified, the best suited driver will be picked
+	// either from GraphDriverPriority or from the platform dependent
```
vrothberg, line 218:
```suggestion
	// either from GraphDriverPriority, if specified, or from the platform dependent
```

docs/containers-storage.conf.5.md
```diff
@@ -59,6 +59,11 @@ A common use case for this field is to provide a local storage directory when us
   container storage run dir (default: "/run/containers/storage")
 Default directory to store all temporary writable content created by container storage programs. The rootless runroot path supports environment variable substitutions (ie. `$HOME/containers/storage`)
 
+**driver_priority**=[]
+  Priority list for the storage drivers that will be tried in case no driver is set.
```
vrothberg, line 63: "will be tried" is a bit vague.  Please be specific when a certain driver will be used and when not.

dcermak:
> > One question about this change though: when invoking podman now, I **always** get the following error:
> > ```
> > ERRO[0000] The storage 'driver' option must be set in /home/dan/.config/containers/storage.conf to guarantee proper operation
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which is also suboptimal from a UX perspective. Does it really need to be an error? The driver is after all set in the store itself.
> 
> @giuseppe @rhatdan WDYT? I am surprised as well. Shouldn't Podman / storage be smart enough to pick a default driver if none is set?

The error comes from here fyi: https://github.com/containers/storage/blob/46e94559ce9dec8bf0ba4fb04c2df4664bb728f5/types/options.go#L384 and I suspect it has been there for a while (git blame unfortunately only goes back to a commit from 2021)

docs/containers-storage.conf.5.md
dcermak: Does it make more sense now?

drivers/driver.go
```diff
@@ -337,7 +338,14 @@ func New(name string, config Options) (Driver, error) {
...
-	for _, name := range priority {
+
+	// use the supplied priority list unless it is empty
+	prioList := config.DriverPriority
+	if len(prioList) == 0 {
+		prioList = priority
+	}
+
+	for _, name := range prioList {
 		if name == "vfs" {
```
giuseppe, line 349: if vfs is set in the `config.DriverPriority` then I expect it to be honored.

I think it should be changed to `name == "vfs" && len(config.DriverPriority) == 0`

giuseppe (commented):
could we keep the error only when DriverPriority is empty (and maybe change it to a warning)?

rhatdan:
I believe at some point we forced users to specify their driver, because they kept on failing over to VFS in rootless mode, and then complained when they ran out of disk space in their homedir.

Forcing them to specify overlay and failing, caused us to know to have them install fuse-overlayfs before the kernel supported rootless overlay.

Now that rootless overlay will work for almost everyone the overlay default should work, and we might not need to require them to enter any longer.

rhatdan:
@dcermak Would like to get this in as we are about to release a new version of containers/storage.  Are you able to address the comments?

dcermak:
> could we keep the error only when DriverPriority is empty (and maybe change it to a warning)?

Given @rhatdan's comment for the historical context, I'd honestly suggest to remove this error completely. WDYT?

rhatdan:
Could we do the check iff driver is not specified and priority is not specified, then fail.

dcermak:
> Could we do the check iff driver is not specified and priority is not specified, then fail.

We do have an internal priority list as a fallback and also in `New` we try more or less every driver until one of them works: https://github.com/containers/storage/blob/fc91849352e5de23e94a01a659131355e3a31017/drivers/driver.go#L332

Thus I don't consider that warning really valid or is there a case where the user really should define a driver?

rhatdan:
I am not crazy about the internal list, since it is not easily understood by humans, so I would rather fail and tell them (or the distribution, specify one rather then having it randomly chosen potentially based on which software they have installe.

If the user or distribution specifies a search list of the driver then we should not warn.

dcermak:
@rhatdan Makes sense, I've adjusted the code accordingly, modified the error to be just a warning and tweaked the message as well to sound less scary.

rhatdan:
LGTM
@vrothberg @giuseppe PTAL

giuseppe (approved):
LGTM

PR #1464, opened 2023-01-11
Compile regex on demand not in init

rhatdan:
Every app is paying the price for these compilations on startup so make only the users of the functions pay the price.

ValidateDNSSearch is not used anywhere so remove it.

### DIFF
--- a/internal/opts/opts.go
+++ b/internal/opts/opts.go
@@ -3,13 +3,7 @@ package opts
 import (
 	"fmt"
 	"net"
-	"regexp"
 	"strings"
-)
-
-var (
-	alphaRegexp  = regexp.MustCompile(`[a-zA-Z]`)
-	domainRegexp = regexp.MustCompile(`^(:?(:?[a-zA-Z0-9]|(:?[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9]))(:?\.(:?[a-zA-Z0-9]|(:?[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])))*)\.?\s*$`)
 )
 
 // ListOpts holds a list of values and a validation function.
@@ -134,7 +128,7 @@ func (o *NamedListOpts) Name() string {
 	return o.name
 }
 
-//MapOpts holds a map of values and a validation function.
+// MapOpts holds a map of values and a validation function.
 type MapOpts struct {
 	values    map[string]string
 	validator ValidatorFctType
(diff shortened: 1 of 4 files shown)


rhatdan:
@alexlarsson @vrothberg PTAL

alexlarsson:
Maybe we can use https://github.com/charlievieth/reonce instead?

alexlarsson:
Using this it would be less work to go a wider replacement of all regexp.MustCompile global vars.

internal/opts/opts.go
```diff
@@ -222,26 +216,6 @@ func ValidateIPAddress(val string) (string, error) {
 	return "", fmt.Errorf("%s is not an ip address", val)
 }
 
-// ValidateDNSSearch validates domain for resolvconf search configuration.
```
vrothberg, line 225: Why remove the code?  Please put it in a different commit in case the code is dead.

pkg/stringid/stringid.go
```diff
@@ -19,15 +19,25 @@ import (
...
 
 var (
-	validShortID = regexp.MustCompile("^[a-f0-9]{12}$")
-	validHex     = regexp.MustCompile(`^[a-f0-9]{64}$`)
+	validShortID *regexp.Regexp
+	validHex     *regexp.Regexp
 
 	rngLock sync.Mutex
 	rng     *rand.Rand // A RNG with seeding properties we control. It can only be accessed with randLock held.
+	once    sync.Once
```
vrothberg, line 27:
```suggestion
	onceRegex    sync.Once
```

Just to add some context to the variable.

vrothberg:
> Maybe we can use https://github.com/charlievieth/reonce instead?

It doesn't look maintained.

rhatdan:
I like the idea of reonce, and would love to see that get into the mainstream package, but I am afraid to add an unmaintained package.

@alexlarsson @vrothberg @giuseppe @flouthoc @nalind @umohnani8 PTAL

ashley-cui (commented):
LGTM

TomSweeneyRedHat:
LGTM

alexlarsson:
While i agree in theory that "unmaintained" packages are bad, i'm not sure that package is actuallyt unmaintained. It got a commit in august 2022, for example. And fundamentally, the package is extremely simple, all it does is forward some calls wrapped around a call to sync.Once. I don't see why this would really need to change much.

Honestly, reimplementing reonce inside containers/common is probably going to end up being less code than open-coding all the regexes in the codebase in a similar way to this merge request.

PR #1465, opened 2023-01-12
Create pkg/regexp to better handle init regex

rhatdan:
The new Reexp structure, will handle the initialization of regexp and then will do the MustCompile within a sync.Once field.

### DIFF
--- a/pkg/regexp/regexp.go
+++ b/pkg/regexp/regexp.go
@@ -1,0 +1,214 @@
+package regexp
+
+import (
+	"io"
+	"regexp"
+	"sync"
+)
+
+// Regexp is a wrapper struct used for wrapping MustCompile regex expressions
+// used as global variables. Using this stucture helps speed the startup time
+// of apps that want to use global regex variables. This library initializes them on
+// first use as opposed to the start of the executable.
+type Regexp struct {
+	once   sync.Once
+	regexp *regexp.Regexp
+	val    string
+}
+
+func Delayed(val string) Regexp {
+	re := Regexp{
+		val: val,
+	}
+	if precompile {
+		re.regexp = regexp.MustCompile(re.val)
+	}
+	return re
+}
+
+func (re *Regexp) compile() {
+	if precompile {
+		return
+	}
+	re.once.Do(func() {
+		re.regexp = regexp.MustCompile(re.val)
+	})
+}
+
+func (re *Regexp) Expand(dst []byte, template []byte, src []byte, match []int) []byte {
+	re.compile()
+	return re.regexp.Expand(dst, template, src, match)
+}
+
+func (re *Regexp) ExpandString(dst []byte, template string, src string, match []int) []byte {
+	re.compile()
+	return re.regexp.ExpandString(dst, template, src, match)
+}
+func (re *Regexp) Find(b []byte) []byte {
+	re.compile()
+	return re.regexp.Find(b)
+}
+
+func (re *Regexp) FindAll(b []byte, n int) [][]byte {
+	re.compile()
+	return re.regexp.FindAll(b, n)
+}
+
+func (re *Regexp) FindAllIndex(b []byte, n int) [][]int {
+	re.compile()
+	return re.regexp.FindAllIndex(b, n)
+}
+
+func (re *Regexp) FindAllString(s string, n int) []string {
+	re.compile()
+	return re.regexp.FindAllString(s, n)
+}
+
+func (re *Regexp) FindAllStringIndex(s string, n int) [][]int {
+	re.compile()
+	return re.regexp.FindAllStringIndex(s, n)
+}
+
+func (re *Regexp) FindAllStringSubmatch(s string, n int) [][]string {
+	re.compile()
+	return re.regexp.FindAllStringSubmatch(s, n)
+}
+
+func (re *Regexp) FindAllStringSubmatchIndex(s string, n int) [][]int {
+	re.compile()
+	return re.regexp.FindAllStringSubmatchIndex(s, n)
+}
+
+func (re *Regexp) FindAllSubmatch(b []byte, n int) [][][]byte {
+	re.compile()
+	return re.regexp.FindAllSubmatch(b, n)
+}
+
+func (re *Regexp) FindAllSubmatchIndex(b []byte, n int) [][]int {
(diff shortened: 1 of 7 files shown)


rhatdan:
@alexlarsson @mtrmac @vrothberg @giuseppe @flouthoc @nalind PTAL

alexlarsson:
looks great to me. Should potentially be somewhere not in containers/storage though... Maybe containers/common?

giuseppe:
LGTM, I like the idea.  The only risk is that the regex using the new package can panic at any point during the program execution (potentially causing db corruptions), while before we'd catch the error immediately; but I don't see an immediate way to address it. Maybe an env variable that causes the regexp to be compiled immediately and enable it in the CI?

alexlarsson:
The reonce package has:

> The reoncetest build tag forces reonce to immediately compile regexes and panic on error. 

If we did this, the CI could enable this.

rhatdan:
@alexlarsson @giuseppe added regex_precompile build tag

giuseppe (approved):
LGTM

pkg/regexp/regexp.go
```diff
@@ -0,0 +1,210 @@
...
+
+import (
+	"io"
+	"regexp"
+	"sync"
+)
+
+// Regexp is a wrapper struct used for wrapping MustCompile regex expressions
+// used init. Using this stucture should help speed up startup time of apps
+// that want to initialize their regex at init time.
```
vrothberg, line 11:
```suggestion
// that do not want to initialize their regex at init time.
```

vrothberg (commented):
Other than the nit LGTM.

I am not sure about the precompile build tag.  If users want to precompile, they can use the stdlib directly.

pkg/regexp/regexp_dontprecompile.go
```diff
@@ -0,0 +1,6 @@
+//go:build !regexp_precompile
+// +build !regexp_precompile
+
+package regexp
+
+var precompile = false
```
alexlarsson, line 6: should this be const?

pkg/regexp/regexp.go
```diff
@@ -0,0 +1,210 @@
+package regexp
```
mtrmac, line 1: I’m not sure about using the same name as the standard library package, the two are not interoperable in any sense. OTOH there’s a certain elegance to this.

pkg/regexp/regexp.go
```diff
@@ -0,0 +1,210 @@
...
+
+import (
+	"io"
+	"regexp"
+	"sync"
+)
+
+// Regexp is a wrapper struct used for wrapping MustCompile regex expressions
+// used init. Using this stucture should help speed up startup time of apps
```
mtrmac, line 10: I’m not sure what “used init” is saying.

pkg/regexp/regexp.go
```diff
@@ -0,0 +1,210 @@
...
+		val: val,
+	}
+	if precompile {
+		re.compile()
+	}
+	return re
+}
+
+func (re *Regexp) compile() {
+	re.once.Do(func() {
```
mtrmac, line 29: Don’t do this if `precompile`?

On the one hand, the current version makes run-time profiling of the `precompile` version representative of the other one.

On the other hand, making compilation conditional here would allow an easy experiment to see if precompilation, _in total_, pays off in a particular situation.

pkg/regexp/regexp.go
```diff
@@ -0,0 +1,210 @@
...
+// Regexp is a wrapper struct used for wrapping MustCompile regex expressions
+// used init. Using this stucture should help speed up startup time of apps
+// that want to initialize their regex at init time.
+type Regexp struct {
+	once   sync.Once
+	regexp *regexp.Regexp
+	val    string
+}
+
+func New(val string) Regexp {
```
mtrmac, line 18: Name this `Delayed`, or something like that, to make it very clear at the call site that something unusual is going on and that this is not the standard library `regexp` package?

Alternatively, that could be expressed in the subpackage name.

mtrmac (commented):
Makes sense overall, some design points to perhaps discuss (but I don’t feel strongly about any of them).

pkg/regexp/regexp_dontprecompile.go
mtrmac: Yes (in both cases); this decision should have zero run-time cost.

pkg/regexp/regexp.go
mtrmac: Sure, that works great.

pkg/regexp/regexp.go
rhatdan: Ok, so you want the re.once.Do command not to be executed.

pkg/regexp/regexp.go
rhatdan: I just switched to Delayed()
I like regexp, for the conflict, to basically convince users to use it instead.  It should work fine even if the users compile the regex within a function and not a global.

pkg/regexp/regexp.go
```diff
@@ -0,0 +1,214 @@
...
+
+import (
+	"io"
+	"regexp"
+	"sync"
+)
+
+// Regexp is a wrapper struct used for wrapping MustCompile regex expressions
+// used as global variables. Using this stucture helps speed the startup time
+// of apps that want to use global regex variable. This librry initialized on
```
mtrmac, line 11:
```suggestion
// of apps that want to use global regex variables. This library initializes them on
```

And possibly “… that want to use global regex variables, especially with containers/storage/pkg/reexec.“

pkg/regexp/regexp.go
mtrmac: The only conflict I could see would be for callers of `regexp.QuoteMeta` … but that might also be something to nudge users away from.

mtrmac (commented):
LGTM; there are still some discussions going on but I’m fine with any outcome of them

pkg/regexp/regexp.go
alexlarsson: Well, for dynaminc use (i.e. create a regexp in a function) you never want the lazy stuff.

mtrmac: That’s a good point.

PR #1470, opened 2023-01-16
[1.43] unshare: fix creating a userns when running as root

giuseppe:
this check would prevent the function to ever be used when running as root, since we won't check what capabilities are currently available to the process.
    
(cherry picked from commit 1af3928e9bf16d9c39d3d60bf3ec6bb7167989a6)


rhatdan:
LGTM

PR #1482, opened 2023-01-26
test helper: maxArgs = 0 is actually a meaningful value

nalind:
We do have commands that don't take any arguments, and quietly ignoring arguments when they get them is not a good idea.

PR #1151, opened 2022-02-22
Fix error handling of a failure to create an image

mtrmac:
… found by go-staticcheck. Also avoid one other warning.

PR #1483, opened 2023-01-26
Complete "pulling up" of images in updateNames()

nalind:
When updateNames() copies an image's record from a read-only store into the read-write store, copy the accompanying data as well.

Add fields for setting data items at creation-time to LayerOptions, ImageOptions, and ContainerOptions to make this easier for us and our consumers.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -107,13 +107,13 @@ type rwContainerStore interface {
 	// stopReading releases locks obtained by startReading.
 	stopReading()
 
-	// Create creates a container that has a specified ID (or generates a
+	// create creates a container that has a specified ID (or generates a
 	// random one if an empty value is supplied) and optional names,
 	// based on the specified image, using the specified layer as its
 	// read-write layer.
 	// The maps in the container's options structure are recorded for the
 	// convenience of the caller, nothing more.
-	Create(id string, names []string, image, layer, metadata string, options *ContainerOptions) (*Container, error)
+	create(id string, names []string, image, layer string, options *ContainerOptions) (*Container, error)
 
 	// updateNames modifies names associated with a  container based on (op, names).
 	updateNames(id string, names []string, op updateNameOperation) error
@@ -651,7 +651,10 @@ func (r *containerStore) SetFlag(id string, flag string, value interface{}) erro
 }
 
 // Requires startWriting.
-func (r *containerStore) Create(id string, names []string, image, layer, metadata string, options *ContainerOptions) (container *Container, err error) {
+func (r *containerStore) create(id string, names []string, image, layer string, options *ContainerOptions) (container *Container, err error) {
+	if options == nil {
+		options = &ContainerOptions{}
+	}
 	if id == "" {
 		id = stringid.GenerateRandomID()
 		_, idInUse := r.byid[id]
@@ -662,12 +665,6 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
 	}
 	if _, idInUse := r.byid[id]; idInUse {
 		return nil, ErrDuplicateID
-	}
-	if options.MountOpts != nil {
-		options.Flags[mountOptsFlag] = append([]string{}, options.MountOpts...)
-	}
-	if options.Volatile {
-		options.Flags[volatileFlag] = true
 	}
 	names = dedupeNames(names)
 	for _, name := range names {
@@ -686,7 +683,7 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
 		Names:          names,
 		ImageID:        image,
 		LayerID:        layer,
-		Metadata:       metadata,
+		Metadata:       options.Metadata,
 		BigDataNames:   []string{},
 		BigDataSizes:   make(map[string]int64),
 		BigDataDigests: make(map[string]digest.Digest),
@@ -696,16 +693,42 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
 		GIDMap:         copyIDMap(options.GIDMap),
 		volatileStore:  options.Volatile,
 	}
+	if options.MountOpts != nil {
+		container.Flags[mountOptsFlag] = append([]string{}, options.MountOpts...)
+	}
+	if options.Volatile {
+		container.Flags[volatileFlag] = true
+	}
 	r.containers = append(r.containers, container)
-	r.byid[id] = container
-	// This can only fail on duplicate IDs, which shouldn’t happen — and in that case the index is already in the desired state anyway.
-	// Implementing recovery from an unlikely and unimportant failure here would be too risky.
+	// This can only fail on duplicate IDs, which shouldn’t happen — and in
+	// that case the index is already in the desired state anyway.
+	// Implementing recovery from an unlikely and unimportant failure here
+	// would be too risky.
 	_ = r.idindex.Add(id)
+	r.byid[id] = container
 	r.bylayer[layer] = container
 	for _, name := range names {
 		r.byname[name] = container
 	}
+	defer func() {
+		if err != nil {
+			// now that the in-memory structures know about the new
+			// record, we can use regular Delete() to clean up if
+			// anything breaks from here on out
+			if e := r.Delete(id); e != nil {
+				logrus.Debugf("while cleaning up partially-created container %q we failed to create: %v", id, e)
+			}
+		}
+	}()
 	err = r.saveFor(container)
+	if err != nil {
+		return nil, err
+	}
+	for _, item := range options.BigData {
+		if err = r.SetBigData(id, item.Key, item.Data); err != nil {
+			return nil, err
+		}
+	}
 	container = copyContainer(container)
 	return container, err
 }
--- a/images.go
+++ b/images.go
@@ -137,10 +137,10 @@ type rwImageStore interface {
 	// stopWriting releases locks obtained by startWriting.
 	stopWriting()
 
-	// Create creates an image that has a specified ID (or a random one) and
+	// create creates an image that has a specified ID (or a random one) and
 	// optional names, using the specified layer as its topmost (hopefully
 	// read-only) layer.  That layer can be referenced by multiple images.
-	Create(id string, names []string, layer, metadata string, created time.Time, searchableDigest digest.Digest) (*Image, error)
+	create(id string, names []string, layer string, options ImageOptions) (*Image, error)
 
 	// updateNames modifies names associated with an image based on (op, names).
 	// The values are expected to be valid normalized
@@ -688,7 +688,7 @@ func (r *imageStore) SetFlag(id string, flag string, value interface{}) error {
 }
 
 // Requires startWriting.
-func (r *imageStore) Create(id string, names []string, layer, metadata string, created time.Time, searchableDigest digest.Digest) (image *Image, err error) {
+func (r *imageStore) create(id string, names []string, layer string, options ImageOptions) (image *Image, err error) {
 	if !r.lockfile.IsReadWrite() {
 		return nil, fmt.Errorf("not allowed to create new images at %q: %w", r.imagespath(), ErrStoreIsReadOnly)
 	}
@@ -709,30 +709,32 @@ func (r *imageStore) Create(id string, names []string, layer, metadata string, c
 			return nil, fmt.Errorf("image name %q is already associated with image %q: %w", name, image.ID, ErrDuplicateName)
 		}
 	}
-	if created.IsZero() {
-		created = time.Now().UTC()
-	}
-
 	image = &Image{
 		ID:             id,
-		Digest:         searchableDigest,
-		Digests:        nil,
+		Digest:         options.Digest,
+		Digests:        copyDigestSlice(options.Digests),
 		Names:          names,
+		NamesHistory:   copyStringSlice(options.NamesHistory),
 		TopLayer:       layer,
-		Metadata:       metadata,
+		Metadata:       options.Metadata,
 		BigDataNames:   []string{},
 		BigDataSizes:   make(map[string]int64),
 		BigDataDigests: make(map[string]digest.Digest),
-		Created:        created,
-		Flags:          make(map[string]interface{}),
+		Created:        options.CreationDate,
+		Flags:          copyStringInterfaceMap(options.Flags),
+	}
+	if image.Created.IsZero() {
+		image.Created = time.Now().UTC()
 	}
 	err = image.recomputeDigests()
 	if err != nil {
 		return nil, fmt.Errorf("validating digests for new image: %w", err)
 	}
 	r.images = append(r.images, image)
-	// This can only fail on duplicate IDs, which shouldn’t happen — and in that case the index is already in the desired state anyway.
-	// Implementing recovery from an unlikely and unimportant failure here would be too risky.
+	// This can only fail on duplicate IDs, which shouldn’t happen — and in
+	// that case the index is already in the desired state anyway.
+	// Implementing recovery from an unlikely and unimportant failure here
+	// would be too risky.
 	_ = r.idindex.Add(id)
 	r.byid[id] = image
 	for _, name := range names {
@@ -742,7 +744,28 @@ func (r *imageStore) Create(id string, names []string, layer, metadata string, c
 		list := r.bydigest[digest]
 		r.bydigest[digest] = append(list, image)
 	}
+	defer func() {
+		if err != nil {
+			// now that the in-memory structures know about the new
+			// record, we can use regular Delete() to clean up if
+			// anything breaks from here on out
+			if e := r.Delete(id); e != nil {
+				logrus.Debugf("while cleaning up partially-created image %q we failed to create: %v", id, e)
+			}
+		}
+	}()
 	err = r.Save()
+	if err != nil {
+		return nil, err
+	}
+	for _, item := range options.BigData {
+		if item.Digest == "" {
+			item.Digest = digest.Canonical.FromBytes(item.Data)
+		}
+		if err = r.setBigData(image, item.Key, item.Data, item.Digest); err != nil {
+			return nil, err
+		}
+	}
 	image = copyImage(image)
 	return image, err
 }
@@ -965,9 +988,6 @@ func imageSliceWithoutValue(slice []*Image, value *Image) []*Image {
 
 // Requires startWriting.
 func (r *imageStore) SetBigData(id, key string, data []byte, digestManifest func([]byte) (digest.Digest, error)) error {
-	if key == "" {
-		return fmt.Errorf("can't set empty name for image big data item: %w", ErrInvalidBigDataName)
-	}
 	if !r.lockfile.IsReadWrite() {
 		return fmt.Errorf("not allowed to save data items associated with images at %q: %w", r.imagespath(), ErrStoreIsReadOnly)
 	}
@@ -975,10 +995,7 @@ func (r *imageStore) SetBigData(id, key string, data []byte, digestManifest func
 	if !ok {
 		return fmt.Errorf("locating image with ID %q: %w", id, ErrImageUnknown)
 	}
-	err := os.MkdirAll(r.datadir(image.ID), 0700)
-	if err != nil {
-		return err
-	}
+	var err error
 	var newDigest digest.Digest
 	if bigDataNameIsManifest(key) {
 		if digestManifest == nil {
@@ -989,6 +1006,18 @@ func (r *imageStore) SetBigData(id, key string, data []byte, digestManifest func
 		}
 	} else {
 		newDigest = digest.Canonical.FromBytes(data)
+	}
+	return r.setBigData(image, key, data, newDigest)
+}
+
+// Requires startWriting.
+func (r *imageStore) setBigData(image *Image, key string, data []byte, newDigest digest.Digest) error {
+	if key == "" {
+		return fmt.Errorf("can't set empty name for image big data item: %w", ErrInvalidBigDataName)
+	}
+	err := os.MkdirAll(r.datadir(image.ID), 0700)
+	if err != nil {
+		return err
 	}
 	err = ioutils.AtomicWriteFile(r.datapath(image.ID, key), data, 0600)
 	if err == nil {
--- a/layers.go
+++ b/layers.go
@@ -112,33 +112,33 @@ type Layer struct {
 	Created time.Time `json:"created,omitempty"`
 
 	// CompressedDigest is the digest of the blob that was last passed to
-	// ApplyDiff() or Put(), as it was presented to us.
+	// ApplyDiff() or create(), as it was presented to us.
 	CompressedDigest digest.Digest `json:"compressed-diff-digest,omitempty"`
 
 	// CompressedSize is the length of the blob that was last passed to
-	// ApplyDiff() or Put(), as it was presented to us.  If
+	// ApplyDiff() or create(), as it was presented to us.  If
 	// CompressedDigest is not set, this should be treated as if it were an
 	// uninitialized value.
 	CompressedSize int64 `json:"compressed-size,omitempty"`
 
 	// UncompressedDigest is the digest of the blob that was last passed to
-	// ApplyDiff() or Put(), after we decompressed it.  Often referred to
+	// ApplyDiff() or create(), after we decompressed it.  Often referred to
 	// as a DiffID.
 	UncompressedDigest digest.Digest `json:"diff-digest,omitempty"`
 
 	// UncompressedSize is the length of the blob that was last passed to
-	// ApplyDiff() or Put(), after we decompressed it.  If
+	// ApplyDiff() or create(), after we decompressed it.  If
 	// UncompressedDigest is not set, this should be treated as if it were
 	// an uninitialized value.
 	UncompressedSize int64 `json:"diff-size,omitempty"`
 
 	// CompressionType is the type of compression which we detected on the blob
-	// that was last passed to ApplyDiff() or Put().
+	// that was last passed to ApplyDiff() or create().
 	CompressionType archive.Compression `json:"compression,omitempty"`
 
 	// UIDs and GIDs are lists of UIDs and GIDs used in the layer.  This
 	// field is only populated (i.e., will only contain one or more
-	// entries) if the layer was created using ApplyDiff() or Put().
+	// entries) if the layer was created using ApplyDiff() or create().
 	UIDs []uint32 `json:"uidset,omitempty"`
 	GIDs []uint32 `json:"gidset,omitempty"`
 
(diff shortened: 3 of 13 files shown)


nalind:
(I forgot this depends on #1482.)

rhatdan:
LGTM 
Isn't this an API breaking change?

rhatdan:
@flouthoc @mtrmac @vrothberg @giuseppe @saschagrunert @umohnani8 PTAL

store.go
```diff
@@ -2962,6 +3019,13 @@ func (s *store) Image(id string) (*Image, error) {
 	if done, err := s.readAllImageStores(func(store roImageStore) (bool, error) {
 		image, err := store.Get(id)
 		if err == nil {
+			if !store.(*imageStore).lockfile.IsReadWrite() {
```
mtrmac, line 3022: I’d rather prefer `store != s.imageStore` instead of this condition. `lockfile` is shared across all users in the _process_, even across `imageStore` and `store` instances, so this implicitly relies on `getLockfile` only allowing a single lock to be only referenced read-write / read-only consistently. That’s a bit remote dependency. (And, pedantically, the Windows lock implementation always returns `IsReadWrite() == false`.)

---

Actually, this _should_ AFAICS be equivalent to `if false`, because `s.imageStore` is the first store used by `readAllImageStores`. So maybe just explicitly document that aspect of `allImageStores` and `readAllImageStores`, and point out that we rely on this here in a comment.

store.go
```diff
@@ -2117,20 +2151,43 @@ func (s *store) updateNames(id string, names []string, op updateNameOperation) e
...
-			if len(deduped) > 1 {
-				// Do not want to create image name in R/W storage
-				deduped = deduped[1:]
+			// "pull up" the image so that we can change its names list
+			options := ImageOptions{
+				Metadata:     i.Metadata,
+				CreationDate: i.Created,
+				Digest:       i.Digest,
+				Digests:      copyDigestSlice(i.Digests),
 			}
```
mtrmac, line 2169: (Note to self: I didn’t check that this copies _everything_ relevant.)

containers.go
```diff
@@ -113,7 +113,7 @@ type rwContainerStore interface {
 	// read-write layer.
 	// The maps in the container's options structure are recorded for the
 	// convenience of the caller, nothing more.
-	Create(id string, names []string, image, layer, metadata string, options *ContainerOptions) (*Container, error)
+	Create(id string, names []string, image, layer string, options *ContainerOptions) (*Container, error)
```
mtrmac, line 116: In all three C/I/L stores: Now that we are touching the method at all, maybe rename it to be lowercase. That would make it clearer that this is an internal API we can change at any time.

images.go
```diff
@@ -709,23 +709,31 @@ func (r *imageStore) Create(id string, names []string, layer, metadata string, c
...
 		BigDataNames:   []string{},
 		BigDataSizes:   make(map[string]int64),
 		BigDataDigests: make(map[string]digest.Digest),
-		Created:        created,
+		Created:        options.CreationDate,
 		Flags:          make(map[string]interface{}),
 	}
+	for _, item := range options.BigData {
+		if item.Digest == "" {
+			item.Digest = digest.Canonical.FromBytes(item.Data)
```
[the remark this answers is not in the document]
nalind: The latter, definitely.  The original level of indirection was intended to allow the callback to be supplied from above every time we needed to be able to recalculate the "right" digest, but it makes the resulting code path to hard to test.

images.go
```diff
@@ -709,23 +709,31 @@ func (r *imageStore) Create(id string, names []string, layer, metadata string, c
...
+		}
+		if err = r.setBigData(image, item.Key, item.Data, item.Digest); err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
nalind: It would, but I figured teaching the new `GarbageCollect()` method to handle it in #1484 was simpler.  I left the "incomplete" flag in there?  It's a relic of a design that predates `GarbageCollect()`, so I'll go take it out.

images.go
```diff
@@ -990,6 +998,15 @@ func (r *imageStore) SetBigData(id, key string, data []byte, digestManifest func
 	} else {
 		newDigest = digest.Canonical.FromBytes(data)
 	}
+	return r.setBigData(image, key, data, newDigest)
+}
+
+// Requires startWriting.
+func (r *imageStore) setBigData(image *Image, key string, data []byte, newDigest digest.Digest) error {
+	err := os.MkdirAll(r.datadir(image.ID), 0700)
```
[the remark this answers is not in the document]
nalind: Removing it.

images.go
```diff
@@ -990,6 +998,15 @@ func (r *imageStore) SetBigData(id, key string, data []byte, digestManifest func
 	} else {
 		newDigest = digest.Canonical.FromBytes(data)
 	}
+	return r.setBigData(image, key, data, newDigest)
+}
+
+// Requires startWriting.
+func (r *imageStore) setBigData(image *Image, key string, data []byte, newDigest digest.Digest) error {
```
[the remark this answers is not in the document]
nalind: Good point, moving it.

containers.go
nalind: No objection here, changing it.

store.go
nalind: Hmm, `NamesHistory` probably needs to be copied, given that we "hide" the read-only version if we've got a writable image record that shares the same image ID, and information about the image apparently disappearing after a "pull up" was the problem I set out to fix.  Adding it.

store.go
nalind: It shouldn't be equivalent to `if false`.  The `lockfile` for a given `imageStore` struct is initialized with its `ro` field set to `false` for a writable store, and `true` for a read-only store, and supposedly that never changes.  We depend on this code path to hide the read-only version of an image that still has a name on it that we intended to remove; one of the new tests exercises that scenario.

mtrmac: My reading of the code is
- `s.imageStore` is only ever initialized from `newImageStore` , which gets a read-write lock (and fails with an error if the same lock is already used read-only)
- Elements of `s.roImageStores` are only ever initialized from `newROImageStore`, similarly clearly read-only.
- `store.readAllImageStores` processes them in the order of `store.allImageStores`, i.e. read-write first, read-only afterwards
- So, if the image exists in the (only) writable store, we find it in the first iteration, and immediately exit the loop (by returning `done = true` from the closure).
  - We will never have the opportunity to find a duplicate image in any read-only store, i.e. the newly added code is never reached.
  - Alternatively, even if that code were reached (which cannot be the case), the second `s.imageStore.Get` of the read-write store must report an error, because we have already processed `s.imageStore` as the first item, and didn’t find the image there (and the store is still locked)

Which part is incorrect?

---

… does this hinge on the difference between `Get(id)` and `Get(image.ID)`, in a situation where the `id` parameter is not actually an ID but a name or an ID prefix?

Oh. That was totally non-obvious to me.

- primarily because the documentation of `Store.Image` just names the parameter `id` and doesn’t say anything about names — but yes, we do use that method to look up names.
- secondarily the new “no, we didn’t” comment hints at, but doesn’t _quite say_, that the first lookup was probably by a name and that name now doesn’t apply.

Do I understand the intent correctly now, or is there even more that I’m missing?

store.go
```diff
@@ -1337,13 +1367,16 @@ func (s *store) CreateImage(id string, names []string, layer, metadata string, o
 
 	var res *Image
 	err := s.writeToImageStore(func() error {
-		creationDate := time.Now().UTC()
-		if options != nil && !options.CreationDate.IsZero() {
-			creationDate = options.CreationDate
+		if options == nil {
+			options = &ImageOptions{}
 		}
```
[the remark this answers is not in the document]
nalind: Our consumers tend to expect that a function that takes a pointer won't crash if it's passed `nil`.  In keeping with that rationale, `containerStore.create()` is going to get updated, `layerStore.Put()` is correctly handling the case, and `layerStore.create()` only passes the pointer down, so it's fine.
`imageStore.Create()` doesn't take a pointer, which admittedly makes it unlike the others, and its caller has to pass it a struct, so it's doing that work.  It should use its own temporary though, as you note below.

store.go
```diff
@@ -1337,13 +1367,16 @@ func (s *store) CreateImage(id string, names []string, layer, metadata string, o
...
+		if options.CreationDate.IsZero() {
+			options.CreationDate = time.Now().UTC()
+		}
+		options.Metadata = metadata
```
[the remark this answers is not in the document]
nalind: Changing it, as above.

store.go
```diff
@@ -1469,6 +1502,7 @@ func (s *store) CreateContainer(id string, names []string, image, layer, metadat
 	if options.HostGIDMapping {
 		options.GIDMap = nil
 	}
+	options.Metadata = metadata
```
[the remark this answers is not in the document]
nalind: Changing it, as above.

store.go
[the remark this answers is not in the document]
mtrmac: I’m sorry, I’m afraid I wasn’t paying enough attention to the non-pointer parameter of `imageStre.Create()`.

The plan to update `containerStore.create` looks good.

layers.go
```diff
@@ -1687,6 +1695,10 @@ func (r *layerStore) SetBigData(id, key string, data io.Reader) error {
 	if !ok {
 		return fmt.Errorf("locating layer with ID %q to write bigdata: %w", id, ErrLayerUnknown)
 	}
+	return r.setBigData(layer, key, data)
+}
+
+func (r *layerStore) setBigData(layer *Layer, key string, data io.Reader) error {
```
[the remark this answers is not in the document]
nalind: Good point, moving it.

store.go
nalind: Yup, will add to the comment that it's a lookup-by-name case.

images.go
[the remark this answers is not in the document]
nalind: No worries, I've been playing a bit of catch-up myself.  Anyway, for some reason I was thinking we'd trend toward routinely calling `GarbageCollect()` at startup in places that, unlike cri-o in a cluster, don't delegate that work to a different process.  That may have been optimistic.  I guess I need to add an "incomplete" state to images.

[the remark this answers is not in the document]
mtrmac: I feel fairly strongly that it is necessary to add deleting data on error paths if this function fails but does not crash.

WRT deleting data on crashes, I’m notably less certain… I can imagine that being a manual process that we don’t automate in principle, and it _might_ be good enough for all I know… especially because automatic incomplete cleanup adds quite a bit of complexity to `startReading`. OTOH images have deterministic IDs, so leaving around data that is _fairly_ likely to conflict with future image creation is risky.

---

One thing to possibly consider is to not add the `BigData` option to container creation, if there there are no current known users and it’s being added only for consistency. That would avoid the need to add a `maxLayerStoreCleanupIterations`-like loop.

Or…, alternatively, making the logic in the three stores _100%_ the same (and perhaps even factoring it out) is a thing to consider. But that would increase the scope quite a bit.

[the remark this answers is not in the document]
nalind: Oh, this actually looks simpler than I had thought: if we fail to save those data items, removing the directory where we would have stored the big data items is most of the cleanup we need to do, and everything else is in-memory changes that can be undone pretty cleanly.

layers.go
```diff
@@ -1301,6 +1304,11 @@ func (r *layerStore) Put(id string, parentLayer *Layer, names []string, mountLab
 		BigDataNames:       []string{},
 		volatileStore:      moreOptions.Volatile,
 	}
+	for _, item := range moreOptions.BigData {
+		if err := r.setBigData(layer, item.Key, item.Data); err != nil {
```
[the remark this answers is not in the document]
nalind: Moving that further down, and adding a cleanup of that directory to the things we try to clean up in the anonymous deferred cleanup function.

containers.go
```diff
@@ -696,6 +696,11 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
 		GIDMap:         copyIDMap(options.GIDMap),
 		volatileStore:  options.Volatile,
 	}
+	for _, item := range options.BigData {
+		if err = r.SetBigData(id, item.Key, item.Data); err != nil {
```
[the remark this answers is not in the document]
nalind: On failure, we should be cleaning it up as best we can.  On crash, I'm not so sure.  The remnant directory is the only thing that would be left over if we failed to save the container's record.  `GarbageCollect()` is already going to catch that, so maybe we'd be better off arranging to call it from a few places.

layers.go
```diff
@@ -112,33 +112,33 @@ type Layer struct {
 	Created time.Time `json:"created,omitempty"`
 
 	// CompressedDigest is the digest of the blob that was last passed to
-	// ApplyDiff() or Put(), as it was presented to us.
+	// ApplyDiff() or put(), as it was presented to us.
```
mtrmac, line 115: `Put()` is no more (a great cleanup), so these references should probably point to `create()`, throughout.

store.go
mtrmac: Thanks!

WDYT about 
```suggestion
			if store != s.imageStore
```
here? It seems to me that the code assumes that there is only one read-write store, and that it’s `s.imageStore`, already.

layers.go
nalind: Oof, of course you're right.  Fixing.

store.go
nalind: It's not the call to a lockfile's `IsReadWrite()` method that we're using in most other places to check for writability... but it avoids a type assertion.  Changing it.

images.go
```diff
@@ -709,27 +710,58 @@ func (r *imageStore) Create(id string, names []string, layer, metadata string, c
...
-		Created:        created,
-		Flags:          make(map[string]interface{}),
+		Created:        options.CreationDate,
+		Flags:          copyStringInterfaceMap(options.Flags),
 	}
 	err = image.recomputeDigests()
 	if err != nil {
 		return nil, fmt.Errorf("validating digests for new image: %w", err)
 	}
+	defer func() {
```
mtrmac, line 735: The `containerStore` discussion applies here as well.

layers.go
```diff
@@ -1309,17 +1306,19 @@ func (r *layerStore) Put(id string, parentLayer *Layer, names []string, mountLab
...
 
 	succeeded := false
 	cleanupFailureContext := ""
 	defer func() {
 		if !succeeded {
-			// On any error, try both removing the driver's data as well
-			// as the in-memory layer record.
+			// On any error, try removing the driver's data as well as the in-memory
+			// layer record and the associated big data items..
+			if err2 := os.RemoveAll(r.datadir(id)); err2 != nil && !errors.Is(err2, os.ErrNotExist) {
```
mtrmac, line 1316: `r.Delete` already removes that AFAICS; is doing this separately necessary?

images.go
```diff
@@ -709,27 +710,58 @@ func (r *imageStore) Create(id string, names []string, layer, metadata string, c
...
+			}
+			for _, name := range names {
+				delete(r.byname, name)
+			}
+			delete(r.byid, id)
+			r.idindex.Delete(id)
+			r.images = imageSliceWithoutValue(r.images, image)
+		}
+	}()
+	for _, item := range options.BigData {
```
mtrmac, line 757: Absolutely non-blocking: A thing that _might_ make sense here, because unlike containers, images often have deterministic IDs, is to `os.RemoveAll(r.datadir(id))` here; that would ensure only the requested data is written.

OTOH it’s an extra syscall, and I’m not sure we ever write a `BigData` item that we wouldn’t want to write on a repeated attempt to create the same image, so I’m not all sure it would be useful.

If users are instructed to trigger `GarbageCollect` frequently enough, the `RemoveAll` wouldn’t buy us much.

rhatdan:
@nalind Still working on this one?

rhatdan:
Do you think this will be ready before podman 4.5?

store.go
```diff
@@ -562,6 +562,15 @@ type LayerOptions struct {
 	UncompressedDigest digest.Digest
 	// True is the layer info can be treated as volatile
 	Volatile bool
+	// BigData is a set of items which should be stored with the layer.
+	BigData []LayerBigDataOption
+	// Flags is a set of named flags and their values to store with the layer.
+	Flags map[string]interface{}
```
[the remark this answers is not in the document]
nalind: This doesn't go as far as making them generally editable, just settable at the time the record is created.  There have been points where we've found ourselves wanting to be able to attach bits of information to records of one type or another, where coordinating them in the one general `Metadata` field would have been too much overhead, and this could have been enough.

Adding notes to the various options fields to note that flags can only be set at create-time, but that could change at some point.

images.go
nalind: Same.

containers.go
```diff
@@ -696,6 +699,26 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
...
+			if e := os.RemoveAll(r.datadir(container.ID)); e != nil && !errors.Is(e, os.ErrNotExist) {
+				logrus.Debugf("while cleaning up data directory for container %q we failed to create: %v", id, e)
+			}
+			for _, name := range names {
+				delete(r.byname, name)
+			}
+			delete(r.bylayer, layer)
+			r.idindex.Delete(id)
+			delete(r.byid, id)
+			r.containers = containerSliceWithoutValue(r.containers, container)
```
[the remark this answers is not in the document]
nalind: Hmm, I like using `Delete()` as a cleanup.  Switching to doing that, and to set `Flags` in the just-created container structure rather than modifying a passed-in `options.Flags` map and then copying the whole map into the just-created container structure.

In order to correctly clean up indexes that it has already modified, it looks like `layerStore.PutAdditionalLayer()` should have been using `Delete()` for cleanup as well, changing it.

store.go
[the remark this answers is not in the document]
mtrmac: Documenting the flag limitations LGTM.

containers.go
```diff
@@ -696,20 +693,57 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
...
+	for _, item := range options.BigData {
+		if err = r.SetBigData(id, item.Key, item.Data); err != nil {
+			return nil, err
+		}
+	}
 	container = copyContainer(container)
 	return container, err
 }
 
+func containerSliceWithoutValue(slice []*Container, value *Container) []*Container {
```
mtrmac, line 736: (This is now unused, and that breaks CI.)

containers.go
```diff
@@ -696,20 +693,57 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
...
 	r.bylayer[layer] = container
 	for _, name := range names {
 		r.byname[name] = container
 	}
+	defer func() {
+		if err != nil {
+			// now that the in-memory structures know about the new
+			// record, we can use regular Delete() to clean up if
+			// anything breaks from here on out
+			if e := r.Delete(container.ID); e != nil && !errors.Is(e, os.ErrNotExist) {
```
mtrmac, line 718: The error from `Delete` on no match (which should never happen?) is `ErrContainerUnknown`, and I can’t see how it could fail with `ErrNotExist` (`RemoveAll` should not return that).

images.go
```diff
@@ -742,7 +745,28 @@ func (r *imageStore) Create(id string, names []string, layer, metadata string, c
 		list := r.bydigest[digest]
 		r.bydigest[digest] = append(list, image)
 	}
+	defer func() {
+		if err != nil {
+			// now that the in-memory structures know about the new
+			// record, we can use regular Delete() to clean up if
+			// anything breaks from here on out
+			if e := r.Delete(image.ID); e != nil && !errors.Is(e, os.ErrNotExist) {
```
mtrmac, line 753: The error from `Delete` is `ErrImageUnknown`, and I can’t see how it could fail with `ErrNotExist` (`RemoveAll` should not return that).

layers.go
```diff
@@ -1200,16 +1197,19 @@ func (r *layerStore) PutAdditionalLayer(id string, parentLayer *Layer, names []s
 		r.byuncompressedsum[layer.UncompressedDigest] = append(r.byuncompressedsum[layer.UncompressedDigest], layer.ID)
 	}
 	if err := r.saveFor(layer); err != nil {
-		if err2 := r.driver.Remove(id); err2 != nil {
-			logrus.Errorf("While recovering from a failure to save layers, error deleting layer %#v: %v", id, err2)
+		if e := r.Delete(layer.ID); e != nil && !errors.Is(e, os.ErrNotExist) {
```
mtrmac, line 1200: The error from `Delete` is ErrLayerUnknown .

layers.go
```diff
@@ -1295,128 +1296,127 @@ func (r *layerStore) Put(id string, parentLayer *Layer, names []string, mountLab
...
-				}
-				logrus.Errorf("While recovering from a failure (%s), error deleting layer %#v: %v", cleanupFailureContext, layer.ID, err2)
+		if err != nil {
+			// now that the in-memory structures know about the new
+			// record, we can use regular Delete() to clean up if
+			// anything breaks from here on out
+			if cleanupFailureContext == "" {
+				cleanupFailureContext = "unknown: cleanupFailureContext not set at the failure site"
+			}
+			if e := r.Delete(layer.ID); e != nil && !errors.Is(e, os.ErrNotExist) {
```
mtrmac, line 1327: The error from `Delete` if a layer is not found (which should not happen?) is `ErrLayerUnknown`, and I can’t see what is expected to return `ErrNotExist`.

Is this targeting `driver.Remove()`? And if so, should that use `os.IsNotExist` (or its modern replacement)?

layers.go
```diff
@@ -1295,128 +1296,127 @@ func (r *layerStore) Put(id string, parentLayer *Layer, names []string, mountLab
...
+				logrus.Debugf("cleaning up partially-created layer %q we failed to create (%s): %v", id, cleanupFailureContext, e)
```
mtrmac, line 1328: Non-blocking I guess: The `Error`-level logging here was added to allow diagnosing an unexpected failure/inconsistency retroactively, from system logs, because historically finding the root cause of “layer layer exists in metadata, but not on the filesystem” failures has been very hard for me; so I’d prefer this to remain quite loud.

layers.go
```diff
@@ -1295,128 +1296,127 @@ func (r *layerStore) Put(id string, parentLayer *Layer, names []string, mountLab
...
-			cleanupFailureContext = "creating a layer from template"
+		if err = r.driver.CreateFromTemplate(id, moreOptions.TemplateLayer, templateIDMappings, parent, parentMappings, &opts, writeable); err != nil {
+			cleanupFailureContext = fmt.Sprintf("creating a layer from template layer %q", moreOptions.TemplateLayer)
 			return nil, -1, fmt.Errorf("creating copy of template layer %q with ID %q: %w", moreOptions.TemplateLayer, id, err)
 		}
 		oldMappings = templateIDMappings
 	} else {
 		if writeable {
-			if err := r.driver.CreateReadWrite(id, parent, &opts); err != nil {
+			if err = r.driver.CreateReadWrite(id, parent, &opts); err != nil {
```
mtrmac, line 1360: I’d much prefer for the `succeeded` variable to remain, because it’s less error-prone exactly WRT exactly these kinds of changes.

Someone not familiar with the whole of the function making a local edit would quite likely, by habit, do `if err := …; err != nil { return … err }`, because that form with `:=` is typically the one that works regardless of the surrounding code.

We have quite a few error paths, and getting all of them always correct seems harder to me than handling the one isolated success path specially.

I’d weakly prefer using the `succeeded` in the container+image stores as well, for the same general reason, although the case for that is admittedly much weaker in those simpler functions.

layers.go
```diff
@@ -1759,6 +1763,9 @@ func (r *layerStore) tspath(id string) string {
 // layerHasIncompleteFlag returns true if layer.Flags contains an incompleteFlag set to true
 // The caller must hold r.inProcessLock for reading.
 func layerHasIncompleteFlag(layer *Layer) bool {
+	if layer.Flags == nil {
+		return false
+	}
 	// layer.Flags[…] is defined to succeed and return ok == false if Flags == nil
```
mtrmac, line 1769: Non-blocking nit: If the condition is added explicitly above, this comment can be dropped. (I don’t care at all which one remains, just that having both is unnecessary.)

mtrmac (commented):
Overall logic looks good.

I don’t know what the `Delete … ErrNotExist` logic is looking for.

containers.go
nalind: Oof, dropping it.

containers.go
nalind: Right, removing that `errors.Is()` call.

images.go
nalind: Right, removing that `errors.Is()` call.

layers.go
nalind: Right, removing that `errors.Is()` call.

layers.go
nalind: `Remove()` can return, well, any error.  But we can check for a not-exist error right where we call `Remove()`, so moving this check there.

layers.go
nalind: Changing it back to `Errorf`.

layers.go
nalind: Oh, I missed that this wasn't using named result values like `create()` does for images and containers, which would have ensured that the deferred cleanup function always sees the `err` value that's being returned from `create()`.  Adding names for them.

layers.go
nalind: Dropping the comment, then.

layers.go
mtrmac: Good idea, that’s more robust.

layers.go
mtrmac: Named returns only ensure that the `defer` sees the error value if:
- the error is written into a variable with the expected name (usually true), and
- it happens in the top-level scope of the function. Compare https://github.com/containers/storage/pull/1151 .

Also see the other new review comment…

layers.go
```diff
@@ -1200,16 +1197,19 @@ func (r *layerStore) PutAdditionalLayer(id string, parentLayer *Layer, names []s
...
+			logrus.Errorf("While recovering from a failure to save layers, error deleting layer %#v: %v", id, e)
 		}
 		return nil, err
 	}
 	return copyLayer(layer), nil
 }
 
 // Requires startWriting.
-func (r *layerStore) Put(id string, parentLayer *Layer, names []string, mountLabel string, options map[string]string, moreOptions *LayerOptions, writeable bool, flags map[string]interface{}, diff io.Reader) (*Layer, int64, error) {
+func (r *layerStore) create(id string, parentLayer *Layer, names []string, mountLabel string, options map[string]string, moreOptions *LayerOptions, writeable bool, diff io.Reader) (layer *Layer, size int64, err error) {
```
mtrmac, line 1209: I completely forgot about that, but making `layer` a named return value apparently breaks the `defer` cleanup handler, because `return nil, -1, …` means “set `layer = nil`; then execute the `defer` handler”, and the handler will crash on trying to access `layer.ID`.

This was IIRC surprising to me, and a bit hard to find, in https://github.com/containers/storage/commit/ce3c7ffbd2e644f86d74cc7a51a35adb20fca6ed .

So I have developed quite an aversion to named return values…

Demo:
```go
package main

import "fmt"

type object struct{}

func unnamed() *object {
	i := &object{}
	defer func() { fmt.Printf("unnamed defer: %v\n", i) }()
	return nil
}

func named() (i *object) {
	i = &object{}
	defer func() { fmt.Printf("named defer: %v\n", i) }()
	return nil
}

func main() {
	fmt.Printf("unnamed return: %v\n", unnamed())
	fmt.Printf("named return: %v\n", named())
}
```

containers.go
```diff
@@ -696,16 +693,42 @@ func (r *containerStore) Create(id string, names []string, image, layer, metadat
...
 	r.bylayer[layer] = container
 	for _, name := range names {
 		r.byname[name] = container
 	}
+	defer func() {
+		if err != nil {
+			// now that the in-memory structures know about the new
+			// record, we can use regular Delete() to clean up if
+			// anything breaks from here on out
+			if e := r.Delete(container.ID); e != nil {
```
mtrmac, line 718: This won’t work with `container` a named return value.

images.go
```diff
@@ -742,7 +744,28 @@ func (r *imageStore) Create(id string, names []string, layer, metadata string, c
 		list := r.bydigest[digest]
 		r.bydigest[digest] = append(list, image)
 	}
+	defer func() {
+		if err != nil {
+			// now that the in-memory structures know about the new
+			// record, we can use regular Delete() to clean up if
+			// anything breaks from here on out
+			if e := r.Delete(image.ID); e != nil {
```
mtrmac, line 752: This won’t work with `image` a named return value.

layers.go
nalind: Fortunately we have a local variable with the same value in it (we even use it in the error message on the next line, doh), so the change needed isn't a large one.

images.go
nalind: Fortunately we have a local variable with the same value in it (we even use it in the error message on the next line, doh), so the change needed isn't a large one.

layers.go
mtrmac: > it'll see the results that are being returned.

Oops, so I was arguing _exactly the opposite_ of my claim in this review comment in the other review comment.

I’m sorry about that.

mtrmac (approved):
LGTM.

PR #1487, opened 2023-01-27
idmap: new package

giuseppe:
move the idmap utils from the overlay driver to a separate package.

### DIFF
--- a/drivers/overlay/check.go
+++ b/drivers/overlay/check.go
@@ -12,6 +12,7 @@ import (
 	"syscall"
 
 	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/idmap"
 	"github.com/containers/storage/pkg/idtools"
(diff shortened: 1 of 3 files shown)


rhatdan (approved):
LGTM

rhatdan:
@nalind @vrothberg @saschagrunert @flouthoc @umohnani8 @mtrmac PTAL

mtrmac (commented):
Nothing here says why this is being done. So, why? (This is not a “no”, this is “in 5 years someone might well be asking what we were thinking today”.)

---

Does this need to be a public stable API of c/storage, or would `internal/idmap` do?

giuseppe:
I'd like to use it from Podman, so it needs to be a public API.  My WIP on Podman is currently here: https://github.com/giuseppe/libpod/tree/idmap-rootfs

mtrmac:
Thanks for the explanation. LGTM.

PR #1489, opened 2023-01-30
pkg/idmap: add boilerplate for unsupported

giuseppe:
we need it to build on freebsd.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #1499, opened 2023-02-01
CI: check that cross-compilation works

nalind:
Try to avoid the need for PRs like #1489 by catching them ahead of time.


edsantiago:
Hardly seems worth a ask-map for this repo, but here ya go 
cirrus-map-pr1499

rhatdan:
LGTM

PR #1501, opened 2023-02-03
Migrate from github.com/tchap/go-patricia to github.com/tchap/go-patricia/v2

nalind:
The API is the same, but it's friendlier to how Go modules like to do versioning.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #1504, opened 2023-02-08
[release-1.38] Make random value generation and use more reliable

TomSweeneyRedHat:
[release-1.38] Make random value generation and use more reliable

Backport the commits in https://github.com/containers/storage/pull/1065
to address https://bugzilla.redhat.com/show_bug.cgi?id=2166429

Make sure the values returned by newLastWriterID() are unique
... by including the time, PID, and a per-process counter.

Pad the rest with rando values for continuity.

This feels a bit like overkill, OTOH adding the PID might be
useful for debugging, so there's that.

Signed-off-by: Miloslav Trmač <mitr@redhat.com>
Signed-off-by: tomsweeneyredhat <tsweeney@redhat.com>


TomSweeneyRedHat:
Based on discussions in https://bugzilla.redhat.com/show_bug.cgi?id=2166429#c4, it appears this fix is not needed this far back, especially since it's having issues.  Closing.  Thanks @mtrmac for the sleuthing!

PR #1509, opened 2023-02-09
Stop using golang.org/x/net/context

mtrmac:
As of Go >= 1.9, the implementation of the used parts of x/net/context just forwards to the standard library.

So, use the standard library directly, and avoid the x/net dependency entirely. This should save a tiny bit of space in users like Podman.


rhatdan:
LGTM
Woohoo shrinkage.

mtrmac:
It’s like 30 kB of the object file, i.e much less in the whole binary… the big gains are elsewhere. I just noticed that we are actually linking this compatibility shim…

vrothberg (approved):
LGTM, nice!

PR #1510, opened 2023-02-16
options: reject empty runroot or graphroot

giuseppe:
make sure that the runroot and graphroot are not empty otherwise we will end up using an empty directory as path prefix and create files in the current directory.

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -152,20 +152,24 @@ func defaultStoreOptionsIsolated(rootless bool, rootlessUID int, storageConf str
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

TomSweeneyRedHat:
LGTM

rhatdan:
Does Podman work if there is no storage.conf at all?

giuseppe:
> Does Podman work if there is no storage.conf at all?

if there is no storage at all, we use the default settings that are non empty strings.

PR #1514, opened 2023-02-22
Don't hard code SELinux labels into code

rhatdan:
These labels can be changed based on installed policy.

I am working on allowing containers within containers, but this will require different SELinux label types other then container_file_t.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -47,8 +47,7 @@ var (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


giuseppe (commented):
LGTM

rhatdan:
@flouthoc @nalind @mtrmac @umohnani8 @saschagrunert @vrothberg PTAL

vrothberg (approved):
LGTM

Issue #1511, opened 2023-02-21
Cannot run podman after upgrading to 4.4.1


mfocko:
### Issue Description

I've upgraded my VPS and after the reboot I could not start any of the podman containers.

It was an upgrade from `4.2.1 → 4.4.1`.

Posibbly related to containers/podman#15641 and containers/podman#17230

### Steps to reproduce the issue

Steps to reproduce the issue
1. run any podman command

### Describe the results you received

    Error: filesystem type 0x13372fc1 reported for /var/lib/containers/storage is not supported with 'overlay': backing file system is unsupported for this graph driver

### Describe the results you expected

podman running correctly

### podman info output

```yaml
host:
  arch: amd64
  buildahVersion: 1.27.0
  cgroupControllers: []
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.1.5-1.fc37.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.5, commit: '
  cpuUtilization:
    idlePercent: 97.02
    systemPercent: 0.29
    userPercent: 2.69
  cpus: 8
  distribution:
    distribution: fedora
    version: "37"
  eventLogger: journald
  hostname: poincare
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 13001
      size: 1
    - container_id: 2
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 13001
      size: 1
    - container_id: 2
      host_id: 100000
      size: 65536
  kernel: 5.10.159.2
  linkmode: dynamic
  logDriver: journald
  memFree: 3479576576
  memTotal: 4294967296
  networkBackend: cni
  ociRuntime:
    name: crun
    package: crun-1.8-1.fc37.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.8
      commit: 0356bf4aff9a133d655dc13b1d9ac9424706cac4
      rundir: /run/user/1000/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +WASM:wasmedge +YAJL
  os: linux
  remoteSocket:
    exists: true
    path: /run/user/1000/podman/podman.sock
  security:
    apparmorEnabled: false
[...]

PR #1512, opened 2023-02-22
overlay: allow unknown backing fs with mountProgram

giuseppe:
if a mountProgram is specified, let it deal with an unknown backing file system instead of failing early.  The error is kept when we use native overlay.

Closes: https://github.com/containers/storage/issues/1511

PR #1515, opened 2023-02-23
[1.45] overlay: allow unknown backing fs with mountProgram

giuseppe:
if a mountProgram is specified, let it deal with an unknown backing file system instead of failing early.  The error is kept when we use native overlay.

Closes: https://github.com/containers/storage/issues/1511

(cherry picked from commit b9f1697c1714f4ff2141ef3f308154f3968e6fda)

backport of: https://github.com/containers/storage/pull/1512

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -314,7 +314,10 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 	}
 	fsName, ok := graphdriver.FsNames[fsMagic]
 	if !ok {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
This looks like it should be in ZeroDay?  Thoughts?  If so, we need a BZ, if not, then this should be pulled for now, as this branch is frozen otherwise.

TomSweeneyRedHat:
and oops, just saw the comment in #1512.  WIll get a BZ spun up.

PR #1523, opened 2023-02-27
chunked: report the correct size

giuseppe:
report the correct size for a layer that was pulled using the chunked package.


rhatdan (approved):
LGTM

saschagrunert (approved):

PR #1529, opened 2023-03-09
Fix typos in error messages

nalind:
Two error messages suggest that podman-system-migrate is a binary that can be run, when the command is "podman system migrate".


saschagrunert (approved):

rhatdan:
LGTM

PR #1432, opened 2022-11-14
overlay: allow overlaying on top of zfs

BtbN:
OpenZFS works fine as backing fs for overlayfs after the following commits:
https://github.com/openzfs/zfs/commit/e015d6cc0b60d4675c9b6d2433eed2c8ef0863e8
https://github.com/openzfs/zfs/commit/86db35c447aa3f4cc848497d78d54ec9c985d1ed
https://github.com/openzfs/zfs/commit/dbf6108b4df92341eea40d0b41792ac16eabc514

Right now this only works with OpenZFS master, but it will find its way to 2.2 eventually, and right now testing it is entirely blocked by this check.

PR #1535, opened 2023-03-15
Cirrus: Migrate Ubuntu to Debian CI VM Images

cevich:
Ref: https://github.com/containers/automation_images/pull/267

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,15 +17,15 @@ env:
     ####
     #### Cache-image names to test with (double-quotes around names are critical)
     ###
-    FEDORA_NAME: "fedora-37" ### 20230120t152650z-f37f36u2204
-    UBUNTU_NAME: "ubuntu-2204" ### 20230120t152650z-f37f36u2204
+    FEDORA_NAME: "fedora-37"
+    DEBIAN_NAME: "debian-12"
 
     # GCE project where images live
     IMAGE_PROJECT: "libpod-218412"
     # VM Image built in containers/automation_images
-    IMAGE_SUFFIX: "c20230120t152650z-f37f36u2204"
+    IMAGE_SUFFIX: "c20230330t153101z-f37f36d12"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
-    UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
+    DEBIAN_CACHE_IMAGE_NAME: "debian-${IMAGE_SUFFIX}"
 
     ####
     #### Command variables to help avoid duplication
@@ -67,7 +67,7 @@ linux_testing: &linux_testing
     always:
         df_script: '${_DFCMD} || true'
         rh_audit_log_script: '${_RAUDITCMD} || true'
-        ubuntu_audit_log_script: '${_UAUDITCMD} || true'
+        debian_audit_log_script: '${_UAUDITCMD} || true'
         journal_log_script: '${_JOURNALCMD} || true'
 
 
@@ -95,13 +95,13 @@ fedora_testing_task: &fedora_testing
 
 
 # aufs was dropped between 20.04 and 22.04, can't test it
-ubuntu_testing_task: &ubuntu_testing
+debian_testing_task: &debian_testing
     <<: *linux_testing
-    alias: ubuntu_testing
+    alias: debian_testing
     name: *std_test_name
     env:
-        OS_NAME: "${UBUNTU_NAME}"
-        VM_IMAGE: "${UBUNTU_CACHE_IMAGE_NAME}"
+        OS_NAME: "${DEBIAN_NAME}"
+        VM_IMAGE: "${DEBIAN_CACHE_IMAGE_NAME}"
     # Not all $TEST_DRIVER combinations valid for all $VM_IMAGE types.
     matrix:
         - env:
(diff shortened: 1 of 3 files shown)


cevich:
Just when I thought I was done building CI VM images, it seems [Debian is missing XFS utils.](https://cirrus-ci.com/task/6146873996607488) :thinking:

cevich:
@rhatdan & @nalind PTAL when you have a moment.  I disabled the ZFS tests due to Debian image-build complications.  If you think that's something we really need, I can fuss with the images more.

rhatdan:
LGTM
I will let @nalind comment on the ZFS

TomSweeneyRedHat:
LGTM and would like a @nalind head nod

nalind:
Yes, I would like to continue testing zfs.  We ship the code, people try to use it, we only recently did the work to start testing it in CI, and I don't want to sacrifice that improvement in order to make a different one.

The additional setup steps seem to look more or less like:
```bash
sed -i -r 's/^(deb.*)/\1 contrib/g' /etc/apt/sources.list
apt-get update
apt-get install zfsutils zstd
```

That drags in zfs-dkms, which builds the kernel module.

cevich:
Thanks for researching the needed steps nalin.  I'll incorporate those into the VM Image build scripts, and crank out a set of new images for this.

cevich:
@nalind any idea why the module isn't loading? It's [definitely being installed](https://cirrus-ci.com/task/6401599514345472?logs=main#L2191) (log).  Is `/lib/modules/6.1.0-7-amd64/updates/dkms/` the correct path for `modprobe` to find it?

nalind:
Is there something odd going on with the handling of the kernel in this VM image?  [The log](https://cirrus-ci.com/task/6401599514345472?logs=main#L1696) and `dkms status` says that the zfs module was built for `6.1.0-7-amd64`, but when I bring it up, `uname -r` says it's running `6.1.0-7-cloud-amd64`.

cevich:
Oh! Good catch!  I wonder if that's some special kernel google placed.

```
    debian: Module build for kernel 6.1.0-7-cloud-amd64 was skipped since the
    debian: kernel headers for this kernel do not seem to be installed.
```

Maybe installing the `cloud` kernel headers is what's needed?  I can give that a try.

cevich:
@nalind Thanks for help with the kernel module. I think I got that fixed now and the ZFS tests look to be running much better.  There's just [a single test failure](https://cirrus-ci.com/task/6310059769069568?logs=build_and_test#L105-L108).  Any idea what might be the trouble?

```
[+1006s] --- FAIL: TestOverlayEcho (0.24s)
[+1006s]     graphtest_unix.go:508: unlinkat /tmp.PSLnHvVlHo/tmp/storage-graphtest-336150358/overlay/3ee87b84f68a559f8458e08dc743e3015e027264d5bf40f13c332a4254062ee7/merged/subdir1/subdir2/subdir3/subdir4/subdir5/subdir6/subdir7/subdir8/subdir9: invalid argument
[+1006s] FAIL
[+1006s] FAIL	github.com/containers/storage/drivers/overlay	11.888s
```

nalind:
Reproducing it, `journalctl` shows me `overlayfs: upper fs does not support RENAME_WHITEOUT.` being logged at about the same time.
We stopped rejecting overlay-on-zfs outright, but https://github.com/containers/storage/pull/1432#issue-1448290513 notes that the support that overlay needs from zfs is not present in the current 2.1.9 release, and is instead slated to be in a future 2.2.x release.  We need to start exercising this functionality as part of checking if an underlying filesystem is suitable for use with overlay.  Opened #1545 to do so.

cevich:
@nalind it looks like all the tests are passing, but the final cleanup has some fart:

```
[+1151s] cannot destroy 'tmp.mNwHJCa6kq/tmp': dataset is busy
```

nalind:
Cherry-picking the tip of https://github.com/containers/storage/pull/1548 should sort that out.

cevich:
Thanks again Nalin.

cevich:
@nalind I rebased to bring in #1548 but [the tests are still complaining](https://cirrus-ci.com/task/5955236884381696): `[+1158s] cannot destroy 'tmp.U8HgWAHExr/tmp': dataset is busy` :cry: Is this a real failure or maybe something we can just ignore because it's CI?

nalind:
There's a leftover `sleep 1000s` from the device-mapper driver's `TestDevmapperMountLeaks()` test holding the zfs dataset mount open in a different moiunt namespace.  I guess that process can be killed, which cleans up the namespace and its mounts, pretty quickly, by adding a
```bash
kill $(lsns -J -t mnt --output-all | jq '.namespaces[]|select(.command=="sleep 1000s").pid')
```
or more politely, by unmounting the dataset from the namespace, by adding
```
umount -l /proc/$(lsns -J -t mnt --output-all | jq '.namespaces[]|select(.command=="sleep 1000s").pid')/root/"$zpool/tmp"
```
to the cleanup logic.  The latter still takes a couple of seconds before zfs will accept that the dataset is no longer busy.

cevich:
@nalind the VFS task is throwing the same error.  Did I put [the fix in the wrong place](https://github.com/containers/storage/pull/1535/files#diff-0ca52cf2d14756ee50e1b99c5b46dfc1c45a1a796746ebe6a6f406055dba929e)?

nalind:
Yeah, the `teardown()` function is by individual `bats` tests, which is how we're running integration tests.
The lingering process is spawned when `go test` is in `drivers/devmapper`, and we're running the unit tests after the integration tests.
I'd suggest adding the extra bit of cleanup around line 73 of `contrib/cirrus/build_and_test.sh`.

cevich:
Yay \o/

All green now :smile:

rhatdan:
LGTM

PR #1538, opened 2023-03-22
fix(quota): added missed ClearQuota method to fix go-builds without cgo

distorhead:
Fixed building of the containers/storage for linux without usage of CGO (tags="linux exclude_disk_quota" & CGO_ENABLED=0): added missing method for the stub Controller implementation.

Refs https://github.com/containers/storage/commit/891593970ae79687f139871f2ae105582060a04b.


rhatdan:
Thanks @distorhead

PR #1545, opened 2023-03-30
supportsOverlay: ensure that whiteout creation works

nalind:
In newer kernel versions, the kernel overlay driver will attempt to atomically create whiteout when items are removed from merged directories.
We need to be able to detect when the underlying filesystem isn't able to do that at startup-time (in this case, zfs before its 2.2 release), so that we don't end up thinking we're fine, but mounting writeable layers which reject attempts to delete things.


TomSweeneyRedHat:
LGTM
but would like a head nod from @rhatdan @giuseppe @flouthoc  or @vrothberg

flouthoc (approved):
LGTM

Issue #1546, opened 2023-03-31
Podman v4.3.1 to 4.4.4 introduced breaking change for unsupported filesystems


tis-rpage:
### Issue Description

On 4.3.1 and earlier, unsupported filesystems, such as `0x53464846` for WSLFS would not abort the command and would complete the action.

### Steps to reproduce the issue

Steps to reproduce the issue
1. Run an unsupported filesystem, like WSL1 WLFS
2. Install podman 4.3.1
3. sudo podman images
4. view successful image list
5. Install podman 4.4.4
6. sudo podman images
7. view nothing, because error handling aborted on the unsupported filesystem that used to be functional yet unsupported

### Describe the results you received

```
podman images
Error: filesystem type 0x53464846 reported for /var/lib/containers/storage is not supported with 'overlay': backing file system is unsupported for this graph driver
```

### Describe the results you expected

```
podman images
REPOSITORY                                                        TAG                                           IMAGE ID      CREATED        SIZE
<none>                                                            <none>                                        b177afe92189  7 months ago   559 MB
<none>                                                            <none>                                        10a9f9f6f494  7 months ago   559 MB
quay.io/keycloak/keycloak                                         19.0.1                                        49756595a9cb  7 months ago   559 MB
gcr.io/k8s-minikube/kicbase                                       v0.0.33                                       b7ab23e98277  8 months ago   1.15 GB
```

### podman info output

```yaml
podman info --log-level=trace
INFO[0000] podman filtering at log level trace
DEBU[0000] Called info.PersistentPreRunE(podman info --log-level=trace)
DEBU[0000] Using conmon: "/usr/bin/conmon"
DEBU[0000] Initializing boltdb state at /var/lib/containers/storage/libpod/bolt_state.db
DEBU[0000] Using graph driver overlay
DEBU[0000] Using graph root /var/lib/containers/storage
DEBU[0000] Using run root /run/containers/storage
DEBU[0000] Using static dir /var/lib/containers/storage/libpod
DEBU[0000] Using tmp dir /run/libpod
DEBU[0000] Using volume path /var/lib/containers/storage/volumes
DEBU[0000] Using transient store: false
DEBU[0000] Set libpod namespace to ""
DEBU[0000] [graphdriver] trying provided driver "overlay"
DEBU[0000] Unknown filesystem type 0x53464846 reported for /var/lib/containers/storage
[...]

PR #1547, opened 2023-03-31
overlay: change error to debug log for unsupported fs

giuseppe:
The overlay driver previously raised an error when encountering an unsupported filesystem. This commit changes the error message to a debug log, allowing the overlay driver to continue its operation even with unsupported filesystems, without causing a failure.

[NO NEW TESTS NEEDED]

Introduced-by: https://github.com/containers/storage/pull/1374

Closes: https://github.com/containers/storage/issues/1546

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -314,9 +314,6 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 	}
 	fsName, ok := graphdriver.FsNames[fsMagic]
 	if !ok {
-		if opts.mountProgram == "" {
-			return nil, fmt.Errorf("filesystem type %#x reported for %s is not supported with 'overlay': %w", fsMagic, filepath.Dir(home), graphdriver.ErrIncompatibleFS)
-		}
 		fsName = "<unknown>"
 	}
 	backingFs = fsName


rhatdan:
Won't overlay break?  Or is there a check later on?

mtrmac:
Could we, for the record / future readers, describe a bit more what “unsupported” means here?
- Is it that some filesystems definitely won’t work, and we shouldn’t even try?
- Is it that only a small well-enumerated (although possibly changing) list can ever work, and most others are hopeless?
- Is it that only some of them are likely tested, and others might work but are not recommended?
- Is it that product vendors might make a commercial decision to test/support only some combinations? (In that case those vendors might want to apply a downstream patch again to reject the other combinations — hint hint :) )

Going from a hard error all the way to almost-never-visible debug log means replacing one extreme by another; that’s of course _possible_ but a bit surprising to me. Would we consider a warning-level log?

(I am totally not an expert and I don’t care _what_ is decided here — but I would very much like to leave a written record of what are the constraints and considerations, for future maintainers.)

aarontomlin:
> Won't overlay break? Or is there a check later on?

Yes it will. I think it is much cleaner to terminate since the filesystem is not considered to be supported.

giuseppe:
we are breaking valid use cases as the issue that was opened today, I don't think we should do that only because we've not tested overlay on a specific file system.

We can trade the error for a warning, but that will still show up, while we probably need this information only when troubleshooting if something went wrong.

Luap99:
Using a white list seems wrong in any case, every time a new fs is added even if it supports overlay it will require us to patch all c/storage users. And then how are we going to keep this in sync with the kernel versions? This is not maintainable at all.

Also looking at the change it doesn't even do what you want. `graphdriver.FsNames` is a map with "some" known file systems. There is no mention in that map that is only contains supported overlay file systems. It even list overlayfs or nfs which are not supported by overlayfs.

Therefore even dropping it to debug level is not correct. IMO we should just revert it.

aarontomlin (commented):
LGTM but may require additional code changes elsewhere since the filesystem is unsupported (or "unknown").

giuseppe:
> Therefore even dropping it to debug level is not correct. IMO we should just revert it.

you are right.  Let's just revert it.  I've updated the PR

giuseppe:
CI is green

mtrmac:
> Therefore even dropping it to debug level is not correct. IMO we should just revert it.

That’s a good point in favor of not keeping the current code.

I’d still, as an outsider, like the experts to clarify, _and write down_, what “supported” means/was intended to mean here, but  I’m fine with merging the PR as is without that.

rhatdan:
LGTM

PR #1548, opened 2023-03-31
DriverTestSetQuota: Put() after Get()

nalind:
Convention for tests is that they `Put()` any layers they `Get()`.  Fix this omission in `DriverTestSetQuota()` and change the layer name from "zfsTest" to "quotaTest", because we have quota support on more than just zfs.


rhatdan:
LGTM
@vrothberg @saschagrunert @flouthoc @giuseppe @umohnani8 PTAL

giuseppe (approved):
/lgtm

PR #1549, opened 2023-04-03
store: add support to split filesystem using `imagestore`

flouthoc:
Allow storage users to split the filesystem of containers vs image store,  `imagestore` if configured will pull images in image storage instead of the `graphRoot` while keeping the other parts still in the originally configured `graphRoot`.

### DIFF
--- a/cmd/containers-storage/main.go
+++ b/cmd/containers-storage/main.go
@@ -43,6 +43,7 @@ func main() {
 		flags := mflag.NewFlagSet(command, eh)
 		flags.StringVar(&options.RunRoot, []string{"-run", "R"}, options.RunRoot, "Root of the runtime state tree")
 		flags.StringVar(&options.GraphRoot, []string{"-graph", "g"}, options.GraphRoot, "Root of the storage tree")
+		flags.StringVar(&options.ImageStore, []string{"-image-store"}, options.ImageStore, "Root of the separate image store")
 		flags.BoolVar(&options.TransientStore, []string{"-transient-store"}, options.TransientStore, "Transient store")
 		flags.StringVar(&options.GraphDriverName, []string{"-storage-driver", "s"}, options.GraphDriverName, "Storage driver to use ($STORAGE_DRIVER)")
 		flags.Var(opts.NewListOptsRef(&options.GraphDriverOptions, nil), []string{"-storage-opt"}, "Set storage driver options ($STORAGE_OPTS)")
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -322,6 +322,7 @@ func getBuiltinDriver(name, home string, options Options) (Driver, error) {
 type Options struct {
 	Root                string
 	RunRoot             string
+	ImageStore          string
 	DriverPriority      []string
 	DriverOptions       []string
 	UIDMaps             []idtools.IDMap
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -112,6 +112,7 @@ type Driver struct {
 	name             string
 	home             string
 	runhome          string
+	imageStore       string
 	uidMaps          []idtools.IDMap
 	gidMaps          []idtools.IDMap
 	ctr              *graphdriver.RefCounter
@@ -304,6 +305,16 @@ func isNetworkFileSystem(fsMagic graphdriver.FsMagic) bool {
 // If overlay filesystem is not supported on the host, a wrapped graphdriver.ErrNotSupported is returned as error.
 // If an overlay filesystem is not supported over an existing filesystem then a wrapped graphdriver.ErrIncompatibleFS is returned.
 func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) {
+	// If custom --imagestore is selected never
+	// ditch the original graphRoot, instead add it as
+	// additionalImageStore so its images can still be
+	// read and used.
+	if options.ImageStore != "" {
+		graphRootAsAdditionalStore := fmt.Sprintf("AdditionalImageStore=%s", options.ImageStore)
+		options.DriverOptions = append(options.DriverOptions, graphRootAsAdditionalStore)
+		// complete base name with driver name included
+		options.ImageStore = filepath.Join(options.ImageStore, "overlay")
+	}
 	opts, err := parseOptions(options.DriverOptions)
 	if err != nil {
 		return nil, err
@@ -328,6 +339,12 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 	// Create the driver home dir
 	if err := idtools.MkdirAllAs(path.Join(home, linkDir), 0755, 0, 0); err != nil {
 		return nil, err
+	}
+
+	if options.ImageStore != "" {
+		if err := idtools.MkdirAllAs(path.Join(options.ImageStore, linkDir), 0755, 0, 0); err != nil {
+			return nil, err
+		}
 	}
 
 	if err := idtools.MkdirAllAs(runhome, 0700, rootUID, rootGID); err != nil {
@@ -421,6 +438,7 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 	d := &Driver{
 		name:             "overlay",
 		home:             home,
+		imageStore:       options.ImageStore,
 		runhome:          runhome,
 		uidMaps:          options.UIDMaps,
 		gidMaps:          options.GIDMaps,
@@ -807,15 +825,22 @@ func (d *Driver) Status() [][2]string {
 // Metadata returns meta data about the overlay driver such as
 // LowerDir, UpperDir, WorkDir and MergeDir used to store data.
 func (d *Driver) Metadata(id string) (map[string]string, error) {
-	dir := d.dir(id)
+	dir, imagestore, _ := d.dir2(id)
 	if _, err := os.Stat(dir); err != nil {
 		return nil, err
 	}
+	workDirBase := dir
+	if imagestore != "" {
+		if _, err := os.Stat(dir); err != nil {
+			return nil, err
+		}
+		workDirBase = imagestore
+	}
 
 	metadata := map[string]string{
-		"WorkDir":   path.Join(dir, "work"),
-		"MergedDir": path.Join(dir, "merged"),
-		"UpperDir":  path.Join(dir, "diff"),
+		"WorkDir":   path.Join(workDirBase, "work"),
+		"MergedDir": path.Join(workDirBase, "merged"),
+		"UpperDir":  path.Join(workDirBase, "diff"),
 	}
 
 	lowerDirs, err := d.getLowerDirs(id)
@@ -930,7 +955,7 @@ func (d *Driver) Create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 }
 
 func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, disableQuota bool) (retErr error) {
-	dir := d.dir(id)
+	dir, imageStore, _ := d.dir2(id)
 
 	uidMaps := d.uidMaps
 	gidMaps := d.gidMaps
@@ -958,8 +983,19 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, disable
 	if err := idtools.MkdirAllAndChownNew(path.Dir(dir), 0755, idPair); err != nil {
 		return err
 	}
+	workDirBase := dir
+	if imageStore != "" {
+		workDirBase = imageStore
+		if err := idtools.MkdirAllAndChownNew(path.Dir(imageStore), 0755, idPair); err != nil {
+			return err
+		}
+	}
 	if parent != "" {
-		st, err := system.Stat(filepath.Join(d.dir(parent), "diff"))
+		parentBase, parentImageStore, _ := d.dir2(parent)
+		if parentImageStore != "" {
+			parentBase = parentImageStore
+		}
+		st, err := system.Stat(filepath.Join(parentBase, "diff"))
 		if err != nil {
 			return err
 		}
@@ -979,12 +1015,22 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, disable
 	if err := idtools.MkdirAllAndChownNew(dir, 0700, idPair); err != nil {
 		return err
 	}
+	if imageStore != "" {
+		if err := idtools.MkdirAllAndChownNew(imageStore, 0700, idPair); err != nil {
+			return err
+		}
+	}
 
 	defer func() {
 		// Clean up on failure
 		if retErr != nil {
 			if err2 := os.RemoveAll(dir); err2 != nil {
 				logrus.Errorf("While recovering from a failure creating a layer, error deleting %#v: %v", dir, err2)
+			}
+			if imageStore != "" {
+				if err2 := os.RemoveAll(workDirBase); err2 != nil {
+					logrus.Errorf("While recovering from a failure creating a layer, error deleting %#v: %v", dir, err2)
+				}
 			}
 		}
 	}()
@@ -1014,20 +1060,31 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, disable
 	if d.options.forceMask != nil {
 		perms = *d.options.forceMask
 	}
+
 	if parent != "" {
-		st, err := system.Stat(filepath.Join(d.dir(parent), "diff"))
+		parentDir, parentImageStore, _ := d.dir2(parent)
+		base := parentDir
+		if parentImageStore != "" {
+			base = parentImageStore
+		}
+		st, err := system.Stat(filepath.Join(base, "diff"))
 		if err != nil {
 			return err
 		}
 		perms = os.FileMode(st.Mode())
 	}
 
-	if err := idtools.MkdirAs(path.Join(dir, "diff"), perms, rootUID, rootGID); err != nil {
+	if err := idtools.MkdirAs(path.Join(workDirBase, "diff"), perms, rootUID, rootGID); err != nil {
 		return err
 	}
 
 	lid := generateID(idLength)
-	if err := os.Symlink(path.Join("..", id, "diff"), path.Join(d.home, linkDir, lid)); err != nil {
+
+	linkBase := path.Join("..", id, "diff")
+	if imageStore != "" {
+		linkBase = path.Join(imageStore, "diff")
+	}
+	if err := os.Symlink(linkBase, path.Join(d.home, linkDir, lid)); err != nil {
 		return err
 	}
 
@@ -1036,7 +1093,7 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, disable
 		return err
 	}
 
-	if err := idtools.MkdirAs(path.Join(dir, "work"), 0700, rootUID, rootGID); err != nil {
+	if err := idtools.MkdirAs(path.Join(workDirBase, "work"), 0700, rootUID, rootGID); err != nil {
 		return err
 	}
 	if err := idtools.MkdirAs(path.Join(dir, "merged"), 0700, rootUID, rootGID); err != nil {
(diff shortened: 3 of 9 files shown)


rhatdan:
```
	imgStoreRoot := s.pullStore
	if imgStoreRoot == "" {
		imgStoreRoot = s.graphRoot
	}
```

flouthoc:
`imagestore` SGTM I'll amend this.

pkg/config/config.go
```diff
@@ -124,6 +124,11 @@ type OptionsConfig struct {
 	// for shared image content
 	AdditionalImageStores []string `toml:"additionalimagestores,omitempty"`
 
+	// Image Store is the location of image store which is seperated from the
```
giuseppe, line 127: typo in seperated

giuseppe: no space between `Image` and `Store`

pkg/config/config.go
```diff
@@ -124,6 +124,11 @@ type OptionsConfig struct {
...
+	// container store. Usually this is not recommended unless users wants
+	// seperate store for image and containers.
```
giuseppe, line 129: typo in seperate

giuseppe (commented):
Will the RO layer be found when `Image Store` is used?

e.g. another client is configured to not use the `Image Store`, and will create all the layers in the regular store.  How would it work when the `Image Store` is set

flouthoc:
> Will the RO layer be found when `Image Store` is used?
> 
> e.g. another client is configured to not use the `Image Store`, and will create all the layers in the regular store. How would it work when the `Image Store` is set

I tried replaying this scenario, no `RO layer` will be not present if user switches to `--imagestore` in between, since `--imagestore` switches to a completely new image store. I such cases if users wants to use layers from previous store he should use `additional-image-store` and use original store as an `additional-image-store` manually.

giuseppe:
> > Will the RO layer be found when `Image Store` is used?
> > e.g. another client is configured to not use the `Image Store`, and will create all the layers in the regular store. How would it work when the `Image Store` is set
> 
> I tried replaying this scenario, no `RO layer` will be not present if user switches to `--imagestore` in between, since `--imagestore` switches to a completely new image store. In such cases if users wants to use layers from previous store he should use `additional-image-store` and use original store as an `additional-image-store` manually.

I'd prefer the opposite to happen.  If a `--imagestore` is specified, then images are pulled there and it is internally handled as an additional image store.  The lookup would work in the same way it does now, first looking in the current layers then looking it up in the additional image stores.

When I build from a Containerfile, where would the layers end up?

flouthoc:
@giuseppe I made a modification where when `--image-store` is used exisintg `graphRoot` is automatically used as an secondary `additionalStore` internal allowing users to use images both from newly added `image-store` and the original `graphRoot`, I have a added an `integration` test: `split-store - use graphRoot as an additional store by default` which tests this behavior. Do you think this suffice the use-case mentioned above.

flouthoc:
Tagging @mrunalp for inputs.

mrunalp:
Looks fine to me. Did we test with two disks?

flouthoc:
I think it should work without any issues but let me test it out with two separate disk to confirm.

flouthoc:
Okay I did test it with `imagestore` on a different disk than my original graphroot and it works fine. I used an external usb stick as `imagestore` and my current disk as original `graphroot`.

```console
# Check if we mounted external usb
[fl@fedora ~]$ lsblk
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
loop0         7:0    0  32.3M  1 loop /var/lib/snapd/snap/snapd/12704
loop1         7:1    0  55.4M  1 loop /var/lib/snapd/snap/core18/2128
loop2         7:2    0  13.9M  1 loop /var/lib/snapd/snap/universal-ctags/552
sdb           8:16   1  14.3G  0 disk 
├─sdb1        8:17   1  14.3G  0 part /run/media/fl/12a7bc12-59b6-4aa2-b40a-76f67faf28b7
└─sdb2        8:18   1     1M  0 part /run/media/fl/UEFI_NTFS
zram0       252:0    0     8G  0 disk [SWAP]
nvme0n1     259:0    0 476.9G  0 disk 
├─nvme0n1p1 259:1    0   600M  0 part /boot/efi
├─nvme0n1p2 259:2    0     1G  0 part /boot
└─nvme0n1p3 259:3    0 475.4G  0 part /home

# Check if external use contains directory for podman-storage
[fl@fedora ~]$ ls  /run/media/fl/12a7bc12-59b6-4aa2-b40a-76f67faf28b7
lost+found  podman-storage

# Check our storage.conf
[fl@fedora ~]$ cat /home/fl/.config/containers/storage.conf
[storage]

imagestore = "/run/media/fl/12a7bc12-59b6-4aa2-b40a-76f67faf28b7/podman-storage"

# First set imagestore to `none` and pull ubuntu
# Second set imagestore to `usb` and pull alpine

# podman images should show image from both original graph root and usb-stick
[fl@fedora bin]$ ./podman images
WARN[0000] The storage 'driver' option should be set in /home/fl/.config/containers/storage.conf. A driver was picked automatically. 
REPOSITORY                TAG         IMAGE ID      CREATED      SIZE        R/O
docker.io/library/alpine  latest      9ed4aefc74f6  11 days ago  7.34 MB     false
docker.io/library/ubuntu  latest      08d22c0ceb15  4 weeks ago  80.3 MB     true

# run container where image is not on imagestore but on original graph root
[fl@fedora bin]$ ./podman run --pull=never -it --rm ubuntu sh
WARN[0000] The storage 'driver' option should be set in /home/fl/.config/containers/storage.conf. A driver was picked automatically. 
# exit
[fl@fedora bin]$ 

# run container where image is on imagestore but not on original graph root
[fl@fedora bin]$ ./podman run --pull=never -it --rm alpine sh
WARN[0000] The storage 'driver' option should be set in /home/fl/.config/containers/storage.conf. A driver was picked automatically. 
/ # exit
[fl@fedora bin]$ 
```

flouthoc:
@giuseppe @mrunalp ^

rhatdan:
You need to update the containers-storage.conf.5.md

store.go
```diff
@@ -777,6 +778,16 @@ func GetStore(options types.StoreOptions) (Store, error) {
...
+	// If custom --imagestore is selected never
+	// ditch the original graphRoot, instead add it as
+	// additionalImageStore so its images can still be
+	// read and used.
+	if options.ImageStore != "" {
+		if options.GraphDriverName == "overlay" {
+			graphRootAsAdditionalStore := fmt.Sprintf("AdditionalImageStore=%s", options.GraphRoot)
+			options.GraphDriverOptions = append(options.GraphDriverOptions, graphRootAsAdditionalStore)
+		}
+	}
```
giuseppe, line 790: this logic must go in the `overlay` driver, not in the store

giuseppe (commented):
I've added:

```
imagestore = "/var/lib/images"
```

to the `storage.conf` file under `storage.options`, then I've tried a `podman pull $IMAGE` but the image is pulled into the graphroot as if the option doesn't change the behavior.

Have you tried setting it from the `storage.conf` file?

giuseppe:
if I run a container though, I see this mount:

```
overlay on /var/lib/images/overlay/1d3e94d927c36596ca191cea1f4ea0077c1e27b18f51696ee6dd223afabddd2d/merged type overlay (rw,nodev,relatime,context="system_u:object_r:container_file_t:s0:c75,c349",lowerdir=/var/lib/images/overlay/l/5S44LLQZ3VTWR42FB2ELAYQDAY,upperdir=/var/lib/images/overlay/1d3e94d927c36596ca191cea1f4ea0077c1e27b18f51696ee6dd223afabddd2d/diff,workdir=/var/lib/images/overlay/1d3e94d927c36596ca191cea1f4ea0077c1e27b18f51696ee6dd223afabddd2d/work,metacopy=on,volatile)
```

shouldn't the upperdir and workdir be under `$GRAPHROOT`?

flouthoc:
> if I run a container though, I see this mount:
> 
> ```
> overlay on /var/lib/images/overlay/1d3e94d927c36596ca191cea1f4ea0077c1e27b18f51696ee6dd223afabddd2d/merged type overlay (rw,nodev,relatime,context="system_u:object_r:container_file_t:s0:c75,c349",lowerdir=/var/lib/images/overlay/l/5S44LLQZ3VTWR42FB2ELAYQDAY,upperdir=/var/lib/images/overlay/1d3e94d927c36596ca191cea1f4ea0077c1e27b18f51696ee6dd223afabddd2d/diff,workdir=/var/lib/images/overlay/1d3e94d927c36596ca191cea1f4ea0077c1e27b18f51696ee6dd223afabddd2d/work,metacopy=on,volatile)
> ```
> 
> shouldn't the upperdir and workdir be under `$GRAPHROOT`?

Let me shift that to `GRAPHROOT`

giuseppe (changes requested):
some issues I've seen:

- if the directory is missing I only get a `Error: no such file or directory` which is not very clear.  We need to add more information to the error.  Or could we just create the directory?
- if I create the directory manually, it goes a bit farther, but it fails with:
```
  Error: copying system image from manifest list: writing blob: adding layer with blob "sha256:f56be85fc22e46face30e2c3de3f7fe7c15f8fd7c4e5add29d7f64b87abdaa09": readlink /var/lib/images/overlay/f1417ff83b319fbdae6dd9cd6d8c9c88002dcd75ecf6ec201c8c6894681cf2b5/diff: no such file or directory
```
- it creates some metadata in the `$GRAPHROOT` directory as well.  Is this expected?


If we want this to behave internally like an "additional store" we should not write any metadata in the `$GRAPHROOT` for the image we are pulling

cmd/containers-storage/main.go
```diff
@@ -43,6 +43,7 @@ func main() {
 		flags := mflag.NewFlagSet(command, eh)
 		flags.StringVar(&options.RunRoot, []string{"-run", "R"}, options.RunRoot, "Root of the runtime state tree")
 		flags.StringVar(&options.GraphRoot, []string{"-graph", "g"}, options.GraphRoot, "Root of the storage tree")
+		flags.StringVar(&options.ImageStore, []string{"-image-store"}, options.ImageStore, "Root of the seperate image store")
```
TomSweeneyRedHat, line 46:
```suggestion
		flags.StringVar(&options.ImageStore, []string{"-image-store"}, options.ImageStore, "Root of the separate image store")
```

TomSweeneyRedHat: FWIW, I type seperate all the time, too for separate.....

store.go
```diff
@@ -1020,7 +1043,11 @@ func (s *store) getLayerStoreLocked() (rwLayerStore, error) {
 	if err := os.MkdirAll(rlpath, 0700); err != nil {
 		return nil, err
 	}
-	glpath := filepath.Join(s.graphRoot, driverPrefix+"layers")
+	imgStoreRoot := s.imageStoreDir
+	if imgStoreRoot == "" {
+		imgStoreRoot = s.graphRoot
+	}
+	glpath := filepath.Join(imgStoreRoot, driverPrefix+"layers")
```
TomSweeneyRedHat, line 1050: I think it's good, but just in case.  Any chance a symlink could sneak into one of these join commands that might cause havoc?

TomSweeneyRedHat:
@flouthoc looks like you need a rebase

store.go
flouthoc: It is same as original graphroot logic, even graphroot can accept a symlink, since its not a workload provided input i think it should not cause any issues :)

drivers/overlay/overlay.go
```diff
@@ -1342,10 +1379,15 @@ func (d *Driver) Get(id string, options graphdriver.MountOpts) (string, error) {
...
-	dir, inAdditionalStore := d.dir2(id)
+	dir, splitStore, inAdditionalStore := d.dir2(id)
 	if _, err := os.Stat(dir); err != nil {
 		return "", err
 	}
+	if splitStore != "" {
+		if _, err := os.Stat(splitStore); err != nil {
+			return "", err
+		}
+	}
```
giuseppe, line 1390: I don't think we need to error out here if it is not in the store

drivers/driver.go
```diff
@@ -322,6 +322,7 @@ func getBuiltinDriver(name, home string, options Options) (Driver, error) {
 type Options struct {
 	Root                string
 	RunRoot             string
+	SplitStore          string
```
giuseppe, line 325: could we rename it to `ImageStore`?

giuseppe (changes requested):
I've tried the following commands and I still see errors:

```
# find /var/lib/images
find: ‘/var/lib/images’: No such file or directory
#  bin/podman pull alpine
# find /var/lib/images/
/var/lib/images/
/var/lib/images/overlay
/var/lib/images/overlay/l
/var/lib/images/overlay/l/WQZSZTPLMUGVUISAU65BJKGNRI
/var/lib/images/overlay/.has-mount-program
/var/lib/images/overlay/backingFsBlockDev
/var/lib/images/overlay/f1417ff83b319fbdae6dd9cd6d8c9c88002dcd75ecf6ec201c8c6894681cf2b5
/var/lib/images/overlay/f1417ff83b319fbdae6dd9cd6d8c9c88002dcd75ecf6ec201c8c6894681cf2b5/link
/var/lib/images/overlay/f1417ff83b319fbdae6dd9cd6d8c9c88002dcd75ecf6ec201c8c6894681cf2b5/merged
/var/lib/images/overlay/f1417ff83b319fbdae6dd9cd6d8c9c88002dcd75ecf6ec201c8c6894681cf2b5/empty
/var/lib/images/overlay-images
/var/lib/images/overlay-images/images.lock
/var/lib/images/overlay-images/images.json
/var/lib/images/overlay-images/9ed4aefc74f6792b5a804d1d146fe4b4a2299147b0f50eaf2b08435d7b38c27e
/var/lib/images/overlay-images/9ed4aefc74f6792b5a804d1d146fe4b4a2299147b0f50eaf2b08435d7b38c27e/=c2hhMjU2OjllZDRhZWZjNzRmNjc5MmI1YTgwNGQxZDE0NmZlNGI0YTIyOTkxNDdiMGY1MGVhZjJiMDg0MzVkN2IzOGMyN2U=
/var/lib/images/overlay-images/9ed4aefc74f6792b5a804d1d146fe4b4a2299147b0f50eaf2b08435d7b38c27e/=bWFuaWZlc3Qtc2hhMjU2OjEyNGM3ZDI3MDc5MDRlZWE3NDMxZmZmZTkxNTIyYTAxZTVhODYxYTYyNGVlMzFkMDMzNzJjYzFkMTM4YTMxMjY=
/var/lib/images/overlay-images/9ed4aefc74f6792b5a804d1d146fe4b4a2299147b0f50eaf2b08435d7b38c27e/=bWFuaWZlc3Qtc2hhMjU2OmI2Y2EyOTBiNmI0Y2RjY2E1YjNkYjNmZmEzMzhlZTAyODVjMTE3NDRiNGE2YWJhYTk2Mjc3NDZlZTMyOTFkOGQ=
/var/lib/images/overlay-images/9ed4aefc74f6792b5a804d1d146fe4b4a2299147b0f50eaf2b08435d7b38c27e/manifest
/var/lib/images/overlay-images/9ed4aefc74f6792b5a804d1d146fe4b4a2299147b0f50eaf2b08435d7b38c27e/=c2lnbmF0dXJlLWI2Y2EyOTBiNmI0Y2RjY2E1YjNkYjNmZmEzMzhlZTAyODVjMTE3NDRiNGE2YWJhYTk2Mjc3NDZlZTMyOTFkOGQ=
/var/lib/images/overlay-layers
/var/lib/images/overlay-layers/layers.lock
/var/lib/images/overlay-layers/layers.json
/var/lib/images/overlay-layers/f1417ff83b319fbdae6dd9cd6d8c9c88002dcd75ecf6ec201c8c6894681cf2b5.tar-split.gz
/var/lib/images/overlay-layers/volatile-layers.json
# bin/podman run --rm -ti alpine sh
Error: creating container storage: creating read-write layer with ID "d1631cddf7219c47a7bd3648aac89ade7484f66dd1f3e95b687d287129c5a7d0": Stat /var/lib/images/overlay/f1417ff83b319fbdae6dd9cd6d8c9c88002dcd75ecf6ec201c8c6894681cf2b5/diff: no such file or directory
```

and if I try pulling the image with a podman without the patch (so that it goes into `$GRAPHROOT`), I get this error:

```
# podman pull alpine
WARN[0000] Failed to decode the keys ["storage.imagestore"] from "/etc/containers/storage.conf" 
Resolved "alpine" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/alpine:latest...
Getting image source signatures
Copying blob f56be85fc22e done  
WARN[0002] Trying to create a layer "f1417ff83b319fbdae6dd9cd6d8c9c88002dcd75ecf6ec201c8c6894681cf2b5" while directory "/var/lib/containers/storage/overlay/f1417ff83b319fbdae6dd9cd6d8c9c88002dcd75ecf6ec201c8c6894681cf2b5" already exists; removing it first 
Copying config 9ed4aefc74 done  
Writing manifest to image destination
Storing signatures
9ed4aefc74f6792b5a804d1d146fe4b4a2299147b0f50eaf2b08435d7b38c27e
```

giuseppe:
thanks for addressing the comments.  It seems to work well now

rhatdan:
@flouthoc could you open a PR against Podman to make sure this does not break anything, then once we get this merged into containers/storage, lets create a test in Podman to make sure that imageStore does not get broken in the future.

rhatdan:
Once this passes on Podman I will merge.

flouthoc:
@rhatdan Podman PR: https://github.com/containers/podman/pull/18224

rhatdan:
LGTM

drivers/overlay/overlay.go
```diff
@@ -807,15 +825,22 @@ func (d *Driver) Status() [][2]string {
...
 // LowerDir, UpperDir, WorkDir and MergeDir used to store data.
 func (d *Driver) Metadata(id string) (map[string]string, error) {
-	dir := d.dir(id)
+	dir, imagestore, _ := d.dir2(id)
 	if _, err := os.Stat(dir); err != nil {
 		return nil, err
 	}
+	workDirBase := dir
+	if imagestore != "" {
+		if _, err := os.Stat(dir); err != nil {
```
mtrmac, line 834: Why is this checking `dir`? Based on the `Stat` just a few lines above, this is not going to fail.

drivers/overlay/overlay.go
```diff
@@ -807,15 +825,22 @@ func (d *Driver) Status() [][2]string {
...
+		workDirBase = imagestore
+	}
 
 	metadata := map[string]string{
-		"WorkDir":   path.Join(dir, "work"),
-		"MergedDir": path.Join(dir, "merged"),
-		"UpperDir":  path.Join(dir, "diff"),
+		"WorkDir":   path.Join(workDirBase, "work"),
+		"MergedDir": path.Join(workDirBase, "merged"),
+		"UpperDir":  path.Join(workDirBase, "diff"),
```
mtrmac, line 843: This is all using the values dependent on `s.imageStore`.

But `create` is using `workDirBase/"work"`, but `dir/"merged"`. `UpdateLayerIDMap` is using `dir/"work"`. Everything seems to be using `dir/"diff"`.

What’s going on here?

If it is hard to keep track of the right directory, introduce methods like `workDir(imageID)` and have _everything_ call them.

drivers/overlay/overlay.go
```diff
@@ -1121,22 +1178,26 @@ func (d *Driver) getLower(parent string) (string, error) {
...
+	p, _, _ := d.dir2(id)
 	return p
 }
 
-func (d *Driver) dir2(id string) (string, bool) {
+func (d *Driver) dir2(id string) (string, string, bool) {
 	newpath := path.Join(d.home, id)
+	imageStore := ""
+	if d.imageStore != "" {
+		imageStore = path.Join(d.imageStore, id)
```
mtrmac, line 1189: So we now have an “image store” variable, which is an individual per-image value, but `d.imageStore` is a per-store global.

And this function creates this `{s.imageStore}/{id}` value, but never checks it for any purpose. Why?
- If `{d.home}/{id}` exists, this function returns that as the first return value
- `{d.home}/{id}` does not exist but one of the “additional image stores” contain an `/{id}` path, this function returns that additional one.

What role does `{s.imageStore}/{id}` play here?

Before, this function fairly clearly returned “which directory contains the image data if the image exists, or where the data should be created if it doesn’t exist anywhere; is the image in an additional store”.

Now it returns … ???

In particular the various parent lookups only check in the `{s.imageStore}/{id}` location, outright failing if the image is not there. But this function does not check _at all_ whether anything exists in that location.

mtrmac (commented):
I am probably too sleep-deprived but I just can’t understand how this works, starting with `dir2`.

Why are all callers of `dir2` special-casing this? I would expect a _single_ location in this package to contain some code to find an existing image, which callers can then use without any conditionals. And I would expect a _single_ location in this package to contain some code to determine where to create new images or store more data, which callers can then use without any conditionals.

This is clearly not that, and I can’t understand how this is intended to work.

Again, maybe it is just me. But please someone take another look.

flouthoc:
I'll open a new PR to address these refactor comments.

nalind:
> I am probably too sleep-deprived but I just can’t understand how this works, starting with `dir2`.
> 
> Why are all callers of `dir2` special-casing this? I would expect a _single_ location in this package to contain some code to find an existing image, which callers can then use without any conditionals. And I would expect a _single_ location in this package to contain some code to determine where to create new images or store more data, which callers can then use without any conditionals.
> 
> This is clearly not that, and I can’t understand how this is intended to work.
> 
> Again, maybe it is just me. But please someone take another look.

Looking at it now, I don't think I fully understand it, either.  With the option set, when the overlay driver creates and removes layers under the graph root, continuing to record things like the list of lowers there, it also seems to be creating and removing directories under the image store location, and it's neither obvious nor commented why it puts things in one place or the other.  If my mostly-guesswork view of what the overlay driver is doing is correct, then the overlay driver is not correctly setting disk quotas.
I definitely don't understand what's supposed to be happening differently in the vfs driver.

PR #1551, opened 2023-04-04
overlay: disable native idmap with mount-program

giuseppe:
never create the native idmap mount when a mount program is used.


flouthoc (approved):
LGTM

rhatdan:
LGTM

PR #1554, opened 2023-04-05
Improve AtomicFileWriter's support for Win and Mac

n1hility:
As discussed on containers/common#1397, this PR is a step towards addressing corruption issues with container and machine configuration files that can occur with an improper shutdown.

Add ExplicitCommit option to provide calling code the ability to preserve original files when an application error occurs, not just an error that occurs during writing (e.g. a marshaller failure)

Change Close() impl to safely handle redundant calls (for defer usage safety)

Introduces two flush points, one just after data is written that supports the previous code's Linux approach of using fdatasync(), and another for Mac/Win/Non-Linux Unix that needs to full flush after metadata updates are staged. 

Using this patch, common's usage for writting the container config file would look like this:

```golang

opts := &ioutils.AtomicFileWriterOptions{ExplicitCommit: true}
configFile, err := ioutils.NewAtomicFileWriterWithOpts(path, 0o644, opts)
if err != nil {
	return err
}
defer configFile.Close()

enc := toml.NewEncoder(configFile)
if err := enc.Encode(c); err != nil {
	return err
}

// If no errors commit the changes to the config file
return configFile.Commit()

```


pkg/ioutils/fswriters.go
```diff
@@ -17,6 +17,20 @@ type AtomicFileWriterOptions struct {
 	// On successful return from Close() this is set to the mtime of the
 	// newly written file.
 	ModTime time.Time
+	// Specifies whether Commit() must be explicitly called to write state
+	// to the destination. This allows an application to preserve the original
+	// file when an error occurs during processing (and not just during write)
+	// The default is false, which will auto-commit on Close
+	ExplicitCommit bool
```
vrothberg, line 24: I think I'm missing a point and wonder why `ExplicitCommit` and `Commit()` are needed.  I somehow expected changes to `Close()` only.  @n1hility, can you elaborate on that?

vrothberg (commented):
@Luap99 @giuseppe @nalind PTAL

pkg/ioutils/fswriters.go
vrothberg: Thanks, makes sense :+1:

n1hility: Great summary, I couldn't have said it better. I would have preferred to make a hard move and make this commit() approach the only option instead of a flag like in my PR, since the previous auto-commit behavior is likely to cause problems in most usage patterns. However, it would be compatibility breaking, and I wasnt sure how much usage of this API there is in the wild.

vrothberg: I think that's a great way of keeping the API stable, @n1hility.  Thanks!

rhatdan:
LGTM

PR #719, opened 2020-09-15
Devmapper fixes and cleanups

kolyshkin:
This is a port of a few upstream PRs:
 * https://github.com/moby/moby/pull/34573 devicemapper: remove container rootfs mountPath after umount
 * https://github.com/moby/moby/pull/36438 devmapper/Remove(): use Rmdir, ignore errors
 * https://github.com/moby/moby/pull/36307 devmapper cleanup improvements
 * https://github.com/moby/moby/pull/40056 devmapper: fix unit test

plus a change to use `mount.Unmount` instead of `unix.Unmount` for the sake of better errors.

The port of https://github.com/moby/moby/pull/34573 is pretty important, as it may prevent those nasty EBUSY errors. Others are in the "nice to have" category.

Please see individual commits and/or the above PRs for details.

PR #1563, opened 2023-04-10
Extend driver.ListLayers()

nalind:
* Implement ListLayers() for the aufs, btrfs, and devicemapper drivers, along with a unit test for them.
* Stop filtering out directories with names that aren't 64-hex chars the overlay and vfs ListLayers() implementations, which is more a convention than a hard rule.
* Close() a dangling ReadCloser in NaiveCreateFromTemplate.
* Have layerStore.Wipe() try to remove remaining listed layers after it removes the layers that the layerStore knew of.
* Switch from using plain defer to using t.Cleanup() to handle deleting layers that tests create, have the addManyLayers() test function do so as well.
* Remove vfs.CopyDir, which near as I can tell isn't referenced anywhere.

### DIFF
--- a/drivers/graphtest/testutil.go
+++ b/drivers/graphtest/testutil.go
@@ -13,10 +13,12 @@ import (
 	"os"
 	"path"
 	"sort"
+	"testing"
 
 	graphdriver "github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/stringid"
+	"github.com/stretchr/testify/require"
 )
 
 func randomContent(size int, seed int64) []byte {
@@ -282,13 +284,14 @@ func addLayerFiles(drv graphdriver.Driver, layer, parent string, i int) error {
 	return nil
 }
 
-func addManyLayers(drv graphdriver.Driver, baseLayer string, count int) (string, error) {
+func addManyLayers(t testing.TB, drv graphdriver.Driver, baseLayer string, count int) (string, error) {
 	lastLayer := baseLayer
 	for i := 1; i <= count; i++ {
 		nextLayer := stringid.GenerateRandomID()
 		if err := drv.Create(nextLayer, lastLayer, nil); err != nil {
 			return "", err
 		}
+		t.Cleanup(func() { removeLayer(t, drv, nextLayer) })
 		if err := addLayerFiles(drv, nextLayer, lastLayer, i); err != nil {
 			return "", err
 		}
@@ -350,4 +353,10 @@ func readDir(dir string) ([]os.DirEntry, error) {
 	}
 
 	return b, nil
+}
+
+// removeLayer tries to remove the layer
+func removeLayer(t testing.TB, driver graphdriver.Driver, name string) {
+	err := driver.Remove(name)
+	require.NoError(t, err)
 }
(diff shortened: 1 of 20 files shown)


rhatdan:
LGTM
@giuseppe @vrothberg @saschagrunert @umohnani8 @mtrmac @flouthoc PTAL

giuseppe:
the test failure seems related to the PR

drivers/overlay/overlay.go
```diff
@@ -1719,26 +1721,29 @@ func (d *Driver) Exists(id string) bool {
 	return err == nil
 }
 
-// List layers (not including additional image stores)
+// ListLayers(), when implemented, is used when checking for errors.
```
mtrmac, line 1724: AFAICS it is actually used for `Wipe` (only?); this comment, and similar comments in other graph drivers, seem to be a bit misleading.

mtrmac (commented):
Just a very minimal pass over the driver changes — I don’t understand the subtleties of the graph drivers, and I didn’t carefully read the graph test parts.

drivers/overlay/overlay.go
mtrmac: I’m fine with keeping it as is, then.

(Or, alternatively, just document the driver’s responsibility, and let callers use it for any purpose they can think of. That’s what the `ProtoDriver`’s interface documentation does already.)

drivers/graphtest/testutil.go
```diff
@@ -351,3 +357,14 @@ func readDir(dir string) ([]os.DirEntry, error) {
 
 	return b, nil
 }
+
+// removeLayer tries to remove the layer, retrying several times if an EBUSY
+// error is returned
+func removeLayer(t testing.TB, driver graphdriver.Driver, name string) {
+	err := driver.Remove(name)
+	for i := 100 * time.Millisecond; i <= 1000*time.Millisecond && err != nil && errors.Is(err, syscall.EBUSY); i += (100 * time.Millisecond) {
```
mtrmac, line 365: (I’m totally unfamiliar with this) BTW is it understood why retries are necessary? I’d naively expect the tests to be one of the more deterministic environments with no unexpected concurrent accessors, so what is keeping the volume busy if it’s not the tests?

Are the tests somehow triggering hard-to-cleanup concurrent goroutines/threads/processes that they don’t wait for, maybe?

mtrmac: And if it is understood, it might be worth documenting somewhere around here.

mtrmac: There’s enough mentions of `EBUSY` throughout the codebase that I totally believe that there’s _something_ (or multiple somethings) going on — I just could never find any specific description of what is actually the reason. The closest is https://github.com/containers/storage/pull/719 / https://github.com/moby/moby/pull/36438 , which also isn’t really specific (and seems to suggest that the error is a genuine external user, not something that could be successfully retried).

So I’m mostly trying to get as much of the knowledge, if any, documented and explicit.
If this is working as designed and just requires a retry for good reasons, great, let’s document that.
If we know what is not working as designed, but needs a workaround short-term, fine, let’s document that.
If the cause is unknown but we know that something is failing, oh well, documenting that would still be nice.

If we are lucky, somebody will, sometime, find and fix the cause, and then look at this kind of loop and wonder whether it can be removed — and documenting why it was added would have been helpful.

PR #1564, opened 2023-04-11
drivers: chown root mount to root in the userns

giuseppe:
chown the root mount to root in the userns so that it can be used by the container.

### DIFF
--- a/drivers/vfs/driver.go
+++ b/drivers/vfs/driver.go
@@ -304,7 +304,16 @@ func (d *Driver) SupportsShifting() bool {
 // UpdateLayerIDMap updates ID mappings in a from matching the ones specified
 // by toContainer to those specified by toHost.
 func (d *Driver) UpdateLayerIDMap(id string, toContainer, toHost *idtools.IDMappings, mountLabel string) error {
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

flouthoc (approved):
LGTM

drivers/vfs/driver.go
```diff
@@ -304,7 +304,16 @@ func (d *Driver) SupportsShifting() bool {
...
-	return d.updater.UpdateLayerIDMap(id, toContainer, toHost, mountLabel)
+	if err := d.updater.UpdateLayerIDMap(id, toContainer, toHost, mountLabel); err != nil {
+		return err
+	}
+	dir := d.dir(id)
+	rootIDs, err := toHost.ToHost(idtools.IDPair{UID: 0, GID: 0})
+	if err != nil {
+		return err
+	}
+	return os.Chown(dir, rootIDs.UID, rootIDs.GID)
```
nalind, line 315: This is the parent of the "diff" directory, whose permissions are what we see when the filesystem is mounted.

nalind: You're right, that's my mistake - I thought this was also in the overlay driver.

PR #1568, opened 2023-04-12
Split up DeleteContainer

nalind:
When deleting a container, delete the layer first, as a distinct step, so that if we fail to do so, we don't just go ahead and remove the container record, which will prevent the caller from being able to try again.

### DIFF
--- a/store.go
+++ b/store.go
@@ -2542,63 +2542,45 @@ func (s *store) DeleteContainer(id string) error {
 			return ErrNotAContainer
 		}
 
-		errChan := make(chan error)
-		var wg sync.WaitGroup
-
+		// delete the layer first, separately, so that if we get an
+		// error while trying to do so, we don't go ahead and delete
+		// the container record that refers to it, effectively losing
+		// track of it
 		if rlstore.Exists(container.LayerID) {
-			wg.Add(1)
-			go func() {
-				errChan <- rlstore.Delete(container.LayerID)
-				wg.Done()
-			}()
+			if err := rlstore.Delete(container.LayerID); err != nil {
+				return err
+			}
 		}
-		wg.Add(1)
-		go func() {
-			errChan <- s.containerStore.Delete(id)
-			wg.Done()
-		}()
+
+		var wg multierror.Group
+		wg.Go(func() error { return s.containerStore.Delete(id) })
 
 		middleDir := s.graphDriverName + "-containers"
-		gcpath := filepath.Join(s.GraphRoot(), middleDir, container.ID)
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
+
+		wg.Go(func() error {
+			gcpath := filepath.Join(s.GraphRoot(), middleDir, container.ID)
 			// attempt a simple rm -rf first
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@giuseppe @flouthoc @saschagrunert @vrothberg @mtrmac @umohnani8 PTAL

mtrmac (commented):
The _mechanics_ of this change LGTM. I don’t know nearly enough to review the intent/design. Feel free to just let someone else review this instead of reading the rambling below…

---

- If we need to keep track of the container’s layer, don’t we need to similarly keep track of the  two directories that are being cleaned up? What is the conceptual difference between the two?
- And from the exact opposite angle, a failure deleting the layer, or a crash in the middle of deleting the layer, will leave a container object with a dangling layer reference. Is that good enough / is there going to be another cleanup mechanism for that? Or do we need to persistently record the intent to delete the container (compare `incompleteFlag` for layers)?

IIRC container IDs are not deterministic, so leaving around a partially-deleted and unusable container might not (?) be as bad as leaving around a partially-deleted layer, and we might not need that complexity. And from a quick skim the code seems to handle a dangling `container.LayerID` well enough, at least locally within c/storage (I didn’t read the users).

So, this PR seems to be trading off one kind of broken result (a corrupt layer that can’t be deleted, and has no user) for another (a corrupt layer + container, both of which can’t be deleted). I’m fine with either outcome, it’s not obvious to me why we would strongly prefer one over the other.

Is this in context of some (CRI-O?) sync loop that manages containers and knows to retry deleting containers, but doesn’t know about layers, for example? (And if so, do we worry about it trying, and failing, infinitely?)

At the very least, making this less concurrent can only improve reliability — so I’m fine with the change as is.

Still, if there are any other concerns or trade-offs that you have considered, I’d very much like to learn.

nalind:
An untracked layer can create problems if/when we want to delete the base image sometime later, but you're right that there's no some-things-were-deleted-but-some-things-weren't state that isn't going to be a problem in some way.

mtrmac:
Thanks, that’s a good point — users ~must know about `podman ps` and `podman images`, and if that’s all that is necessary, without having to think about layers at all, to clean up after a transient failure to delete a layer, and to successfully delete the image, that’s valuable.

LGTM, then.

---

I guess I don’t have a mental model for why would layer deletion fail _trasiently_, i.e. why it would be possible to successfully retry later; AFAICS that’s the only scenario where this makes a difference. Is this targeting some specific failure scenario?

(An I/O error cured with `fsck` run in the meantime? Something in the graph driver being busy due to administrator’s Podman-unmanaged processes?)

nalind:
I might be overcorrecting from getting EBUSY while working on #1563, though that may itself have been a hallucination.

mtrmac:
Thanks, EBUSY really is an error that should, eventually, just go away and allow retries. _And_ it is much more likely to affect the graph driver than the two other directories.

mtrmac (approved):

TomSweeneyRedHat:
LGTM

vrothberg (commented):
LGTM

PR #1569, opened 2023-04-13
Initial implementation of consistency checks

nalind:
Add initial Check() and Repair() methods to Stores.

Check() checks for inconsistencies between the layers which the lower-level storage driver claims to know about and the ones which we know we're managing.  It checks that layers referenced by layers, images, and containers are known to us and that images referenced by containers are known to us.  It checks that data which we store alongside layers, images, and containers is still present, and to the extent which we store other information about that data (frequenly just the size of the data), verifies that it matches recorded expectations. Lastly, it checks that layers which are part of images (and which we therefore know what they should have in them) have the expected content, and nothing else.

Repair() removes any containers, images, and layers which have any errors associated with them.  This is destructive, so its use should be considered and deliberate.

### DIFF
--- a/check.go
+++ b/check.go
@@ -1,0 +1,1086 @@
+package storage
+
+import (
+	"archive/tar"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+
+	drivers "github.com/containers/storage/drivers"
+	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/ioutils"
+	"github.com/containers/storage/types"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	// ErrLayerUnaccounted describes a layer that is present in the lower-level storage driver,
+	// but which is not known to or managed by the higher-level driver-agnostic logic.
+	ErrLayerUnaccounted = types.ErrLayerUnaccounted
+	// ErrLayerUnreferenced describes a layer which is not used by any image or container.
+	ErrLayerUnreferenced = types.ErrLayerUnreferenced
+	// ErrLayerIncorrectContentDigest describes a layer for which the contents of one or more
+	// files which were added in the layer appear to have changed.  It may instead look like an
+	// unnamed "file integrity checksum failed" error.
+	ErrLayerIncorrectContentDigest = types.ErrLayerIncorrectContentDigest
+	// ErrLayerIncorrectContentSize describes a layer for which regenerating the diff that was
+	// used to populate the layer produced a diff of a different size.  We check the digest
+	// first, so it's highly unlikely you'll ever see this error.
+	ErrLayerIncorrectContentSize = types.ErrLayerIncorrectContentSize
+	// ErrLayerContentModified describes a layer which contains contents which should not be
+	// there, or for which ownership/permissions/dates have been changed.
+	ErrLayerContentModified = types.ErrLayerContentModified
+	// ErrLayerDataMissing describes a layer which is missing a big data item.
+	ErrLayerDataMissing = types.ErrLayerDataMissing
+	// ErrLayerMissing describes a layer which is the missing parent of a layer.
+	ErrLayerMissing = types.ErrLayerMissing
+	// ErrImageLayerMissing describes an image which claims to have a layer that we don't know
+	// about.
+	ErrImageLayerMissing = types.ErrImageLayerMissing
+	// ErrImageDataMissing describes an image which is missing a big data item.
+	ErrImageDataMissing = types.ErrImageDataMissing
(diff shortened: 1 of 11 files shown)


rhatdan:
Nice work.
LGTM
@giuseppe  @vrothberg @saschagrunert @flouthoc @umohnani8 @mtrmac PTAL

giuseppe (approved):
LGTM

check.go
```diff
@@ -0,0 +1,1086 @@
...
+				}
+			}
+		}
+		return false, nil
+	}); errored {
+		return CheckReport{}, err
+	}
+
+	// Iterate over each container in turn.
+	if done, err := s.readContainerStore(func() (bool, error) {
```
mtrmac, line 573: Should this be nested inside the container image lock, for consistency?

check.go
```diff
@@ -0,0 +1,1086 @@
...
+						if int64(len(data)) != container.BigDataSizes[key] {
+							err = fmt.Errorf("container %s: data item %q: %w", id, key, ErrContainerDataIncorrectSize)
+							report.Containers[id] = append(report.Containers[id], err)
+							return
+						}
+					}()
+				}
+			}
+			// Look at the container's base image.  If the image has errors, the image's errors
+			// are the container's errors.
```
mtrmac, line 607: I’m not sure this is very useful — if we have a corrupt base layer, 20 application images build with that base layer, and 100 containers running those images, 100 individual reports of that layer being wrong seem rather redundant. _Especially_ if there were multiple reported errors per layer, or per image.

Maybe just a single “references an image/layer with errors” error would be sufficient.

check.go
```diff
@@ -0,0 +1,1086 @@
...
+				if len(report.Images[container.ImageID]) > 0 {
+					report.Containers[id] = append(report.Containers[id], report.Images[container.ImageID]...)
+				}
+				if len(report.ROImages[container.ImageID]) > 0 {
+					report.Containers[id] = append(report.Containers[id], report.ROImages[container.ImageID]...)
+				}
+			}
+			// Count the container's layer as referenced.
+			if container.LayerID != "" {
+				referencedLayers[container.LayerID] = true
```
mtrmac, line 622: What if this layer reference is dangling? What if this points to a read-only layer?

check.go
```diff
@@ -0,0 +1,1086 @@
...
+		}
+		layers, err := store.Layers()
+		if err != nil {
+			return true, err
+		}
+		for _, layer := range layers {
+			maximumAge := defaultMaximumUnreferencedLayerAge
+			if options.LayerUnreferencedMaximumAge != nil {
+				maximumAge = *options.LayerUnreferencedMaximumAge
+			}
```
mtrmac, line 645: This can be moved outside of the loops.

check.go
```diff
@@ -0,0 +1,1086 @@
...
+		}
+		return false, nil
+	}); errored {
+		return CheckReport{}, err
+	}
+
+	// If the driver can tell us about which layers it knows about, we should have previously
+	// examined all of them.  Any that we didn't are probably just wasted space.
+	// Note: if the driver doesn't support enumerating layers, it returns ErrNotSupported.
+	if err := s.startUsingGraphDriver(); err != nil {
```
mtrmac, line 665: Hum, should the graph lock be held for the lifetime of this function, to ensure a consistent view?

check.go
```diff
@@ -0,0 +1,1086 @@
...
+		}
+		// we've heard of both of them; the one that's on the end of a longer chain goes first
+		return depth(layersToDelete[i]) > depth(layersToDelete[j]) // closer-to-a-notional-base layers get removed later
+	})
+	// Now delete the layers that haven't been removed along with images.
+	for _, id := range layersToDelete {
+		if _, ok := deletedLayers[id]; ok {
+			continue
+		}
+		for _, reportedErr := range report.Layers[id] {
```
mtrmac, line 774: This could try to remove the same layer multiple times, is that intentional?

I’d expect a single attempt, at the driver or higher level depending on `isUnaccounted`.

mtrmac (commented):
This is a fairly brief single-pass skim; please treat it mostly as unbaked low-priority suggestions.

store.go
```diff
@@ -1215,7 +1222,18 @@ func (s *store) writeToImageStore(fn func() error) error {
 	return fn()
 }
 
-// writeToContainerStore is a convenience helper for working with store.getContainerStore():
+// readContainerStore is a convenience helper for working with store.containerStore:
+// It locks the store for reading, checks for updates, and calls fn(), which can then access store.containerStore.
+// It returns the return value of fn, or its own error initializing the store.
+func (s *store) readContainerStore(fn func() (bool, error)) (bool, error) {
+	if err := s.containerStore.startReading(); err != nil {
+		return true, err
```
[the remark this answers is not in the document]
mtrmac: https://github.com/containers/storage/pull/1577

PR #1571, opened 2023-04-14
types: try harder to use overlay with rootless

giuseppe:
if there are no configuration files present, attempt to use overlay for rootless if fuse-overlayfs is installed or if the kernel is >= 5.13.

Closes: https://github.com/containers/storage/issues/1570

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -305,7 +305,11 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
 	}
 	if opts.GraphDriverName == "" {
 		if len(systemOpts.GraphDriverPriority) == 0 {
-			opts.GraphDriverName = "vfs"
+			if canUseRootlessOverlay(opts.GraphRoot, opts.RunRoot) {
+				opts.GraphDriverName = overlayDriver
+			} else {
+				opts.GraphDriverName = "vfs"
+			}
 		} else {
 			opts.GraphDriverPriority = systemOpts.GraphDriverPriority
 		}
--- a/types/options_linux.go
+++ b/types/options_linux.go
@@ -1,5 +1,13 @@
 package types
 
+import (
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"golang.org/x/sys/unix"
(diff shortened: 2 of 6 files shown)


types/options_linux.go
```diff
@@ -12,3 +20,33 @@ const (
 var (
 	defaultOverrideConfigFile = "/etc/containers/storage.conf"
 )
+
+// canUseRootlessOverlay returns true if the overlay driver can be used for rootless containers
+func canUseRootlessOverlay(home, runhome string) bool {
+	// we check first for fuse-overlayfs since it is cheaper.
+	if path, _ := exec.LookPath("fuse-overlayfs"); path != "" {
+		//return true
```
rhatdan, line 28: Why is this commented out?

giuseppe: thanks for spotting it, it is a debugging artifact

rhatdan:
LGTM
@nalind @vrothberg @saschagrunert @flouthoc @umohnani8 @mtrmac PTAL

vrothberg (approved):
LGTM
@giuseppe any chance we can get a regression test here on in podman for it?

vrothberg (approved):
LGTM, thanks!

flouthoc (approved):
LGTM

nalind:
This is changing buildah from using vfs to using overlay when it's upgraded, including without a wipe in between.

types/options.go
```diff
@@ -305,7 +305,11 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
 	}
 	if opts.GraphDriverName == "" {
 		if len(systemOpts.GraphDriverPriority) == 0 {
-			opts.GraphDriverName = "vfs"
+			if canUseRootlessOverlay(opts.GraphRoot, opts.RunRoot) {
```
flouthoc, line 308: Should this check if last usage is already using `vfs` ? I think can be verified by checking if `GraphRoot` already contains a `vfs` directory or not. If yes then assume user is already using `vfs` and do not switch to overlay ?

PR #1572, opened 2023-04-14
Don't use bytes.NewBuffer to read data

mtrmac:
The documentation says
> The new Buffer takes ownership of buf, and the
> caller should not use buf after this call.

so use the more directly applicable, and simpler, bytes.Reader instead, to avoid this potentially risky use.


rhatdan:
LGTM

rhatdan:
@nalind @vrothberg @saschagrunert @flouthoc @umohnani8 @giuseppe  PTAL

nalind:
LGTM.

flouthoc (approved):
LGTM

vrothberg (approved):

PR #1577, opened 2023-04-17
Use Go 1.18 generics in the store’s locking helpers

mtrmac:
… avoiding quite a bit of the previously-required boilerplate.

This is probably easier to review as individual commits, and maybe with whitespace changes ignored.

@nalind PTAL

### DIFF
--- a/store.go
+++ b/store.go
@@ -1266,9 +1266,10 @@ func (s *store) readContainerStore(fn func() (bool, error)) (bool, error) {
 // writeToContainerStore is a convenience helper for working with store.containerStore:
 // It locks the store for writing, checks for updates, and calls fn(), which can then access store.containerStore.
 // It returns the return value of fn, or its own error initializing the store.
-func (s *store) writeToContainerStore(fn func() error) error {
+func writeToContainerStore[T any](s *store, fn func() (T, error)) (T, error) {
 	if err := s.containerStore.startWriting(); err != nil {
-		return err
+		var zeroRes T // A zero value of T
+		return zeroRes, err
 	}
 	defer s.containerStore.stopWriting()
 	return fn()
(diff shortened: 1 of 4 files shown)


giuseppe (approved):
LGTM

store.go
```diff
@@ -2088,12 +2076,11 @@ func (s *store) ContainerSize(id string) (int64, error) {
 		return -1, err
 	}
 
-	var res int64 = -1
-	err = s.writeToContainerStore(func() error { // Yes, s.containerStore.BigDataSize requires a write lock.
+	return writeToContainerStore(s, func() (int64, error) { // Yes, rcstore.BigDataSize requires a write lock.
```
nalind, line 2079: Where did `rcstore` come from?

store.go
```diff
@@ -2103,117 +2090,98 @@ func (s *store) ContainerSize(id string) (int64, error) {
...
 		if store.Exists(id) {
-			res = true
-			return true, nil
+			return true, true, nil
 		}
-		return false, nil
+		return false, false, nil
 	}); done {
-		return res
+		return res // false if err != nil
```
nalind, line 2169: Which `err` is this referring to, here and in various similar comments where the closure is calling `Exists()`?

nalind:
It is, thanks.  Generally looks good, though I'm mildly surprised that it doesn't make the logic that much shorter.
Defining a package-local `void` type when `struct{}` is widely used just makes me do a double-take when I read it.

store.go
mtrmac: Thanks, fixed. (It’s a mistake in rebasing, `rcstore` used to exist back when this branch was originally prepared.)

store.go
mtrmac: The one that was assigned to `_` when calling the locking wrapper…

It would be possible to explain that that in those comments, but at that point it seems simpler to just make the code explicit `if err { return false }; if found { return true }` now. So updated that way.

mtrmac:
Updated: I have changed my mind on `readContainerStore` returning an extra `bool`. In _most_ callers it causes extra boilerplate, but a few really do need to distinguish between a locking failure, item not found, and valid data. Also the symmetry with the other two `read*` helpers is a bit easier to follow in the more complex callers.

nalind:
> > Defining a package-local `void` type when `struct{}` is widely used just makes me do a double-take when I read it.
> 
> I chose this because I thought the slight readability benefit, making it very clear that there is nothing there, is worth it — and because I didn’t want to keep typing that `struct{}{}` — but that’s a very weakly held aesthetic opinion.
> 
> I fully realize that this is a fairly goofy approach, and I’m perfectly willing to drop that part.

Yes, please go ahead and drop that part.  I don't have an issue with returning an unused variable from a generic convenience function.

mtrmac:
`void` dropped.

store.go
```diff
@@ -1869,44 +1866,41 @@ func (s *store) Metadata(id string) (string, error) {
...
-			return true, nil
+			return size, true, nil
 		}
-		return false, nil
+		return -1, false, nil
 	}); done {
-		return res, err
+		if err != nil {
+			return -1, err
+		}
```
nalind, line 1891: If I'm reading it right, `err` is always `nil` here.

nalind:
One nit in `ImageBigDataSize`, otherwise LGTM.

store.go
mtrmac: `err` can be set if `readAllImageStores` fails in `imageStore.startReading`, e.g. on an I/O error.

rhatdan:
LGTM

PR #1578, opened 2023-04-19
store,imagestore: treat `graphRoot` as `rwImageStore` to support `Delete`

flouthoc:
PR https://github.com/containers/storage/pull/1549 added support for split-filesystem using `--imagestore` and when `--imagestore` is used current `graphRoot` is mounted as an additional store.

In such case the `graphRoot` must be mounted as an `rw` additional image store so users can still `Delete` images from the `graphRoot` if they want to .

### DIFF
--- a/store.go
+++ b/store.go
@@ -676,6 +676,7 @@ type store struct {
 	autoNsMinSize   uint32
 	autoNsMaxSize   uint32
 	imageStore      rwImageStore
+	rwImageStores   []rwImageStore
 	roImageStores   []roImageStore
 	containerStore  rwContainerStore
 	digestLockRoot  string
@@ -940,9 +941,22 @@ func (s *store) load() error {
 
 	for _, store := range driver.AdditionalImageStores() {
 		gipath := filepath.Join(store, driverPrefix+"images")
-		ris, err := newROImageStore(gipath)
-		if err != nil {
-			return err
+		var ris roImageStore
+		if s.imageStoreDir != "" && store == s.graphRoot {
+			// If --imagestore was set and current store
+			// is `graphRoot` then mount it as a `rw` additional
+			// store instead of `readonly` additional store.
+			imageStore, err := newImageStore(gipath)
+			if err != nil {
+				return err
+			}
+			s.rwImageStores = append(s.rwImageStores, imageStore)
+			ris = imageStore
+		} else {
+			ris, err = newROImageStore(gipath)
+			if err != nil {
+				return err
+			}
 		}
 		s.roImageStores = append(s.roImageStores, ris)
 	}
@@ -2479,8 +2493,28 @@ func (s *store) DeleteLayer(id string) error {
 func (s *store) DeleteImage(id string, commit bool) (layers []string, err error) {
 	layersToRemove := []string{}
 	if err := s.writeToAllStores(func(rlstore rwLayerStore) error {
-		if s.imageStore.Exists(id) {
-			image, err := s.imageStore.Get(id)
+		// Perform delete image on primary imageStore but if image
+		// is not found on primary store then try delete on any additional
+		// write-able image store.
+		imageStores := []rwImageStore{s.imageStore}
+		imageStores = append(imageStores, s.rwImageStores...)
+		imageFound := false
+		primaryImageStore := true
+		for _, imageStore := range imageStores {
+			if !imageStore.Exists(id) {
+				primaryImageStore = false
+				continue
+			}
+			imageFound = true
+			if !primaryImageStore {
+				// This is an additional writeable image store
+				// so we must perform lock
+				if err := imageStore.startWriting(); err != nil {
+					return err
+				}
+				defer imageStore.stopWriting()
+			}
+			image, err := imageStore.Get(id)
 			if err != nil {
 				return err
 			}
@@ -2496,7 +2530,7 @@ func (s *store) DeleteImage(id string, commit bool) (layers []string, err error)
 			if container, ok := aContainerByImage[id]; ok {
 				return fmt.Errorf("image used by %v: %w", container, ErrImageUsedByContainer)
 			}
-			images, err := s.imageStore.Images()
+			images, err := imageStore.Images()
 			if err != nil {
 				return err
 			}
@@ -2518,7 +2552,7 @@ func (s *store) DeleteImage(id string, commit bool) (layers []string, err error)
 				}
 			}
 			if commit {
-				if err = s.imageStore.Delete(id); err != nil {
+				if err = imageStore.Delete(id); err != nil {
 					return err
 				}
 			}
@@ -2563,7 +2597,8 @@ func (s *store) DeleteImage(id string, commit bool) (layers []string, err error)
 				layersToRemoveMap[layer] = struct{}{}
 				layer = parent
 			}
-		} else {
+		}
+		if !imageFound {
 			return ErrNotAnImage
 		}
 		if commit {
--- a/tests/split-store.bats
+++ b/tests/split-store.bats
@@ -89,10 +89,21 @@ load helpers
 	[[ "$output" =~ "Name: $name" ]]
 	[[ "$output" =~ "Data: random1" ]]
 	[[ "$output" =~ "Data: random2" ]]
-	# Since this image is being read from the readonly graph root
-	# so it must show that
-	[[ "$output" =~ "Read Only: true" ]]
 
 	# shutdown store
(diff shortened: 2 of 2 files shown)


tests/split-store.bats
```diff
@@ -89,9 +89,6 @@ load helpers
 	[[ "$output" =~ "Name: $name" ]]
 	[[ "$output" =~ "Data: random1" ]]
 	[[ "$output" =~ "Data: random2" ]]
-	# Since this image is being read from the readonly graph root
-	# so it must show that
-	[[ "$output" =~ "Read Only: true" ]]
```
giuseppe, line 94: Could we test it is writeable instead?

giuseppe (approved):
LGTM just one nit

tests/split-store.bats
flouthoc: Modified the test to verify if we can actually delete image from `graphRoot` while `--image-store` is still set.

giuseppe (approved):
LGTM

flouthoc:
@rhatdan @nalind PTAL

store.go
```diff
@@ -2479,8 +2493,28 @@ func (s *store) DeleteLayer(id string) error {
...
+		// write-able image store.
+		imageStores := []rwImageStore{s.imageStore}
+		imageStores = append(imageStores, s.rwImageStores...)
+		imageFound := false
+		primaryImageStore := true
+		for _, imageStore := range imageStores {
+			if !imageStore.Exists(id) {
+				primaryImageStore = false
+				continue
+			}
```
TomSweeneyRedHat, line 2507: If we go through this for loop 3 times, and the 2nd iteration we hit the continue , but the third we find it, imageFound will be set to true.  Is that valid in that circumstance?  (Stretching my store memory greatly).

flouthoc: Yes that sounds valid.

rhatdan:
LGTM

store.go
```diff
@@ -940,9 +941,22 @@ func (s *store) load() error {
...
 	for _, store := range driver.AdditionalImageStores() {
 		gipath := filepath.Join(store, driverPrefix+"images")
-		ris, err := newROImageStore(gipath)
-		if err != nil {
-			return err
+		var ris roImageStore
+		if s.imageStoreDir != "" && store == s.graphRoot {
+			// If --imagestore was set and current store
+			// is `graphRoot` then mount it as a `rw` additional
+			// store instead of `readonly` additional store.
```
mtrmac, line 948: The code says this already.

It would be much more useful to have a comment saying _why_ this is necessary.

store.go
```diff
@@ -676,6 +676,7 @@ type store struct {
 	autoNsMinSize   uint32
 	autoNsMaxSize   uint32
 	imageStore      rwImageStore
+	rwImageStores   []rwImageStore
```
mtrmac, line 679: If we need to special-case this situation, fine.

This `rwImageStores`:
- has only zero or one element
- does not directly correspond to any user-visible UI feature

So to me, that’s an a surprising way to record the data. It seems to me to be an unnecessary generalization just complicating the code right now.

_At the very least_ this field needs documentation as to how it relates to `imageStore`/`roImageStores` (it _looks_ as if it is some kind of direct sibling to `imageStore`/`roImageStore`, which is not the case). And I’d prefer the relevant stores to be somehow directly modeled here — “primary image store” / “graph-root image store” / “additional image stores as configured by user”, or something like that.

---

To my way of thinking, it’s surprising, inconsistent and hard to follow that the responsibility for `options.ImageStore` is so diffuse. This `store` object is responsible for creating the directories based on `options.ImageStore`, but somehow the individual graph driver is responsible for turning `options.ImageStore` into an entry in `AdditionalImageStores`, but it _must_ use precisely the `s.graphRoot` path, or the upper-level `store` breaks.

This is not at all a clean division of responsibility between the top-level store and the graph driver. I feel rather confident in predicting that it will lead to bugs.

(To be fair, all of the “additional * store” has similar design entanglement between `store` and the overlay driver; but this is expanding it to a much larger degree.)

---

Either this whole image store feature is a UI convenience over the existing “additional read-only stores” feature, and in that case it should exist ~entirely in the top-level objects close to the UI. (I think this should be the case, but that’s not true after this PR.) And then the graph driver should be _completely unaware_ of the `options.ImageStore` option.

Or it is really a cross-cutting feature across the stores and the drivers, and in that case I think it should be _explicitly modeled_ as Go struct fields, and methods that can fail. The top-level store should, at the very least, _know_ whether the graph driver implements the “graphRoot additional image store” (maybe by the driver implementing a method that returns a value about that). And then, again, the store can just instruct the driver to use an additional image store, without the driver explicitly checking `options.ImageStore`.

It’s a strong “smell” of unclean design to me that `store` and the drivers (only one of them!), within the same codebase, communicate via `fmt.Sprintf("AdditionalImageStore=%s", options.ImageStore)`. Why use text when we we have Go struct fields?

store.go
```diff
@@ -2479,8 +2493,28 @@ func (s *store) DeleteLayer(id string) error {
...
+		imageStores = append(imageStores, s.rwImageStores...)
+		imageFound := false
+		primaryImageStore := true
+		for _, imageStore := range imageStores {
+			if !imageStore.Exists(id) {
+				primaryImageStore = false
+				continue
+			}
+			imageFound = true
+			if !primaryImageStore {
```
mtrmac, line 2509: Shouldn’t this just be `if imageStore != s.imageStore`?

- That’s _much_ much more explicit about what we need to check for. The locking logic _really_ doesn’t care about which store was used first, or whether the image exists.
- It’s much shorter to implement
- This one is actually incorrect: If an image exists in all stores, `primaryImageStore` is never set to `false`, and the other stores incorrectly be used without locking.

store.go
```diff
@@ -2479,8 +2493,28 @@ func (s *store) DeleteLayer(id string) error {
...
-			image, err := s.imageStore.Get(id)
+		// Perform delete image on primary imageStore but if image
+		// is not found on primary store then try delete on any additional
+		// write-able image store.
+		imageStores := []rwImageStore{s.imageStore}
+		imageStores = append(imageStores, s.rwImageStores...)
+		imageFound := false
+		primaryImageStore := true
+		for _, imageStore := range imageStores {
+			if !imageStore.Exists(id) {
```
mtrmac, line 2504: This is using `imageStore`, if it is not `s.imageStore`, completely without any locking; that’s incorrect.

store.go
```diff
@@ -2479,8 +2493,28 @@ func (s *store) DeleteLayer(id string) error {
 func (s *store) DeleteImage(id string, commit bool) (layers []string, err error) {
 	layersToRemove := []string{}
 	if err := s.writeToAllStores(func(rlstore rwLayerStore) error {
-		if s.imageStore.Exists(id) {
-			image, err := s.imageStore.Get(id)
+		// Perform delete image on primary imageStore but if image
+		// is not found on primary store then try delete on any additional
+		// write-able image store.
+		imageStores := []rwImageStore{s.imageStore}
+		imageStores = append(imageStores, s.rwImageStores...)
```
mtrmac, line 2500: (There’s enough precedent in this file for `for _, store := range append([]someStore{value}, otherValues…)`, without the extra variable.)

store.go
```diff
@@ -2479,8 +2493,28 @@ func (s *store) DeleteLayer(id string) error {
 func (s *store) DeleteImage(id string, commit bool) (layers []string, err error) {
 	layersToRemove := []string{}
 	if err := s.writeToAllStores(func(rlstore rwLayerStore) error {
-		if s.imageStore.Exists(id) {
-			image, err := s.imageStore.Get(id)
+		// Perform delete image on primary imageStore but if image
+		// is not found on primary store then try delete on any additional
+		// write-able image store.
```
mtrmac, line 2498: This comment would, typically, also be of the “code says it already” kind — except that this one is factually incorrect: the code, as is, deletes the image from _all_ image stores where it is found.

store.go
```diff
@@ -2479,8 +2493,28 @@ func (s *store) DeleteLayer(id string) error {
...
+		imageStores := []rwImageStore{s.imageStore}
+		imageStores = append(imageStores, s.rwImageStores...)
+		imageFound := false
+		primaryImageStore := true
+		for _, imageStore := range imageStores {
```
mtrmac, line 2503: I would prefer just `store`, or maybe `is` or whatever, as the loop variable. It’s a _risk_ that `s.imageStore` and `imageStore` both exist, and mean different things.

flouthoc:
I'll open a new PR to address these refactor comments.

PR #1581, opened 2023-04-27
pkg/rm: attempt os.RemoveAll() first

giuseppe:
the mount.RecursiveUnmount() call can be expensive, since it needs to parse the mount table to reconstruct the mount tree.

Attempt the simple case first, and do the more expensive removal only if necessary.


giuseppe:
@kolyshkin PTAL

giuseppe:
rebased on top of https://github.com/containers/storage/pull/1585

kolyshkin (approved):
I can't think of why anyone else (including me) haven't done this yet.

LGTM

PR #1589, opened 2023-05-03
store: lock additional store before validating image

flouthoc:
* We must lock store before doing any operation if its not the default imagestore.
* Refactor and address some comments from: https://github.com/containers/storage/pull/1578

How to verify ?
Tests from https://github.com/containers/storage/blob/main/tests/split-store.bats must pass as-is.

### DIFF
--- a/store.go
+++ b/store.go
@@ -2470,28 +2470,22 @@ func (s *store) DeleteLayer(id string) error {
 func (s *store) DeleteImage(id string, commit bool) (layers []string, err error) {
 	layersToRemove := []string{}
 	if err := s.writeToAllStores(func(rlstore rwLayerStore) error {
-		// Perform delete image on primary imageStore but if image
-		// is not found on primary store then try delete on any additional
-		// write-able image store.
-		imageStores := []rwImageStore{s.imageStore}
-		imageStores = append(imageStores, s.rwImageStores...)
+		// Delete image from all available imagestores configured to be used.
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind @mtrmac @vrothberg PTAL

store.go
```diff
@@ -2473,25 +2473,21 @@ func (s *store) DeleteImage(id string, commit bool) (layers []string, err error)
 		// Perform delete image on primary imageStore but if image
 		// is not found on primary store then try delete on any additional
 		// write-able image store.
```
mtrmac, line 2475: I think this comment is still inaccurate, https://github.com/containers/storage/pull/1578#discussion_r1174138580 .

mtrmac (commented):
Thanks! Locking changes LGTM.

I still don’t understand, and can’t find any documentation, why we need this extra write functionality _only for deletes_ at all, (https://github.com/containers/storage/pull/1578#pullrequestreview-1396335642 ), but I guess I’l leave that to the involved parties to deal with.

store.go
flouthoc: Fixed.

flouthoc:
> why we need this extra write functionality only for deletes at all

@mtrmac I'll extend this for `tag/untag` as well, in a new PR. Following PR is just to address the locking race ( which seemed urgent) before anybody starts using this from upstream.

PR #1594, opened 2023-05-04
drivers/quota.openDir() log errno if C.opendir() fails

nalind:
If `C.opendir()` fails, include `errno` along with the path of the directory that we failed to open.


rhatdan:
LGTM

PR #1596, opened 2023-05-10
check: suppress certain errors when pull_options might cause them

nalind:
Files hard linked in from an OSTree repository won't tend to have the right ownership, permissions, or timestamps on them, so we have to accept that they'll frequently not match what we have on record when we're using one to speed up pulling images.

### DIFF
--- a/check.go
+++ b/check.go
@@ -74,6 +74,13 @@ type CheckOptions struct {
 	ContainerData               bool           // check that associated "big" data items are present and can be read
 }
 
+// checkIgnore is used to tell functions that compare the contents of a mounted
+// layer to the contents that we'd expect it to have to ignore certain
+// discrepancies
+type checkIgnore struct {
+	ownership, timestamps, permissions bool
+}
+
 // CheckMost returns a CheckOptions with mostly just "quick" checks enabled.
 func CheckMost() *CheckOptions {
 	return &CheckOptions{
@@ -125,10 +132,20 @@ func RepairEverything() *RepairOptions {
 // Check returns a list of problems with what's in the store, as a whole.  It can be very expensive
 // to call.
 func (s *store) Check(options *CheckOptions) (CheckReport, error) {
-	var ignoreChownErrors bool
+	var ignore checkIgnore
 	for _, o := range s.graphOptions {
-		if strings.Contains(o, "ignore_chown_errors") {
-			ignoreChownErrors = true
+		if strings.Contains(o, "ignore_chown_errors=true") {
+			ignore.ownership = true
+		}
+		if strings.HasPrefix(o, "force_mask=") {
+			ignore.permissions = true
+		}
+	}
+	for o := range s.pullOptions {
+		if strings.Contains(o, "use_hard_links") {
+			if s.pullOptions[o] == "true" {
+				ignore.timestamps = true
+			}
 		}
 	}
 
@@ -409,7 +426,7 @@ func (s *store) Check(options *CheckOptions) (CheckReport, error) {
 						return
 					}
 					// Every departure from our expectations is an error.
-					diffs := compareCheckDirectory(expectedCheckDirectory, actualCheckDirectory, ignoreChownErrors)
+					diffs := compareCheckDirectory(expectedCheckDirectory, actualCheckDirectory, ignore)
 					for _, diff := range diffs {
 						err := fmt.Errorf("%slayer %s: %s, %w", readWriteDesc, id, diff, ErrLayerContentModified)
 						if isReadWrite {
@@ -805,26 +822,25 @@ func (s *store) Repair(report CheckReport, options *RepairOptions) []error {
 }
 
 // compareFileInfo returns a string summarizing what's different between the two checkFileInfos
-func compareFileInfo(a, b checkFileInfo, ignoreChownErrors bool) string {
-	if a.typeflag == b.typeflag && a.uid != b.uid && a.gid != b.gid && a.size != b.size &&
-		(os.ModeType|os.ModePerm)&a.mode != (os.ModeType|os.ModePerm)&b.mode {
-		return ""
-	}
+func compareFileInfo(a, b checkFileInfo, ignore checkIgnore) string {
 	var comparison []string
 	if a.typeflag != b.typeflag {
 		comparison = append(comparison, fmt.Sprintf("filetype:%v￫%v", a.typeflag, b.typeflag))
 	}
-	if a.uid != b.uid && ignoreChownErrors {
+	if a.uid != b.uid && !ignore.ownership {
 		comparison = append(comparison, fmt.Sprintf("uid:%d￫%d", a.uid, b.uid))
 	}
-	if a.gid != b.gid && ignoreChownErrors {
+	if a.gid != b.gid && !ignore.ownership {
 		comparison = append(comparison, fmt.Sprintf("gid:%d￫%d", a.gid, b.gid))
 	}
 	if a.size != b.size {
 		comparison = append(comparison, fmt.Sprintf("size:%d￫%d", a.size, b.size))
 	}
-	if (os.ModeType|os.ModePerm)&a.mode != (os.ModeType|os.ModePerm)&b.mode {
+	if (os.ModeType|os.ModePerm)&a.mode != (os.ModeType|os.ModePerm)&b.mode && !ignore.permissions {
 		comparison = append(comparison, fmt.Sprintf("mode:%04o￫%04o", a.mode, b.mode))
+	}
+	if a.mtime != b.mtime && !ignore.timestamps {
+		comparison = append(comparison, fmt.Sprintf("mtime:0x%x￫0x%x", a.mtime, b.mtime))
 	}
 	return strings.Join(comparison, ",")
 }
@@ -835,6 +851,7 @@ type checkFileInfo struct {
 	uid, gid int
 	size     int64
 	mode     os.FileMode
+	mtime    int64 // unix-style whole seconds
 }
 
 // checkDirectory is a node in a filesystem record, possibly the top
@@ -845,7 +862,7 @@ type checkDirectory struct {
 }
 
 // newCheckDirectory creates an empty checkDirectory
-func newCheckDirectory(uid, gid int, size int64, mode os.FileMode) *checkDirectory {
+func newCheckDirectory(uid, gid int, size int64, mode os.FileMode, mtime int64) *checkDirectory {
 	return &checkDirectory{
 		directory: make(map[string]*checkDirectory),
 		file:      make(map[string]checkFileInfo),
@@ -855,6 +872,7 @@ func newCheckDirectory(uid, gid int, size int64, mode os.FileMode) *checkDirecto
 			gid:      gid,
 			size:     size,
 			mode:     mode,
+			mtime:    mtime,
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


check.go
```diff
@@ -805,7 +812,7 @@ func (s *store) Repair(report CheckReport, options *RepairOptions) []error {
 }
 
 // compareFileInfo returns a string summarizing what's different between the two checkFileInfos
-func compareFileInfo(a, b checkFileInfo, ignoreChownErrors bool) string {
+func compareFileInfo(a, b checkFileInfo, ignoreChownErrors, ignoreTimestamps, ignorePermissions bool) string {
 	if a.typeflag == b.typeflag && a.uid != b.uid && a.gid != b.gid && a.size != b.size &&
 		(os.ModeType|os.ModePerm)&a.mode != (os.ModeType|os.ModePerm)&b.mode {
```
mtrmac, line 817: Shouldn’t most of the `!=`s here be `==`?

check.go
```diff
@@ -125,12 +125,19 @@ func RepairEverything() *RepairOptions {
...
 func (s *store) Check(options *CheckOptions) (CheckReport, error) {
-	var ignoreChownErrors bool
+	var ignoreChownErrors, ignoreTimestamps, ignorePermissions bool
 	for _, o := range s.graphOptions {
 		if strings.Contains(o, "ignore_chown_errors") {
 			ignoreChownErrors = true
 		}
 	}
+	for o := range s.pullOptions {
+		if strings.Contains(o, "ostree_repos") || strings.Contains(o, "use_hard_links") {
```
mtrmac, line 135: The other caller consumes `use_hard_links` via `parseBooleanPullOption` (apparently allowing `"false"`), not just a presence check.

---

(It seems tempting to suggest that the individual `store.pullOptions` values should somehow only be accessed via dedicated accessors, to make the primary consumers and `check.go` clearly connected in the call graph… except that that would require a `c/storage/internal/$something` and probably isn’t worth it for the three options that currently exist.)

mtrmac (commented):
Mostly a quick local pass, I didn’t carefully re-read what `pkg/chunked` does and I might well be missing something about when the `ignore*` flags should be set.

Note the `!=` vs `==` question.

check.go
nalind: Yes, I think so.

check.go
```diff
@@ -805,7 +812,7 @@ func (s *store) Repair(report CheckReport, options *RepairOptions) []error {
 }
 
 // compareFileInfo returns a string summarizing what's different between the two checkFileInfos
-func compareFileInfo(a, b checkFileInfo, ignoreChownErrors bool) string {
+func compareFileInfo(a, b checkFileInfo, ignoreChownErrors, ignoreTimestamps, ignorePermissions bool) string {
 	if a.typeflag == b.typeflag && a.uid != b.uid && a.gid != b.gid && a.size != b.size &&
 		(os.ModeType|os.ModePerm)&a.mode != (os.ModeType|os.ModePerm)&b.mode {
 		return ""
```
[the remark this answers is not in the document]
nalind: Actually, I'll drop this bit.  If we benchmark this at some point, we can consider adding it back.

rhatdan:
@giuseppe PTAL

check.go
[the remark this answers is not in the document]
nalind: Even worse, I think this was masking a bug or two.

check.go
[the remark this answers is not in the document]
nalind: ID mapping does need to be taken account when SupportsShifting() is false, I'll add something for that.  I think you're right about not needing to relax checks on ownership and permissions for these cases.

rhatdan:
@mtrmac PTAL

check.go
```diff
@@ -125,10 +133,18 @@ func RepairEverything() *RepairOptions {
...
 // to call.
 func (s *store) Check(options *CheckOptions) (CheckReport, error) {
-	var ignoreChownErrors bool
+	var ignore checkIgnore
 	for _, o := range s.graphOptions {
 		if strings.Contains(o, "ignore_chown_errors") {
-			ignoreChownErrors = true
+			ignore.ownership = true
+		}
+		if strings.Contains(o, "force_mask") {
```
mtrmac, line 141: (Overlay also sets `force_mask` if `isNetworkFileSystem`. At some point it may be easiest to let the driver determine the right `checkIgnore` value.

Automatically-setting `force_mask` if `isNetworkFileSystem` only happens with a warning, so not handling that here seems fair enough.)

check.go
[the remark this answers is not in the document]
mtrmac: AFAICT setting `ostree_repos` without `use_hard_links` should preserve timestamps, so maybe that doesn’t need to be considered here at all.

check.go
```diff
@@ -398,6 +414,21 @@ func (s *store) Check(options *CheckOptions) (CheckReport, error) {
...
+					canUseShifting := true // mirror the logic in canUseShifting()
+					if !s.graphDriver.SupportsShifting() {
+						canUseShifting = false
+					}
+					if len(layer.UIDMap) != 0 && !idtools.IsContiguous(layer.UIDMap) {
+						canUseShifting = false
+					}
+					if len(layer.GIDMap) != 0 && !idtools.IsContiguous(layer.GIDMap) {
+						canUseShifting = false
+					}
```
mtrmac, line 427: Is there a reason not to actually call `canUseShifting` as is?

If the code can’t be shared, maybe also a comment in `canUseShifting` to also update this location might be warranted.

check.go
```diff
@@ -899,11 +930,20 @@ func (c *checkDirectory) add(path string, typeflag byte, uid, gid int, size int6
 	if components[len(components)-1] == "" {
 		components = components[:len(components)-1]
 	}
-	if typeflag == tar.TypeLink || typeflag == tar.TypeRegA {
+	if components[0] == "." {
+		components = components[1:]
+	}
+	if typeflag == tar.TypeRegA {
 		typeflag = tar.TypeReg
```
mtrmac, line 937: `archive/tar.Reader.next` can possibly turn `TypeRegA` into `TypeDir` based on trailing slash in `hdr.Name`. Is that applicable here?

check.go
```diff
@@ -971,6 +1011,30 @@ func (c *checkDirectory) header(hdr *tar.Header) {
 			c.remove(path.Join(dir, base[len(archive.WhiteoutPrefix):]))
 		}
 	} else {
+		if hdr.Typeflag == tar.TypeLink {
+			// look up the attributes of the target of the hard link
+			// n.b. by convention, Linkname is always relative to the
+			// root directory of the archive, which is not always the
+			// same as being relative to hdr.Name
+			directory := c
+			for _, component := range strings.Split(hdr.Linkname, "/") {
```
mtrmac, line 1020: This seems incorrect for
- targets that end with `/.` (the attributes are not set)
- targets with with `".."` elements, I think.
- targets that link to a “subdirectory within a file”, e.g. for an existing non-directory `a/b` and a link target `a/b/c` (incorrectly uses attributes of the last matching element)

Layers with the “subdirectory within a file” cases should be impossible to extract, so I’m not sure how much we care. If a subdirectory were corrupted on disk and replaced with a file, that should be detected when checking the link target’s tar header, shouldn’t it?

A `filepath.Clean` here would be consistent with how `pkg/archive.createTarFile` implicitly does one via `filepath.Join`.

check.go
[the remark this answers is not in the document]
nalind: Yeah, it does preserve them. `use_hard_links` by itself looks like the right condition to use here.

check.go
nalind: These either come from `Next()`, which won't give us a header with `TypeRegA` in it, or from `tar.FileInfoHeader()`, which either sets a different type flag or returns an error, so I guess not.

check.go
nalind: Adding a `path.Clean()` here, which should strip any final "/."s off of the `Linkname` value.
If a subdirectory were replaced with a file in a layer, we'd notice the difference due to its type being different from what we expected, if nothing else.
We should skip ".." components, too, because `path.Clean()` should be taking out any such components that aren't at the beginning of the path, but since it's a relative path, it isn't going to remove them if the path starts with "../".

check.go
```diff
@@ -894,31 +918,39 @@ func newCheckDirectoryFromDirectory(dir string) (*checkDirectory, error) {
 }
 
 // add adds an item to a checkDirectory
-func (c *checkDirectory) add(path string, typeflag byte, uid, gid int, size int64, mode os.FileMode) {
+func (c *checkDirectory) add(path string, typeflag byte, uid, gid int, size int64, mode os.FileMode, mtime int64) {
```
mtrmac, line 921: (Absolutely non-blocking: The UID…mTime fields seem to always come from a `*tar.Header`, so it might be possible to just pass that pointer instead of all the individual fields.

This works just fine as is.)

nalind: There's a lot more going on in a `tar.Header`, and while I'm wary of checking most of it, I'm particularly concerned with comparing timestamps with sub-second precision, which would be the natural thing to do with a header's `ModTime` field.  As it is, ISTR one of the older Linux filesystems only being precise to within two seconds, but I couldn't find a reference confirming that for anything other than FAT.

store.go
```diff
@@ -1319,10 +1319,12 @@ func (s *store) writeToAllStores(fn func(rlstore rwLayerStore) error) error {
 	return fn(rlstore)
 }
 
-// canUseShifting returns ???
+// canUseShifting returns true if we can use mount-time arguments (shifting) to
+// avoid having to create a mapped top layer for a base image when we want to
+// use it to create a container using ID mappings.
 // store must be locked for writing.
```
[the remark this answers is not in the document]
nalind: In practice, most the callers in `store` methods have a write lock on the layer store, and perhaps other locks, when they call into this function.  The caller in `Check()` needs to be updated to do the same.

check.go
mtrmac: > There's a lot more going on in a `tar.Header`

That’s a very fair point; and adding an _entirely new_ struct would be overkill.

---

> As it is, ISTR one of the older Linux filesystems only being precise to within two seconds, 

I think that was only FAT, at least the very first Minix filesystem already had 1 second.

(POSIX requires one second… but the kernel, for a long time, didn’t even provide causality: https://bugzilla.redhat.com/show_bug.cgi?id=244697 .)

store.go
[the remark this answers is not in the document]
mtrmac: (I understand very little about how the graph driver locking is supposed to work; my work mostly stopped at the C/I/L metadata store layer.)

Yes… right now `check` is holding a read lock on all layers, which is incompatible, but as long as a single `store` always uses the same graph driver and root, it seems that should be safe to get a read-write-locked layer store just to determine this value before starting the actual check.

check.go
nalind: > (POSIX requires one second… but the kernel, for a long time, didn’t even provide causality: https://bugzilla.redhat.com/show_bug.cgi?id=244697 .)

I don't have the proper words for how I'm reacting to learning this.

store.go
```diff
@@ -1216,6 +1216,54 @@ func writeToLayerStore[T any](s *store, fn func(store rwLayerStore) (T, error))
...
+// If all layer stores are processed without setting done == true, it returns ({}, false, nil).
+//
+// Typical usage:
+//
+//	if res, done, err := s.readOrWriteAllLayerStores(store, func(…) {
+//		…
+//	}; done {
+//		return res, err
+//	}
+func readOrWriteAllLayerStores[T any](s *store, fn func(store roLayerStore) (T, bool, error)) (T, bool, error) {
```
mtrmac, line 1238: (I was surprised, but this fits very well.)

mtrmac (approved):
LGTM.

mtrmac:
I’m afraid I somehow lost this comment yesterday:
- The `use_hard_links="false"` option seems to be parsed correctly, https://github.com/containers/storage/pull/1596#discussion_r1191510564 .

\*shrug\* this PR is a good improvement already.

nalind:
> I’m afraid I somehow lost this comment yesterday:
> 
>     * The `use_hard_links="false"` option seems to be parsed correctly, [check: suppress certain errors when pull_options might cause them #1596 (comment)](https://github.com/containers/storage/pull/1596#discussion_r1191510564) .

Tweaked that just a bit.

TomSweeneyRedHat:
LGTM and happy green test buttons.  Would like a head nod from @giuseppe, @vrothberg or @rhatdan too

vrothberg (approved):
LGTM

PR #1601, opened 2023-05-17
drivers/overlay.get(): only use the relative workdir for mountOverlayFrom()

nalind:
When we convert the work directory's location to a relative path for passing to mount-in-a-subprocess, don't change the value that we'll subsequently use when removing the "work/incompat/volatile" subdirectory while still in the parent process.  Fixes https://github.com/containers/buildah/issues/4794.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -202,6 +202,8 @@ func checkSupportVolatile(home, runhome string) (bool, error) {
 			if err = cachedFeatureRecord(runhome, feature, usingVolatile, ""); err != nil {
 				return false, fmt.Errorf("recording volatile-being-used status: %w", err)
 			}
+		} else {
+			usingVolatile = false
 		}
 	}
 	return usingVolatile, nil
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@mtrmac @vrothberg @giuseppe @flouthoc @saschagrunert PTAL

flouthoc (approved):
LGTM

drivers/overlay/overlay.go
```diff
@@ -202,6 +202,8 @@ func checkSupportVolatile(home, runhome string) (bool, error) {
 			if err = cachedFeatureRecord(runhome, feature, usingVolatile, ""); err != nil {
 				return false, fmt.Errorf("recording volatile-being-used status: %w", err)
 			}
+		} else {
```
mtrmac, line 205: AFAICS `doesVolatile` _never_ returns `false, nil`, currently. Shouldn’t it? Then this new error path could actually trigger a complete failure on unexpected problems.

nalind: I think it's an attempt to include a description of why it returns `false` in cases where it does so, though that suggests that it doesn't need to be returning two values, when returning just one, like `doesSupportNativeDiff()` does, would offer less room for ambiguity.

mtrmac: That makes sense — I have really no opinion on logging the failure or not.

I also know absolutely nothing about whether continuing without volatile support on failure (and recording the outcome in cache) is the right thing to do.

I have just locally noticed that the code path above, with `usingVolatile == false`, is dead code — and so we never call `cachedFeatureRecord(…"volatile", false)`. The second part might well be intended and correct.

nalind: In this specific case, the "volatile" option is being used to squeeze a little more speed out of writes to the mounted layer, as compared to not using the option.  I think the downside of not caching a `false` is having to do the test again on every startup if the kernel doesn't support the option at all.

PR #1603, opened 2023-05-18
store: ensure lockfile is updated before writes

giuseppe:
The lockfile's write record is now updated prior to the actual write operation.  This ensures that, in the event of an unexpected termination, other processes are correctly notified of an initiated write operation.


giuseppe:
@mtrmac PTAL

rhatdan:
@nalind PTAL

mtrmac (commented):
Good catch!

Code LGTM, but please add a comment in each instance, so that this is not inadvertently reverted.

giuseppe:
> Good catch!
> 
> Code LGTM, but please add a comment in each instance, so that this is not inadvertently reverted.

added the comment

mtrmac (commented):
LGTM. Thanks!

rhatdan:
LGTM

Issue #1394, opened 2022-10-17
Project Quota map is not thread-safe


Ezetowers:
# Summary
While using cri-o as container runtime, a [race condition](https://github.com/cri-o/cri-o/issues/6321) was found by the Golang compiler and the service crashed. It seems to be the error is associated with the lack of locking when the [quota map is accessed in the projectquota.go file](https://github.com/containers/storage/blob/main/drivers/quota/projectquota.go#L80) in this library.

Since this is a race condition, I am not able to create a set of steps to reproduce this issue, but if you think this problem belongsto this lib I can create a PR adding proper locking to the quota maps in the projectquota.go file

Kind Regards
Ezequiel

PR #1604, opened 2023-05-18
drivers/quota.Control: lock a mutex when called

nalind:
The "quotas" map and "next project ID" field don't take well to being created and updated concurrently in multiple threads.  Throw a lock around the whole object when any of its exported methods are called.  Fixes containers/container-libs#161.

### DIFF
--- a/drivers/quota/projectquota.go
+++ b/drivers/quota/projectquota.go
@@ -57,6 +57,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sync"
 	"syscall"
 	"unsafe"
 
@@ -80,6 +81,7 @@ type Quota struct {
 // Control - Context to be used by storage driver (e.g. overlay)
 // who wants to apply project quotas to container dirs
 type Control struct {
+	sync.Mutex        // Lock()ed during every call to an exported method
 	backingFsBlockDev string
 	nextProjectID     uint32
 	quotas            map[string]uint32
@@ -191,7 +193,8 @@ func NewControl(basePath string) (*Control, error) {
 // SetQuota - assign a unique project id to directory and set the quota limits
 // for that project id
 func (q *Control) SetQuota(targetPath string, quota Quota) error {
-
+	q.Lock()
+	defer q.Unlock()
 	projectID, ok := q.quotas[targetPath]
 	if !ok {
 		projectID = q.nextProjectID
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
At the local object level LGTM.

(Internal users don’t need this because the internal user is the overlay driver, called under the layer store lock — but CRI-O is calling `driver.ReadWriteDiskUsage`.)

---

Oh, actually external callers have _direct access_ to the graph driver via `store.GraphDriver`??! What does that mean for https://github.com/containers/storage/pull/1214 ? Do we need to re-add that, and add other synchronization on all the public methods of all drivers?

Does the `ReadWriteDiskUsage` need to care about concurrent layer updates / removals? I.e. should the actual locking happen at the `ReadWriteDiskUsage` API level, to synchronize with all other layer updates by the graph driver?

Or is the right decision to pretend that the public access to the graph driver mostly doesn’t exist, it’s the caller’s responsibility not to do that, and we treat `ReadWriteDiskUsage` as a special case highly-optimized for concurrency? If that’s the intent, I think it would be useful to have widespread comments pointing that special case out.

mtrmac:
(I appreciate that this fixes #1394 as is, and it’s better to have this PR than to have nothing. It’s just… a whole new can of worms I was not aware of.

Even if the individual disk usage computation was fine vs. concurrent updates/removals, and it would be fine to return such mostly-invalid data, I’m at least worried about things like `Driver.dir` not being _aware_ of the need to correctly work outside of the usual locks.

nalind:
It's not exposed through the `Store` object.  I think this is about library consumers calling into the `quota` package directly, which is something podman has apparently been doing for a while.

nalind:
Rebased.

mtrmac:
The CRI-O backtrace in the original report points at https://github.com/cri-o/cri-o/blob/80228a5d6a4b26e9ed2363ca6a936f5fb604c607/internal/lib/stats/stats_server.go#L189-L197 , which uses a published method of `Store`: https://github.com/containers/storage/blob/9e425af19496438e8f50f25494d2771166a3c2b5/store.go#L154 .

(The Podman user I can find is https://github.com/containers/podman/blob/b15510694b81929dbc107d354fdb0a83d0bfaa2a/libpod/runtime_volume_common.go#L181-L197 , but that’s a private single-use `quota.Control`, so I can’t see that that one requires locking.)

mtrmac:
WDYT about marking `Store.GraphDriver` deprecated?

That would require giving CRI-O a new `Store` API to get the container size, and that could lock via `readAllLayerStores`. And something to expose `Driver.Metadata` for Podman’s inspect functions.

jankaluza:

PR #1606, opened 2023-05-19
Don't silently ignore errors decoding mountpoints.json

mtrmac:
Other `Unmarshal` calls have already been modified similarly.

This “should not matter” in practice, but we are seeing incorrect mount counts in https://github.com/containers/podman/issues/17216 .

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -864,33 +864,35 @@ func (r *layerStore) loadMounts() error {
 		return err
 	}
 	layerMounts := []layerMountPoint{}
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@nalind @flouthoc @giuseppe @vrothberg @saschagrunert PTAL

saschagrunert (approved):

vrothberg (approved):
LGTM

PR #1607, opened 2023-05-19
Fail on unmount/rmdir in overlay.Driver.Put

mtrmac:
… instead of only logging a debug-level failure.

We are seeing layers being mounted while our metadata claims that they are unmounted, and this seems to be a way how that could happen without leaving a visible trace.

It's unclear to me why this silently succeeds; this behavior is very long-standing.

Motivated by https://github.com/containers/podman/issues/17216 , where we seem to be seeing mounted layers with zero count in `mount points.json`.

⚠️ I don’t really know what I’m doing, there might be a good reason for the current behavior. Alternatively, I’d like the `Debug` logs to be elevated to at least a `Warning`, possibly outright `Error`. @nalind @giuseppe PTAL.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1770,11 +1770,13 @@ func (d *Driver) Put(id string) error {
 	if !unmounted {
 		if err := unix.Unmount(mountpoint, unix.MNT_DETACH); err != nil && !os.IsNotExist(err) {
 			logrus.Debugf("Failed to unmount %s overlay: %s - %v", id, mountpoint, err)
+			return fmt.Errorf("unmounting %q: %w", mountpoint, err)
 		}
 	}
 
 	if err := unix.Rmdir(mountpoint); err != nil && !os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


mtrmac:
I have filed https://github.com/containers/podman/pull/18634 to trigger a Podman test run at least, to see if there is something very obvious I’m missing.

rhatdan:
@giuseppe PTAL

giuseppe (approved):
I don't know how these failures could happen, since it is performing a lazy umount, but I agree it is better to know if something goes wrong so we know what to look at.

LGTM

mtrmac:
> I don't know how these failures could happen, since it is performing a lazy umount

I don’t know that either, but I _am_ seeing EINVAL on unmount in https://github.com/containers/podman/pull/18634 . (At this point I have no idea whether that’s a long-standing bug to be fixed, a failure that should be ignored, or a short-term fallout of some unrelated bug.)

rhatdan:
LGTM

PR #1612, opened 2023-05-23
WIP: [1.37] Always close pgzip writer on `applyDiffWithOptions`

saschagrunert:
There are cases where the writer will not be closed because we return earlier, which may cause goroutines leaking in CRI-O.

Refers to https://issues.redhat.com/browse/OCPBUGS-10290

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -1554,8 +1554,9 @@ func (r *layerStore) applyDiffWithOptions(to string, layerOptions *LayerOptions,
 	tsdata := bytes.Buffer{}
 	compressor, err := pgzip.NewWriterLevel(&tsdata, pgzip.BestSpeed)
 	if err != nil {
-		compressor = pgzip.NewWriter(&tsdata)
+		return -1, err
 	}
+	defer compressor.Close()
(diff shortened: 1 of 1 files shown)


shitou12345678:
We has the same problem https://github.com/containers/storage/issues/1623, hoping it can be resolved as soon as possible.

rhatdan:
@saschagrunert Still working on this one?

saschagrunert:
@rhatdan yes we're waiting for customer feedback on it.

layers.go
```diff
@@ -1554,8 +1554,9 @@ func (r *layerStore) applyDiffWithOptions(to string, layerOptions *LayerOptions,
 	tsdata := bytes.Buffer{}
 	compressor, err := pgzip.NewWriterLevel(&tsdata, pgzip.BestSpeed)
 	if err != nil {
-		compressor = pgzip.NewWriter(&tsdata)
+		return -1, err
 	}
+	defer compressor.Close()
```
mtrmac, line 1559: As @shitou12345678 points out in #1623 , this is incorrect: `.Close` triggers compression of `compressor.currentBuffer`, and writes a trailer. That must happen before consuming `tsdata`.

mtrmac:
Compare #1641 .

Issue #1614, opened 2023-05-24
ci: golangci-lint is not working


kolyshkin:
Found this in CI (e.g. https://cirrus-ci.com/task/4877586900189184):

```
make TAGS=regex_precompile local-validate && make lint && make clean
make -C tests/tools
make[1]: Entering directory '/go/src/github.com/containers/storage/tests/tools'
export \
	VERSION=v1.51.2 \
	URL=https://raw.githubusercontent.com/golangci/golangci-lint \
	BINDIR=build && \
curl -sfL $URL/$VERSION/install.sh | sh -s $VERSION
golangci/golangci-lint info checking GitHub for tag 'v1.51.2'
golangci/golangci-lint info found version: 1.51.2 for v1.51.2/linux/amd64
golangci/golangci-lint info installed build/golangci-lint
make[1]: Leaving directory '/go/src/github.com/containers/storage/tests/tools'
fatal: Not a valid object name main
make -C tests/tools
make[1]: Entering directory '/go/src/github.com/containers/storage/tests/tools'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/go/src/github.com/containers/storage/tests/tools'
tests/tools/build/golangci-lint run --build-tags=" libdm_no_deferred_remove  "
level=warning msg="[linters_context] running gomoddirectives failed: failed to get module file: command go list: exit status 1: go: list -m cannot be used with GO111MODULE=off\n: if you are not using go modules it is suggested to disable this linter"
rm -f -f containers-storage containers-storage.* docs/*.1 docs/*.5
```

and it does not work locally either:

```console
[kir@kir-rhat storage]$ make lint
make -C tests/tools
make[1]: Entering directory '/home/kir/go/src/github.com/containers/storage/tests/tools'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/home/kir/go/src/github.com/containers/storage/tests/tools'
tests/tools/build/golangci-lint run --build-tags="  libsubid "
ERRO Running error: no such linter "cyclop"       
make: *** [Makefile:90: lint] Error 3
```

kolyshkin:
I'll work on fixing this.

PR #1615, opened 2023-05-24
Simplify Makefiles, make golangci-lint work again

kolyshkin:
This simplifies Makefiles, removing some old hacks, and makes golangci-linter work again. Please see individual commits for details.

Fixes: #1614

TODO (not in this PR):
 - re-add the disabled linters, going back to (at least) golangci-lint default set;
 - get rid of tests/tools entirely.


kolyshkin:
OK, golangci-lint is now working: https://cirrus-ci.com/task/6600612519870464 (check "Run test").

TomSweeneyRedHat:
@cevich PTAL

Makefile
```diff
@@ -3,20 +3,31 @@ export GOPROXY=https://proxy.golang.org
...
 	local-test-integration \
 	local-test-unit \
 	local-validate \
-	lint \
-	vendor
+	test-integration \
+	test-unit \
+	validate \
+	vendor \
+	vendor-in-container
```
TomSweeneyRedHat, line 27: I'm not sure we need to include vendor-in-container.

cevich: Ya we do, it's a phony target - Meaning, it doesn't create a file called `vendor-in-container`.

cevich:
Wow @kolyshkin this is really nice work and a very welcome set of improvements.  The only thing I'm slightly unsure of is the `GO111MODULES` change...

@TomSweeneyRedHat Do we need to maintain compatibility with (very) old golang due to backport possibility for some ancient (still supported) release?  I'm guessing not (they're probably pinned to specific commits), but wanted to double-check.

TomSweeneyRedHat:
Thanks for the eyeballs @cevich 

As far as the old ancient Go goes, it's not outside of the realm of possibilities that we'd need to backport to an older version, but I'd say it's not very likely.  I wouldn't make older go a gating factor for this nor any other PR.

cevich (approved):
Thanks Tom.  K, LGTM then.

kolyshkin:
> Wow @kolyshkin this is really nice work and a very welcome set of improvements. The only thing I'm slightly unsure of is the `GO111MODULES` change...

This is not needed since about go 1.14 (which is kinda ancient), and we've already made changes like this in other github.com/containers/* repos, e.g. https://github.com/containers/common/pull/993

rhatdan:
LGTM

PR #1616, opened 2023-05-24
.cirrus.yml: fix a warning

kolyshkin:
Cirrus CI emits a bunch of warnings like this one:

> task "success" depends on task "fedora-38 vfs", but their only_if conditions are different

~~To fix, use "skip" instead of "only_if". The difference between the two is described at
https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution~~

This was probably a copy-paste leftover from another repo, which uses CI:DOCS
magic to skip some tests. In this repo, it is not used, so remove the condition.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -56,7 +56,6 @@ gce_instance:
 linux_testing: &linux_testing
     depends_on:
         - lint
-    only_if: $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
     gce_instance:  # Only need to specify differences from defaults (above)
         image_name: "${VM_IMAGE}"
 


flouthoc (approved):
LGTM

TomSweeneyRedHat:
@cevich PTAL
LGTM, but want a head nod from Chris

cevich:
Oof, this is confusing: None of the other testing tasks have this condition on them.  WDYT about just removing it entirely?

kolyshkin:
> Oof, this is confusing: None of the other testing tasks have this condition on them. WDYT about just removing it entirely?

I barely remember this was made so some (expensive) tests can be skipped if only documentation changes are being made.

Found this in a neighbor repo:

https://github.com/containers/podman/blob/47ac6c4cbe1fea2f78b1f72937b43a8c6be28420/CONTRIBUTING.md?plain=1#L180C1-L182

I guess, yes, if this test skipping feature is not needed here, we can remove it entirely.

cevich:
Yeah, this and several other magic strings are used extensively in the podman and a few other repos.  But in this repo, I count a total of 7 (closed) PRs using it, out of thousands.  So I think it's safe to just remove it.  That way no maintainer will ever question it's apparently partial implementation.  Thanks for the effort though, trying to clean up all those annoying annotations.  I appreciate the help.

kolyshkin:
> Yeah, this and several other magic strings are used extensively in the podman and a few other repos. But in this repo, I count a total of 7 (closed) PRs using it, out of thousands. So I think it's safe to just remove it. That way no maintainer will ever question it's apparently partial implementation. Thanks for the effort though, trying to clean up all those annoying annotations. I appreciate the help.

updated; ptal

kolyshkin:
CI failure seems to be unrelated

cevich (approved):
LGTM.  Thanks again.

cevich:
I just hit "re-run" on all the failures.

PR #1618, opened 2023-05-24
options: keep using prior drivers if found

flouthoc:
There is no need for `vfs` to be the default storage driver since kernel
`>= 5.13` supports `overlay` natively however there is use-case for users
who don't had any configs and they started using `vfs` in a default manner following check is a hack to keep `buildah` and `podman` working for such users.

See: https://github.com/containers/storage/pull/1571 for prior discussions.

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -10,6 +10,8 @@ import (
 	"time"
 
 	"github.com/BurntSushi/toml"
+	drivers "github.com/containers/storage/drivers"
+	_ "github.com/containers/storage/drivers/register"
 	cfg "github.com/containers/storage/pkg/config"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/sirupsen/logrus"
@@ -310,10 +312,21 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
 	}
 	if opts.GraphDriverName == "" {
 		if len(systemOpts.GraphDriverPriority) == 0 {
-			if canUseRootlessOverlay(opts.GraphRoot, opts.RunRoot) {
-				opts.GraphDriverName = overlayDriver
-			} else {
-				opts.GraphDriverName = "vfs"
+			driversMap := drivers.ScanPriorDrivers(opts.GraphRoot)
+
+			for _, name := range drivers.Priority {
+				if _, prior := driversMap[name]; prior {
+					opts.GraphDriverName = name
+					break
+				}
+			}
+
+			if opts.GraphDriverName == "" {
+				if canUseRootlessOverlay(opts.GraphRoot, opts.RunRoot) {
+					opts.GraphDriverName = overlayDriver
+				} else {
+					opts.GraphDriverName = "vfs"
+				}
 			}
 		} else {
 			opts.GraphDriverPriority = systemOpts.GraphDriverPriority
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -342,12 +342,12 @@ func New(name string, config Options) (Driver, error) {
 	}
 
 	// Guess for prior driver
-	driversMap := scanPriorDrivers(config.Root)
+	driversMap := ScanPriorDrivers(config.Root)
 
 	// use the supplied priority list unless it is empty
 	prioList := config.DriverPriority
 	if len(prioList) == 0 {
-		prioList = priority
+		prioList = Priority
 	}
 
 	for _, name := range prioList {
@@ -419,12 +419,12 @@ func isDriverNotSupported(err error) bool {
 }
 
 // scanPriorDrivers returns an un-ordered scan of directories of prior storage drivers
-func scanPriorDrivers(root string) map[string]bool {
+func ScanPriorDrivers(root string) map[string]bool {
 	driversMap := make(map[string]bool)
 
 	for driver := range drivers {
 		p := filepath.Join(root, driver)
-		if _, err := os.Stat(p); err == nil && driver != "vfs" {
+		if _, err := os.Stat(p); err == nil {
 			driversMap[driver] = true
 		}
 	}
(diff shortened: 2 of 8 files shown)


flouthoc:
@nalind @giuseppe PTAL , this is a hack to prevent breaking old users, was not able to think of any other cheap way to prevent this break.

types/options.go
```diff
@@ -315,6 +315,15 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
 			} else {
 				opts.GraphDriverName = "vfs"
 			}
+
+			// Note: This is hack is to prevent old users from breaking.
+			// Reason: There can be users who defaulted to `vfs` and have
+			// already generated working content with `vfs` driver in such
+			// case `storage` will not silently default to `overlay`.
+			_, err := os.Stat(filepath.Join(opts.GraphRoot, "vfs"))
```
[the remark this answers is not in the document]
flouthoc: Yes sounds better to me.

nalind:
There's a whole `scanPriorDrivers()` function in `drivers/drivers.go`.  Can we find a way to refrain from bypassing it?

flouthoc:
> There's a whole scanPriorDrivers() function in drivers/drivers.go. Can we find a way to refrain from bypassing it?

Ah it does a similar check, I think its a good idea if we can find a way so `c/storage` does not bypass it while checking config. Let me take a look.

giuseppe:
we may need to tweak `scanPriorDrivers()` though as it has an explicit check to skip `vfs`, looking at the only caller of the function it should be fine to just drop the `driver != "vfs"` check

flouthoc:
@giuseppe @nalind PTAL, it seems modifying behavior of `New(` could again break API users, so instead I think adding a similar check in types is safer way. Which will be only triggered if no selection of driver was set by user and `c/storage` is trying to find a default driver, in such case it will first check if any `prior` driver was set.

types/options.go
```diff
@@ -310,10 +311,27 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
...
 		if len(systemOpts.GraphDriverPriority) == 0 {
-			if canUseRootlessOverlay(opts.GraphRoot, opts.RunRoot) {
-				opts.GraphDriverName = overlayDriver
-			} else {
-				opts.GraphDriverName = "vfs"
+			driversMap := drivers.ScanPriorDrivers(opts.GraphRoot)
+
+			// use the supplied priority list unless it is empty
+			prioList := systemOpts.GraphDriverPriority
+			if len(prioList) == 0 {
```
giuseppe, line 318: how can this be longer than 0 if we just checked for it at line 313?

flouthoc: This check is not needed. I missed it while copying it from drivers.

flouthoc:
cc @edsantiago I think this will unblock buildah CI.

giuseppe (approved):
LGTM

nalind:
`types/options.go` will need to add `_ "github.com/containers/storage/drivers/register"` to its list of imports, otherwise its first call to `ScanPriorDrivers()` appears to be iterating through the map of registered drivers before that map is populated.  LGTM once that's changed.

rhatdan:
LGTM

Luap99:
This change is causing a huge bloat for rootlessport (size increase is 3 MB) in podman. Making storage/types depend on the drivers causes a new gigantic transitive dependency chain.

But more importantly this now causes build issues because it tries to build the devmapper driver even on systems without the device-mapper-devel installed, of course the correct thing is to use `exclude_graphdriver_devicemapper` but that wasn't necessary before which means likely other users who only depend on `github.com/containers/storage/types` will run into the same problem.

And to clarify rootlessport does not depend on `c/storage/types` directly. It only needs `c/common/pkg/config` but this package depends on `c/storage/types` which means everyone depending only on `c/common/pkg/config` and not the storage drivers will not suffer a almost 3 MB bloat in binary size.

Is there any way we can move this check into another package? Shouldn't this be decided in `driver.New()` and not in types?

rhatdan:
@flouthoc PTAL 

We should have never allowed the change in types/options.go

Lets revert this change and rework it.

```
@@ -10,6 +10,8 @@ import (
	"time"

	"github.com/BurntSushi/toml"
	drivers "github.com/containers/storage/drivers"
	_ "github.com/containers/storage/drivers/register"
	cfg "github.com/containers/storage/pkg/config"
	"github.com/sirupsen/logrus"
```

flouthoc:
@rhatdan We can move `getRootlessStorageOpts` to a different pkg , that should resolve this.

flouthoc:
I think moving entire check to `Drivers.New(` would break the API which will return empty driver when no driver is selected in the config and if we default to only `vfs` then we will have to hack `Drivers.New` to support native overlay again.

A simpler solution could be to create a function similar to `ScanPriorDrivers` for `types/options` which will only work on `text`  filtering from a list of drivers ( e.g `[]string{"overlay", "vfs" ... }` ) instead of going through a map of registered drivers.

This will remove `types/options.go` dependency on `drivers` and `drivers` can still keep using older code.

@containers/storage-maintainers @Luap99  WDYT ?

flouthoc:
@Luap99 @containers/storage-maintainers This is patch which i want to propose https://github.com/containers/storage/commit/0c6fcf47000811a4cff2527f0b1795d4c36871ca

flouthoc:
Nvm I think there is another simpler patch suggested by @giuseppe , https://github.com/containers/storage/commit/d7a6662e428d375925d520dde3cfcdd85140dc57 on the branch https://github.com/flouthoc/storage/tree/use-readdir-prior, this should do the trick as well.

flouthoc:
@Luap99 Following PR should address this https://github.com/containers/storage/pull/1637

PR #1621, opened 2023-05-25
ci: enable govet and gosimple linters, fix found bugs

kolyshkin:
Found/fixed 3 issues with govet and 3 issues with gosimple.

### DIFF
--- a/pkg/regexp/regexp.go
+++ b/pkg/regexp/regexp.go
@@ -11,7 +11,7 @@ import (
 // of apps that want to use global regex variables. This library initializes them on
 // first use as opposed to the start of the executable.
 type Regexp struct {
-	once   sync.Once
+	once   *sync.Once
 	regexp *regexp.Regexp
 	val    string
 }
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM
@nalind @giuseppe @vrothberg @mtrmac @umohnani8 @saschagrunert PTAL

saschagrunert (approved):

nalind:
LGTM.

pkg/regexp/regexp.go
```diff
@@ -22,7 +22,10 @@ func Delayed(val string) Regexp {
 	}
 	if precompile {
 		re.regexp = regexp.MustCompile(re.val)
+	} else {
+		re.once = &sync.Once{}
 	}
+
```
mtrmac, line 28: Having multiple `Regexp` values pointing to a shared `once` doesn’t make sense.

Instead, make `type Regexp` a `*struct`, and avoid the copy that way.

mtrmac: Filed #1636 .

Issue #1619, opened 2023-05-24
RFC: remove devmapper driver


kolyshkin:
It seems that no one is using devicemapper storage driver nowadays, since overlayfs is so much better.

Can it be deprecated and removed? I see that moby/moby did that recently.

rhatdan:
Makes sense to me.
@nalind @rhvgoyal @vrothberg @giuseppe @mtrmac @saschagrunert WDYT?

giuseppe:
yes, I agree, let's drop old unmaintained stuff

saschagrunert:
I agree we should remove it if nobody maintains it any more.

vrothberg:
SGTM :+1:

PR #1622, opened 2023-05-26
Remove devicemapper storage driver

kolyshkin:
Fixes: #1619


kolyshkin:
Surely, we need a new major version after this PR.

rhatdan:
I think we would need to go to podman 5.0 before we merge this in.

TomSweeneyRedHat:
Thoughts on migration from v4.0 to v5.0 when/if this comes into play?

kolyshkin:
> I think we would need to go to podman 5.0 before we merge this in.

Podman is not compiling in devmapper driver since v0.9.2 (https://github.com/containers/podman/pull/1445). The driver can be compiled in (I checked that), but no one knows if it works at all (I haven't checked).

rhatdan:
Yes I agree this is  not an issue for Podman. So I say we move forward on removing it.

saschagrunert (approved):

kolyshkin:
Need to remove it from the major users first:

* podman: is not compiling in devicemapper since v0.9.2 (containers/podman#1445).
* cri-o: https://github.com/cri-o/cri-o/issues/7002 / https://github.com/cri-o/cri-o/pull/7003
* buildah: https://github.com/containers/buildah/pull/4832/

kolyshkin:
Currently, the only user is cri-o, so it needs to be removed from it first.

The biggest roadblock seems to be https://github.com/cri-o/cri-o/pull/7003#issuecomment-1640106231. It's hard to tell if there are cri-o users who use devmapper (I hope not, but I don't know how to verify it). @haircommander maybe you can shed some light?

rhatdan:
I would be shocked if there were any. Only possible way would be on an ancient 3.11 version, which will never get a new containers/storage version.

haircommander:
@fidencio @littlejawa last I remembered there was a kata case for devmapper, are we still using it there?

littlejawa:
Talked with @fidencio about this - it seems our dependency is gone. We have an alternative to device mapper, so it can be removed.

kolyshkin:
> Talked with @fidencio about this - it seems our dependency is gone. We have an alternative to device mapper, so it can be removed.

Awesome! I will revive this

PR #1627, opened 2023-06-01
chunked: generate tar-split as part of zstd:chunked

giuseppe:
change the file format to store the tar-split as part of the zstd:chunked image.  This will allow clients to rebuild the entire tarball without having to download it fully.

also store the uncompressed digest for the tarball, so that it can be stored into the storage database.

Needs: https://github.com/containers/image/pull/1976

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1952,6 +1952,9 @@ func (d *Driver) ApplyDiffFromStagingDirectory(id, parent, stagingDirectory stri
 	if err := os.RemoveAll(diff); err != nil && !os.IsNotExist(err) {
 		return err
 	}
+
+	diffOutput.UncompressedDigest = diffOutput.TOCDigest
+
 	return os.Rename(stagingDirectory, diff)
 }
 
--- a/layers.go
+++ b/layers.go
@@ -2392,8 +2392,26 @@ func (r *layerStore) ApplyDiffFromStagingDirectory(id, stagingDirectory string, 
 	layer.UncompressedDigest = diffOutput.UncompressedDigest
 	layer.UncompressedSize = diffOutput.Size
 	layer.Metadata = diffOutput.Metadata
-	if err = r.saveFor(layer); err != nil {
-		return err
+	if len(diffOutput.TarSplit) != 0 {
+		tsdata := bytes.Buffer{}
+		compressor, err := pgzip.NewWriterLevel(&tsdata, pgzip.BestSpeed)
+		if err != nil {
+			compressor = pgzip.NewWriter(&tsdata)
+		}
+		if err := compressor.SetConcurrency(1024*1024, 1); err != nil { // 1024*1024 is the hard-coded default; we're not changing that
+			logrus.Infof("setting compression concurrency threads to 1: %v; ignoring", err)
+		}
+		if _, err := compressor.Write(diffOutput.TarSplit); err != nil {
+			compressor.Close()
+			return err
+		}
+		compressor.Close()
+		if err := os.MkdirAll(filepath.Dir(r.tspath(layer.ID)), 0o700); err != nil {
+			return err
+		}
+		if err := ioutils.AtomicWriteFile(r.tspath(layer.ID), tsdata.Bytes(), 0o600); err != nil {
+			return err
+		}
 	}
 	for k, v := range diffOutput.BigData {
 		if err := r.SetBigData(id, k, bytes.NewReader(v)); err != nil {
@@ -2402,6 +2420,9 @@ func (r *layerStore) ApplyDiffFromStagingDirectory(id, stagingDirectory string, 
 			}
 			return err
 		}
+	}
+	if err = r.saveFor(layer); err != nil {
+		return err
 	}
 	return err
 }
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -523,7 +523,7 @@ func unmarshalToc(manifest []byte) (*internal.TOC, error) {
 		for iter.ReadArray() {
 			for field := iter.ReadObject(); field != ""; field = iter.ReadObject() {
 				switch field {
-				case "type", "name", "linkName", "digest", "chunkDigest", "chunkType":
+				case "type", "name", "linkName", "digest", "chunkDigest", "chunkType", "modtime", "accesstime", "changetime":
 					count += len(iter.ReadStringAsSlice())
 				case "xattrs":
 					for key := iter.ReadObject(); key != ""; key = iter.ReadObject() {
@@ -574,7 +574,7 @@ func unmarshalToc(manifest []byte) (*internal.TOC, error) {
 					m.UID = iter.ReadInt()
 				case "GID":
 					m.GID = iter.ReadInt()
-				case "ModTime":
+				case "modtime":
 					time, err := time.Parse(time.RFC3339, byteSliceAsString(iter.ReadStringAsSlice()))
 					if err != nil {
 						return nil, err
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -516,14 +516,14 @@ func unmarshalToc(manifest []byte) (*internal.TOC, error) {
 
 	iter := jsoniter.ParseBytes(jsoniter.ConfigFastest, manifest)
 	for field := iter.ReadObject(); field != ""; field = iter.ReadObject() {
-		if field != "entries" {
+		if strings.ToLower(field) != "entries" {
 			iter.Skip()
 			continue
 		}
 		for iter.ReadArray() {
 			for field := iter.ReadObject(); field != ""; field = iter.ReadObject() {
-				switch field {
-				case "type", "name", "linkName", "digest", "chunkDigest", "chunkType", "modtime", "accesstime", "changetime":
+				switch strings.ToLower(field) {
+				case "type", "name", "linkname", "digest", "chunkdigest", "chunktype", "modtime", "accesstime", "changetime":
 					count += len(iter.ReadStringAsSlice())
 				case "xattrs":
 					for key := iter.ReadObject(); key != ""; key = iter.ReadObject() {
@@ -548,31 +548,31 @@ func unmarshalToc(manifest []byte) (*internal.TOC, error) {
 
 	iter = jsoniter.ParseBytes(jsoniter.ConfigFastest, manifest)
 	for field := iter.ReadObject(); field != ""; field = iter.ReadObject() {
-		if field == "version" {
+		if strings.ToLower(field) == "version" {
 			toc.Version = iter.ReadInt()
 			continue
 		}
-		if field != "entries" {
+		if strings.ToLower(field) != "entries" {
 			iter.Skip()
 			continue
 		}
 		for iter.ReadArray() {
 			var m internal.FileMetadata
 			for field := iter.ReadObject(); field != ""; field = iter.ReadObject() {
-				switch field {
+				switch strings.ToLower(field) {
 				case "type":
 					m.Type = getString(iter.ReadStringAsSlice())
 				case "name":
 					m.Name = getString(iter.ReadStringAsSlice())
-				case "linkName":
+				case "linkname":
 					m.Linkname = getString(iter.ReadStringAsSlice())
 				case "mode":
 					m.Mode = iter.ReadInt64()
 				case "size":
 					m.Size = iter.ReadInt64()
-				case "UID":
+				case "uid":
 					m.UID = iter.ReadInt()
-				case "GID":
+				case "gid":
 					m.GID = iter.ReadInt()
 				case "modtime":
 					time, err := time.Parse(time.RFC3339, byteSliceAsString(iter.ReadStringAsSlice()))
@@ -592,23 +592,23 @@ func unmarshalToc(manifest []byte) (*internal.TOC, error) {
 						return nil, err
 					}
 					m.ChangeTime = &time
-				case "devMajor":
+				case "devmajor":
 					m.Devmajor = iter.ReadInt64()
-				case "devMinor":
+				case "devminor":
 					m.Devminor = iter.ReadInt64()
 				case "digest":
 					m.Digest = getString(iter.ReadStringAsSlice())
 				case "offset":
 					m.Offset = iter.ReadInt64()
-				case "endOffset":
+				case "endoffset":
 					m.EndOffset = iter.ReadInt64()
-				case "chunkSize":
+				case "chunksize":
 					m.ChunkSize = iter.ReadInt64()
-				case "chunkOffset":
+				case "chunkoffset":
 					m.ChunkOffset = iter.ReadInt64()
-				case "chunkDigest":
+				case "chunkdigest":
 					m.ChunkDigest = getString(iter.ReadStringAsSlice())
-				case "chunkType":
+				case "chunktype":
 					m.ChunkType = getString(iter.ReadStringAsSlice())
 				case "xattrs":
 					m.Xattrs = make(map[string]string)
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -150,22 +150,32 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, 
 // readZstdChunkedManifest reads the zstd:chunked manifest from the seekable stream blobStream.  The blob total size must
 // be specified.
 // This function uses the io.github.containers.zstd-chunked. annotations when specified.
-func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable, blobSize int64, annotations map[string]string) ([]byte, int64, error) {
+func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable, blobSize int64, annotations map[string]string) ([]byte, []byte, int64, error) {
 	footerSize := int64(internal.FooterSizeSupported)
 	if blobSize <= footerSize {
-		return nil, 0, errors.New("blob too small")
+		return nil, nil, 0, errors.New("blob too small")
 	}
 
 	manifestChecksumAnnotation := annotations[internal.ManifestChecksumKey]
 	if manifestChecksumAnnotation == "" {
-		return nil, 0, fmt.Errorf("manifest checksum annotation %q not found", internal.ManifestChecksumKey)
+		return nil, nil, 0, fmt.Errorf("manifest checksum annotation %q not found", internal.ManifestChecksumKey)
 	}
 
 	var offset, length, lengthUncompressed, manifestType uint64
 
+	var offsetTarSplit, lengthTarSplit, lengthUncompressedTarSplit uint64
+	tarSplitChecksumAnnotation := ""
+
 	if offsetMetadata := annotations[internal.ManifestInfoKey]; offsetMetadata != "" {
 		if _, err := fmt.Sscanf(offsetMetadata, "%d:%d:%d:%d", &offset, &length, &lengthUncompressed, &manifestType); err != nil {
-			return nil, 0, err
+			return nil, nil, 0, err
+		}
+
+		if tarSplitInfoKeyAnnotation, found := annotations[internal.TarSplitInfoKey]; found {
+			if _, err := fmt.Sscanf(tarSplitInfoKeyAnnotation, "%d:%d:%d", &offsetTarSplit, &lengthTarSplit, &lengthUncompressedTarSplit); err != nil {
+				return nil, nil, 0, err
+			}
+			tarSplitChecksumAnnotation = annotations[internal.TarSplitChecksumKey]
 		}
 	} else {
 		chunk := ImageSourceChunk{
@@ -174,39 +184,39 @@ func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable
 		}
 		parts, errs, err := blobStream.GetBlobAt([]ImageSourceChunk{chunk})
 		if err != nil {
-			return nil, 0, err
+			return nil, nil, 0, err
 		}
 		var reader io.ReadCloser
 		select {
 		case r := <-parts:
 			reader = r
 		case err := <-errs:
-			return nil, 0, err
+			return nil, nil, 0, err
 		}
 		footer := make([]byte, footerSize)
 		if _, err := io.ReadFull(reader, footer); err != nil {
-			return nil, 0, err
+			return nil, nil, 0, err
 		}
 
 		offset = binary.LittleEndian.Uint64(footer[0:8])
 		length = binary.LittleEndian.Uint64(footer[8:16])
 		lengthUncompressed = binary.LittleEndian.Uint64(footer[16:24])
 		manifestType = binary.LittleEndian.Uint64(footer[24:32])
-		if !isZstdChunkedFrameMagic(footer[32:40]) {
-			return nil, 0, errors.New("invalid magic number")
+		if !isZstdChunkedFrameMagic(footer[48:56]) {
+			return nil, nil, 0, errors.New("invalid magic number")
 		}
 	}
 
 	if manifestType != internal.ManifestTypeCRFS {
-		return nil, 0, errors.New("invalid manifest type")
+		return nil, nil, 0, errors.New("invalid manifest type")
 	}
 
 	// set a reasonable limit
 	if length > (1<<20)*50 {
-		return nil, 0, errors.New("manifest too big")
+		return nil, nil, 0, errors.New("manifest too big")
 	}
 	if lengthUncompressed > (1<<20)*50 {
-		return nil, 0, errors.New("manifest too big")
+		return nil, nil, 0, errors.New("manifest too big")
 	}
 
 	chunk := ImageSourceChunk{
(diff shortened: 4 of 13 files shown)


pkg/chunked/compressor/compressor.go
```diff
@@ -202,7 +205,18 @@ func writeZstdChunkedStream(destFile io.Writer, outMetadata map[string]string, r
 	// total written so far.  Used to retrieve partial offsets in the file
 	dest := ioutils.NewWriteCounter(destFile)
 
-	tr := tar.NewReader(reader)
+	metaPackerBuffer := bytes.NewBuffer(nil)
```
mtrmac, line 208: Would it make sense to inline the compression here, instead of building a large in-memory buffer only to compress it? Some layers are gigantic.

pkg/chunked/internal/compression.go
```diff
@@ -167,13 +170,41 @@ func WriteZstdChunkedManifest(dest io.Writer, outMetadata map[string]string, off
...
+		return err
+	}
+	compressedTarSplit := compressedTarSplitBuffer.Bytes()
+
+	tarSplitDigester := digest.Canonical.Digester()
+	tarSplitChecksum := tarSplitDigester.Hash()
+	if _, err := tarSplitChecksum.Write(compressedTarSplit); err != nil {
+		return err
+	}
+	outMetadata[TarSplitChecksumKey] = tarSplitDigester.Digest().String()
```
mtrmac, line 192: This seems to amount to `digest.Canonical.FromBytes()`. If so, please use that.

(Alternatively, compute the digest on the fly while creating the data.)

pkg/chunked/internal/compression.go
```diff
@@ -167,13 +170,41 @@ func WriteZstdChunkedManifest(dest io.Writer, outMetadata map[string]string, off
...
 	binary.LittleEndian.PutUint64(manifestDataLE, manifestOffset)
-	binary.LittleEndian.PutUint64(manifestDataLE[8:], uint64(len(compressedManifest)))
-	binary.LittleEndian.PutUint64(manifestDataLE[16:], uint64(len(manifest)))
-	binary.LittleEndian.PutUint64(manifestDataLE[24:], uint64(ManifestTypeCRFS))
-	copy(manifestDataLE[32:], ZstdChunkedFrameMagic)
+	binary.LittleEndian.PutUint64(manifestDataLE[8*1:], uint64(len(compressedManifest)))
+	binary.LittleEndian.PutUint64(manifestDataLE[8*2:], uint64(len(manifest)))
+	binary.LittleEndian.PutUint64(manifestDataLE[8*3:], uint64(ManifestTypeCRFS))
+	binary.LittleEndian.PutUint64(manifestDataLE[8*4:], uint64(tarSplitOffset))
+	binary.LittleEndian.PutUint64(manifestDataLE[8*5:], uint64(len(compressedTarSplit)))
```
mtrmac, line 206: The reader seems to consume only the `TarSplitInfoKey` etc. annotation. Why does this need to be written into the footer as well?

pkg/chunked/storage_linux.go
```diff
@@ -1579,6 +1584,9 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions) (gra
 	if totalChunksSize > 0 {
 		logrus.Debugf("Missing %d bytes out of %d (%.2f %%)", missingPartsSize, totalChunksSize, float32(missingPartsSize*100.0)/float32(totalChunksSize))
 	}
+
+	output.UncompressedDigest = c.uncompressedDigest
```
mtrmac, line 1588: Actually, this seems to be a major problem.

We rely on `UncompressedDigest` for layer deduplication; if we accept this value just provided by the caller at face value, it seems to me that an attacker could create a Zstd:chunked layer with malicious contents, and _claim_ in the metadata that the `UncompressedDigest` is a value that matches a known widely-used base image. That could then cause the malicious contents to be used for that layer, affecting pulls of other (correctly-signed) containers.

For manifest digest matching (and therefore signature validation) rule is that we only use digests that we compute ourselves (i.e. the compressed-layer digests we always compute and match; we also compute uncompressed-layer digests ourselves, and record the compressed/uncompressed pairs _locally_ to allow safe layer reuse later).

Ugh… I suppose we would need to have `ApplyDiffFromStagingDirectory` actually recreate the uncompressed tarball and to compute the full digest? That seems awfully expensive.

pkg/chunked/internal/compression.go
```diff
@@ -167,13 +170,41 @@ func WriteZstdChunkedManifest(dest io.Writer, outMetadata map[string]string, off
...
+	}
+	compressedTarSplit := compressedTarSplitBuffer.Bytes()
+
+	tarSplitDigester := digest.Canonical.Digester()
+	tarSplitChecksum := tarSplitDigester.Hash()
+	if _, err := tarSplitChecksum.Write(compressedTarSplit); err != nil {
+		return err
+	}
+	outMetadata[TarSplitChecksumKey] = tarSplitDigester.Digest().String()
+	tarSplitOffset := manifestOffset + uint64(len(compressedManifest)) + 8
```
mtrmac, line 193: Would it be easy enough to get rid of the `+8` magic?

The caller of this function already has an `ioutils.WriteCounter`; it can pass that into this function, and AFAICS we could get rid of the `offset` variable entirely. Then, just call `appendZstdSkippableFrame`, and read the current offset to put into the annotation.

pkg/chunked/compressor/compressor.go
```diff
@@ -6,13 +6,16 @@ package compressor
...
+	"bytes"
 	"encoding/base64"
 	"io"
 
 	"github.com/containers/storage/pkg/chunked/internal"
 	"github.com/containers/storage/pkg/ioutils"
 	"github.com/opencontainers/go-digest"
 	"github.com/vbatts/tar-split/archive/tar"
+	"github.com/vbatts/tar-split/tar/asm"
+	"github.com/vbatts/tar-split/tar/storage"
```
mtrmac, line 18: This is, sadly, going to be added into every single c/image/copy and c/image/docker user…

I don’t immediately see a way to avoid that, I’m afraid. But if anyone else does see a way, I’d love that.

pkg/chunked/storage_linux.go
giuseppe: indeed.  Thanks for spotting it!

What do you think if we change the logic for `digestingReader` to not raise an error when the digest is computed but just use the correct digest instead of the one we store (that could be the compressed digest)?  Maybe we could limit this behavior only for the partial layers?

giuseppe: I was thinking of something like the following for c/image:

```diff
diff --git a/copy/blob.go b/copy/blob.go
index f45b97f5..293ed4f1 100644
--- a/copy/blob.go
+++ b/copy/blob.go
@@ -129,7 +129,9 @@ func (ic *imageCopier) copyBlobFromStream(ctx context.Context, srcReader io.Read
        }
 
        if digestingReader.validationFailed { // Coverage: This should never happen.
-               return types.BlobInfo{}, fmt.Errorf("Internal error writing blob %s, digest verification failed but was ignored", srcInfo.Digest)
+               uploadedInfo.Digest = digestingReader.digester.Digest()
+               logrus.Warningf("Digest verification failed for blob %s, using computed %s", srcInfo.Digest, uploadedInfo.Digest)
+               return uploadedInfo, nil
        }
        if stream.info.Digest != "" && uploadedInfo.Digest != stream.info.Digest {
                return types.BlobInfo{}, fmt.Errorf("Internal error writing blob %s, blob with digest %s saved with digest %s", srcInfo.Digest, stream.info.Digest, uploadedInfo.Digest)
diff --git a/copy/digesting_reader.go b/copy/digesting_reader.go
index 901d1082..028c86fc 100644
--- a/copy/digesting_reader.go
+++ b/copy/digesting_reader.go
@@ -54,9 +54,7 @@ func (d *digestingReader) Read(p []byte) (int, error) {
                actualDigest := d.digester.Digest()
                if actualDigest != d.expectedDigest {
                        d.validationFailed = true
-                       return 0, fmt.Errorf("Digest did not match, expected %s, got %s", d.expectedDigest, actualDigest)
                }
-               d.validationSucceeded = true
        }
        return n, err
 }
```

so we won't have to pay the high price of computing the uncompressed digest at pull time (which would probably defeat all the advantages of a partial pull)

mtrmac: I don’t think that code path is involved in a partial pull at all. What am I missing?!

(Also, `validationFailed` is when we _positively know_ the digest does not match, because we have read all of the blob; I don’t see why it would ever be OK to continue in that situation.)

drivers/overlay/overlay.go
```diff
@@ -1952,6 +1953,15 @@ func (d *Driver) ApplyDiffFromStagingDirectory(id, parent, stagingDirectory stri
 	if err := os.RemoveAll(diff); err != nil && !os.IsNotExist(err) {
 		return err
 	}
+
+	keys := make([]string, 0, len(diffOutput.BigData))
+	for k := range diffOutput.BigData {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
mtrmac, line 1961: `keys` seems to be unused.

layers.go
```diff
@@ -160,6 +160,9 @@ type Layer struct {
 	// convenience of the caller.  They can be large, and are only in
 	// memory when being read from or written to disk.
 	BigDataNames []string `json:"big-data-names,omitempty"`
+
+	// Incremental is true if this layer was created by a partial pull.
+	Incremental bool `json:"incremental,omitempty"`
```
mtrmac, line 165: “was created by a partial pull” does not help consumers of the value much to understand what needs doing. Does this mean “if `UncompressedDigest` is not actually an uncompressed digest”?

Basically this would force every user of `store.LayersByUncompressedDigest` to now start checking `Incremental` on the result to see whether calling `.Diff` is safe.

That _might_ be good enough, but it also feels unclean.

Alternatively, we could track _exactly_ what we know, set `.UncompressedDigest` to `""`, and to add a new `ChunkedTOCDigest` field (and a new `LayersByChunkedTOCDigest` function). That seems more correct, OTOH also probably triggering more invasive changes in callers.


---

Either way, c/image’s `TryReusingBlob*` would need to be taught about TOC digests, and trigger a search by the TOC digest value. … And at which layer would we turn the blob annotation in to a TOC digest?

Do we need, instead of `LayersByChunkedTOCDigest`, some kind of `ChunkedLayersByManifestAnnotations`?

rhatdan:
LGTM
@mtrmac PTAL

giuseppe:
these fixes are independent of the work in c/image, except for [chunked: add function to retrieve TOC digest](https://github.com/containers/storage/pull/1627/commits/91a1fe92125406a296002d3573e7ab84e2dbfa72)

rhatdan:
LGTM
@mtrmac It is up to you

giuseppe:
rebased

giuseppe:
I'd like to use these fixes for the new composefs integration work.  Could we move forward and if there is anything to change/fix we could address it later?

layers.go
```diff
@@ -2386,8 +2386,25 @@ func (r *layerStore) ApplyDiffFromStagingDirectory(id, stagingDirectory string,
 	layer.UncompressedDigest = diffOutput.UncompressedDigest
 	layer.UncompressedSize = diffOutput.Size
 	layer.Metadata = diffOutput.Metadata
-	if err = r.saveFor(layer); err != nil {
-		return err
+	if diffOutput.TarSplit != nil {
```
nalind, line 2389: I think `len(diffOutput.TarSplit) != 0` is more idiomatic.  I don't think it's ever `nil` for zstd chunked.

layers.go
```diff
@@ -2386,8 +2386,25 @@ func (r *layerStore) ApplyDiffFromStagingDirectory(id, stagingDirectory string,
...
+		tsdata := bytes.Buffer{}
+		compressor, err := pgzip.NewWriterLevel(&tsdata, pgzip.BestSpeed)
+		if err != nil {
+			compressor = pgzip.NewWriter(&tsdata)
+		}
+		if err := compressor.SetConcurrency(1024*1024, 1); err != nil { // 1024*1024 is the hard-coded default; we're not changing that
+			logrus.Infof("setting compression concurrency threads to 1: %v; ignoring", err)
+		}
+		if _, err := compressor.Write(diffOutput.TarSplit); err != nil {
+			return err
```
nalind, line 2399: `compressor` doesn't get `Close()`d in this branch.

pkg/chunked/cache_linux.go
```diff
@@ -559,22 +559,22 @@ func unmarshalToc(manifest []byte) (*internal.TOC, error) {
 		for iter.ReadArray() {
 			var m internal.FileMetadata
 			for field := iter.ReadObject(); field != ""; field = iter.ReadObject() {
-				switch field {
+				switch strings.ToLower(field) {
```
nalind, line 562: Why does this location switch to treating field names returned by `iter.ReadObject()` in a case-insensitive fashion while the previous loop over the same values doesn't?

pkg/chunked/compression_linux.go
```diff
@@ -174,87 +184,125 @@ func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable
...
+	manifest, err := readBlob(length)
+	if err != nil {
+		return nil, nil, 0, err
 	}
 
-	manifestDigester := digest.Canonical.Digester()
-	manifestChecksum := manifestDigester.Hash()
-	if _, err := manifestChecksum.Write(manifest); err != nil {
-		return nil, 0, err
+	decodedBlob, err := decodeAndValidateBlob(manifest, lengthUncompressed, manifestChecksumAnnotation)
```
nalind, line 266: If `err` is not `nil`, why does the error only get returned after possibly attempting to decode a tarsplit blob?

pkg/chunked/compression_linux.go
```diff
@@ -174,87 +184,125 @@ func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable
...
+
+		decodedTarSplit, err = decodeAndValidateBlob(tarSplit, lengthUncompressedTarSplit, tarSplitChecksumAnnotation)
+		if err != nil {
+			return nil, nil, 0, err
+		}
+	}
+	return decodedBlob, decodedTarSplit, int64(offset), err
+}
+
+func decodeAndValidateBlob(blob []byte, lengthUncompressed uint64, expectedChecksum string) ([]byte, error) {
```
nalind, line 282: Does `expectedChecksum` refer to the checksum of the blob before or after decompression?  Please add a comment or rename the variable.

pkg/chunked/compression_linux.go
```diff
@@ -174,87 +184,125 @@ func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable
...
+	blobDigester := digest.Canonical.Digester()
```
nalind, line 283: Is `expectedChecksum` mandated or guaranteed to be a sha256 digest?

pkg/chunked/compression_linux.go
```diff
@@ -174,87 +184,125 @@ func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable
...
 
 	b := make([]byte, 0, lengthUncompressed)
-	if decoded, err := decoder.DecodeAll(manifest, b); err == nil {
-		return decoded, int64(offset), nil
+	if decoded, err := decoder.DecodeAll(blob, b); err == nil {
+		return decoded, nil
 	}
-
-	return manifest, int64(offset), nil
+	return blob, nil
```
nalind, line 307: If I'm reading this right, the caller can't tell if `DecodeAll()` failed for any reason, since in either case `decodeAndValidateBlob()` returns a slice and a `nil` error.

pkg/chunked/storage_unsupported.go
```diff
@@ -15,3 +15,9 @@ import (
 func GetDiffer(ctx context.Context, store storage.Store, blobSize int64, annotations map[string]string, iss ImageSourceSeekable) (graphdriver.Differ, error) {
 	return nil, errors.New("format not supported on this architecture")
 }
+
+// GetTOCDigest returns the digest of the TOC as recorded in the annotations.
+// This is an experimental feature and may be changed/removed in the future.
+func GetTOCDigest(annotations map[string]string) (*digest.Digest, error) {
+	return nil, errors.New("format not supported on this architecture")
```
nalind, line 22: Looking around in this package, this should be complaining about the OS.

pkg/chunked/cache_linux.go
giuseppe: thanks for spotting it.  I've changed also the previous loop to use the case-insensitive values.

pkg/chunked/compression_linux.go
giuseppe: no need for that, I've fixed it

pkg/chunked/compression_linux.go
giuseppe: changed the variable name to be clearer that it is the digest for the uncompressed blob

pkg/chunked/compression_linux.go
giuseppe: we were previously using uncompressed digests so this code was handling both cases.

Anyway, there is no real reason for doing it now since we always compress it, so I've added a patch to drop support for uncompressed manifests.

pkg/chunked/storage_unsupported.go
giuseppe: changed the error message to the more generic "format not support on this system"

pkg/chunked/compression_linux.go
giuseppe: I've generalized it to use the same algorithm as `expectedChecksum` so we don't have to hardcode sha256

layers.go
giuseppe: fixed

giuseppe:
all comments adressed, and the CI is green

pkg/chunked/compression_linux.go
```diff
@@ -174,87 +184,125 @@ func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable
...
+		select {
+		case r := <-parts:
+			reader = r
+		case err := <-errs:
+			return nil, err
+		}
+
+		blob := make([]byte, len)
+		if _, err := io.ReadFull(reader, blob); err != nil {
+			return nil, err
```
nalind, line 253: This branch doesn't close `reader`.

pkg/chunked/compressor/compressor.go
```diff
@@ -383,7 +426,20 @@ func writeZstdChunkedStream(destFile io.Writer, outMetadata map[string]string, r
 	}
 	zstdWriter = nil
 
-	return internal.WriteZstdChunkedManifest(dest, outMetadata, uint64(dest.Count), metadata, level)
+	if err := tarSplitData.zstd.Flush(); err != nil {
```
nalind, line 429: This branch doesn't close `tarSplitData.zstd`, and a number of branches above it don't close `zstdWriter`.

nalind:
Mostly LGTM, just nits.  `restartCompression()` in compressor.go is calling `zstdWriter.Flush()` after `zstdWriter.Close()`, which looks redundant.

layers.go
```diff
@@ -2392,8 +2392,26 @@ func (r *layerStore) ApplyDiffFromStagingDirectory(id, stagingDirectory string,
...
 	layer.Metadata = diffOutput.Metadata
-	if err = r.saveFor(layer); err != nil {
-		return err
+	if len(diffOutput.TarSplit) != 0 {
+		tsdata := bytes.Buffer{}
+		compressor, err := pgzip.NewWriterLevel(&tsdata, pgzip.BestSpeed)
+		if err != nil {
+			compressor = pgzip.NewWriter(&tsdata)
+		}
+		if err := compressor.SetConcurrency(1024*1024, 1); err != nil { // 1024*1024 is the hard-coded default; we're not changing that
```
TomSweeneyRedHat, line 2401: consider making `1024*1024` a constant and stick it somewhere under /types.  This does appear to be the only instance at the moment, so a consideration for later and not holding for this PR.

TomSweeneyRedHat:
LGTM other than @nalind's comments

pkg/chunked/compressor/compressor.go
giuseppe: there is a func for zstdWriter to clean it up if it is not closed (and thus set to nil afterwards):

```
	defer func() {
		if zstdWriter != nil {
			zstdWriter.Close()
		}
	}()
```

is that sufficient?

I've added the same pattern for tarSplitData

giuseppe:
> Mostly LGTM, just nits. `restartCompression()` in compressor.go is calling `zstdWriter.Flush()` after `zstdWriter.Close()`, which looks redundant.

thanks, I've added a separate patch to drop it since it was already this way (so in case of regressions, it is easier to bisect it)

layers.go
giuseppe: that is the default used by github.com/klauspost/pgzip/

rhatdan:
LGTM

drivers/overlay/overlay.go
```diff
@@ -1952,6 +1952,9 @@ func (d *Driver) ApplyDiffFromStagingDirectory(id, parent, stagingDirectory stri
 	if err := os.RemoveAll(diff); err != nil && !os.IsNotExist(err) {
 		return err
 	}
+
+	diffOutput.UncompressedDigest = diffOutput.TOCDigest
```
mtrmac, line 1956: The net effect of this is that
- the `Layer.UncompresssedDigest` field now has two different alternative semantics
- Which one is used is ~indeterminate (it depends on whether a partial pull was used, which is partially registry-controlled!). So the caller has no idea what the value is. And there is no record of that at all.

That might be fine, but I don’t see an _obvious_ argument. E.g. what if a blob with the same digest were both a valid-enough chunked manifest and a valid-enough tar layer?

Why is this safe?

(Extra credit: it seems natural to “just” add an alternative field and not set `Layer.UncompressedDigest` on the partial pull paths at all. But that is clearly an API break, c/image expects `UncompressedDigest` to be set. Is there a migration path that doesn’t break API without an exorbitant cost?)

Compare previous https://github.com/containers/storage/pull/1627#discussion_r1218299872 .

mtrmac (commented):
Catching up for https://github.com/containers/image/pull/1980 … quite possibly the questions have been resolved since. Follow https://github.com/containers/image/pull/1980 for progress.

pkg/chunked/compression_linux.go
```diff
@@ -174,87 +184,126 @@ func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable
...
+			return nil, nil, 0, err
 		}
 
 		offset = binary.LittleEndian.Uint64(footer[0:8])
 		length = binary.LittleEndian.Uint64(footer[8:16])
 		lengthUncompressed = binary.LittleEndian.Uint64(footer[16:24])
 		manifestType = binary.LittleEndian.Uint64(footer[24:32])
-		if !isZstdChunkedFrameMagic(footer[32:40]) {
-			return nil, 0, errors.New("invalid magic number")
+		if !isZstdChunkedFrameMagic(footer[48:56]) {
```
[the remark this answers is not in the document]
giuseppe: thanks for spotting this.

I've opened a PR to fix it and add some tests: https://github.com/containers/storage/pull/1697

I am not 100% sure what to do with the footer data, this code path is not used by us at the moment, since we store this information in the OCI annotations for the layer (and if the OCI annotations are missing we don't even attempt a partial pull).
On the other hand it seems nice to have this information somewhere in the layer file itself, considering it is not a lot of data anyway.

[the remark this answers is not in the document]
mtrmac: I think plausible alternatives are:
- Keep it as a frozen artifact, supporting only the old features. _If_ the old features are sufficient to pull/push useful images.
- Design _some_ kind of forward-compatibility mechanism. Maybe a new magic value, and now including a version identifier. And the design would need to allow adding more fields, ie.e. the magic/version ID would need to be at a predictable place regardless of version. And it wouldn’t be a replacement for the annotations because we need the layer’s metadata to be authenticated without having to read all of the layer. If we are not consuming the data at all, all of this seems to me to be very likely to keep breaking, and just not worth the effort. 
- Redesign so that this data does not _need_ extending: only include data about the manifest in here, and add all new fields in the manifest instead. That seems to me to be the cleanest way to have the layer blob be ~self-describing… OTOH in principle it would be possible for an attacker to point at one manifest in this footer and another in the annotation, so I’m not sure there’s that much of a benefit to being self-describing, either.
- Completely drop support, and rely on annotations only? I didn’t look into the history, whether we can do that.

pkg/chunked/internal/compression.go
```diff
@@ -167,13 +176,20 @@ func WriteZstdChunkedManifest(dest io.Writer, outMetadata map[string]string, off
...
 	manifestDataLE := make([]byte, FooterSizeSupported)
 	binary.LittleEndian.PutUint64(manifestDataLE, manifestOffset)
-	binary.LittleEndian.PutUint64(manifestDataLE[8:], uint64(len(compressedManifest)))
-	binary.LittleEndian.PutUint64(manifestDataLE[16:], uint64(len(manifest)))
-	binary.LittleEndian.PutUint64(manifestDataLE[24:], uint64(ManifestTypeCRFS))
-	copy(manifestDataLE[32:], ZstdChunkedFrameMagic)
+	binary.LittleEndian.PutUint64(manifestDataLE[8*1:], uint64(len(compressedManifest)))
+	binary.LittleEndian.PutUint64(manifestDataLE[8*2:], uint64(len(manifest)))
+	binary.LittleEndian.PutUint64(manifestDataLE[8*3:], uint64(ManifestTypeCRFS))
+	copy(manifestDataLE[8*4:], ZstdChunkedFrameMagic)
```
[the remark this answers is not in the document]
mtrmac: Fixed in https://github.com/containers/storage/pull/1697 , at least to use all of the allocated space.

pkg/chunked/storage_linux.go
```diff
@@ -156,13 +179,20 @@ func makeZstdChunkedDiffer(ctx context.Context, store storage.Store, blobSize in
 		return nil, err
 	}
 
+	tocDigest, err := digest.Parse(annotations[internal.ManifestChecksumKey])
+	if err != nil {
+		return nil, fmt.Errorf("parse TOC digest %q: %w", annotations[internal.ManifestChecksumKey], err)
+	}
```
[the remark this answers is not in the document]
mtrmac: I… don’t understand what I was trying to say here :/

drivers/overlay/overlay.go
mtrmac: This was disambiguated in https://github.com/containers/storage/pull/1719 , and c/image now understands `TOCDigest` layers where `UncompressedDigest` is unavailable.

layers.go
```diff
@@ -2392,8 +2392,26 @@ func (r *layerStore) ApplyDiffFromStagingDirectory(id, stagingDirectory string,
...
+		compressor.Close()
+		if err := os.MkdirAll(filepath.Dir(r.tspath(layer.ID)), 0o700); err != nil {
+			return err
+		}
+		if err := ioutils.AtomicWriteFile(r.tspath(layer.ID), tsdata.Bytes(), 0o600); err != nil {
+			return err
+		}
 	}
 	for k, v := range diffOutput.BigData {
 		if err := r.SetBigData(id, k, bytes.NewReader(v)); err != nil {
```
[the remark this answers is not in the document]
mtrmac: https://github.com/containers/storage/pull/1844 proposes reverting the move.

pkg/chunked/compression_linux.go
```diff
@@ -174,87 +184,126 @@ func readZstdChunkedManifest(ctx context.Context, blobStream ImageSourceSeekable
...
+		decodedTarSplit, err = decodeAndValidateBlob(tarSplit, lengthUncompressedTarSplit, tarSplitChecksumAnnotation)
+		if err != nil {
+			return nil, nil, 0, err
+		}
 	}
+	return decodedBlob, decodedTarSplit, int64(offset), err
+}
 
-	d, err := digest.Parse(manifestChecksumAnnotation)
+func decodeAndValidateBlob(blob []byte, lengthUncompressed uint64, expectedUncompressedChecksum string) ([]byte, error) {
```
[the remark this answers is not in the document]
mtrmac: Fixed in https://github.com/containers/storage/pull/1844.

pkg/chunked/storage_linux.go
```diff
@@ -135,6 +138,26 @@ func copyFileContent(srcFd int, destFile string, dirfd int, mode os.FileMode, us
...
+// This is an experimental feature and may be changed/removed in the future.
+func GetTOCDigest(annotations map[string]string) (*digest.Digest, error) {
+	if tocDigest, ok := annotations[estargz.TOCJSONDigestAnnotation]; ok {
+		d, err := digest.Parse(tocDigest)
+		if err != nil {
+			return nil, err
+		}
+		return &d, nil
+	}
+	if tocDigest, ok := annotations[internal.ManifestChecksumKey]; ok {
```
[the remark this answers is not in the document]
mtrmac: This was fixed here; https://github.com/containers/storage/pull/1844 also makes `GetTOCDigest` unambiguous.

PR #1631, opened 2023-06-06
quota: unlink tmp file before creating it

giuseppe:
Before creating the tmp file, we unlink it first to avoid an EEXIST error.


drivers/quota/projectquota.go
```diff
@@ -428,7 +428,11 @@ func makeBackingFsDev(home string) (string, error) {
 	backingFsBlockDevTmp := backingFsBlockDev + ".tmp"
 	// Re-create just in case someone copied the home directory over to a new device
 	if err := unix.Mknod(backingFsBlockDevTmp, unix.S_IFBLK|0o600, int(stat.Dev)); err != nil {
-		return "", fmt.Errorf("failed to mknod %s: %w", backingFsBlockDevTmp, err)
+		// On errors, try again after unlinking any potential leftover.
```
edsantiago, line 431: Wouldn't it be safer to only try this on EEXIST?

rhatdan: Yes only do this if error.Is(err, EEXIST)

giuseppe: changed to be retried only on EEXIST

TomSweeneyRedHat:
LGTM

rhatdan:
LGTM

PR #1637, opened 2023-06-12
types/options: check prior drivers using readdir

flouthoc:
Fetch prior drivers if any using `ReadDir` and looking for traces of `%s-images` dirs if present already on the system where `%s` is the prior driver.


flouthoc:
@giuseppe @nalind @rhatdan @containers/storage-maintainers PTAL

types/options.go
```diff
@@ -312,12 +310,15 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
...
-				if _, prior := driversMap[name]; prior {
-					opts.GraphDriverName = name
-					break
+			dirEntries, err := os.ReadDir(opts.GraphRoot)
+			if err == nil {
+				for _, entry := range dirEntries {
+					if strings.HasSuffix(entry.Name(), "-images") {
+						res := strings.Split(entry.Name(), "-images")
+						if len(res) > 0 {
+							opts.GraphDriverName = res[0]
```
Luap99, line 319: add a break here

types/options.go
```diff
@@ -312,12 +310,15 @@ func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOpti
...
-
-			for _, name := range drivers.Priority {
-				if _, prior := driversMap[name]; prior {
-					opts.GraphDriverName = name
-					break
+			dirEntries, err := os.ReadDir(opts.GraphRoot)
+			if err == nil {
+				for _, entry := range dirEntries {
+					if strings.HasSuffix(entry.Name(), "-images") {
+						res := strings.Split(entry.Name(), "-images")
```
Luap99, line 317: I think using TrimSuffix() is better than using Split().

Luap99 (commented):
Thanks

giuseppe (approved):
LGTM

Issue #1623, opened 2023-05-30
It might use defer close pgzip writer for avoiding returning  earlier without closing


shitou12345678:
In the function `applyDiffWithOptions` of https://github.com/containers/storage/blob/main/layers.go#L2292, pgzip.NewWriter does not close with defer, if there is a error making the function return earlier, this Writer can not close. It will cause the goroutines residue. So we should use `defer compressor.Close()` instead.

shitou12345678:
Maybe it can not use `defer compressor.Close()`, because if using defer compressor.Close(), the data stored in r.tspath(layer.ID) in the code if err := ioutils.AtomicWriteFile(r.tspath(layer.ID), tsdata.Bytes(), 0600) may be incomplete.

rhatdan:
@mtrmac PTAL

mtrmac:
Should be fixed by #1641 .

PR #1641, opened 2023-06-13
Audit data flow in applyDiffWithOptions

mtrmac:
This is a superset of #1612 / #1623 , also fixing an unlikely-to-hit race in use of `pkg/tarlog`.


vrothberg (approved):
LGTM

PR #1646, opened 2023-06-19
composefs integration in the overlay driver

giuseppe:
ComposeFS has been integrated into the overlay driver.  The EROFS filesystem is now used to mount the file system metadata.  Each layer is mounted individually, but we can change this later and merge multiple layers into a single file system.

Filesystem-level integrity verification is now enabled through the implementation of fs-verity. This enhancement ensures the integrity of both the composefs blob and data files, providing an additional layer of security. Note that this implementation is currently considered a best-effort attempt since there is no control yet on how to configure it as well as there is no support from the overlay driver in the kernel (it is being worked on upstream).

It requires the `composefs-from-json` tool installed and accessible in the path from the https://github.com/containers/composefs project.

### DIFF
--- a/drivers/overlay/composefs_supported.go
+++ b/drivers/overlay/composefs_supported.go
@@ -1,0 +1,66 @@
+//go:build linux && composefs && cgo
+// +build linux,composefs,cgo
+
+package overlay
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"sync"
+
+	"github.com/containers/storage/pkg/loopback"
+	"golang.org/x/sys/unix"
+)
+
+var (
+	composeFsHelperOnce sync.Once
+	composeFsHelperPath string
+	composeFsHelperErr  error
+)
+
+func getComposeFsHelper() (string, error) {
+	composeFsHelperOnce.Do(func() {
+		composeFsHelperPath, composeFsHelperErr = exec.LookPath("composefs-from-json")
+	})
+	return composeFsHelperPath, composeFsHelperErr
+}
+
+func composeFsSupported() bool {
+	_, err := getComposeFsHelper()
+	return err == nil
+}
+
+func generateComposeFsBlob(toc []byte, destFile string) error {
+	writerJson, err := getComposeFsHelper()
+	if err != nil {
+		return fmt.Errorf("failed to find composefs-from-json: %w", err)
+	}
+
+	fd, err := unix.Openat(unix.AT_FDCWD, destFile, unix.O_WRONLY|unix.O_CREAT|unix.O_TRUNC|unix.O_EXCL|unix.O_CLOEXEC, 0o644)
+	if err != nil {
+		return fmt.Errorf("failed to open output file: %w", err)
+	}
+	outFd := os.NewFile(uintptr(fd), "outFd")
+
+	defer outFd.Close()
+	cmd := exec.Command(writerJson, "--format=erofs", "--out=/proc/self/fd/3", "/proc/self/fd/0")
+	cmd.ExtraFiles = []*os.File{outFd}
+	cmd.Stdin = bytes.NewReader(toc)
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to convert json to erofs: %w", err)
+	}
+	return nil
+}
+
+func mountErofsBlob(blobFile, mountPoint string) error {
+	loop, err := loopback.AttachLoopDevice(blobFile)
+	if err != nil {
+		return err
+	}
+	defer loop.Close()
+
+	return unix.Mount(loop.Name(), mountPoint, "erofs", unix.MS_RDONLY, "ro")
+}
--- a/drivers/overlay/composefs_supported.go
+++ b/drivers/overlay/composefs_supported.go
@@ -5,12 +5,18 @@ package overlay
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"sync"
+	"syscall"
+	"unsafe"
 
 	"github.com/containers/storage/pkg/loopback"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
 
@@ -32,6 +38,43 @@ func composeFsSupported() bool {
 	return err == nil
 }
 
+func enableVerity(description string, fd int) error {
+	enableArg := unix.FsverityEnableArg{
+		Version:        1,
+		Hash_algorithm: unix.FS_VERITY_HASH_ALG_SHA256,
+		Block_size:     4096,
+	}
+
+	_, _, e1 := syscall.Syscall(unix.SYS_IOCTL, uintptr(fd), uintptr(unix.FS_IOC_ENABLE_VERITY), uintptr(unsafe.Pointer(&enableArg)))
+	if e1 != 0 && !errors.Is(e1, unix.EEXIST) {
+		return fmt.Errorf("failed to enable verity for %q: %w", description, e1)
+	}
+	return nil
+}
+
+func enableVerityRecursive(path string) error {
+	walkFn := func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if !d.Type().IsRegular() {
+			return nil
+		}
+
+		f, err := os.Open(path)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+
+		if err := enableVerity(path, int(f.Fd())); err != nil {
+			return err
+		}
+		return nil
+	}
+	return filepath.WalkDir(path, walkFn)
+}
+
 func generateComposeFsBlob(toc []byte, destFile string) error {
 	writerJson, err := getComposeFsHelper()
 	if err != nil {
@@ -44,14 +87,35 @@ func generateComposeFsBlob(toc []byte, destFile string) error {
 	}
 	outFd := os.NewFile(uintptr(fd), "outFd")
 
-	defer outFd.Close()
-	cmd := exec.Command(writerJson, "--format=erofs", "--out=/proc/self/fd/3", "/proc/self/fd/0")
-	cmd.ExtraFiles = []*os.File{outFd}
-	cmd.Stdin = bytes.NewReader(toc)
-	cmd.Stderr = os.Stderr
-	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("failed to convert json to erofs: %w", err)
+	fd, err = unix.Open(fmt.Sprintf("/proc/self/fd/%d", outFd.Fd()), unix.O_RDONLY|unix.O_CLOEXEC, 0)
+	if err != nil {
+		outFd.Close()
+		return fmt.Errorf("failed to dup output file: %w", err)
 	}
+	newFd := os.NewFile(uintptr(fd), "newFd")
+	defer newFd.Close()
+
+	err = func() error {
+		// a scope to close outFd before setting fsverity on the read-only fd.
+		defer outFd.Close()
+
+		cmd := exec.Command(writerJson, "--format=erofs", "--out=/proc/self/fd/3", "/proc/self/fd/0")
+		cmd.ExtraFiles = []*os.File{outFd}
+		cmd.Stderr = os.Stderr
+		cmd.Stdin = bytes.NewReader(toc)
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("failed to convert json to erofs: %w", err)
+		}
+		return nil
+	}()
+	if err != nil {
+		return err
+	}
+
+	if err := enableVerity("manifest file", int(newFd.Fd())); err != nil && !errors.Is(err, unix.ENOTSUP) && !errors.Is(err, unix.ENOTTY) {
+		logrus.Warningf("%s", err)
+	}
+
 	return nil
 }
 
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1502,6 +1502,16 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		}
 	}
 
+	idmappedMountProcessPid := -1
+	if needsIDMapping {
+		pid, cleanupFunc, err := idmap.CreateUsernsProcess(options.UidMaps, options.GidMaps)
+		if err != nil {
+			return "", err
+		}
+		idmappedMountProcessPid = int(pid)
+		defer cleanupFunc()
+	}
+
 	// For each lower, resolve its path, and append it and any additional diffN
 	// directories to the lowers list.
 	for _, l := range splitLowers {
@@ -1601,18 +1611,12 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 
 	if needsIDMapping {
 		var newAbsDir []string
+		idMappedMounts := make(map[string]string)
+
 		mappedRoot := filepath.Join(d.home, id, "mapped")
 		if err := os.MkdirAll(mappedRoot, 0o700); err != nil {
 			return "", err
 		}
-
-		pid, cleanupFunc, err := idmap.CreateUsernsProcess(options.UidMaps, options.GidMaps)
-		if err != nil {
-			return "", err
-		}
-		defer cleanupFunc()
-
-		idMappedMounts := make(map[string]string)
 
 		// rewrite the lower dirs to their idmapped mount.
 		c := 0
@@ -1623,7 +1627,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 			if !found {
 				root = filepath.Join(mappedRoot, fmt.Sprintf("%d", c))
 				c++
-				if err := idmap.CreateIDMappedMount(mappedMountSrc, root, int(pid)); err != nil {
+				if err := idmap.CreateIDMappedMount(mappedMountSrc, root, idmappedMountProcessPid); err != nil {
 					return "", fmt.Errorf("create mapped mount for %q on %q: %w", mappedMountSrc, root, err)
 				}
 				idMappedMounts[mappedMountSrc] = root
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1934,7 +1934,7 @@ func (d *Driver) ApplyDiffWithDiffer(id, parent string, options *graphdriver.App
 		IgnoreChownErrors: d.options.ignoreChownErrors,
 		WhiteoutFormat:    d.getWhiteoutFormat(),
 		InUserNS:          unshare.IsRootless(),
-	})
+	}, nil)
 	out.Target = applyDir
 	return out, err
 }
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -82,6 +82,8 @@ const (
 	lowerFile  = "lower"
 	maxDepth   = 500
 
+	zstdChunkedManifest = "zstd-chunked-manifest"
+
 	// idLength represents the number of random characters
 	// which can be used to create the unique link identifier
 	// for every layer. If this value is too long then the
@@ -780,6 +782,10 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 }
 
 func (d *Driver) useNaiveDiff() bool {
+	if d.useComposeFs() {
+		return true
+	}
+
 	useNaiveDiffLock.Do(func() {
 		if d.options.mountProgram != "" {
 			useNaiveDiffOnly = true
@@ -1512,12 +1518,66 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		defer cleanupFunc()
 	}
 
+	erofsLayers := filepath.Join(workDirBase, "erofs-layers")
+	if err := os.MkdirAll(erofsLayers, 0o700); err != nil {
+		return "", err
+	}
+
+	skipIDMappingLayers := make(map[string]string)
+
+	composeFsLayers := []string{}
+
+	erofsMounts := []string{}
+	defer func() {
+		for _, m := range erofsMounts {
+			defer unix.Unmount(m, unix.MNT_DETACH)
+		}
+	}()
+
+	maybeAddErofsMount := func(lowerID string, i int) (string, error) {
+		erofsBlob := d.getErofsBlob(lowerID)
+		_, err = os.Stat(erofsBlob)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return "", nil
+			}
+			return "", err
+		}
+		logrus.Debugf("overlay: using erofs blob %s for lower %s", erofsBlob, lowerID)
+
+		dest := filepath.Join(erofsLayers, fmt.Sprintf("%d", i))
+		if err := os.MkdirAll(dest, 0o700); err != nil {
+			return "", err
+		}
+
+		if err := mountErofsBlob(erofsBlob, dest); err != nil {
+			return "", err
+		}
+		erofsMounts = append(erofsMounts, dest)
+		composeFsPath, err := d.getDiffPath(lowerID)
+		if err != nil {
+			return "", err
+		}
+		composeFsLayers = append(composeFsLayers, composeFsPath)
+		skipIDMappingLayers[composeFsPath] = composeFsPath
+		return dest, nil
+	}
+
+	diffDir := path.Join(workDirBase, "diff")
+
+	if dest, err := maybeAddErofsMount(id, 0); err != nil {
+		return "", err
+	} else if dest != "" {
+		diffDir = dest
+	}
+
 	// For each lower, resolve its path, and append it and any additional diffN
 	// directories to the lowers list.
-	for _, l := range splitLowers {
+	for i, l := range splitLowers {
 		if l == "" {
 			continue
 		}
+
 		lower := ""
 		newpath := path.Join(d.home, l)
 		if st, err := os.Stat(newpath); err != nil {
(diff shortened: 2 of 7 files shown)


drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,91 @@
...
+		buf := toc
+		for len(buf) > 0 {
+			n, err := unix.Write(fd, buf)
+			if err != nil {
+				return fmt.Errorf("failed to write to memfd: %w", err)
+			}
+			buf = buf[n:]
+		}
+
+		cmd := exec.Command(writerJson, "--format=erofs", fmt.Sprintf("--out=/proc/self/fd/%d", outFd), fmt.Sprintf("/proc/self/fd/%d", fd))
```
cgwalters, line 117: I'm a bit confused why we're not just passing `"--out=" + destFile` here?

And instead of reading from a memfd copy of the input, wouldn't it be nicer and actually more efficient to use a plain old `pipe()` here? 

(See also https://github.com/containers/composefs/pull/155#issuecomment-1597227020 - if we followed that suggestion then we could just stream cpio to the helper here)

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		return nil
+	}()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func mountErofsBlob(blobFile, mountPoint string) error {
```
cgwalters, line 135: Any reason not to use `mount.composefs`?

drivers/overlay/composefs_supported.go
```diff
@@ -77,6 +124,11 @@ func generateComposeFsBlob(toc []byte, destFile string) error {
 	if err != nil {
 		return err
 	}
+
+	if err := enableVerity("manifest file", newFd); err != nil {
+		logrus.Warningf("%s", err)
```
cgwalters, line 129: I don't think we should warn here if we get e.g. `EOPNOTSUPP` - that'd just be annoying for everyone using filesystems which don't yet support it (like xfs, but that's in progress).

cgwalters:
> Only images that are using the zstd:chunked and eStargz format can be
used in this way since the metadata is stored in the image itself.

Wait...why?  What metadata?

pkg/chunked/storage_linux.go
```diff
@@ -1317,6 +1317,42 @@ func (c *chunkedDiffer) findAndCopyFile(dirfd int, r *internal.FileMetadata, cop
 	return false, nil
 }
 
+func makeEntriesFlat(dirfd int, mergedEntries []internal.FileMetadata) ([]internal.FileMetadata, error) {
+	var new []internal.FileMetadata
+
+	hashes := make(map[string]string)
+	for i := range mergedEntries {
+		if mergedEntries[i].Type != TypeReg {
+			continue
```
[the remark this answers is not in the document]
alexlarsson: symlinks are stored in the composfs directly, only regular files need a backing file.

drivers/overlay/composefs_supported.go
alexlarsson: That would be nice. It has some optimizations, like mounting with noacl (which is faster) when there are not acl:s in the erofs image. (This info is in the composefs header at the start of the image file)

cgwalters:
OK, yes I see this JSON format was introduced in https://github.com/containers/storage/commit/a0ff89e09dcf3f345c31642c8820ed74934e06f3

A few things here.  I think a few choices in this JSON format are suboptimal:

- using `string` for symlink targets requires them to be valid Unicode; I would bet money that there are container images with non-Unicode filenames.
- using `string` for the xattr names has a similar conceptual problem (although non-Unicode xattr names are probably very rare and weird if they exist).  But xattr *values* is more problematic, because file capabilities aka `security.capability` are not valid Unicode.
- I've already mentioned that IMO parsing human-readable dates here is just unnecessarily inefficient

So I would still propose that this JSON format should stay as a c/storage thing, and we have a different format in composefs upstream.

(Or, c/storage could switch to a different format, because I believe this tooling in c/storage is still experimental)

alexlarsson:
> * But xattr _values_ is more problematic, because file capabilities aka `security.capability` are not valid Unicode.

Some xattrs also contain a terminating zero byte. selinux spring to mind, although i may be misremembering.

cgwalters:
Another thing about the times: EROFS seems to (as you'd expect) make "ctime == mtime == atime", because for a read-only system these must all be the same; (and it further optimizes the common case of squashing all timestamps in an image to a common build time).

Another way to say this is that it doesn't support any of these times having different values, and hence if we care about this use case (as is implicit in these being different values in the JSON here) we have a problem.  I somehow doubt there are container images out in the world where something actually *depends* on having the atime being honored even if it's encoded in the tar stream.

cgwalters:
Well, I don't want to spend too much time going down this rabbit hole, but it sure looks to me like the default GNU tar option won't serialize atime/ctime, and GNU tar is commonly used to make "base images".  And it looks to me like the same is true of the current c/storage code; I don't see it assigning to `AccessTime`, though the code tries to support [unpacking it at least](https://github.com/containers/storage/blob/c989f8f2d4ee28bf3d181beffaa1c09341545758/pkg/archive/archive.go#L732).

Anyways so...do we really need to care about atime/ctime?  I am doubtful.  But regardless again, it can't be in `composefs-from-json` as there is no support for it.

drivers/overlay/composefs_supported.go
giuseppe: Probably I am just overthinking it and it is not worth the extra-complexity, but in this way I can open a read-only file descriptor as well (to use for setting the fs-verity flag) and make sure both the fds point to the same file.

cgwalters:
Another concern I have with all of this is that we're going to be passing this JSON to C code, which pretty quickly hands it off to libcomposefs (also in C) - and I don't think the current libcomposefs C library is written in a "tries hard to be resistant to malicious input" style.

And this JSON may contain malicious input.

(I see that `composefs-from-json` does some sandboxing on its own.  But that's kind of weird because it's designed to be run from a container runtime...why can't we just have the caller do sandboxing?  Among other things, the calling container runtime is in a better position to e.g. set up a full read-only mount namespace, only allowing writing to the target place)

But more generally, we should probably try to better define the security boundaries here.  In general I think the [chromium "Rule of 2"](https://chromium.googlesource.com/chromium/src/+/master/docs/security/rule-of-2.md) is a good idea.  OK, in this case we are at least using the Go JSON parser first, then passing that pre-validated buffer to C, but still.

giuseppe:
the xattrs are encoded in base64 so we can use a string: https://github.com/containers/storage/blob/main/pkg/chunked/compressor/compressor.go#L378

and `composefs-from-json` handles that: https://github.com/containers/composefs/blob/main/tools/composefs-from-json.c#L347

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,91 @@
...
+	return ComposeFsHelperPath, ComposeFsHelperErr
+}
+
+func composeFsSupported() bool {
+	_, err := getComposeFsHelper()
+	return err == nil
+}
+
+func generateComposeFsBlob(toc []byte, destFile string) error {
+	outFd, err := unix.Openat(unix.AT_FDCWD, destFile, unix.O_WRONLY|unix.O_CREAT|unix.O_TRUNC|unix.O_EXCL, 0644)
```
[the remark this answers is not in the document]
giuseppe: that would prevent to pass the fd to the `composefs-from-json` helper process

drivers/overlay/composefs_supported.go
giuseppe: `mount.composefs` expects a `basedir` option, which I am not using at this moment since this will be picked later when doing the overlay mount.  So I really need to just mount the EROFS file system here

drivers/overlay/composefs_supported.go
giuseppe: yes good idea.  I've dropped that

hsiangkao:
> do we really need to care about atime/ctime? I am doubtful.

`atime` and `ctime` are much of filesystem internal state and volatile. I'm not sure if they are useful (`ctime`  was once considered but it could be changed due to inode metadata changes, so I and Android guys decided to shift it to `mtime` finally which also keep in sync with tarball. CRFS[1] also only keeps `mtime`). Apart from these, we also have `crtime`.

Anyway, if there could be further use cases, some xattrs could be used to keep that instead but I think use cases of `ctime`, `atime` are quite limited.

[1] https://github.com/google/crfs/blob/master/stargz/stargz.go#L127

rhatdan:
@alexlarsson @cgwalters @nalind @mtrmac PTAL

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/containers/storage/pkg/loopback"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+var (
+	ComposeFsHelperOnce sync.Once
+	ComposeFsHelperPath string
+	ComposeFsHelperErr  error
```
nalind, line 25: Do these variables need to be exported?  I don't see them being used from outside of the package.

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		f, err := os.Open(path)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+
+		if err := enableVerity(path, int(f.Fd())); err != nil {
+			return fmt.Errorf("file %s: %w", path, err)
```
nalind, line 70: An error returned by `enableVerity` will already include the path name of the file, so there's nothing to be gained by wrapping the error to get it again.

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,144 @@
...
+
+	newFd, err := unix.Open(fmt.Sprintf("/proc/self/fd/%d", outFd), unix.O_RDONLY, 0)
+	if err != nil {
+		return fmt.Errorf("failed to dup output file: %w", err)
+	}
+	defer unix.Close(newFd)
+
+	writerJson, err := getComposeFsHelper()
+	if err != nil {
+		return fmt.Errorf("failed to find composefs-from-json: %w", err)
```
nalind, line 91: This branch doesn't close `outFd`.

drivers/overlay/overlay.go
```diff
@@ -2001,6 +2136,16 @@ func (d *Driver) ApplyDiff(id, parent string, options graphdriver.ApplyDiffOpts)
 	return directory.Size(applyDir)
 }
 
+func (d *Driver) getComposeFsPath(id string) string {
+	dir := d.dir(id)
+	return path.Join(dir, ".cfs")
```
nalind, line 2141: Is there a requirement that this directory not show up if someone runs `ls` on its parent directory?

nalind:
> OK, yes I see this JSON format was introduced in [a0ff89e](https://github.com/containers/storage/commit/a0ff89e09dcf3f345c31642c8820ed74934e06f3)
> 
> A few things here. I think a few choices in this JSON format are suboptimal:
> 
>     * using `string` for symlink targets requires them to be valid Unicode; I would bet money that there are container images with non-Unicode filenames.

Strictly speaking, Go does not require `string` values to be valid UTF-8.

cgwalters:
> Strictly speaking, Go does not require string values to be valid UTF-8.

Yeah, you're right; as of lately I've been writing less Go and more Rust and there's a strong separation between arbitrary bytes and Unicode data there.

That said...I think my point is still somewhat valid here because this JSON is going to be used to interchange data with other processes written in different languages (as we are here), and when one is doing that it's important to try to nail these things down more.  

And overall I just don't think JSON makes sense for this use case.

giuseppe:
I am not opposing changing the representation but we are using the same format as CRFS and estargz, one advantage is that these files work almost out of the box with the containerd `stargz-snapshotter`.

I thought about using the composefs format itself to describe the files in the layer, all we needed was two extra pointers to refer the offset in the compressed stream and its size; but I think that is out of the table for now with the EROFS integration.

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,145 @@
...
+		buf := toc
+		for len(buf) > 0 {
+			n, err := unix.Write(fd, buf)
+			if err != nil {
+				return fmt.Errorf("failed to write to memfd: %w", err)
+			}
+			buf = buf[n:]
+		}
+
+		cmd := exec.Command(writerJson, "--format=erofs", fmt.Sprintf("--out=/proc/self/fd/%d", outFd), fmt.Sprintf("/proc/self/fd/%d", fd))
```
nalind, line 119: This doesn't appear to add either `outFd` or `fd`, or wrappers for them, to `cmd.ExtraFiles`.  Will the child process not close them before exec()ing the helper, or are they somehow being assigned to 0 and 1?

giuseppe: good question!  The code seems to work fine, in fact, I am getting the manifest generated; it seems that the Go runtime doesn't really close the FDs when they are opened directly with `unix.Open`.  In any case, it is an error, so I've added explicitly the files we need to `ExtraFiles`

drivers/overlay/overlay.go
nalind: I don't follow.  `dir` is the parent of the layer's `diff` directory, if it has one, so `.cfs` shouldn't be visible in mounted layers?

cgwalters:
> I am not opposing changing the representation but we are using the same format as CRFS and estargz, one advantage is that these files work almost out of the box with the containerd stargz-snapshotter.

Ah, I see.  But in practice...since the c/storage flow is using zstd, are these two things *in practice* compatible?  If I wanted to pull an image prepared with CRFS/estargz and then ship it via zstd it requires a transformation anyways, at which point we don't need to use their JSON format.

Anyways, what's most important to me here is to nail down more precisely what the boundaries and interfaces are between higher level tools (ostree, c/storage) and composefs.  As of right now in composefs we have *three* interfaces:

- `mkcomposefs`: Takes an input real directory
- `libcomposefs`: C library
- `composefs-from-json`: Accepts this CRFS json, also implicitly acts as "Go composefs bindings" due to the Go-oriented things like base64 strings.

I am arguing this should really shrink to *one* core interface, which is that `mkcomposefs` (a binary) accepts a sane streaming format on stdin, and also as as convenience accepts an input directory.

For expedience now, we could move the core functionality that lives in `composefs-from-json` into `mkcomposefs --from-crfs-json` or so?

giuseppe:
> Anyways, what's most important to me here is to nail down more precisely what the boundaries and interfaces are between higher level tools (ostree, c/storage) and composefs. As of right now in composefs we have _three_ interfaces:
> 
> * `mkcomposefs`: Takes an input real directory
> * `libcomposefs`: C library
> * `composefs-from-json`: Accepts this CRFS json, also implicitly acts as "Go composefs bindings" due to the Go-oriented things like base64 strings.

I'd consider `mkcomposefs` just a debugging tool, and we can move the `composefs-from-json` program into c/storage, either a separate program or embed it directly into the storage library, so that the only interface is the `libcomposefs` C library.

cgwalters:
I have a 30%-working start to teaching `mkcomposefs` to accept a serialized input stream on a pipe and then `composefs-from-json` can actually just be written in Go, see https://github.com/cgwalters/composefs/commit/faea356b908bb47957b74cc3f810fe7f3dada6c5

I'm also OK with having the C library be primary.  But at a practical level then, would we try to install it as a shared library for real (with all the symbol versioning, ABI commitments etc.) or just change c/storage to statically link it, as we're doing with ostree now?

cgwalters:
> If there are no blockers, let's merge this PR and iterate again

From my PoV (not a regular c/storage reviewer obviously), yes that makes sense to me.

> So if we keep a composefs-from-json binary, the current PR would still work?

For sure let's keep `composefs-from-json` for now!  

But let's aim to Soon (weeks?) try to close in on the key decision of exactly what the interface to composefs is (binary or C shlib, or both) - and bigger picture, try to converge on sharing more code here between c/storage and ostree (and other users).

cgwalters:
One thing I think is somewhat important here and is much easier after merging this is setting up a CI job which enables composefs for c/storage and runs through at least some of the basic podman tests.

rhatdan:
To merge, you need to remove the draft flag.

drivers/overlay/overlay.go
giuseppe: I've simplified the code and not create the `.cfs` subdirectory.

It is not yet added to the composefs tools, so this was just a preparatory work.  At this point it is not even clear if that is going to be used at all anyway, since overlay will support "data only" layers that are used only for metacopy operations but not to merge the content of directories.

giuseppe:
> To merge, you need to remove the draft flag.

I had noticed a comment from Nalin that I didn't deal with before.  I've fixed it in the last version pushed

giuseppe:
it is ready now

rhatdan:
@alexlarsson @cgwalters @nalind @mtrmac PTAL

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		unix.Close(outFd)
+		return fmt.Errorf("failed to dup output file: %w", err)
+	}
+	defer unix.Close(newFd)
+
+	err = func() error {
+		defer unix.Close(outFd)
+
+		fd, err := unix.MemfdCreate("json-toc", unix.MFD_ALLOW_SEALING)
```
cgwalters, line 98: Not blocking but I still don't understand why a memfd here instead of a plain `pipe`.  Generally for one-way data transfer too like this an `O_TMPFILE` also works and has the advantage that from the Go perspective it's just a regular `File`.

cgwalters (approved):
I'm not really an "official" reviewer here but at this point I think it's undergone enough review to land because this is an off-by-default thing.

What seems important here to me again is to have this tested out with podman in some place other than Giuseppe's development box, and that gets easier after it's merged.

pkg/chunked/storage_linux.go
```diff
@@ -1324,6 +1332,35 @@ func (c *chunkedDiffer) findAndCopyFile(dirfd int, r *internal.FileMetadata, cop
...
+func makeEntriesFlat(mergedEntries []internal.FileMetadata) ([]internal.FileMetadata, error) {
+	var new []internal.FileMetadata
+
+	hashes := make(map[string]string)
+	for i := range mergedEntries {
+		if mergedEntries[i].Type != TypeReg {
+			continue
+		}
+		if mergedEntries[i].Digest == "" {
+			continue
```
alexlarsson, line 1344: Should missing digests maybe be an error

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,82 @@
...
+	return err == nil
+}
+
+func generateComposeFsBlob(toc []byte, destFile string) error {
+	writerJson, err := getComposeFsHelper()
+	if err != nil {
+		return fmt.Errorf("failed to find composefs-from-json: %w", err)
+	}
+
+	outFd, err := unix.Openat(unix.AT_FDCWD, destFile, unix.O_WRONLY|unix.O_CREAT|unix.O_TRUNC|unix.O_EXCL, 0o644)
```
alexlarsson, line 83: unix.O_CLOEXEC ?

pkg/chunked/storage_linux.go
giuseppe: the files without digest are the empty files and they are simply ignored from the output.  I can improve the check though

drivers/overlay/composefs_supported.go
giuseppe: changed to a regular pipe

drivers/overlay/composefs_supported.go
giuseppe: I need to pass the fd to the writer-json process, it cannot be O_CLOEXEC

cgwalters:
, 
> unix.O_WRONLY|unix.O_CREAT|unix.O_TRUNC|unix.O_EXCL, 0o644)
> I need to pass the fd to the writer-json process, it cannot be O_CLOEXEC

Go will dup the fd when passing via the Command struct extra files

giuseppe:
> ,
>  unix.O_WRONLY|unix.O_CREAT|unix.O_TRUNC|unix.O_EXCL, 0o644)
>  I need to pass the fd to the writer-json process, it cannot be O_CLOEXEC
> Go will dup the fd when passing via the Command struct extra files

I've changed it to be `O_EXCL` and use ExtraFiles

pkg/chunked/storage_linux.go
alexlarsson: Well, You might have for whatever reason a buggy json file with non-empty files without digests. Might be good to report that.

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,66 @@
...
+	cmd.Stdin = bytes.NewReader(toc)
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to convert json to erofs: %w", err)
+	}
+	return nil
+}
+
+func mountErofsBlob(blobFile, mountPoint string) error {
+	loop, err := loopback.AttachLoopDevice(blobFile)
```
alexlarsson, line 123: This is not setting the loopback device read-only. I think we should do that. (composefs does)

alexlarsson: And, ideally it should pass "noacl" if the composefs header flag that says there are no acls in the file is set (this will improve performance quite a lot).

drivers/overlay/composefs_supported.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return err
+	}
+	defer loop.Close()
+
+	return unix.Mount(loop.Name(), mountPoint, "erofs", unix.MS_RDONLY, "")
```
alexlarsson, line 129: I.e. here, we should have `noacl` option if possible.

drivers/overlay/overlay.go
```diff
@@ -1551,6 +1611,30 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
...
+			return "", err
+		}
+		lowerID := filepath.Base(filepath.Dir(linkContent))
+		erofsMount, err := maybeAddErofsMount(lowerID, i+1)
+		if err != nil {
+			return "", err
+		}
+		if erofsMount != "" {
+			if needsIDMapping {
+				if err := idmap.CreateIDMappedMount(erofsMount, erofsMount, idmappedMountProcessPid); err != nil {
```
alexlarsson, line 1626: In mount.composefs I do the idmap as part of the erofs mount, using the mount_setattr syscall. Maybe its hard to do that here though. It needs the new mount api.

drivers/overlay/composefs_supported.go
```diff
@@ -32,6 +38,43 @@ func composeFsSupported() bool {
...
+func enableVerity(description string, fd int) error {
+	enableArg := unix.FsverityEnableArg{
+		Version:        1,
+		Hash_algorithm: unix.FS_VERITY_HASH_ALG_SHA256,
+		Block_size:     4096,
+	}
+
+	_, _, e1 := syscall.Syscall(unix.SYS_IOCTL, uintptr(fd), uintptr(unix.FS_IOC_ENABLE_VERITY), uintptr(unsafe.Pointer(&enableArg)))
+	if e1 != 0 {
+		return fmt.Errorf("failed to enable verity for %q: %w", description, e1)
```
alexlarsson, line 50: This should silently accept EEXIST (fs-verity already enabled for the file).

alexlarsson: I mean, couldn't this happen if a layer file is hard-linked from another layer?

alexlarsson:
I added a few comments, but overall I think this looks good. The main issue I see right now is that it depends on the composefs-from-json helper, which is currenly not installed from composefs. We have to chose how to do this. Alternatives:

1) Install and support if in composefs
2) Import the code into mkcomposefs and use that
3) Call libcomposefs from containers/storage, reimplementing the json stuff.

I don't know how painful 3 is, calling C is always icky. However, we do already parse the json in the go code, so not having to do that twice is a win.

drivers/overlay/overlay.go
giuseppe: yes, I need some refactoring since we already have that code in `pkg/idmap`, for now I've left a TODO in the code

drivers/overlay/composefs_supported.go
giuseppe: yes, this could happen with hard link deduplication

giuseppe:
I've looked into the comments that do not require touching other packages, I can take care of other stuff later with more targetted PR.  The composefs integration though works only with single layered images at the moment, since multiple data layers conflict with each other, so this still needs the "data only" feature in overlay and we can avoid creating the whiteout files as part of them

alexlarsson:
Data only overlayfs support has landed and will be in 6.5 btw.

drivers/overlay/composefs_supported.go
alexlarsson: But if an OCI layer tar has the ACL xattr set, historically that would have been extracted to the lowerdir file, and when userspace looked at the file it would have seen the ACL, no?

alexlarsson: Honestly, I have no idea if ACLs work in the wild at all though.

alexlarsson:
Given this is not enabled by default currently i think it makes a lot to merge this as is.

giuseppe:
> I don't know how painful 3 is, calling C is always icky. However, we do already parse the json in the go code, so not having to do that twice is a win.

I've a WIP branch with a CGo implementation: https://github.com/giuseppe/storage/tree/chunked-libcomposefs-Cgo so we could replace the current `composefs-from-json` dependency.

I'd need to do some sandboxing and ideally re-execute Podman.

I've dropped the `noacl` for now, let's take care of this optimization later.  The code in overlay doesn't know about the TOC file and the zstd:chunked metadata, so we need to check this information in the chunked package and then just reuse from overlay.

> Given this is not enabled by default currently i think it makes a lot to merge this as is.

I am trying to reduce the number of changes so we can finally merge this PoC and iterate again from here, and also don't want to get back from PTO straight to merge conflicts :-)

drivers/overlay/composefs_supported.go
giuseppe: I've dropped it for now, let's be safe and handle the optimization later.  I'd need to add the check in pkg/chunked and store somewhere that the image has no ACLs.

rhatdan:
LGTM

giuseppe:
opened a PR to handle the `noacl` mount flag: https://github.com/containers/storage/pull/1663

PR #1647, opened 2023-06-19
drivers: add EROFS filesystem magic

hsiangkao:
Composefs use cases use on-disk EROFS for metadata.

I hope more people could be interested in EROFS development and make it better together, hopefully...


cgwalters (approved):

rhatdan:
LGTM

PR #1650, opened 2023-06-27
Fix error if continueWrite/continueRead pipe open fails

dvdsk:
Hi all,  

I was reading the source to learn more about how unshare is used in podman. It seems error message that was on this line was copied from a few lines above by accident. 

Since the guidlines say "*No Pull Request (PR) is too small!*" here it is.


rhatdan:
Thanks @dvdsk 
LGTM

giuseppe (approved):
/lgtm

PR #1651, opened 2023-06-27
pkg/regexp: make sure that &Regexp implements the interfaces

nalind:
Making our exported Regexp type a pointer breaks code that expects to be able to take its address and call methods on the result.
Should fix vendor test error at https://cirrus-ci.com/task/6751444020232192?logs=main#L222.


rhatdan:
LGTM

PR #1652, opened 2023-06-28
Remove use of fillGo18FileTypeBits

rhatdan:
Fixes for https://github.com/containers/buildah/pull/4772


rhatdan:
@nalind Is this what you intended?

pkg/archive/archive.go
```diff
@@ -358,7 +358,6 @@ func FileInfoHeader(name string, fi os.FileInfo, link string) (*tar.Header, erro
 	if err != nil {
 		return nil, err
 	}
-	hdr.Mode = fillGo18FileTypeBits(int64(chmodTarEntry(os.FileMode(hdr.Mode))), fi)
```
flouthoc, line 361: I think  `hdr.Mode` should still be set.

```diff
+ hdr.Mode = int64(chmodTarEntry(os.FileMode(hdr.Mode)))
```

flouthoc:
I think we should still set the mode without `file` and `dir` bits ?

```diff
-hdr.Mode = fillGo18FileTypeBits(int64(chmodTarEntry(os.FileMode(hdr.Mode))), fi)
+hdr.Mode = int64(chmodTarEntry(os.FileMode(hdr.Mode)))
```

nalind:
LGTM, but does it fix the issue?

cevich:
> LGTM @cevich would it be possible to pull this commit in the branch for running the CI ?

You can manually with `hack/get_ci_vm.sh <task name>`, just fixup the system how you need, and run the necessary `*_script:` commands directly from `.cirrus.yml` and they **should** do the right thing (this is not a guarantee).

Another option is the `re-run in terminal` trick on an already finished branch run.  I'm assuming you're talking about buildah CI?  Check out https://cirrus-ci.com/github/containers/buildah/main (or replace `main` with whatever branch you like).  Pick a recent green ***NON-CRON-INVOCATION*** job and poke away.  
(the "cron-invocation" jobs are monitored)

PR #1653, opened 2023-06-28
[RFC] overlay: make sure directories omitted from layers have the right permissions

nalind:
When extracting layer contents, if we find ourselves needing to implicitly create a directory (due to its not being included in the layer diff), try to give it the permissions, ownership, attributes, and datestamp of the corresponding directory from the layer which would be stacked below it.
    
When a layer omits any of the directories which contain items which that layer adds or modifies, this should prevent the default values that we would use from overriding those which correspond to the same directory
in a lower layer, which could later be mistaken as an indication that one or more of those was intentionally changed, forcing the directory to be pulled up.

### DIFF
--- a/pkg/tarbackfill/tarbackfill.go
+++ b/pkg/tarbackfill/tarbackfill.go
@@ -1,0 +1,159 @@
+package tarbackfill
+
+import (
+	"archive/tar"
+	"io"
+	"path"
+	"strings"
+)
+
+// Reader wraps a tar.Reader so that if an item which would be read from it is
+// in a directory which is not included in the archive, a specified Backfiller
+// interface's Backfill() method will be called to supply a tar.Header which
+// will be inserted into the stream just ahead of that item.
+type Reader struct {
+	*tar.Reader
+	backfiller      Backfiller
+	seen            map[string]struct{}
+	queue           []*tar.Header
+	currentIsQueued bool
+	err             error
+}
+
+// Backfiller is a wrapper for Backfill, which can supply headers to insert
+// into an archive which is on its way to being extracted.
+type Backfiller interface {
+	// Backfill either returns an entry for the passed-in path, nil if
+	// no entry should be added to the stream, or an error if something
+	// unexpected happened.
+	Backfill(string) (*tar.Header, error)
+}
+
+// NewReaderWithBackfiller creates a new Reader reading from r, asking the
+// passed-in Backfiller for information about parent directories which it
+// hasn't seen yet.
+func NewReaderWithBackfiller(r *tar.Reader, backfiller Backfiller) *Reader {
+	reader := &Reader{
+		Reader:     r,
+		backfiller: backfiller,
+		seen:       make(map[string]struct{}),
+	}
+	return reader
+}
+
+// Next returns either the next item from the archive we're filtering, or a
+// synthesized entry for a directory that arguably should have been in that
+// archive.
+func (r *Reader) Next() (*tar.Header, error) {
+	// Drain the queue first.
+	if len(r.queue) > 0 {
+		next, queue := r.queue[0], r.queue[1:]
+		r.queue = queue
+		r.currentIsQueued = len(r.queue) > 0
+		return next, nil
+	}
+	// If we've hit the end of the archive, we've hit the end of the archive.
+	r.currentIsQueued = false
+	if r.err != nil {
+		return nil, r.err
+	}
+	// Check what's next in the archive.
+	hdr, err := r.Reader.Next()
+	if err != nil {
+		r.err = err
+	}
+	if hdr == nil {
+		return hdr, err
+	}
+	for {
+		// Trim off an initial or final path separator.
+		name := strings.Trim(hdr.Name, "/")
+		if hdr.Typeflag == tar.TypeDir {
+			// Trim off an initial or final path separator, and
+			// note that we won't need to supply it later.
+			r.seen[name] = struct{}{}
+		}
+		// Figure out which directory this item is directly in.
+		p := name
+		dir, _ := path.Split(name)
+		var newHdr *tar.Header
+		for dir != p {
+			var bfErr error
+			dir = strings.Trim(dir, "/")
+			// If we already saw that directory, no need to interfere (further).
+			if _, ok := r.seen[dir]; dir == "" || ok || dir == name {
+				return hdr, err
+			}
+			// Ask the backfiller what to do here.
+			newHdr, bfErr = r.backfiller.Backfill(dir)
+			if bfErr != nil {
+				r.err = bfErr
+				return nil, bfErr
+			}
+			if newHdr == nil {
+				dir, _ = path.Split(dir)
+				continue
+			}
+			// Make sure the Name looks right, then queue up the current entry.
+			newHdr.Format = tar.FormatPAX
+			newHdr.Name = strings.Trim(newHdr.Name, "/")
+			if newHdr.Typeflag == tar.TypeDir {
+				// We won't need to supply it later.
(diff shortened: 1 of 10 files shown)


rhatdan:
@vrothberg @mtrmac @giuseppe @saschagrunert PTAL

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Next returns either the next item from the archive we're filtering, or a
+// synthesized entry for a directory that arguably should have been in that
+// archive.
+func (r *Reader) Next() (*tar.Header, error) {
+	// Drain the queue first.
+	if len(r.queue) > 0 {
+		next, queue := r.queue[0], r.queue[1:]
+		r.queue = queue
+		r.currentIsQueued = len(r.queue) > 0
```
mtrmac, line 52: Shouldn’t this actually be true for this just-dequeued entry as well, i.e. just `true` unconditionally on this path?

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+// Reader wraps a tar.Reader so that if an item which would be read from it is
+// in a directory which is not included in the archive, a specified Backfiller
+// interface's Backfill() method will be called to supply a tar.Header which
+// will be inserted into the stream just ahead of that item.
+type Reader struct {
+	*tar.Reader
+	backfiller      Backfiller
+	seen            map[string]struct{}
```
mtrmac, line 17: (Some documentation, or maybe naming this `seenDirs`, would be helpful for readers.)

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
...
+// archive.
+func (r *Reader) Next() (*tar.Header, error) {
+	// Drain the queue first.
+	if len(r.queue) > 0 {
+		next, queue := r.queue[0], r.queue[1:]
+		r.queue = queue
+		r.currentIsQueued = len(r.queue) > 0
+		return next, nil
+	}
+	// If we've hit the end of the archive, we've hit the end of the archive.
```
mtrmac, line 55: Is this intended to apply to the `r.err` check one line below instead?

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
...
+		if hdr.Typeflag == tar.TypeDir {
+			// Trim off an initial or final path separator, and
+			// note that we won't need to supply it later.
+			r.seen[name] = struct{}{}
+		}
+		// Figure out which directory this item is directly in.
+		p := name
+		dir, _ := path.Split(name)
+		var newHdr *tar.Header
+		for dir != p {
```
mtrmac, line 80: I don’t understand this condition. `dir` is a parent or ancestor of `p == name`, and `p` never changes, so how can this ever be `false` (other than I guess at the very first iteration, if `name` is `/'`)?

Best I can guess, this loop actually terminates by setting, and noticing, `r.seen[""]` — but I’m not sure, I’m quite possibly reading the code wrong.

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
...
+			newHdr.Name = strings.Trim(newHdr.Name, "/")
+			if newHdr.Typeflag == tar.TypeDir {
+				// We won't need to supply it later.
+				r.seen[newHdr.Name] = struct{}{}
+				newHdr.Name += "/"
+			}
+			r.queue = append([]*tar.Header{hdr}, r.queue...)
+			hdr = newHdr
+			r.currentIsQueued = true
+			dir, _ = path.Split(dir)
```
mtrmac, line 108: Parts of this seem ~redundant with the outer infinite loop.

Or another way to put it, it seems to me that the inner loop could iterate through `dir, _ = path.Split(dir)` just to find a backfill match; and the inner loop could iterate through a series of backfilled-ancestor tar headers.

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+	}
+}
+
+// Read will either read from a real entry in the archive, or pretend very hard
+// that an entry we inserted had no content.
+func (r *Reader) Read(b []byte) (int, error) {
+	if r.currentIsQueued {
+		return 0, nil
```
mtrmac, line 117: https://pkg.go.dev/archive/tar#Reader.Read documents `0, io.EOF` on both special file types, and empty regular files.

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
...
+			hdr = newHdr
+			r.currentIsQueued = true
+			dir, _ = path.Split(dir)
+		}
+	}
+}
+
+// Read will either read from a real entry in the archive, or pretend very hard
+// that an entry we inserted had no content.
+func (r *Reader) Read(b []byte) (int, error) {
```
mtrmac, line 115: What … happens … when the parent layer contains a regular file in place of an expected parent directory?

I suppose it doesn’t matter what we do here, because extracting the current layer should fail, shouldn’t it?

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
...
+func NewIOReaderWithBackfiller(reader io.Reader, backfiller Backfiller) io.ReadCloser {
+	rc, wc := io.Pipe()
+	go func() {
+		r := tar.NewReader(reader)
+		tr := NewReaderWithBackfiller(r, backfiller)
+		tw := tar.NewWriter(wc)
+		hdr, err := tr.Next()
+		defer func() {
+			closeErr := tw.Close()
+			io.Copy(wc, reader)
```
mtrmac, line 134: What does this do? If `tw.Close` succeeded, the stream is already terminated by the tar EOF blocks.

And even if that were not true, both the reader and the writer are in an indeterminate (or anyway hard-for-me-to-reason-about) state.

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
...
+			if err != nil {
+				wc.CloseWithError(err)
+			} else if closeErr != nil {
+				wc.CloseWithError(closeErr)
+			} else {
+				wc.Close()
```
mtrmac, line 140: (`Close()` is equivalent to `CloseWithError(nil)`, so this could be:
```go
if err == nil { err = closeErr }
wc.CloseWithError(err)
```

pkg/tarbackfill/tarbackfill.go
```diff
@@ -0,0 +1,159 @@
+package tarbackfill
```
mtrmac, line 1: Should this be a public c/storage API, or would something like `c/storage/internal/tarbackfill` work?

(I don’t have a strong opinion, I just want this to be an intentional decision.)

mtrmac (commented):
Doing this makes sense to me, but oh it’s _a lot_ of code to make that possible. That’s an impressive amount of effort.

(I have only fairly briefly skimmed the tests. But their existence adds a lot of confidence anyway :) )

rhatdan:
@nalind Waiting on you?

cgwalters:
> which could later be mistaken as an indication that one or more of those was intentionally changed, forcing the directory to be pulled up.

Another way to solve this possibly: Add an xattr to the directory like `user.containers.meta` (with serialized extended state, json/bincode/whatever) that would include to start a boolean value like `synthetic: true` that denotes this state, and we know hence when re-serializing to ignore that directory

jankaluza:

PR #1656, opened 2023-06-30
refactor(pattern-matcher): add `CanSkipDir` method in patternMatcher

BlackHole1:
1. Remove deprecated method: `Matches` in `PatternMatcher`
2. Add `CanSkipDir` to `MatchesResult`
3. Improve `MatchesResult`
   1. Remove the `matches` and `excludes` fields (no one is using these two fields)

This PR is mainly aimed at addressing https://github.com/containers/podman/blob/fca3c2ef8476669825564af89b008946bdf87599/pkg/bindings/images/build.go#L699-L701

> For the context regarding the removal of Matches and the modification based on MatchesResult, please refer to: https://github.com/containers/buildah/discussions/4897


BlackHole1:
@rhatdan @giuseppe @vrothberg PTAL

pkg/fileutils/fileutils.go
```diff
@@ -57,63 +57,30 @@ func NewPatternMatcher(patterns []string) (*PatternMatcher, error) {
 	return pm, nil
 }
 
-// Deprecated: Please use the `MatchesResult` method instead.
-// Matches matches path against all the patterns. Matches is not safe to be
-// called concurrently
-func (pm *PatternMatcher) Matches(file string) (bool, error) {
```
vrothberg, line 63: Despite being deprecated, this can break consumers and is hence a breaking change which in turn would require to bump the major version of c/storage to v2.0.0.  Let's not go down this path and leave the API as is.

pkg/fileutils/fileutils.go
```diff
@@ -57,63 +57,30 @@ func NewPatternMatcher(patterns []string) (*PatternMatcher, error) {
...
-	if matched {
-		logrus.Debugf("Skipping excluded path: %s", file)
-	}
-
-	return matched, nil
-}
-
 type MatchResult struct {
-	isMatched         bool
-	matches, excludes uint
```
vrothberg, line 93: Same here.  That's a breaking a change.

BlackHole1:
Thank you for your CR @vrothberg. Modification Completed.

pkg/fileutils/fileutils.go
```diff
@@ -129,16 +137,34 @@ func (pm *PatternMatcher) MatchesResult(file string) (res *MatchResult, err erro
...
 				res.matches++
+				matchPatterns = append(matchPatterns, pattern)
+			} else {
+				res.excludes++
 			}
 		}
 	}
 
-	if res.matches > 0 {
+	if res.isMatched {
```
BlackHole1, line 149: The previous implementation had a bug. When the rule is `["1", "!1"]`, it would also enter this conditional statement, but it should not.

vrothberg (commented):
Thanks! @saschagrunert PTAL

rhatdan:
LGTM

nalind:
The comment in the podman source is incorrect - podman only needs to (and by that I mean _should_) descend into a directory that's covered by an ignore pattern if the directory exactly matches the initial portion of a '!' pattern - it doesn't need to descend into directories that might only match a '!' pattern by matching using wild cards.  I don't think that this will help podman determine if that's the case.

BlackHole1:
Hey, @nalind. I didn't understand what you meant. Could you please provide a few examples to illustrate your point?

nalind:
> Hey, @nalind. I didn't understand what you meant. Could you please provide a few examples to illustrate your point?
> 
> If you could provide an example in the format of a unit test, it would be even better. For example: `{[]string{"/prefix*", "!prefix/path"}, "prefix/path", 0, 1, false, false},`

Specifically, if I have a directory in my build context named "foo" with a file in it named "bar", and my ignores list is ["f*", "!f*/*"], code that wants to do things the way `docker build` does shouldn't descend into "foo" to pick up "foo/bar".  If, on the other hand, the ignores list is ["f*", "!foo/*"], the directory should be scanned, because there's a "!" pattern which has the directory's name, as an exact prefix, after a pattern which would have caused it to be skipped.

In this version of the PR, the last test case in `TestMatchesAmount()` would probably be {[]string{"/prefix*", "!prefix*/path"}, "prefix/path", 0, 1, false, true}`, but take that with the caveat that the goal for this code isn't to be strictly correct, but to be compatible with what `docker build` does, which has some odd corner cases, including a couple that I've only just noticed that buildah's conformance tests weren't exercising.

BlackHole1:
@nalind Ping.

rhatdan:
@nalind PTAL

jankaluza:

PR #1662, opened 2023-07-05
Implement windows lockfile

mikenorgate:
This is an implementation of `lockfile` for Windows, it includes a refactor of the Linux code to remove duplication

Signed-off-by: Mike Norgate <mike.norgate@mythical.games>


rhatdan:
@mtrmac @vrothberg @n1hility @giuseppe @nalind PTAL

giuseppe (approved):
LGTM

pkg/lockfile/lockfile_windows.go
```diff
@@ -102,51 +52,47 @@ func (l *LockFile) GetLastWrite() (LastWrite, error) {
...
+func openHandle(path string, mode int) (uintptr, error) {
+	fd, err := windows.Open(path, mode, windows.S_IWRITE)
+	return uintptr(fd), err
 }
 
-func (l *LockFile) IsReadWrite() bool {
-	return false
+func lockHandle(fd uintptr, lType int16) {
+	ol := new(windows.Overlapped)
+	for windows.LockFileEx(windows.Handle(fd), uint32(lType), reserved, allBytes, allBytes, ol) != nil {
```
n1hility, line 90: In the Windows synchronous case (with no FAIL_IMMEDIATELY flag), all errors would be fatal, and retrying indefinitely would result in a hang. Since the API doesn't provide error propagation, I think it's better to panic if this happens. or at least retry a limited number of times.

This is a little bit different than the Unix side of the equation, which needs to loop through signal interruption, although arguably that code should only be doing that when errno == EINTR

n1hility (changes requested):

n1hility:
i only have one comment/suggestion here and its not a showstopper imo

pkg/lockfile/lockfile_windows.go
mikenorgate: I think panic makes more sense than retrying an arbitrary number of times. I've pushed the change

n1hility (approved):
Thanks! LGTM

pkg/lockfile/lockfile_windows.go
```diff
@@ -5,89 +5,39 @@ package lockfile
...
-func (l *LockFile) AssertLockedForWriting() {
-	// DO NOT provide a variant that returns the current lock state.
-	//
-	// The same caveats as for AssertLocked apply equally.
-	l.AssertLocked() // The current implementation does not distinguish between read and write locks.
-}
+const (
+	F_WRLCK   = windows.LOCKFILE_EXCLUSIVE_LOCK
+	F_RDLCK   = 0
+	O_CLOEXEC = windows.O_CLOEXEC
```
mtrmac, line 16: (On both platforms) these constants should not end up being added to the public API of the subpackage.

mtrmac, line 16: (Aesthetically, I’m a bit unhappy with the generic code turning generic concepts like “read-write lock” into a UNIX-specific flag, only for the Windows implementation to pretend to be UNIX. It seems to me that the generic code could just pass a generic “read-only” flag and let the platform-specific handler select the right API.

But I didn’t actually check whether it would be easy to do it the other way… and the current approach is, I suppose, more readable for us Unix-native maintainers — and it is certainly less invasive. So I guess I don’t feel strongly about this.)

mikenorgate: @mtrmac I've push a commit which removes these constants in favour of a flag. Let me know if you feel this is a better approach

mtrmac: Thanks!

Yes, I think that’s much cleaner.

Two more possible improvements:
- Use an enum for the lock type, instead of a `bool`. OTOH it’s used in few enough places that we can just pay attention.
- Use a platform-specific `fileHandle` type instead of universal `uintptr`; then the Windows code could use a `windows.Handle`.

mtrmac (commented):
Now a full review.

I think the `unlockHandle` and the test parts need addressing; the rest is completely non-blocking subjective aesthetics.

mikenorgate:
@mtrmac I've made some updates based on your comments

pkg/lockfile/lockfile.go
```diff
@@ -128,3 +312,101 @@ func getLockfile(path string, ro bool) (*LockFile, error) {
...
+	}
+
+	return &LockFile{
+		file: path,
+		ro:   ro,
+
+		rwMutex:    &sync.RWMutex{},
+		stateMutex: &sync.Mutex{},
+		lw:         newLastWrite(), // For compatibility, the first call of .Modified() will always report a change.
+		lockType:   lType,
```
mtrmac, line 349: Note to self:
- Isn’t this field only used if `locked`? So it doesn’t actually need to be set here.
- Actually `locked` and `lockType` could be combined.

Something to clean up later…

pkg/lockfile/lockfile.go
```diff
@@ -128,3 +312,101 @@ func getLockfile(path string, ro bool) (*LockFile, error) {
...
+
+		return openLock(path, ro)
+	}
+
+	return fd, &os.PathError{Op: "open", Path: path, Err: err}
+}
+
+// lock locks the lockfile via syscall based on the specified type and
+// command.
+func (l *LockFile) lock(ro bool) {
```
mtrmac, line 386: `lock(lType lockType)` please

pkg/lockfile/lockfile.go
```diff
@@ -54,9 +55,42 @@ type Locker interface {
...
+	// In other cases, they need to be protected using stateMutex.
+	fd fileHandle
+}
+
 var (
-	lockFiles     map[string]*LockFile
-	lockFilesLock sync.Mutex
+	lockFiles           map[string]*LockFile
+	lockFilesLock       sync.Mutex
+	lastWriterIDCounter uint64 // Private state for newLastWriterID
```
mtrmac, line 93: This can also be moved to `lastwrite.go`.

pkg/lockfile/lockfile.go
```diff
@@ -128,3 +312,101 @@ func getLockfile(path string, ro bool) (*LockFile, error) {
...
+		fd, err := openLock(l.file, l.ro)
+		if err != nil {
+			panic(err)
+		}
+		l.fd = fd
+
+		// Optimization: only use the (expensive) syscall when
+		// the counter is 0.  In this case, we're either the first
+		// reader lock or a writer lock.
+		lockHandle(l.fd, ro)
```
mtrmac, line 407: `lockHandle(l.fd, lType)` please

mtrmac (commented):
Thanks!

LGTM, there are just a few more opportunities to use the new enum instead of a `bool`.

pkg/lockfile/lockfile.go
mtrmac: > * Actually `locked` and `lockType` could be combined.

On second thought, that might not be worthwhile, e.g. some of the `lock` functions would need to deal with the “unlocked” value.

pkg/lockfile/lastwrite.go
```diff
@@ -19,6 +20,12 @@ type LastWrite struct {
 	state []byte // Contents of the lock file.
 }
 
+var (
+	lockFiles           map[string]*LockFile
+	lockFilesLock       sync.Mutex
```
mtrmac, line 25: Non-blocking: `lastWriterIDCounter` relates to `LastWrite`, … but these two do belong back with the only user in `lockfile.go` :)

mtrmac (commented):
LGTM. Thanks!

mtrmac (approved):

PR #1663, opened 2023-07-05
overlay: store if there are no ACLs in the layer

giuseppe:
if there are no ACLs, then we mount the EROFS layer with the "noacl" mount option.
    
### DIFF
--- a/drivers/overlay/composefs_supported.go
+++ b/drivers/overlay/composefs_supported.go
@@ -5,6 +5,7 @@ package overlay
 
 import (
 	"bytes"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -128,6 +129,41 @@ func generateComposeFsBlob(toc []byte, composefsDir string) error {
 	return nil
 }
 
+/*
+typedef enum {
+	LCFS_EROFS_FLAGS_HAS_ACL = (1 << 0),
+} lcfs_erofs_flag_t;
+
+struct lcfs_erofs_header_s {
+	uint32_t magic;
+	uint32_t version;
+	uint32_t flags;
+	uint32_t unused[5];
+} __attribute__((__packed__));
+*/
+
+// hasACL returns true if the erofs blob has ACLs enabled
+func hasACL(path string) (bool, error) {
+	const LCFS_EROFS_FLAGS_HAS_ACL = (1 << 0)
+
+	fd, err := unix.Openat(unix.AT_FDCWD, path, unix.O_RDONLY|unix.O_CLOEXEC, 0)
+	if err != nil {
+		return false, err
+	}
+	defer unix.Close(fd)
+	// do not worry about checking the magic number, if the file is invalid
+	// we will fail to mount it anyway
+	flags := make([]byte, 4)
+	nread, err := unix.Pread(fd, flags, 8)
+	if err != nil {
+		return false, err
+	}
+	if nread != 4 {
+		return false, fmt.Errorf("failed to read flags from %q", path)
+	}
+	return binary.LittleEndian.Uint32(flags)&LCFS_EROFS_FLAGS_HAS_ACL != 0, nil
+}
+
 func mountComposefsBlob(dataDir, mountPoint string) error {
 	blobFile := getComposefsBlob(dataDir)
 	loop, err := loopback.AttachLoopDevice(blobFile)
@@ -136,5 +172,14 @@ func mountComposefsBlob(dataDir, mountPoint string) error {
 	}
 	defer loop.Close()
 
-	return unix.Mount(loop.Name(), mountPoint, "erofs", unix.MS_RDONLY, "ro")
+	hasACL, err := hasACL(blobFile)
+	if err != nil {
+		return err
+	}
+	mountOpts := "ro"
+	if !hasACL {
+		mountOpts += ",noacl"
+	}
+
+	return unix.Mount(loop.Name(), mountPoint, "erofs", unix.MS_RDONLY, mountOpts)
 }
--- a/drivers/overlay/composefs_supported.go
+++ b/drivers/overlay/composefs_supported.go
@@ -75,7 +75,16 @@ func enableVerityRecursive(path string) error {
 	return filepath.WalkDir(path, walkFn)
 }
 
-func generateComposeFsBlob(toc []byte, destFile string) error {
+func getComposefsBlob(dataDir string) string {
+	return filepath.Join(dataDir, "composefs.blob")
+}
+
+func generateComposeFsBlob(toc []byte, composefsDir string) error {
+	if err := os.MkdirAll(composefsDir, 0o700); err != nil {
+		return err
+	}
+
+	destFile := getComposefsBlob(composefsDir)
 	writerJson, err := getComposeFsHelper()
 	if err != nil {
 		return fmt.Errorf("failed to find composefs-from-json: %w", err)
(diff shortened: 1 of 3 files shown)


drivers/overlay/composefs_supported.go
```diff
@@ -75,7 +75,13 @@ func enableVerityRecursive(path string) error {
 	return filepath.WalkDir(path, walkFn)
 }
 
-func generateComposeFsBlob(toc []byte, destFile string) error {
+func getErofsBlob(dataDir string) string {
+	return filepath.Join(dataDir, "erofs.blob")
+}
+
+func generateComposeFsBlob(toc []byte, composefsDir string) error {
+	_ = os.MkdirAll(composefsDir, 0o700)
```
cgwalters, line 124: Any reason to ignore errors here?

drivers/overlay/composefs_supported.go
```diff
@@ -80,6 +80,46 @@ func getErofsBlob(dataDir string) string {
 	return filepath.Join(dataDir, "erofs.blob")
 }
 
+type composefsAdditionalData struct {
+	NoACL bool `json:"noACL"`
+}
+
+func getComposefsAdditionalDataFile(dataDir string) string {
+	return filepath.Join(dataDir, "composefs.additional-data")
```
cgwalters, line 88: Not a blocker but I think this would be cleaner as an xattr on the composefs data instead.

Also AIUI we only need this because we're not using `mount.composefs` which has https://github.com/containers/composefs/blob/ac729b579dd332938c6ad43ba83b2b896631dad4/libcomposefs/lcfs-mount.c#L404 right?

I'm a bit confused why we're not using `mount.composefs`.

giuseppe: I was also unsure whether using a file or a xattr and I am fine to change it.

I've preferred a file as a xattr is not covered by fs-verity if we ever need it and it would be a problem for rootless on tmpfs (so with `--transient-store) since `user.` is not supported

drivers/overlay/composefs_supported.go
giuseppe: I'll fix that

drivers/overlay/composefs_supported.go
giuseppe: > I'm a bit confused why we're not using `mount.composefs`.

because `mount.composefs` uses a single erofs mount with multiple data layers under an overlay mount.
In c/storage we are using multiple erofs (one for layer) above multiple data layers under the same overlay mount that includes also the writeable upper layer where necessary.

The model in c/storage doesn't work well yet with multiple layers since the whiteouts in each data layer conflict with other layers, but once the "data only" overlay patches land upstream, we don't need these whiteouts anymore so the data paths are "append only" instead of hiding files below them.  And we need to decide on the prefix to use to avoid polluting the root directory, e.g. `.cfs/ab/123456...` instead of using `ab/123456...` as we do now.

cgwalters: > because mount.composefs uses a single erofs mount with multiple data layers under an overlay mount.
In c/storage we are using multiple erofs (one for layer) above multiple data layers under the same overlay mount that includes also the writeable upper layer where necessary.

Right, OK.  Hmm.  

> . And we need to decide on the prefix to use to avoid polluting the root directory, e.g. .cfs/ab/123456... instead of using ab/123456... as we do now.

Yeah, agree on this.  Would definitely be cleaner to have one whiteout for `.cfs`.

cgwalters: > I've preferred a file as a xattr is not covered by fs-verity if we ever need it 

But this state is locally computed; if we enabled fs-verity, what key would sign it?  There is I guess value in having it be immutable post-creation though.

Anyways backing up a second...maybe what would work a bit better is something like `mount.composefs --inspect /path/to/cfs.img` and that exports JSON that we can read even if we're doing the mounting on our own?  Since we already have this "has acl" flags in the composefs image right?

> and it would be a problem for rootless on tmpfs (so with --transient-store) since user.` is not supported

Ah yes :cry:  ostree uses `user.` xattrs in some modes and the tmpfs thing has been a tremendous pain.

alexlarsson:
None of this seems needed. We already store the LCFS_EROFS_FLAGS_HAS_ACL in the struct lcfs_erofs_header_s at offset 0 in the image.

cgwalters:
> None of this seems needed. We already store the LCFS_EROFS_FLAGS_HAS_ACL in the struct lcfs_erofs_header_s at offset 0 in the image.

Right, I raised the same point in this thread https://github.com/containers/storage/pull/1663#discussion_r1254112039 but basically we're not using `mount.composefs` here.

drivers/overlay/composefs_supported.go
alexlarsson: No need for that. The lcfs_erofs_header_s struct and the flags for wheter an ACL is in the image is public and ABI stable, we can just read it in go.

giuseppe:
fixed the patch to read the flag from the composefs file itself, it is much simpler now

drivers/overlay/composefs_supported.go
```diff
@@ -75,7 +75,13 @@ func enableVerityRecursive(path string) error {
 	return filepath.WalkDir(path, walkFn)
 }
 
-func generateComposeFsBlob(toc []byte, destFile string) error {
+func getErofsBlob(dataDir string) string {
+	return filepath.Join(dataDir, "erofs.blob")
```
alexlarsson, line 81: Maybe we can name it something with composefs or cfs, as it is not any old erofs image.

alexlarsson: Hmm, i guess that is in the directory name

drivers/overlay/composefs_supported.go
```diff
@@ -119,7 +125,13 @@ func generateComposeFsBlob(toc []byte, destFile string) error {
 	return nil
 }
 
-func mountErofsBlob(blobFile, mountPoint string) error {
+func mountComposefsBlob(dataDir, mountPoint string) error {
+	data, err := getComposefsAdditionalData(dataDir)
```
alexlarsson: What is this needed for?

alexlarsson: Ah, its removed in a later commit. Maybe should drop it?

giuseppe: yes, cleaned up the patches

drivers/overlay/composefs_supported.go
giuseppe: renamed to `composefs.blob`

alexlarsson (approved):

rhatdan:
LGTM

drivers/overlay/overlay.go
```diff
@@ -1518,42 +1518,42 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
...
+		for _, m := range composefsMounts {
 			defer unix.Unmount(m, unix.MNT_DETACH)
 		}
 	}()
 
-	maybeAddErofsMount := func(lowerID string, i int) (string, error) {
-		erofsBlob := d.getErofsBlob(lowerID)
-		_, err = os.Stat(erofsBlob)
+	maybeAddComposefsMount := func(lowerID string, i int) (string, error) {
+		composefsBlob := d.getComposefsData(lowerID)
```
mtrmac, line 1538: `s/composefsBlob/composefsDataDir/g`?

drivers/overlay/overlay.go
```diff
@@ -1518,42 +1518,42 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
...
+		_, err = os.Stat(composefsBlob)
 		if err != nil {
 			if os.IsNotExist(err) {
 				return "", nil
 			}
 			return "", err
 		}
-		logrus.Debugf("overlay: using erofs blob %s for lower %s", erofsBlob, lowerID)
+		logrus.Debugf("overlay: using composefs blob %s for lower %s", composefsBlob, lowerID)
```
mtrmac, line 1546: … and something like `s/blob/subdirectory/` here.

mtrmac (commented):
(A very late look I’m afraid)

PR #1664, opened 2023-07-06
pkg/idmap: use unix package for new mount syscalls

Luap99:
I noticed this will looking at the new mount api in podman. This is just a small cleanup to use the unix functions instead of calling the raw syscall directly.


Luap99:
@giuseppe PTAL

rhatdan:
LGTM

giuseppe (approved):
LGTM

PR #1665, opened 2023-07-06
pkg/archive: Ignore file mounts when checking for mount points

dfr:
If we return SkipDir for file mounts, this can generate a diff that deletes all files after the mount point.

### DIFF
--- a/pkg/archive/changes_other.go
+++ b/pkg/archive/changes_other.go
@@ -92,7 +92,10 @@ func collectFileInfo(sourceDir string, idMappings *idtools.IDMappings) (*FileInf
 			return err
 		}
 
-		if s.Dev() != sourceStat.Dev() {
+		// Don't cross mount points. This ignores file mounts to avoid
+		// generating a diff which deletes all files following the
+		// mount.
+		if s.Dev() != sourceStat.Dev() && s.IsDir() {
 			return filepath.SkipDir
 		}
 
--- a/pkg/system/stat_unix.go
+++ b/pkg/system/stat_unix.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"strconv"
 	"syscall"
+
+	"golang.org/x/sys/unix"
 )
 
 // StatT type contains status of a file. It contains metadata
@@ -55,6 +57,10 @@ func (s StatT) Mtim() syscall.Timespec {
 // Dev returns a unique identifier for owning filesystem
 func (s StatT) Dev() uint64 {
 	return s.dev
+}
+
+func (s StatT) IsDir() bool {
+	return (s.mode & unix.S_IFDIR) != 0
 }
 
 // Stat takes a path to a file and returns
--- a/pkg/system/stat_windows.go
+++ b/pkg/system/stat_windows.go
@@ -48,6 +48,10 @@ func (s StatT) Dev() uint64 {
 	return 0
 }
 
+func (s StatT) IsDir() bool {
+	return s.Mode().IsDir()
+}
+
 // Stat takes a path to a file and returns
 // a system.StatT type pertaining to that file.
 //


dfr:
Fixed the windows build break by adding an IsDir method to system.StatT

flouthoc (approved):
LGTM

rhatdan:
@giuseppe @vrothberg @nalind @saschagrunert @mtrmac PTAL

nalind:
LGTM: we want the file tree walk that's calling this function to not skip the rest of the current directory if the item being considered is not a directory.  Might need a rebase.

dfr:
> LGTM: we want the file tree walk that's calling this function to not skip the rest of the current directory if the item being considered is not a directory. Might need a rebase.

Exactly - when I in put the check for crossing mount points, I didn't take file mounts into account probably because file mount support hadn't landed in FreeBSD quite yet. Now that file mounts are supported, this extra check is needed to avoid bad diffs for running containers which tend to have mangled /etc due to the /etc/hosts and/or /etc/resolv.conf mounts.

pkg/archive/changes_other.go
```diff
@@ -92,7 +92,10 @@ func collectFileInfo(sourceDir string, idMappings *idtools.IDMappings) (*FileInf
 			return err
 		}
 
-		if s.Dev() != sourceStat.Dev() {
+		// Don't cross mount points. This ignores file mounts to avoid
+		// generating a diff which deletes all files following the
+		// mount.
+		if s.Dev() != sourceStat.Dev() && s.IsDir() {
 			return filepath.SkipDir
 		}
```
mtrmac, line 100: This will now not just avoid the `SkipDir`, but it will actually add the file mount’s `info` to `parent.children`.

Is that intentional? Is that desirable?

dfr: Originally, I thought that would be a problem but in practice, the mount points don't end up in the diff:
```
# podman run -d --rm quay.io/dougrabson/freebsd-minimal:13.2 sleep 3600
15b273454dbc84e06e40cf4633b650351eb58690af797b737d37d4f85352b791
# podman exec -l df
Filesystem                                                                                                                         512-blocks     Used     Avail Capacity  Mounted on
zboot/containers/f21426f565549a5bbde547a3989d153f805e02902969a4ef052aab48034db6bb                                                   479877240    70992 479806248     0%    /
devfs                                                                                                                                       2        0         2     0%    /dev
fdescfs                                                                                                                                     2        0         2     0%    /dev/fd
/var/run/containers/storage/zfs-containers/15b273454dbc84e06e40cf4633b650351eb58690af797b737d37d4f85352b791/userdata/resolv.conf    522679256 42873008 479806248     8%    /etc/resolv.conf
/var/run/containers/storage/zfs-containers/15b273454dbc84e06e40cf4633b650351eb58690af797b737d37d4f85352b791/userdata/hosts          522679256 42873008 479806248     8%    /etc/hosts
/var/run/containers/storage/zfs-containers/15b273454dbc84e06e40cf4633b650351eb58690af797b737d37d4f85352b791/userdata/.containerenv  522679256 42873008 479806248     8%    /run/.containerenv
# podman diff -l
C /etc
```
This is similar to what I see when I run an alpine image on fedora:
```
# podman run -d --rm alpine sleep 3600
1d63e2e6edd5b21d01aefb0014f75eca2a4c9941682617bec52192e7405a5dff
# podman diff -l
C /etc
```

dfr: On the other hand, if I commit the running image and examine it with skopeo, the new layer contains the container versions of /etc/hosts and /etc/resolv.conf. When I do the same on linux, the new layer contains empty files for /etc/resolv.conf and /run/.containerenv which are the placeholders in the container which get mounted on.

I'm not sure if it will be possible to avoid this for diff and commit of a running container since on FreeBSD the host and container share the global mount namespace. I would quite like to extend FreeBSD jails to support something similar to mount namespaces but that's not going to happen any time soon. In the meantime, this proposed change is at least 'less bad' than the current behaviour.

dfr: Just one more follow up on this thread. I did figure out a way to avoid this problem on FreeBSD and allow 'podman diff' to ignore all the file mounts. Currently, the mount from storage is used directly as the container root - if instead we use nullfs to make a new mountpoint on top of that (this is not an overlay - nullfs behaves like Linux bind), then the container mounts can be made on the nullfs alias and diff can see the underlying storage without the mounts.

I put together a prototype of this and it does work, at the cost of an extra mount. Completing this would need to take into account containers started by an old podman without the extra mount being cleaned up by a newer podman which does support that feature.

PR #1676, opened 2023-08-03
Replace drivers/quota.Quota.quotas with a sync.Map

nalind:
Change `drivers/quota.Quota.quotas` from being a `map[string]uint32` to a `*sync.Map`, since people may be accessing it via the the store's `GraphDriver()`, which provides no locking of its own which would prevent concurrent accesses.


vrothberg (approved):
LGTM
@flouthoc @rhatdan PTAL

flouthoc (approved):
LGTM

PR #1680, opened 2023-08-08
chunked: move GetTOCDigest to a subpackage

giuseppe:
The purpose of this move is to reduce dependencies for the new package.

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -164,26 +164,6 @@ func copyFileContent(srcFd int, destFile string, dirfd int, mode os.FileMode, us
 	return dstFile, st.Size(), nil
 }
 
-// GetTOCDigest returns the digest of the TOC as recorded in the annotations.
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

rhatdan:
@mtrmac @flouthoc PTAL

pkg/chunked/storage_linux.go
```diff
@@ -164,26 +164,6 @@ func copyFileContent(srcFd int, destFile string, dirfd int, mode os.FileMode, us
 	return dstFile, st.Size(), nil
 }
 
-// GetTOCDigest returns the digest of the TOC as recorded in the annotations.
-// This is an experimental feature and may be changed/removed in the future.
-func GetTOCDigest(annotations map[string]string) (*digest.Digest, error) {
```
mtrmac, line 169: This is an API break. c/storage isn’t quite as strict about that as c/image, but still, a forwarding stub here would not be hard to add.

PR #1681, opened 2023-08-09
btrfs: remove an unnecessary local variable

nalind:
We don't need to construct a path ourselves when we already have a method that does it.


flouthoc (approved):
LGTM

PR #1687, opened 2023-08-14
overlay: Put should ignore `ENINVAL` for Unmount

flouthoc:
It seems there are cases where underlying layers can be already unmounted in such case the layer directory is not removed and put fails early with `EINVAL` causing dangling directory for these layers.

Following case may be a hidden underlying issue with `unmounting` of `c/stoage` which needs more investigation however prior discussion shows that this is a long standing issue in `storage` and which was never caught till https://github.com/containers/storage/pull/1607


drivers/overlay/overlay.go
```diff
@@ -1879,7 +1879,9 @@ func (d *Driver) Put(id string) error {
 	if !unmounted {
 		if err := unix.Unmount(mountpoint, unix.MNT_DETACH); err != nil && !os.IsNotExist(err) {
 			logrus.Debugf("Failed to unmount %s overlay: %s - %v", id, mountpoint, err)
-			return fmt.Errorf("unmounting %q: %w", mountpoint, err)
+			if err.Error() != "invalid argument" {
```
giuseppe, line 1882: could we use `Errors.Is` and check it is unix.EINVAL?

giuseppe: probably the `errors.Is` is not even required

vrothberg, line 1882: Could we use `syscall.EINVAL` instead?

vrothberg (approved):
LGTM

PR #1688, opened 2023-08-14
overlay: Diffsize: avoid naive diff

vrothberg:
Computing the diff size for the rootful overlay storage driver used the naive diff.  The reasoning was that we have made use of rootful copy-up. The downside is a mind-blowing performance penalty in comparison to the rootless case.

Hence, drop the naive diff and only compute the size of the directory which is absolutely sufficient for the motivating use case of podman-system-df.  This drops the execution of system-df from 19.3s to 0.11s listing 5 containers and 1 image.

Fixes: github.com/containers/podman/issues/19467

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -2162,10 +2162,6 @@ func (d *Driver) getLowerDiffPaths(id string) ([]string, error) {
 // and its parent and returns the size in bytes of the changes
 // relative to its base filesystem directory.
 func (d *Driver) DiffSize(id string, idMappings *idtools.IDMappings, parent string, parentMappings *idtools.IDMappings, mountLabel string) (size int64, err error) {
-	if d.options.mountProgram == "" && (d.useNaiveDiff() || !d.isParent(id, parent)) {
-		return d.naiveDiff.DiffSize(id, idMappings, parent, parentMappings, mountLabel)
-	}
-
 	p, err := d.getDiffPath(id)
 	if err != nil {
 		return 0, err


vrothberg:
@giuseppe @rhatdan PTAL

flouthoc (approved):
LGTM

rhatdan:
@giuseppe @nalind I merged, but would like to have you comment on this

giuseppe:
that is the patch I've suggested in https://github.com/containers/podman/issues/19467#issuecomment-1677009528 so I am fine with this approach

drivers/overlay/overlay.go
```diff
@@ -2162,10 +2162,6 @@ func (d *Driver) getLowerDiffPaths(id string) ([]string, error) {
 // and its parent and returns the size in bytes of the changes
 // relative to its base filesystem directory.
 func (d *Driver) DiffSize(id string, idMappings *idtools.IDMappings, parent string, parentMappings *idtools.IDMappings, mountLabel string) (size int64, err error) {
-	if d.options.mountProgram == "" && (d.useNaiveDiff() || !d.isParent(id, parent)) {
```
mtrmac, line 2165: A random drive-by: Is it really correct to drop the special handling of `!d.isParent`?

vrothberg: Thanks!

I think it doesn't matter for the current callers but we probably should do a naive diff if it's not the parent.
@giuseppe WDYT?

giuseppe: yes, I also think we should do the naive diff in this case.

Thanks @mtrmac for spotting it

PR #1690, opened 2023-08-17
overlay.Status(): report volatile/shifting support

nalind:
When we're asked for our status, also let the caller know if we support the volatile flag and use of ID-mapped mounts.


rhatdan:
LGTM
@giuseppe @vrothberg @saschagrunert @flouthoc @mtrmac PTAL

mtrmac (commented):
LGTM.

PR #1691, opened 2023-08-21
overlay: Diffsize: naive diff when not parent

vrothberg:
If the specified parent is not ID's parent, use the naive diff.

@mtrmac @giuseppe PTAL

### DIFF
--- a/contrib/cirrus/build_and_test.sh
+++ b/contrib/cirrus/build_and_test.sh
@@ -71,7 +71,10 @@ case $TEST_DRIVER in
         zfs create $zpool/tmp
         TMPDIR="/$zpool/tmp" showrun make STORAGE_DRIVER=$TEST_DRIVER local-test-integration local-test-unit
         # Ensure no datasets are held open prior to `zfs destroy` trap.
-        kill $(lsns -J -t mnt --output-all | jq '.namespaces[]|select(.command=="sleep 1000s").pid')
+        datasets=$(lsns -J -t mnt --output-all | jq '.namespaces[]|select(.command=="sleep 1000s").pid')
+        if [[ -n "$datasets" ]]; then
+            kill $datasets
+        fi
         ;;
     *)
         die "Unknown/Unsupported \$TEST_DRIVER=$TEST_DRIVER (see .cirrus.yml and $(basename $0))"
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -2170,6 +2170,10 @@ func (d *Driver) getLowerDiffPaths(id string) ([]string, error) {
 // and its parent and returns the size in bytes of the changes
 // relative to its base filesystem directory.
 func (d *Driver) DiffSize(id string, idMappings *idtools.IDMappings, parent string, parentMappings *idtools.IDMappings, mountLabel string) (size int64, err error) {
+	if !d.isParent(id, parent) {
+		return d.naiveDiff.DiffSize(id, idMappings, parent, parentMappings, mountLabel)
+	}
+
 	p, err := d.getDiffPath(id)
 	if err != nil {
 		return 0, err


vrothberg:
This reintroduces the perf issue.

vrothberg:
> This reintroduces the perf issue.

False alarm.  I used the wrong branch for testing -.-

rhatdan:
LGTM

mtrmac (approved):
LGTM.

vrothberg:
Does anybody know what's up the Debian job?

> [+2491s] kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
[10:42:58] END - [+2492s] total duration since START

@cevich @giuseppe ?

giuseppe (approved):
LGTM

vrothberg:
Thanks, @edsantiago !

contrib/cirrus/build_and_test.sh
```diff
@@ -71,7 +71,10 @@ case $TEST_DRIVER in
         zfs create $zpool/tmp
         TMPDIR="/$zpool/tmp" showrun make STORAGE_DRIVER=$TEST_DRIVER local-test-integration local-test-unit
         # Ensure no datasets are held open prior to `zfs destroy` trap.
-        kill $(lsns -J -t mnt --output-all | jq '.namespaces[]|select(.command=="sleep 1000s").pid')
+        datasets=$(lsns -J -t mnt --output-all | jq '.namespaces[]|select(.command=="sleep 1000s").pid')
```
mtrmac, line 74: “dataset” seems to be a ZFS term, something that is _held_ by a process but is not a process. `pids`?

(I know ~nothing about ZFS and whether this is the right thing to do, or why it suddenly started breaking after ~4 months.)

edsantiago: @cevich could you weigh in here please? The solution is obvious and already applied, but it would be really nice to give this code a meaningful comment.

cevich: ZFS testing has been a PITA (mostly for me) in this repo. for a long while.  However, apparently zfs-testing is actually needed due to a large user-base (I think it's Giuseppe who strongly desires it).  I didn't write these script bits, but I would guess they came in when Ubuntu was used in CI.  I'm not surprised it's breaking under the much more modern Debian SID.  For Ubuntu it was easy/built-in to the distro, but in Debian it's a kind of add-on similar to EPEL (near as I can tell).  That part alone smells like trouble to me :disappointed:

giuseppe: in any case, ZFS failures for sure do not depend on this change as it affects only the overlay driver

edsantiago: @giuseppe @cevich I interpret these comments as "get rid of that ZFS dataset special case"; but it could also be interpreted as "get rid of all ZFS special cases". I would like to open a cleanup PR to do so. Please clarify which of those I should do (only 3 lines, or *all* zfs lines). Thank you.

cevich: > I've no idea about ZFS, nor desire to keep it :-)

Well darn my bad memory...I'm sure somebody said (strongly) that we need it in CI. 
That was in response to my wanting to trash all ZFS testing.  This is the _only_ repo. that does it.  That's really the only high-level context I can add here :disappointed:

@edsantiago I really don't know/remember anything about these `build_and_test.sh` lines, I don't think I wrote them(?).  I welcome your efforts toward making it cleaner, and if I can help somehow I will.  I was just questioning if we really do need this at all, since it's so much of a corner-case in the overall/multi-project CI matrix.  I think the answer is still "yes"...just wish I remembered who the strong advocate is :disappointed:

giuseppe: If anyone in the community cares enough to open PRs to maintain these other drivers, then it is fine but I don't see the point for us to worry about anything else than overlay.  I am not even talking about RHEL support, even upstream the other backends miss too many features that are present in overlay.

So +1 on "get rid of that "get rid of all ZFS special cases" sounds right if that can simplify our life.

PR #1693, opened 2023-08-22
Bump to v1.49.0 

rhatdan:


vrothberg:
Can elaborate on why we need a new minor release?

vrothberg:
... and why you force merged it without review?

rhatdan:
There only integers we have plenty to go around. Previous update is causing Vendoring to go haywire.  This update will fix that and prevent future updates of older versions from hitting other repos incorrectly.

vrothberg:
So we do releases now to work around vendor bots and all that without reviews?  I am really no fan :(

rhatdan:
We did not release for two months and builtup  a lot of fixes, and bumping takes 10 seconds and affects very little to work around bots issues.

vrothberg:
We just did a patch release (1.48.1) today.  1.49.0 will not be used anywhere and is missing patches (https://github.com/containers/storage/pull/1691).  I assume the sole purpose was to silence the vendor bots but I hope there's a better solution than cutting ghost releases.

mtrmac:
> I assume the sole purpose was to silence the vendor bots but I hope there's a better solution than cutting ghost releases.

Merge the release branch into the main one, as c/image is now regularly doing.

That would also surface the code missing on the main branch, and bring it there.

PR #1712, opened 2023-09-13
overlay, composefs: mount loop device RO

giuseppe:
otherwise it fails when mounting a file that was protected with fs-verity.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #1713, opened 2023-09-13
overlay, composefs: use data-only lower layers

giuseppe:
use the new overlay data-only feature to mount the composefs data directory so there is no need for upper layers to create whiteouts to hide payload files.

The feature was added to Linux 6.5.

### DIFF
--- a/cmd/containers-storage/diff.go
+++ b/cmd/containers-storage/diff.go
@@ -1,12 +1,17 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"os"
 
 	"github.com/containers/storage"
+	graphdriver "github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/chunked"
+	"github.com/containers/storage/pkg/chunked/compressor"
+	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/mflag"
 )
 
@@ -96,6 +101,93 @@ func diff(flags *mflag.FlagSet, action string, m storage.Store, args []string) (
 	return 0, nil
 }
 
+type fileFetcher struct {
+	file *os.File
+}
+
+func sendFileParts(f *fileFetcher, chunks []chunked.ImageSourceChunk, streams chan io.ReadCloser, errors chan error) {
+	defer close(streams)
+	defer close(errors)
+
+	for _, chunk := range chunks {
+		l := io.NewSectionReader(f.file, int64(chunk.Offset), int64(chunk.Length))
+		streams <- ioutils.NewReadCloserWrapper(l, func() error {
+			return nil
+		})
+	}
+}
+
+func (f fileFetcher) GetBlobAt(chunks []chunked.ImageSourceChunk) (chan io.ReadCloser, chan error, error) {
+	streams := make(chan io.ReadCloser)
+	errs := make(chan error)
+	go sendFileParts(&f, chunks, streams, errs)
+	return streams, errs, nil
+}
+
+func applyDiffUsingStagingDirectory(flags *mflag.FlagSet, action string, m storage.Store, args []string) (int, error) {
+	if len(args) < 2 {
+		return 2, nil
+	}
+
+	layer := args[0]
+	sourceDirectory := args[1]
+
+	tOptions := archive.TarOptions{}
+	tr, err := archive.TarWithOptions(sourceDirectory, &tOptions)
+	if err != nil {
+		return 1, err
+	}
+	defer tr.Close()
+
+	tar, err := os.CreateTemp("", "layer-diff-tar-")
+	if err != nil {
+		return 1, err
+	}
+	defer os.Remove(tar.Name())
+	defer tar.Close()
+
+	// we go through the zstd:chunked compressor first so that it generates the metadata required to mount
+	// a composefs image.
+
+	metadata := make(map[string]string)
+	compressor, err := compressor.ZstdCompressor(tar, metadata, nil)
+	if err != nil {
+		return 1, err
+	}
+
+	if _, err := io.Copy(compressor, tr); err != nil {
+		return 1, err
+	}
+	if err := compressor.Close(); err != nil {
+		return 1, err
+	}
+
+	size, err := tar.Seek(0, io.SeekCurrent)
+	if err != nil {
+		return 1, err
+	}
+
+	fetcher := fileFetcher{
+		file: tar,
+	}
+
+	differ, err := chunked.GetDiffer(context.Background(), m, size, metadata, &fetcher)
+	if err != nil {
+		return 1, err
+	}
+
+	var options graphdriver.ApplyDiffOpts
+	out, err := m.ApplyDiffWithDiffer("", &options, differ)
+	if err != nil {
+		return 1, err
+	}
+	if err := m.ApplyDiffFromStagingDirectory(layer, out.Target, out, &options); err != nil {
+		m.CleanupStagingDirectory(out.Target)
+		return 1, err
+	}
+	return 0, nil
+}
+
 func applyDiff(flags *mflag.FlagSet, action string, m storage.Store, args []string) (int, error) {
 	if len(args) < 1 {
 		return 1, nil
@@ -178,5 +270,13 @@ func init() {
 		addFlags: func(flags *mflag.FlagSet, cmd *command) {
 			flags.StringVar(&applyDiffFile, []string{"-file", "f"}, "", "Read from file instead of stdin")
 		},
+	})
+	commands = append(commands, command{
+		names:       []string{"applydiff-using-staging-dir"},
+		optionsHelp: "layerNameOrID directory",
+		usage:       "Apply a diff to a layer using a staging directory",
+		minArgs:     2,
+		maxArgs:     2,
+		action:      applyDiffUsingStagingDirectory,
 	})
 }
--- a/docs/containers-storage-applydiff-using-staging-dir.md
+++ b/docs/containers-storage-applydiff-using-staging-dir.md
@@ -1,0 +1,27 @@
+## containers-storage-applydiff-using-staging-dir 1 "September 2023"
+
+## NAME
+containers-storage applydiff-using-staging-dir - Apply a layer diff to a layer using a staging directory
+
+## SYNOPSIS
+**containers-storage** **applydiff-using-staging-dir** *layerNameOrID* *source*
+
+## DESCRIPTION
+When a layer is first created, it contains no changes relative to its parent
+layer.  The layer can either be mounted read-write and its contents modified
+directly, or contents can be added (or removed) by applying a layer diff.  A
+layer diff takes the form of a (possibly compressed) tar archive with
+additional information present in its headers, and can be produced by running
+*containers-storage diff* or an equivalent.
+
+Differently than **apply-diff**, the command **applydiff-using-staging-dir**
+first creates a staging directory and then moves the final result to the destination.
+
+## EXAMPLE
+**containers-storage applydiff-using-staging-dir 5891b5b522 /path/to/diff**
+
+## SEE ALSO
+containers-storage-apply-diff(1)
+containers-storage-changes(1)
+containers-storage-diff(1)
+containers-storage-diffsize(1)
--- a/drivers/overlay/check.go
+++ b/drivers/overlay/check.go
@@ -275,3 +275,36 @@ func supportsIdmappedLowerLayers(home string) (bool, error) {
 	}()
 	return true, nil
 }
+
+// supportsDataOnlyLayers checks if the kernel supports mounting a overlay file system
+// that uses data-only layers.
+func supportsDataOnlyLayers(home string) (bool, error) {
+	layerDir, err := os.MkdirTemp(home, "compat")
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		_ = os.RemoveAll(layerDir)
+	}()
+
+	mergedDir := filepath.Join(layerDir, "merged")
+	lowerDir := filepath.Join(layerDir, "lower")
+	lowerDirDataOnly := filepath.Join(layerDir, "lower-data")
+	upperDir := filepath.Join(layerDir, "upper")
+	workDir := filepath.Join(layerDir, "work")
+
+	_ = idtools.MkdirAs(mergedDir, 0o700, 0, 0)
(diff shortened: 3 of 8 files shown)


drivers/overlay/overlay.go
```diff
@@ -1750,11 +1750,17 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		absLowers = newAbsDir
 	}
 
+	lowerDirs := strings.Join(absLowers, ":")
+	if len(composeFsLayers) > 0 {
+		composeFsLayersLowerDirs := strings.Join(composeFsLayers, "::")
+		lowerDirs = lowerDirs + "::" + composeFsLayersLowerDirs
+	}
+
```
nalind, line 1758: If I'm reading this right, `lowerDirs` and `absLowers` no longer refer to the same set of locations, which is a problem later if we find ourselves using `absLowers`.

nalind:
How is whether or not we can use a data-only lower layer determined?  Is there a run-time feature test for this?

drivers/overlay/overlay.go
giuseppe: yes after this point absLowers cannot be used anymore as it misses the data only directories.

Do you suggest setting it to `nil` so we will catch in future any error if we attempt to use it or any other suggestions?

giuseppe:
> How is whether or not we can use a data-only lower layer determined? Is there a run-time feature test for this?

I've added a run-time feature test.  Without the data-only lower layers feature we are limited to mount only images with a single layer when using composefs.  I don't think it is worth the extra complexity, so composefs is disabled when the feature is not present.

@alexlarsson PTAL

giuseppe:
we also need: https://github.com/containers/storage/pull/1712

drivers/overlay/overlay.go
rhatdan: I think it would make sense to nil it out.

giuseppe: done, I am setting it to `nil` after usage

drivers/overlay/overlay.go
```diff
@@ -1750,11 +1750,20 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
...
 	}
 
+	lowerDirs := strings.Join(absLowers, ":")
+	if len(composeFsLayers) > 0 {
+		composeFsLayersLowerDirs := strings.Join(composeFsLayers, "::")
+		lowerDirs = lowerDirs + "::" + composeFsLayersLowerDirs
+	}
+	// absLowers is not valid anymore now as we have added composeFsLayers to it, so prevent
+	// its usage.
+	absLowers = nil
```
flouthoc, line 1760: linter says: ineffectual assignment about this

giuseppe: fixed

giuseppe:
ready for review

rhatdan:
LGTM
@nalind @mtrmac @alexlarsson @flouthoc PTAL

nalind (changes requested):
I'd really like a test that ensures that the driver does the right thing when the feature is available and asked for, even if it has to be skipped in CI in the short-term.

giuseppe:
> I'd really like a test that ensures that the driver does the right thing when the feature is available and asked for, even if it has to be skipped in CI in the short-term.

Sure, I can add some tests.  Do you think there is anything in particular we should test for?  Once composefs is enabled, there should not be any visible difference for the users and the existing tests should still work

nalind:
> > I'd really like a test that ensures that the driver does the right thing when the feature is available and asked for, even if it has to be skipped in CI in the short-term.
> 
> Sure, I can add some tests. Do you think there is anything in particular we should test for? Once composefs is enabled, there should not be any visible difference for the users and the existing tests should still work

Do we have tests that exercise using composefs layers as read-only layers?

giuseppe:
> > > I'd really like a test that ensures that the driver does the right thing when the feature is available and asked for, even if it has to be skipped in CI in the short-term.
> > 
> > 
> > Sure, I can add some tests. Do you think there is anything in particular we should test for? Once composefs is enabled, there should not be any visible difference for the users and the existing tests should still work
> 
> Do we have tests that exercise using composefs layers as read-only layers?

I don't think it would be any different than the tests we already have, the only difference with composefs is that there will be more 2x lower layers (for each layer there is one overlay layer with the erofs file system and one for the data)

giuseppe:
I've added a new test to exercise `ApplyDiffFromStagingDirectory` that is needed for composefs.

I've tested it manually with composefs on Rawhide

drivers/overlay/mount.go
```diff
@@ -141,14 +141,27 @@ func mountOverlayFromMain() {
 	// the new value for the list of lowers, because it's shorter.
 	if lowerv != "" {
 		lowers := strings.Split(lowerv, ":")
+		var newLowers []string
+		dataOnly := false
 		for i := range lowers {
```
mtrmac, line 146: (Absolutely non-blocking: `for _, lowerPath := range lowers` (or even `range strings.Split…)` might cut down on the visual noise a bit.)

drivers/overlay/mount.go
```diff
@@ -141,14 +141,27 @@ func mountOverlayFromMain() {
...
+			if lowers[i] == "" {
```
mtrmac, line 147: Non-blocking: If I’m following `composeFsLayersLowerDirs` construction correctly, this
- never happens twice in a row
- never happens for the last element.

If so,
```go
for i, oldLower := range lowers {
   if oldLower != "" {
      // … unix.Open …
      lowers[i] = fmt.Sprintf(…)
  }
}
```
or
```go
for _, lower := range lowers {
  if lower != "" {
      // … unix.Open …
      lower = fmt.Sprintf(…)
  }
  newLowers = append(newLowers, lower)
}
```

would be easier for me to follow.

The way I read the current `dataOnly` logic is “if this more complex logic than a trivial `map` is needed, it probably exists to handle consecutive, or trailing, empty entries”… but AFAICS that isn’t possible. Am I misreading the code?

Warning: I didn’t take the time to understand what the syntax / semantics of the value actually is…

mtrmac (commented):
ACK “I saw nothing obviously suspicious” — not really a LGTM this because I don’t understand the > 400-line `overlay.Driver.Get` and I can’t now take the time to learn (and, certainly, splitting that up is not what this PR should deal with).

I’ll just note that this breaks CI, so can’t be merged as is.

drivers/overlay/mount.go
giuseppe: overlay now supports something like: `a:b::c::d` each directory that has `::` is considered a data-only layer (and at the moment regular layers cannot follow data-only paths but this could be relaxed in future), so `::` can happen multiple times.

mtrmac: Oh, so it’s all my ignorance. Sorry about that.

mtrmac: … and I was imprecise. “multiple times” is not a problem.

I was reading the logic in this PR to turn `a::::b:` into `fd1:fd2` (collapsing a _sequence_ of `:::` or more into a single `::`, and dropping the trailing one).

If that’s not relevant/expected, and it’s just a special `::`_something_ syntax, where preserving the original number/position of `:` chars is the right thing to do, then, yes, I think one of the two proposed loop alternatives (which don’t carry a value across iterations) would be a bit easier to follow.

Again, this is all non-blocking either way.

cmd/containers-storage/diff.go
```diff
@@ -1,12 +1,19 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
```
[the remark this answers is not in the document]
giuseppe: dropped now

cmd/containers-storage/diff.go
```diff
@@ -96,6 +103,101 @@ func diff(flags *mflag.FlagSet, action string, m storage.Store, args []string) (
...
+	defer close(errors)
+
+	for _, chunk := range chunks {
+		f.mutex.Lock()
+		_, err := f.file.Seek(int64(chunk.Offset), 0)
+		if err != nil {
+			errors <- err
+			continue
+		}
+		l := io.LimitReader(f.file, int64(chunk.Length))
```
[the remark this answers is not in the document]
giuseppe: thanks, that is better

drivers/overlay/mount.go
giuseppe: fixed

giuseppe:
took care of the comments, and the CI is green now

rhatdan:
@mtrmac @nalind PTAL

docs/containers-storage-applydiff-using-staging-dir.md
```diff
@@ -0,0 +1,27 @@
...
+directly, or contents can be added (or removed) by applying a layer diff.  A
+layer diff takes the form of a (possibly compressed) tar archive with
+additional information present in its headers, and can be produced by running
+*containers-storage diff* or an equivalent.
+
+Differently than **apply-diff**, the command **applydiff-using-staging-dir**
+first creates a staging directory and then moves the final result to the destination.
+
+## EXAMPLE
+**containers-storage applydiff-using-staging-dir 71841c97e320d6cde.tar.gz SOURCE**
```
nalind, line 21: This example doesn't agree with the help output.

cmd/containers-storage/diff.go
```diff
@@ -96,6 +101,93 @@ func diff(flags *mflag.FlagSet, action string, m storage.Store, args []string) (
...
+	}
+
+	if _, err := io.Copy(compressor, tr); err != nil {
+		return 1, err
+	}
+	if err := compressor.Close(); err != nil {
+		return 1, err
+	}
+
+	size, err := tar.Seek(0, 1)
```
nalind, line 165: Pass either `io.SeekCurrent` or `io.SeekEnd` as the second parameter to `Seek()`.

giuseppe:
can I get another review?

cmd/containers-storage/diff.go
```diff
@@ -96,6 +101,93 @@ func diff(flags *mflag.FlagSet, action string, m storage.Store, args []string) (
...
+	sourceDirectory := args[1]
+
+	tOptions := archive.TarOptions{}
+	tr, err := archive.TarWithOptions(sourceDirectory, &tOptions)
+	if err != nil {
+		return 1, err
+	}
+	defer tr.Close()
+
+	tar, err := os.CreateTemp("", "composefs-")
```
alexlarsson, line 142: I understand this is used for the composefs usecase, but this seems like a weird tempname prefix? layer-diff-tar- maybe?

alexlarsson (approved):
Minor comments, but looks good to me.

drivers/overlay/check.go
```diff
@@ -275,3 +275,36 @@ func supportsIdmappedLowerLayers(home string) (bool, error) {
...
+	upperDir := filepath.Join(layerDir, "upper")
+	workDir := filepath.Join(layerDir, "work")
+
+	_ = idtools.MkdirAs(mergedDir, 0o700, 0, 0)
+	_ = idtools.MkdirAs(lowerDir, 0o700, 0, 0)
+	_ = idtools.MkdirAs(lowerDirDataOnly, 0o700, 0, 0)
+	_ = idtools.MkdirAs(upperDir, 0o700, 0, 0)
+	_ = idtools.MkdirAs(workDir, 0o700, 0, 0)
+
+	opts := fmt.Sprintf("lowerdir=%s::%s,upperdir=%s,workdir=%s,metacopy=on", lowerDir, lowerDirDataOnly, upperDir, workDir)
```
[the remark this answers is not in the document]
giuseppe: but this test ensures the `::` works also in a regular container configuration where there is a writeable layer (even if the are no kernel versions that support that feature only in RO mode).  I can change that though if you think it is not really needed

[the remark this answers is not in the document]
alexlarsson: sure

rhatdan:
@alexlarsson @nalind PTANL

alexlarsson (approved):
lgtm

rhatdan:
@mtrmac @nalind another review?

giuseppe:
rebased, can you please take another look?

giuseppe:
@nalind could you please take another look?

drivers/overlay/mount.go
```diff
@@ -141,14 +141,27 @@ func mountOverlayFromMain() {
 	// the new value for the list of lowers, because it's shorter.
 	if lowerv != "" {
 		lowers := strings.Split(lowerv, ":")
-		for i := range lowers {
-			lowerFd, err := unix.Open(lowers[i], unix.O_RDONLY, 0)
+		var newLowers []string
+		dataOnly := false
+		for _, lowerPath := range lowers {
+			if lowerPath == "" {
+				dataOnly = true
```
nalind, line 148: Should this loop be enforcing the "once we see a data-only layer, they should all be data-only layers" requirement that the kernel has?

tests/apply-diff.bats
```diff
@@ -44,3 +44,61 @@ load helpers
...
+EOF
+
+	# Create a layer.
+	CONTAINERS_STORAGE_CONF=$sconf run storage --debug=false create-layer
+	[ "$status" -eq 0 ]
+	[ "$output" != "" ]
+	layer="$output"
+
+	CONTAINERS_STORAGE_CONF=$sconf run storage --debug=false applydiff-using-staging-dir $layer $SRC
+        cat <<< $output
```
nalind, line 85: Why is this different than the test at line 81?  If $output is empty, doesn't it still succeed?

docs/containers-storage-applydiff-using-staging-dir.md
```diff
@@ -0,0 +1,27 @@
+## containers-storage-applydiff-using-staging-dir 1 "September 2023"
```
nalind, line 1: There should be a reference to this in the SUB-COMMANDS section of "docs/containers-storage.md".

tests/apply-diff.bats
giuseppe: the command doesn't print anything on success.  I'll drop this line as it is not needed.

docs/containers-storage-applydiff-using-staging-dir.md
giuseppe: thanks, added now

drivers/overlay/mount.go
giuseppe: not sure, if the input is not correct we'd get a similar error from `Mount` as we do if we call directly the syscall and we are not doing any validation in the `mountOverlayFromMain` function.

I don't think we need it, but if you prefer to have it, I'll add it

nalind:
LGTM

PR #1714, opened 2023-09-14
Run codespell on code

rhatdan:


types/options.go
```diff
@@ -220,9 +220,9 @@ type StoreOptions struct {
 	// GraphRoot is the filesystem path under which we will store the
 	// contents of layers, images, and containers.
 	GraphRoot string `json:"root,omitempty"`
-	// Image Store is the location of image store which is seperated from the
+	// Image Store is the location of image store which is separated from the
 	// container store. Usually this is not recommended unless users wants
-	// seperate store for image and containers.
+	// separate store for image and containers.
```
edsantiago, line 225: I think this needs more than just spelling fixes? The first correction should IMO also be "separate", not separated. And the second instance is just repetition, not really helpful. Is there a way to phrase that such that it would explain the _why_, not the what?

rhatdan: Ok I took a stab at improving the description of imagestore in all locations.

docs/containers-storage.conf.5.md
```diff
@@ -56,10 +56,9 @@ $ restorecon -R -v /NEWSTORAGEPATH
...
 
 **imagestore**=""
-  image storage path (the default is assumed to be the same as `graphroot`).
-  Path of imagestore different from `graphroot`, by default storage library stores all images in `graphroot` but if `imagestore` is provided it will store newly pulled images in provided `imagestore` but will keep using `graphroot` for everything else. If user is using `overlay` driver then images which were already part of `graphroot` will still be accessible ( Internally storage library will mount `graphroot` as an `additionalImageStore` to allow this behaviour ).
+  image storage path (the default is assumed to be the same as `graphroot`). Path of imagestore different from `graphroot`. By default images in the storage library are stored int the `graphroot`. If `imagestore` is provided newly pulled images will be stored in the `imagestore` location. All other storage continues to be stored in the `graphroot`. When using `overlay` driver images previously stored in the `graphroot` continue to be accessible ( Internally storage library mounts `graphroot` as an `additionalImageStore` to allow this behaviour ).
```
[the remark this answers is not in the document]
nalind: ... stored **int** the...

docs/containers-storage.conf.5.md
```diff
@@ -56,10 +56,9 @@ $ restorecon -R -v /NEWSTORAGEPATH
...
 
 **imagestore**=""
-  image storage path (the default is assumed to be the same as `graphroot`).
-  Path of imagestore different from `graphroot`, by default storage library stores all images in `graphroot` but if `imagestore` is provided it will store newly pulled images in provided `imagestore` but will keep using `graphroot` for everything else. If user is using `overlay` driver then images which were already part of `graphroot` will still be accessible ( Internally storage library will mount `graphroot` as an `additionalImageStore` to allow this behaviour ).
+  image storage path (the default is assumed to be the same as `graphroot`). Path of imagestore different from `graphroot`. By default images in the storage library are stored in the `graphroot`. If `imagestore` is provided, newly pulled images will be stored in the `imagestore` location. All other storage continues to be stored in the `graphroot`. When using `overlay` driver images previously stored in the `graphroot` continue to be accessible ( Internally storage library mounts `graphroot` as an `additionalImageStore` to allow this behaviour ).
```
edsantiago, line 59: I still find this hard to grok. It's missing a couple of commas, and is a little confusing, but it's better than it was, so thank you.

TomSweeneyRedHat, line 59:
```suggestion
 The image storage path (the default is assumed to be the same as `graphroot`). Path of the imagestore, which is different from `graphroot`. By default, images in the storage library are stored in the `graphroot`. If `imagestore` is provided, newly pulled images will be stored in the `imagestore` location. All other storage continues to be stored in the `graphroot`. When using the `overlay` driver, images previously stored in the `graphroot` remain accessible. Internally, the storage library mounts `graphroot` as an `additionalImageStore` to allow this behavior.
```

flouthoc (approved):
LGTM

PR #1723, opened 2023-09-29
pkg/system: reduce retry timeout for EnsureRemoveAll

giuseppe:
reduce the waiting time for attempting again the removal to 10ms instead of the current 100ms.
    
Currently, if we hit the timeout, we add an artificial 100ms delay to the container cleanup, while with this we will react faster when the removal is possible.

### DIFF
--- a/pkg/system/rm.go
+++ b/pkg/system/rm.go
@@ -38,7 +38,7 @@ func EnsureRemoveAll(dir string) error {
 
 	// Attempt to unmount anything beneath this dir first
 	if err := mount.RecursiveUnmount(dir); err != nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
The change looks OK to me. I'm just wondering if, given we're checking more quickly now, should we reduce the retry count to keep from hammering the system.  I was thinking 500 instead of 1000.  IDK, I'm fence-sitting and looking for other opinions.

giuseppe:
I think it is fine as it is so we have more time to recover from temporary failures.  The operations involved are not too expensive so even every 10ms is still a relatively long time

PR #1724, opened 2023-10-04
store: call RecordWrite() before graphDriver Cleanup()

giuseppe:
Move the execution of RecordWrite() before the graphDriver Cleanup(). This addresses a longstanding issue that occurs when the Podman cleanup process is forcely terminated and on some occasions the termination happens after the Cleanup() but before the change is recorded.  This causes that the next user is not notified about the change and will mount the container without the home directory below (the infamous /var/lib/containers/storage/overlay mount). Then when the next time the graphDriver is initialized, the home directory is mounted on top of the existing mounts causing some containers to fail with ENOENT since all files are hidden and some others cannot be cleaned up since their mount directory is covered by the home directory mount.

Closes: https://github.com/containers/podman/issues/18831
Closes: https://github.com/containers/podman/issues/17216
Closes: https://github.com/containers/podman/issues/17042

### DIFF
--- a/store.go
+++ b/store.go
@@ -3407,16 +3407,16 @@ func (s *store) Shutdown(force bool) ([]string, error) {
 		err = fmt.Errorf("a layer is mounted: %w", ErrLayerUsedByContainer)
 	}
 	if err == nil {
-		err = s.graphDriver.Cleanup()
 		// We don’t retain the lastWrite value, and treat this update as if someone else did the .Cleanup(),
 		// so that we reload after a .Shutdown() the same way other processes would.
 		// Shutdown() is basically an error path, so reliability is more important than performance.
 		if _, err2 := s.graphLock.RecordWrite(); err2 != nil {
-			if err == nil {
-				err = err2
-			} else {
-				err = fmt.Errorf("(graphLock.RecordWrite failed: %v) %w", err2, err)
-			}
+			err = fmt.Errorf("(graphLock.RecordWrite failed: %w", err2)
+		}
+		// Do the Cleanup() only after we are sure that the change was recorded with RecordWrite(), so that
+		// the next user picks it.
+		if err == nil {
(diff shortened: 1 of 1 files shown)


giuseppe:
@edsantiago I am quite sure this solves the issue we've seen.

It is marked as Draft, as I am still testing it though to get more confidence,

giuseppe:
@mtrmac @rhatdan @nalind PTAL

rhatdan:
LGTM

giuseppe:
@vrothberg @flouthoc PTAL

flouthoc (approved):
LGTM

vrothberg (commented):
Amazing job tracking this down, @giuseppe !

rhatdan:
Lets open a PR to get this into Podman.

giuseppe:
here it is: https://github.com/containers/podman/pull/20273

store.go
```diff
@@ -3407,16 +3407,16 @@ func (s *store) Shutdown(force bool) ([]string, error) {
...
 		// We don’t retain the lastWrite value, and treat this update as if someone else did the .Cleanup(),
 		// so that we reload after a .Shutdown() the same way other processes would.
 		// Shutdown() is basically an error path, so reliability is more important than performance.
 		if _, err2 := s.graphLock.RecordWrite(); err2 != nil {
-			if err == nil {
-				err = err2
-			} else {
-				err = fmt.Errorf("(graphLock.RecordWrite failed: %v) %w", err2, err)
-			}
+			err = fmt.Errorf("(graphLock.RecordWrite failed: %w", err2)
```
nalind, line 3414: Mismatched parentheses in the format string.

giuseppe: thanks for spotting it, opened a PR: https://github.com/containers/storage/pull/1726

PR #1727, opened 2023-10-12
overlay, composefs: use mkcomposefs to generate composefs

giuseppe:
use mkcomposefs to generate composefs, instead of using the custom composefs-from-json.
    
As part of the integration, retrieve the verity fs measurement after the files are written so that they can be embedded in the composefs image and their payload is validated each time the files are accessed.


giuseppe:
@alexlarsson PTAL

giuseppe:
marked as a draft for now, I'll try to move the fsverity logic into the chunked package so we don't need to carry the digests around

giuseppe:
> marked as a draft for now, I'll try to move the fsverity logic into the chunked package so we don't need to carry the digests around

looked into it, we will still need to propagate the digests anyway since composefs lives in the overlay driver.

Moving the fsverity measurement in the chunked package can be done later.

pkg/chunked/dump/dump.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+const (
+	NOESCAPE_SPACE = 1 << iota
+	ESCAPE_EQUAL
+	ESCAPE_LONE_DASH
+	ESCAPE_STANDARD
+)
+
+func escaped(val string, length int, escape int) string {
```
alexlarsson, line 21: I understand this is a direct port of the C code, but you can really drop the length here, as in golang the string comes with its own length (and you can easily create substrings that also come with their own length).

pkg/chunked/dump/dump.go
```diff
@@ -0,0 +1,219 @@
...
+	noescapeSpace := escape&NOESCAPE_SPACE != 0
+	escapeEqual := escape&ESCAPE_EQUAL != 0
+	escapeLoneDash := escape&ESCAPE_LONE_DASH != 0
+
+	if length < 0 {
+		length = len(val)
+	}
+
+	if escapeLoneDash && length == 1 && val[0] == '-' {
```
alexlarsson, line 30: In golang, this is nicer as `escapeLoneDash && val == "-"`, at least once you drop the length arg.

pkg/chunked/dump/dump.go
```diff
@@ -0,0 +1,219 @@
...
+
+	"github.com/containers/storage/pkg/chunked/internal"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	NOESCAPE_SPACE = 1 << iota
+	ESCAPE_EQUAL
+	ESCAPE_LONE_DASH
+	ESCAPE_STANDARD
```
alexlarsson, line 18: ESCAPE_STANDARD should be "0", not 1<<3.

pkg/chunked/dump/dump.go
```diff
@@ -0,0 +1,219 @@
...
+
+func dumpNode(out io.Writer, links map[string]int, verityDigests map[string]string, entry *internal.FileMetadata) error {
+	path := entry.Name
+	if path == "" {
+		path = "/"
+	} else if path[0] != '/' {
+		path = "/" + path
+	}
+
+	if _, err := fmt.Fprint(out, escaped(path, len(path), ESCAPE_STANDARD)); err != nil {
```
alexlarsson, line 105: Is the writer buffered here? Otherwise all these small writes are gonna perform badly.

alexlarsson: Seems it is a straigh i.Pipe(), which says:

>Reads and Writes on the pipe are matched one to one except when multiple Reads are needed to consume a single Write. That is, each Write to the PipeWriter blocks until it has satisfied one or more Reads from the PipeReader that fully consume the written data. The data is copied directly from the Write to the corresponding Read (or Reads); there is no internal buffering. 

I think adding a bufio.Writer() around the write-pipe here makes a lot of sense.

drivers/overlay/composefs_supported.go
```diff
@@ -71,24 +87,42 @@ func enableVerityRecursive(path string) error {
 		if err := enableVerity(path, int(f.Fd())); err != nil {
 			return err
 		}
+
+		verity, err := measureVerity(path, int(f.Fd()))
```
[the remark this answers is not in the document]
alexlarsson: Ah, i see this is called after succeeding enableVerity. so, strike this.

drivers/overlay/composefs_supported.go
```diff
@@ -109,10 +143,10 @@ func generateComposeFsBlob(toc []byte, composefsDir string) error {
 		// a scope to close outFd before setting fsverity on the read-only fd.
 		defer outFd.Close()
 
-		cmd := exec.Command(writerJson, "--format=erofs", "--out=/proc/self/fd/3", "/proc/self/fd/0")
+		cmd := exec.Command(writerJson, "--from-file", "/proc/self/fd/0", "/proc/self/fd/3")
```
alexlarsson, line 146: "-" works as the input arg too.

giuseppe:
thanks for the review.  I've addressed the comments and pushed a new version

pkg/chunked/dump/dump.go
```diff
@@ -0,0 +1,227 @@
...
+	link := ""
+	if entry.Type == internal.TypeLink {
+		link = "@"
+	}
+
+	rdev := unix.Mkdev(uint32(entry.Devmajor), uint32(entry.Devminor))
+
+	entryTime := entry.ModTime
+	if entryTime == nil {
+		t := time.Now()
```
alexlarsson, line 118: Would it be possible to use some other mtime as a fallback? Like maybe the time created time of the image manifes config? That would make things more reproducible.

alexlarsson: At the very least, pass in a single default mtime used for all files as the fallback.

pkg/chunked/dump/dump.go
```diff
@@ -0,0 +1,227 @@
...
+	if _, err := fmt.Fprintf(out, " %d %s%o %d %d %d %d %d.%d ", entry.Size,
+		link, mode,
+		nlinks, entry.UID, entry.GID, rdev,
+		entryTime.Unix(), entryTime.Nanosecond()); err != nil {
+		return err
+	}
+
+	var payload string
+	if entry.Linkname != "" {
+		payload = entry.Linkname
```
alexlarsson, line 136: Are we sure this is an absolute path?

pkg/chunked/dump/dump.go
giuseppe: hardcoded  to the epoch if not specified

pkg/chunked/dump/dump.go
giuseppe: I've added the same code as for `entry.Name` to ensure it is a full path.

giuseppe:
@alexlarsson I've encountered a small issue (still investigating):

something like the following doesn't seem to preserve the st_nlink:

```
/bin/busybox-hardlink 847 100755 1 0 0 0 1690477944.0 89/f85a8b92bdb5ada51bd8ee0d9d687da1a759a0d80f8d012e7f8cbb3933a3a0 - d4ced2c28e68aa87530507099801f1ab575ee18054847d301bad5a27a4984855
/bin/busybox-hardlink2 0 @100755 1 0 0 0 1690477944.0 /bin/busybox-hardlink - -
/bin/busybox-hardlink3 0 @100755 1 0 0 0 1690477944.0 /bin/busybox-hardlink - -
/bin/busybox-hardlink4 0 @100755 1 0 0 0 1690477944.0 /bin/busybox-hardlink - -
```

```
# bin/podman run --rm quay.io/giuseppe/zstd-chunked:hard-link sh -c 'stat -c "%n -> %h" /bin/busybox-hardlink*' 
/bin/busybox-hardlink -> 1
/bin/busybox-hardlink2 -> 1
/bin/busybox-hardlink3 -> 1
/bin/busybox-hardlink4 -> 1
```

anything obviously wrong?

alexlarsson (approved):
lgtm

rhatdan:
@nalind @mtrmac PTAL

giuseppe:
@flouthoc PTAL

rhatdan:
LGTM

nalind:
_Please_ let the bot handle merges.

cgwalters:
I know this question isn't related to this PR, just reusing it as a handy discussion forum...it looks like right now enabling this logic requires a build tag.  And once the build tag is set, we unconditionally use composefs?

Is there any reason not to switch to having this be an option in `storage.conf` instead, and have the code always be enabled?

giuseppe:
I agree, it is better to drop the build tag.  I've used it when it was just a PoC and that required some custom binary to be present, but now there is no need to hide the feature

PR #1728, opened 2023-10-13
Define better error message for container name conflicts with external storage

cgiradkar:
Updated the error message to suggest user to use --replace option to instruct Podman to replace the existsing external container with a newly created one.

Closes [16759](https://github.com/containers/podman/issues/16759)


cgiradkar:
Making Podman specific changes in Podman repo, hence, closing it here.

PR #1734, opened 2023-10-18
overlay: do not precreate the composefs layers dir

giuseppe:
Let's not pre-create the composefs layers dir every time we mount a layer, but do it only when there is a composefs blob to use.

Closes: https://github.com/containers/storage/issues/1733

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1529,14 +1529,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		defer cleanupFunc()
 	}
 
-	composefsLayers := filepath.Join(workDirBase, "composefs-layers")
-	if err := os.MkdirAll(composefsLayers, 0o700); err != nil {
-		return "", err
-	}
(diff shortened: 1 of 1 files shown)


drivers/overlay/overlay.go
```diff
@@ -1530,9 +1530,6 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	}
 
 	composefsLayers := filepath.Join(workDirBase, "composefs-layers")
```
nalind, line 1532: This definition is now pretty far from the only point where it gets referenced.  Can it be moved?

giuseppe: sure, fixed now

nalind:
LGTM.  Ideally we'd do a `mount --bind --read-only` of the read-only store directory onto itself after it's set up in the "additional-stores" test and unmount it at the end, but I don't see a way to guarantee that it's unmounted if the test fails for any other reason before it hits the end of the test function.
/lgtm
/hold

PR #1739, opened 2023-10-29
archive: fix mode for root dir with ForceMask

giuseppe:
if force_mask is in use, we need to store the root directory permission after we read it from the tar archive.  We were incorrectly reading it from the directory on the filesystem.


vrothberg (approved):
LGTM

flouthoc (approved):
LGTM
/lgtm
/approve

PR #1740, opened 2023-10-30
Consolidate XDG_RUNTIME code

giuseppe:
Closes: https://issues.redhat.com/browse/RUN-1169


giuseppe:
testing the change here: https://github.com/containers/podman/pull/20528

giuseppe:
podman PR passes.

Ready to review

rhatdan:
LGTM
@nalind @flouthoc @vrothberg @saschagrunert @Luap99 @mtrmac PTAL

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +99,98 @@ func stick(f string) error {
...
+	})
+
+	if rootlessConfigHomeDirError != nil {
+		return "", rootlessConfigHomeDirError
+	}
+
+	return rootlessConfigHomeDir, nil
+}
+
+// GetRuntimeDirUser returns the runtime directory
```
nalind, line 139: Is there guidance for someone who's using this package as to when they should use the value returned by this function instead of the one they get from `GetRuntimeDir()`?

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +99,98 @@ func stick(f string) error {
...
 }
+
+var (
+	rootlessConfigHomeDirOnce sync.Once
+	rootlessConfigHomeDir     string
+	rootlessRuntimeDirOnce    sync.Once
+	rootlessRuntimeDir        string
+)
+
+// GetRootlessConfigHomeDir returns the config home directory when running as non root
```
nalind, line 110: Is this supposed to be used as an alternative to `GetConfigHome()`?  When?

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +99,98 @@ func stick(f string) error {
...
+		if cfgHomeDir == "" {
+			home := os.Getenv("HOME")
+			resolvedHome, err := filepath.EvalSymlinks(home)
+			if err != nil {
+				rootlessConfigHomeDirError = fmt.Errorf("cannot resolve %s: %w", home, err)
+				return
+			}
+			tmpDir := filepath.Join(resolvedHome, ".config")
+			st, err := os.Stat(tmpDir)
+			if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() && st.Mode().Perm() >= 0o700 {
```
edsantiago, line 125: The `>=` made my eyes bug out. I understand that `FileMode.Perm` is only the `ugo` portion, but even so this just feels too weird. Is this a common Go idiom? If it isn't, would you mind changing it to `.Perm() & 0o700 == 0o700`?

pkg/homedir/homedir_unix.go
giuseppe: the documentation of GetConfigHome() is very clear on the expectation of that function, and I didn't want to break it.  I am not sure what is the best way to proceed.

Are you okay if we add the extra feature of trying to guess the ConfigHomeDir to the existing `GetConfigHome()` and not add a new function?

pkg/homedir/homedir_unix.go
giuseppe: I can specify it, the main difference is that this function tries to guess the value of `XDG_RUNTIME_DIR` if it is not set

pkg/homedir/homedir_unix.go
giuseppe: that code is coming from https://github.com/containers/podman/commit/f44b05f6fda6ec91bc21088321da739ce26cfe2d

This is a good occasion to fix it, I'll use your version

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +99,104 @@ func stick(f string) error {
...
+
+		uid := strconv.Itoa(rootlessUID)
+		if runtimeDir == "" {
+			tmpDir := filepath.Join("/run", "user", uid)
+			if err := os.MkdirAll(tmpDir, 0o700); err != nil {
+				logrus.Debug(err)
+			}
+			st, err := os.Stat(tmpDir)
+			if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() && (st.Mode().Perm()&0o700 == 0o700) {
+				runtimeDir = tmpDir
```
mtrmac, line 169: The Jira card says “otherwise throw a human-readable error.”

I don’t know what needs adjusting, but at least one of the two does.

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +99,104 @@ func stick(f string) error {
...
+			}
+		}
+		if runtimeDir == "" {
+			tmpDir := filepath.Join(os.TempDir(), fmt.Sprintf("storage-run-%s", uid))
+			if err := os.MkdirAll(tmpDir, 0o700); err != nil {
+				logrus.Debug(err)
+			}
+			st, err := os.Stat(tmpDir)
+			if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() && (st.Mode().Perm()&0o700 == 0o700) {
```
mtrmac, line 178: The Jira card says “otherwise throw a human-readable error.”

I don’t know what needs adjusting, but at least one of the two does.

---

In this case I could see an argument for this being a hard failure because the user is under attack.

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +99,104 @@ func stick(f string) error {
...
+			if home == "" {
+				rootlessRuntimeDirError = errors.New("neither XDG_RUNTIME_DIR nor HOME was set non-empty")
+				return
+			}
+			resolvedHome, err := filepath.EvalSymlinks(home)
+			if err != nil {
+				rootlessRuntimeDirError = fmt.Errorf("cannot resolve %s: %w", home, err)
+				return
+			}
+			runtimeDir = filepath.Join(resolvedHome, "rundir")
```
mtrmac, line 193: `$HOME/rundir`, do we really want that?

pkg/homedir/homedir.go
```diff
@@ -50,3 +36,47 @@ func GetCacheHome() (string, error) {
...
 	return filepath.Join(home, ".cache"), nil
 }
+
+// SetXdgDirs ensures the XDG_RUNTIME_DIR env and XDG_CONFIG_HOME variables are set.
+// containers/image uses XDG_RUNTIME_DIR to locate the auth file, XDG_CONFIG_HOME is
+// use for the containers.conf configuration file.
+func SetXdgDirs(rootless bool, rootlessUID int) error {
+	if !rootless {
+		os.Unsetenv("XDG_RUNTIME_DIR")
+		os.Unsetenv("XDG_CONFIG_HOME")
```
mtrmac, line 46: Yes, the Jira card says “Root should ignore XDG_RUNTIME_DIR and unset it such that other tools (e.g., crun) don't use it either.”. but it provides absolutely no explanation.

Actively interfering with users’ configuration seems unexpected. Please write down in detail what’s the problem and solution, so that the code can be maintained over time.

pkg/homedir/homedir_others.go
```diff
@@ -19,3 +19,16 @@ func GetRuntimeDir() (string, error) {
...
+// GetConfigHome returns $HOME/.config and nil error if XDG_CONFIG_HOME is not set.
+//
+// See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
+func GetConfigHome() (string, error) {
+	return "", errors.New("homedir.GetRootlessConfigHomeDir() is not supported on this system")
+}
+
+// GetRuntimeDirUser is unsupported on non-linux system.
+func GetRuntimeDirUser(rootless bool, rootlessUID int) (string, error) {
+	return "", errors.New("homedir.GetRuntimeDirUser() is not supported on this system")
```
mtrmac, line 33: Similarly here, is there any guidance for callers on other platforms?

mtrmac (commented):
I appreciate this is mostly “just” moving code from Podman, but that is an occasion to revisit. Also, I think the introduction or 1/2/3 project boundaries between this subpackage and the original Podman callers mean that the API semantics and contracts must be _much_ better documented, because with the ~stable API of c/storage (and presumably new non-Podman callers?), future maintainers of this subpackage need to now such things and “just see what Podman does” will be much less tenable.

giuseppe:
> I appreciate this is mostly “just” moving code from Podman, but that is an occasion to revisit. Also, I think the introduction or 1/2/3 project boundaries between this subpackage and the original Podman callers mean that the API semantics and contracts must be _much_ better documented, because with the ~stable API of c/storage (and presumably new non-Podman callers?), future maintainers of this subpackage need to now such things and “just see what Podman does” will be much less tenable.

from how I understood it, the goal of this PR is effectively just to move code from Podman to c/storage so that it can be used from other packages.  If there are small changes to accommodate other users then fine, but I don't think we need to re-evaluate all we did so far.

Currently, I am not even sure why we need this, I am working on it only because there is a Jira card for it.

If you think we don't need it, I am fine to close this PR and explain it in the related Jira card.

mtrmac:
> from how I understood it, the goal of this PR is effectively just to move code from Podman to c/storage so that it can be used from other packages.

With my Skopeo hat on, I don‘t know how I would review a PR to call a function which requires a parameter with an unknown value.

_Somebody_ in our GitHub organization must know what the `rootless` parameter means. It’s hard for me to think that writing that down would be an excessive burden. If really no-one knows or cares, I 👿 …propose ripping that code out everywhere because it serves no known purpose? (Not really, but almost…)

For many of the other comments, quite possibly it’s just the comments that need adjusting.

(I think moving the code to c/storage _is_ a bit of a higher burden, because c/storage mostly (entirely?) wants to keep a stable API, so the functions need to be merged roughly in the shape callers will want to use them. We have much less freedom to just adjust the function and all callers, and doing that cross-repo comes with a notably higher cost.)

---

> Currently, I am not even sure why we need this, I am working on it only because there is a Jira card for it.

@vrothberg filed it. Valentin?

(My _guess_ is
- We almost certainly should share, in c/storage, a “get runtime directory” utility that deals with the `su` / `sudo` breakage by producing a nice error message. That seems clearly useful to me; but we also _need_ to account for exotic platforms in some way. So maybe we should get the added error messages without the bit that makes functions not work at all on Windows.
- For the bit that makes functions not work at all on Windows… I think that’s _conceptually_ correct but if we do that, we ultimately have to replace that code in many callers with _something_. I don’t know what that is, and it seems to me that figuring that out is not much of a priority.
- Similarly, for the functions to be useful, I think they should either return a value, or fail. "`return ""`, deal with this yourself” — how is the caller supposed to do that?
- I don’t know what the “unset variables when not rootless” bit is addressing.
- Moving the code for `XDG_RUNTIME_DIR` fallback to c/storage, so that `Skopeo` etc. fall back the same way, and use the same files for `skopeo login`, would resolve a long-standing inconsistency, and seems desirable, but I don’t remember all of the history. At least the aspect of the current implementation, where this fallback is stored in a separate `rootlessRuntimeDirOnce`, and _not_ overwritten back into the environment, would, I think, be quite an improvement for library use.
- I don’t know what `SetXdgDirs` does / is intended to do so I don’t have an opinion on the code itself, apart from “I can’t LGTM code I don’t understand at all, especially when it clearly contradicts its own documentation.”

)

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +99,104 @@ func stick(f string) error {
...
+)
+
+// GetConfigHome returns XDG_CONFIG_HOME.
+// GetConfigHome returns $HOME/.config and nil error if XDG_CONFIG_HOME is not set.
+//
+// See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
+func GetConfigHome() (string, error) {
+	var rootlessConfigHomeDirError error
+
+	rootlessConfigHomeDirOnce.Do(func() {
```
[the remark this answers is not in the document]
giuseppe: `SetXdgDirs()` sets the value it gets from `GetConfigHome()` so the expectation is that you get the same value from `GetConfigHome()` and from the `XDG_CONFIG_HOME` env variable after `SetXdgDirs()` is used.  We set the env variable so other programs we invoke will use the same one as Podman does.

pkg/homedir/homedir_others.go
```diff
@@ -19,3 +19,16 @@ func GetRuntimeDir() (string, error) {
 func StickRuntimeDirContents(files []string) ([]string, error) {
 	return nil, errors.New("homedir.StickRuntimeDirContents() is not supported on this system")
 }
+
+// GetConfigHome returns XDG_CONFIG_HOME.
+// GetConfigHome returns $HOME/.config and nil error if XDG_CONFIG_HOME is not set.
+//
+// See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
+func GetConfigHome() (string, error) {
+	return "", errors.New("homedir.GetRootlessConfigHomeDir() is not supported on this system")
```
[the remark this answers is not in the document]
giuseppe: I've moved the old code from `homedir.go` to `homedir_others.go` so it doesn't fail on other platforms and behave like today

pkg/homedir/homedir.go
```diff
@@ -50,3 +36,47 @@ func GetCacheHome() (string, error) {
 	}
 	return filepath.Join(home, ".cache"), nil
 }
+
+// SetXdgDirs ensures the XDG_RUNTIME_DIR env and XDG_CONFIG_HOME variables are set.
+// containers/image uses XDG_RUNTIME_DIR to locate the auth file, XDG_CONFIG_HOME is
+// use for the containers.conf configuration file.
+func SetXdgDirs(rootless bool, rootlessUID int) error {
```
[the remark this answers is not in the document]
giuseppe: it is up to the caller to define it.  For Podman it is `euid != 0 || !CAP_SYS_ADMIN`.

Should we hide the `rootless bool, rootlessUID int` details from this function and GetRuntimeDirUser?

pkg/homedir/homedir.go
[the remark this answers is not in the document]
mtrmac: > “ensures the variables… are set” vs. this unsetting them.

This was removed.

I’d still like to see some explanation of what this function is for and who should call it and when.

pkg/homedir/homedir.go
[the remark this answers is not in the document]
mtrmac: > it is up to the caller to define it.

(And how would the caller do that, without any documentation what the parameter means?)

---

Removing the parameter and delegating to `unshare.IsRootless` is not 100% an answer, but it does get us a single consistent value, so that’s 95% there.

pkg/homedir/homedir_unix.go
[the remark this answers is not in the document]
mtrmac: > `SetXdgDirs()` sets the value it gets from `GetConfigHome()`

Ah, great point. I missed that.

vrothberg:
> @vrothberg filed it. Valentin?

I filed the card during/after a meeting where we discussed XDG_ related issues.  IIRC after looking a bug and realizing that the tools don't behave the same and that the code is scattered.  The goal of consolidating the code was to have 1 source of truth to make sure that a given bug must only get fixed in one place.

giuseppe:
what do you think of the new version?

There are no API changes and it doesn't break any functionality on non-linux platforms.

The `SetXdgDirs` is copied from Podman.  If it facilitates the merge of this PR, I can drop it from here and leave it in Podman

mtrmac:
> There are no API changes and it doesn't break any functionality on non-linux platforms.

See the non-resolved comments from the earlier review — these functions now return `""` in some paths.

The Jira card asked for warnings, but that’s not a blocker for this PR as such.

> The `SetXdgDirs` is copied from Podman. If it facilitates the merge of this PR, I can drop it from here and leave it in Podman

I personally can’t LGTM that underdocumented function as it exists today. So splitting that into a separate PR and finding another reviewer would work.

Alternatively, that other reviewer could LGTM this PR as a whole as it is — I’m not an expert in the problem space and I’m perfectly fine with leaving this PR to one.

pkg/homedir/homedir.go
```diff
@@ -50,3 +38,47 @@ func GetCacheHome() (string, error) {
...
+		runtimeDir, err = GetRuntimeDir()
+		if err != nil {
+			return err
+		}
+	}
+	if err := os.Setenv("XDG_RUNTIME_DIR", runtimeDir); err != nil {
+		return fmt.Errorf("cannot set XDG_RUNTIME_DIR: %w", err)
+	}
+
+	if rootless && os.Getenv("DBUS_SESSION_BUS_ADDRESS") == "" {
```
mtrmac, line 66: There’s already an `if !rootless { return}` at the top.

pkg/unshare/unshare_linux.go
```diff
@@ -427,6 +427,15 @@ func IsRootless() bool {
 				isRootless = true
 			}
 		}
+		if !isRootless {
```
mtrmac, line 430: Exactly this check already exists a few lines above — don’t we only need this once?

pkg/homedir/homedir.go
giuseppe: yeah, we don't need this, thanks for spotting it.

Not sure what change in the last version broke the Podman tests, I am still investigating

mtrmac:
Both `GetConfigHome` and `GetRuntimeDir` can now, AFAICS, return `"", nil` on some paths; and almost no callers can deal with that. `dir, err := GetRuntimeDir(); if err != nil { /* fail */ }; … filepath.Join(dir, "some-constant")` is how they are typically used.

giuseppe:
that is how the `GetRuntimeDir` in Podman returned the empty string when running as non-rootless.

We have 3 different versions of the same function, each maintained in a different repo and with some minor differences.

I've no strong opinion on the right or wrong behavior here, but the current version can be replaced in all the places and tests pass.

pkg/homedir/homedir_unix.go
Luap99: yeah I think it would make sense to drop this if we "rework" this anyway.
I cannot imagine anyway wanting this fallback. Also podman sort of depends on runroot actually being tmpfs as the data must be gone after a reboot so this seems like a bad default. In all the issues I looked at in podman I never seen a path like this as the logic should already pick a path under /tmp before that I hardly imagine a case were this fails but $HOME would work.

And lastly I hate any program thinking they have the right to directly write into the my $HOME and not using some standard directory.

mtrmac:
I don’t have nearly as much faith in our test coverage, and either way, test coverage is not really a substitute for a maintainable design.

I don’t actually care strongly about what the behavior _is_, but I want it to be understood and documented, most of all.

It then follows that callers then should use it consistently with documentation. If you want to document “`GetRuntimeDir` returns `""` when run `!unshare.IsRootless`” (or, alternatively, and IMHO more cleanly, `GetRuntimeDir` fails with an error when run `!unshare.IsRootless`”), fine; as long as the callers are updated to actually handle that.

---

Looking at the callers:
- The behavior of `GetConfigHome` returning `"", nil`, is, AFAICS, net new in this PR. Not a single caller can handle that. We can just not introduce that new behavior, AFAICS, and return a proper error. I don’t know what the callers could do otherwise, anyway.
- c/common and c/storage’s current `GetRuntimeDir` never return `"", nil`; and the callers can’t handle that.
- The behavior of returning `"", nil` is preexisting in Podman, and most(?) callers in there _can_ handle that, but almost all of them are _already_ conditionalizing on `unshare.IsRootless`; so it seems much cleaner to me to leave that condition Podman-specific, or preferably to move it all up to the individual callers. (The commit message introducing that behavior talks only about a directory used by `crun`, which also suggests that making this a global behavior is just the wrong place.)

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +87,110 @@ func stick(f string) error {
...
+// The function will try to use the XDG_RUNTIME_DIR env variable if it is set.
+// XDG_RUNTIME_DIR is typically configured via pam_systemd.
+// If XDG_RUNTIME_DIR is not set, GetRuntimeDir will try to find a suitable
+// directory for the current user.
+//
+// See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
+func GetRuntimeDir() (string, error) {
+	var rootlessRuntimeDirError error
+
+	rootless := unshare.IsRootless()
```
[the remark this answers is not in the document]
mtrmac: @Luap99 , @giuseppe : If there are subtle differences like this, please _write them down_. I wouldn’t be surprised if this amounted to 5 dense paragraphs of rare environments and carefully-tuned tradeoffs. Outsiders like me have no chance to maintain the code without that knowledge.

pkg/homedir/homedir_unix.go
mtrmac: Weird things do happen in containers and other restricted environments… and this dates back to 2018 (https://github.com/containers/podman/commit/4086a0f7371dbe5ea104dafe83aadb77b969c0ba ).

I guess I don’t feel _that_ strongly about this part — it is worth discussing but this function can change its fallback logic without changing its API (unless something records the value to persistent storage, and you hinted elsewhere that might be happening?).

giuseppe:
I've changed the code so that:
- `GetConfigHome` never returns the empty string.
- `GetRuntimeDir` never returns the empty string.
- Dropped the code to use the `$HOME` directory as fallback.

Looking more into it, I think we can go farther as a next cleanup and consolidate the `IsRootless()` functions too.  From the kernel PoV the difference is really between "needs a user namespace" to work and "runs as root and capabilities in the initial namespace".  Let's track this separately, opened a PR to see if anything breaks: https://github.com/containers/podman/pull/20601

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +87,100 @@ func stick(f string) error {
...
+			}
+			tmpDir := filepath.Join(resolvedHome, ".config")
+			_ = os.MkdirAll(tmpDir, 0o700)
+			st, err := os.Stat(tmpDir)
+			if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() && isWriteableOnlyByOwner(st.Mode().Perm()) {
+				cfgHomeDir = tmpDir
+			}
+		}
+		if cfgHomeDir == "" {
+			rootlessConfigHomeDirError = errors.New("could not find suitable config directory")
```
mtrmac, line 127: Non-blocking? I think this can be an `else` of the long `if … isWriteableOnlyByOwner` condition, and that would allow producing a useful error message.

mtrmac (commented):
`GetConfigHome`/`GetRuntimeDir` LGTM (noting that Podman callers migrating to this `GetRuntimeDir` will need adjusting for the new `!rootless` behavior).

I can’t speak to `SetXdgDirs`.

pkg/homedir/homedir_unix.go
giuseppe: I can change it as it is anyway clearer to read, but how would you improve the error message?

mtrmac: “directory {tmpDir} already exits and is not used by current user”, optionally listing the expected/actual UIDs. Similarly for the permission.

giuseppe: thanks.  Amended

giuseppe:
@mtrmac I've dropped the patch for `SetXdgDirs` since anyway at the moment it is used only by Podman.  We can reconsider if other tools/packages need it

giuseppe:
@mtrmac are you fine with the last version?

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +86,95 @@ func stick(f string) error {
...
+			return
+		}
+
+		uid := strconv.Itoa(unshare.GetRootlessUID())
+		if runtimeDir == "" {
+			tmpDir := filepath.Join("/run", "user", uid)
+			if err := os.MkdirAll(tmpDir, 0o700); err != nil {
+				logrus.Debug(err)
+			}
+			st, err := os.Stat(tmpDir)
```
mtrmac, line 158: Pre-existing: `Lstat` might be more correct here, but it doesn’t matter as long as `/run/user` is writable by `root` only.

pkg/homedir/homedir_unix.go
```diff
@@ -94,3 +86,95 @@ func stick(f string) error {
...
+			if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() && isWriteableOnlyByOwner(st.Mode().Perm()) {
+				runtimeDir = tmpDir
+			}
+		}
+		if runtimeDir == "" {
+			tmpDir := filepath.Join(os.TempDir(), fmt.Sprintf("storage-run-%s", uid))
+			if err := os.MkdirAll(tmpDir, 0o700); err != nil {
+				logrus.Debug(err)
+			}
+			st, err := os.Stat(tmpDir)
```
mtrmac, line 168: Preexisting, but this _must_ be `Lstat`.

mtrmac (commented):
This code LGTM…

… but there’s one more bit to the consolidation: `c/storage/types/getRootlessRuntimeDir`.

Arguably, the `c/storage/types` one calls `homedir.GetRuntimeDir`, so it is “fine” not to touch it. OTOH it does very similar kinds of fallbacks to `/run/user` and to `/tmp` (and, of course, it diverges, in having a special `systemd` logic); so I think it would make sense to consolidate all of them.

And the `c/storage/types` one is the only one that has unit tests, and operating in `/tmp` is somewhat security-critical… so seems attractive to consolidate it as well, including the unit tests.

I’m afraid I didn’t find that one when just searching for `Getenv(…)`.

I don’t know what’s the best way to handle this. Maybe merge this PR as is, and then consolidate in an immediate follow-up, as a part of the same Jira card? WDYT?

giuseppe:
> Arguably, the `c/storage/types` one calls `homedir.GetRuntimeDir`, so it is “fine” not to touch it. OTOH it does very similar kinds of fallbacks to `/run/user` and to `/tmp` (and, of course, it diverges, in having a special `systemd` logic); so I think it would make sense to consolidate all of them.

I really missed that one, thanks for checking it.  Not sure how we could end up in this mess.

The version in c/storage/types seems just wrong, it accepts the UID as an argument but it is kind of ignored.

Should we just remove this function and modify its callers to use `homedir.GetRuntimeDir()`?

mtrmac:
> I really missed that one, thanks for checking it. Not sure how we could end up in this mess.

I guess every project needed a fallback, but consolidation was nobody’s job so far…

> The version in c/storage/types seems just wrong, it accepts the UID as an argument but it is kind of ignored.

Is it? AFAICS `getRootlessRuntimeDir` uses the value to compute the fallback paths.

---

> Should we just remove this function and modify its callers to use `homedir.GetRuntimeDir()`?

I now remember asking for those unit tests (https://github.com/containers/storage/pull/568 was a _long_ effort), and I’m rather fond of keeping them.

Also non-systems users probably want that systemd exception that originally motivated that PR. OTOH Podman didn’t have it, so maybe that part doesn’t work any more?

types/utils.go
```diff
@@ -71,92 +57,6 @@ func isRootlessRuntimeDirOwner(dir string, env rootlessRuntimeDirEnvironment) bo
 	return err == nil && int(st.UID()) == os.Getuid() && st.Mode()&0o700 == 0o700 && st.Mode()&0o066 == 0o000
 }
 
-// getRootlessRuntimeDirIsolated is an internal implementation detail of getRootlessRuntimeDir to allow testing.
-// Everyone but the tests this is intended for should only call getRootlessRuntimeDir, never this function.
-func getRootlessRuntimeDirIsolated(env rootlessRuntimeDirEnvironment) (string, error) {
```
mtrmac, line 76: I’d prefer adapting the `rootlessRuntimeDirEnvironment`-based implementation… but if it were to be removed, then all of that interface and the default implementation should be removed as well.

types/options.go
```diff
@@ -273,10 +275,16 @@ func isRootlessDriver(driver string) bool {
 func getRootlessStorageOpts(rootlessUID int, systemOpts StoreOptions) (StoreOptions, error) {
 	var opts StoreOptions
 
-	dataDir, rootlessRuntime, err := getRootlessDirInfo(rootlessUID)
+	dataDir, err := homedir.GetDataHome()
```
mtrmac, line 278: This makes the `rootlessUID` value ignored, and that propagates up the call stack… `DefaultStoreOptions` is widely called, and it has at least one caller who calls it with `(false, 0)`.

I would love for the configuration loading code of c/storage to be simplified, but I quickly gave up every time I looked at that spaghetti; doing that in this PR seems way too big a scope expansion.

giuseppe: `getRootlessDirInfo(rootlessUID int) (string, string, error)` also ignored it:

```
      rootlessRuntime, err := GetRootlessRuntimeDir(rootlessUID)
      if err != nil {
               return "", "", err
      }

      home := homedir.Get()
       if home == "" {
               return "", "", fmt.Errorf("neither XDG_DATA_HOME nor HOME was set non-empty: %w", err)
       }
```

only `GetRootlessRuntimeDir()` used it to hardcode in the directory path.

The caller with `(false, 0)` uses only the UIDMap and GIDMap settings so it should not be affected by this change.  I'll vendor it in Podman to see if anything breaks.

giuseppe:
Podman tests pass with this version

types/options.go
mtrmac: > `getRootlessDirInfo(rootlessUID int) (string, string, error)` also ignored it:

That’s not what I see at all?! `rootlessRuntime` is returned on every non-error exit from `getRootlessDirInfo`.

utils.go
```diff
@@ -11,11 +11,6 @@ func ParseIDMapping(UIDMapSlice, GIDMapSlice []string, subUIDMap, subGIDMap stri
 	return types.ParseIDMapping(UIDMapSlice, GIDMapSlice, subUIDMap, subGIDMap)
 }
 
-// GetRootlessRuntimeDir returns the runtime directory when running as non root
-func GetRootlessRuntimeDir(rootlessUID int) (string, error) {
```
mtrmac, line 15: This is an API break… c/storage, is IIRC, not that strict about them as c/image, I don’t know where exactly the line is.

types/options.go
mtrmac: > The caller with `(false, 0)` uses only the UIDMap and GIDMap settings so it should not be affected by this change.

Good point.

Also, looking at `DefaultStoreOptions`, it makes no sense at all to have the two parameters, but then process them only under a `sync.Once`. So I think that should be deprecated in favor of some variant of `DefaultStoreOptionsAutoDetectUID`.

But that is really looking like 2 or 3 separate PRs…

utils.go
giuseppe: yes it is an API break, just I don't think we should keep around this function that accepts a generic UID.  I would expect the result to be the same as `GetRuntimeDir()` and this is not the case now.

I can move it to a separate PR if it helps, but IMO this variant should just disappear.

mtrmac: I really don’t know what the rules for c/storage are.

Personally, it seems to me that we’ve been breaking API similarly… often enough, and I’m not aware of any vocal complaints of wrist slaps; so I think removing this would be reasonably consistent with past practice.

I’m mostly just highlighting this to make it explicit and allow other interested parties to speak up.

mtrmac: Also, I don’t think splitting to a separate PR makes much of a difference here: either the API breakage is acceptable, or it isn’t. And if this PR, either way, removed the only callers of the public `GetRootlessRuntimeDir` (and I can see no out-of-repo users in GitHub.com/containers), removing the functions here is trivially small and can be done here.

(By the “2 or 3 separate PRs” above I was thinking:
- A recent version of this: consolidating `GetConfigHome`+`GetRuntimeDir` from Podman etc., keeping `utils.GetRootlessRuntimeDir` mostly unchanged.
- Deprecating `DefaultStoreOptions` (and other APIs that take explicit UID parameters???), and dealing with non-zero fallout of that.
- Then folding `utils.GetRootlessRuntimeDir` into `homedir.GetRuntimeDir`.

I don’t know whether all of that is immediately worthwhile.)

rhatdan:
Is this ready to merge?

mtrmac:
- Although we don’t seem to have any “real users” of the UID parameter of `getRootlessStorageOpts`, I think just silently ignoring a parameter this deep in the call stack is confusing and makes the whole thing harder to maintain. That parameter should be removed outside-in: the callers, the public API, the implementation here.
- I’m not really happy with dropping the unit tests using `rootlessRuntimeDirEnvironmentTest`, and replacing them with a call to a function with no tests. (I am open to being overridden by others on this, I guess.)

giuseppe:
I think we should just drop `DefaultStoreOptions()` in its current form and only have `DefaultStoreOptionsAutoDetectUID()`.  I forgot why we added it, maybe because Podman had a different way to track the rootless user compared to `pkg/unshare`.  Since we are consolidating on that, there is no reason to pass the UID from the outside.

Do you want me to do this as part of this PR or can be done later?  The issue with `rootlessUID` being partly ignored is already present in the current version

mtrmac:
I agree that dropping `DefaultStoreOptions` (and `UpdateStoreOptions`) in favor of …`AutoDetectUID` (whether keeping the current names, or renaming …`AutoDetectUID` to just `DefaultStoreOptions`, it’s an API break either way) makes perfect sense.

Ideally, the ordering that makes most sense to me is https://github.com/containers/storage/pull/1740#discussion_r1389688527 . Priorities change, so I’d prefer not to make things a bit worse now to make them better only later; but I’ll defer to others if that seems unreasonable.

types/options.go
```diff
@@ -197,15 +199,16 @@ func loadStoreOptions(rootless bool, rootlessUID int) (StoreOptions, error) {
...
-	storeOptions, storeError = loadStoreOptions(rootless, rootlessUID)
+func UpdateStoreOptions() (StoreOptions, error) {
+	storeOptions, storeError = loadStoreOptions(unshare.IsRootless(), unshare.GetRootlessUID())
 	return storeOptions, storeError
 }
 
 // DefaultStoreOptions returns the default storage ops for containers
-func DefaultStoreOptions(rootless bool, rootlessUID int) (StoreOptions, error) {
+func DefaultStoreOptions() (StoreOptions, error) {
+	uid := unshare.GetRootlessUID()
```
mtrmac, line 209: AFAICS this can be computed inside the `once.Do` without changing semantics. Is there some non-trivial reason that’s not the case?

types/options.go
```diff
@@ -197,15 +199,16 @@ func loadStoreOptions(rootless bool, rootlessUID int) (StoreOptions, error) {
...
 	once.Do(func() {
-		storeOptions, storeError = loadStoreOptions(rootless, rootlessUID)
+		storeOptions, storeError = loadStoreOptions(uid != 0, uid)
```
mtrmac, line 211: Why is `loadStoreOptions` called with differently-computed `rootless` parameter on these two code paths? If there is a non-trivial reason, please document that.

(And if the two shouldn’t differ, the values could be computed inside `loadStoreOptions` instead of the callers.)

types/utils.go
```diff
@@ -169,15 +23,15 @@ func expandEnvPath(path string, rootlessUID int) (string, error) {
...
+func DefaultConfigFile() (string, error) {
 	if defaultConfigFileSet {
 		return defaultConfigFile, nil
 	}
 
 	if path, ok := os.LookupEnv(storageConfEnv); ok {
 		return path, nil
 	}
-	if !rootless {
+	if unshare.GetRootlessUID() == 0 {
```
mtrmac, line 34: Again on the `IsRootless` vs. `UID == 0` — all Podman callers AFAICS used to call this with `rootless.IsRootless()`, so to ignorant me this looks like a change. Is that intentional?

… and there is also `defaultStoreOptionsIsolated` conditioned on `rootless && rootlessUID != 0`; so we have _three_ different conditions around config files in this subpackage. Are they really intentionally different?

If the logic should not be just a plain `rootless.IsRootless()`, I’d like to see that condition consolidated into a single function (`usePerUserStorage`??? possibly 1 line of code + 10 lines of comments) called everywhere here, so that the individual places don’t drift.

It’s also quite possible I am completely missing the point.

mtrmac (commented):
I’m still unhappy with dropping the `rootlessRuntimeDirEnvironmentTest` tests.

mtrmac:
> this complicates dependencies since now I've also to deal with buildah and c/image.

Yeah, and unfortunately the timing coincides with the soon-to-begin vendor dance.

> I've made these changes and opened the test PRs. If there are no more objections and the test PRs pass, let's merge as it is becoming kind of difficult to keep rebasing all our repos

When a problem, during investigation, grows larger than anticipated, my go-to tool is to split it into smaller and more granular PRs, slicing the problem into isolated easier-to-handle chunks.

I don’t know how to deal with a PR that grows with the understood problem … or even _faster_ than the understood problem, when issues come up more frequently than resolutions. (Historically, when there was enough business value for the feature, what has happened is that someone else approved and merged that PR. And I realize that included some in-retrospect-critical features. \*shrug\*)

types/utils.go
giuseppe: no you are right.  I've added a `usePerUserStorage` function that explains why we don't want to use directly `IsRootless`

types/options.go
```diff
@@ -109,21 +111,41 @@ func loadDefaultStoreOptions() {
...
+	}
+	return loadStoreOptionsFromConfFile(storageConf)
+}
+
+// usePerUserStorage returns whether the user private storage must be used.
+// We cannot simply use the unshare.IsRootless() condition, because
+// that checks only if the current process needs a user namespace to
+// work and it would break cases where the process is already created
+// in a user namespace (e.g. nested Podman/Buildah) and the desired
+// behavior is to use system paths instead of user private paths.
```
mtrmac, line 128: Thanks for documenting this!

types/options.go
```diff
@@ -109,21 +111,41 @@ func loadDefaultStoreOptions() {
...
+func usePerUserStorage() bool {
+	return unshare.IsRootless() && unshare.GetRootlessUID() != 0
+}
+
+// loadStoreOptionsFromConfFile is an internal implementation detail of DefaultStoreOptions to allow testing.
 // Everyone but the tests this is intended for should only call DefaultStoreOptions, never this function.
-func defaultStoreOptionsIsolated(rootless bool, rootlessUID int, storageConf string) (StoreOptions, error) {
+func loadStoreOptionsFromConfFile(storageConf string) (StoreOptions, error) {
```
mtrmac, line 135: (The idea of `…Isolated` is that it doesn’t depend on the environment, and it can be fully unit-tested in any environment; in this case, that would probably still mean having `userPerUserStorage bool, rootlessUID int` parameters; see [1]

… OTOH now that `getRootlessStorageOpts` effectively can’t be isolated that way any more, and considering that there is only one unit test exercising this, that might be not valuable enough to preserve… )

types/utils_test.go
```diff
@@ -1,276 +1,18 @@
...
-	assert.NilError(t, err)
-	assert.Assert(t, resultDir != tmpPerUserDir, "Rootless runtime dir shouldn't follow race dir.")
-}
-
 func TestDefaultStoreOpts(t *testing.T) {
-	storageOpts, err := defaultStoreOptionsIsolated(true, 1000, "./storage_test.conf")
-
-	expectedPath := filepath.Join(os.Getenv("HOME"), "1000", "containers/storage")
+	storageOpts, err := loadStoreOptionsFromConfFile("./storage_test.conf")
+	expectedPath := filepath.Join(os.Getenv("HOME"), fmt.Sprintf("%d", unshare.GetRootlessUID()), "containers/storage")
```
mtrmac, line 15: [1] `…Isolated` had the `(rootless, UID)` parameters so that this test succeeds in all environments, even when run as root.

That would no longer be the case with these changes, but I also don’t know how much that matters. Maybe an explicit test skip would be enough. (Ideally I’d like the config logic to have good unit test coverage, but that’s clearly not this PR.)

types/options.go
```diff
@@ -109,21 +111,41 @@ func loadDefaultStoreOptions() {
...
+// that checks only if the current process needs a user namespace to
+// work and it would break cases where the process is already created
+// in a user namespace (e.g. nested Podman/Buildah) and the desired
+// behavior is to use system paths instead of user private paths.
+func usePerUserStorage() bool {
+	return unshare.IsRootless() && unshare.GetRootlessUID() != 0
+}
+
+// loadStoreOptionsFromConfFile is an internal implementation detail of DefaultStoreOptions to allow testing.
 // Everyone but the tests this is intended for should only call DefaultStoreOptions, never this function.
```
mtrmac, line 134:
```suggestion
// Everyone but the tests this is intended for should only call loadStoreOptions, never this function.
```
?

types/options.go
```diff
@@ -270,14 +284,22 @@ func isRootlessDriver(driver string) bool {
...
 		return opts, err
 	}
-	opts.RunRoot = rootlessRuntime
+
+	rootlessRuntime, err := homedir.GetRuntimeDir()
+	if err != nil {
+		return opts, err
+	}
+
+	opts.RunRoot = filepath.Join(rootlessRuntime, "containers")
```
mtrmac, line 302: Oh what a trap, I have totally missed this.

The previous code also does `MkdirAll`, is that unnecessary?

giuseppe: thanks for checking it, I've reintroduced it

giuseppe:
tests pass both in Podman and Buildah:

- https://github.com/containers/podman/pull/20528
- https://github.com/containers/buildah/pull/5163

types/utils_test.go
```diff
@@ -1,276 +1,21 @@
...
-	assert.NilError(t, err)
-	assert.Assert(t, resultDir != tmpPerUserDir, "Rootless runtime dir shouldn't follow race dir.")
-}
-
 func TestDefaultStoreOpts(t *testing.T) {
-	storageOpts, err := defaultStoreOptionsIsolated(true, 1000, "./storage_test.conf")
-
-	expectedPath := filepath.Join(os.Getenv("HOME"), "1000", "containers/storage")
+	if !usePerUserStorage() {
+		return
```
mtrmac, line 15: `t.Skip()`, please.

giuseppe: fixed

giuseppe:
can we merge now?

rhatdan:
@mtrmac PTAL

mtrmac:
I’d like the `rootlessRuntimeDirEnvironmentTest` tests to be ported/preserved/updated.

I don’t feel _that_ strongly about that; if anyone else merges this, I won’t object.

PR #1752, opened 2023-11-06
store: Make additionalimagestores optional as layerstore source

idleroamer:
Optionally make additionalImageStore independently maintainable by ensuring that layers are not shared between RO and RW storages so that updating/removing RO storages won't break images in RO ones

Closes: #20603

### DIFF
--- a/store.go
+++ b/store.go
@@ -945,6 +945,7 @@ func (s *store) load() error {
 	s.containerStore = rcs
 
 	for _, store := range driver.AdditionalImageStores() {
+		store = strings.Split(store, "|")[0]
 		gipath := filepath.Join(store, driverPrefix+"images")
 		var ris roImageStore
 		if s.imageStoreDir != "" && store == s.graphRoot {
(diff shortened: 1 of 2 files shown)


store.go
```diff
@@ -1109,6 +1110,14 @@ func (s *store) getROLayerStoresLocked() ([]roLayerStore, error) {
...
 	}
 	for _, store := range s.graphDriver.AdditionalImageStores() {
+		if strings.Contains(store, "|") {
+			parts := strings.Split(store, "|")
+			if len(parts) > 1 && parts[1] == "NO_REUSE" {
+				continue
+			} else {
+				store = strings.Split(store, "|")[0]
+			}
+		}
```
mtrmac, line 1120: I can’t see how this would work; layers in these stores would be effectively invisible, but images in those stores would be visible as usual.

The immediate effect is that all images from a no-reuse store are visible but have unresolvable layers = look broken to all existing tooling.

store.go
```diff
@@ -1109,6 +1110,14 @@ func (s *store) getROLayerStoresLocked() ([]roLayerStore, error) {
 		return nil, err
 	}
 	for _, store := range s.graphDriver.AdditionalImageStores() {
+		if strings.Contains(store, "|") {
+			parts := strings.Split(store, "|")
+			if len(parts) > 1 && parts[1] == "NO_REUSE" {
```
mtrmac, line 1115: We can invent an infinite number of new options. Please don’t add ambiguous in-line syntaxes unless there is really no other alternative.

mtrmac (changes requested):
Copy&pasting from https://github.com/containers/podman/pull/20604 to preserve the comments…

store.go
idleroamer: I have tested the changes in 3.X branch, as part of our older software version which manifested the issue and verified basic operations. Admittedly the codebase has changed substantially and the current patch as you described doesn't work anymore.
I will check if there is still a manageable way to introduce the change in main branch.

mtrmac: Note use cases like `podman history` and `podman image tree`.

idleroamer:
I don't see any way around increasing the complexity to introduce this feature, so I am going to close the PR for now.
I appreciate your support @mtrmac

rhatdan:
You could bring this up for discussion at the next Cabal meeting.

PR #1758, opened 2023-11-14
Bump to v1.51.0 and then v1.51.1-dev

TomSweeneyRedHat:
First step in the vendor dance for Podman v4.8


TomSweeneyRedHat:
Happy Green Test Buttons @containers/image-maintainers PTAL

mtrmac:
The final merge commit is a bit unexpected (this effectively tags a version one dependency update behind current `main`) but it’s clear what is being tagged, and I can’t see how this would break anything. We should get that dependency from c/image anyway…

/lgtm

PR #1761, opened 2023-11-15
Embed text from VERSION for use in Store.Version()

nalind:
Store.Version() has long returned an empty slice of [2]string.  Change that to return "Version" and the text from the top-level VERSION file.


mtrmac (commented):
/lgtm

Nice!

PR #1769, opened 2023-12-06
composefs: enable it at runtime

giuseppe:
drop the build tag for building composefs and convert it to an option in `storage.conf`


pkg/config/config_test.go
```diff
@@ -251,6 +251,15 @@ func TestOverlayOptions(t *testing.T) {
 		t.Fatalf("Expected to find 'skip_mount_home' options, got %v", doptions)
 	}
 
+	options.Overlay.UseComposefs = "true"
+	doptions = GetGraphDriverOptions("overlay", options)
+	if len(doptions) == 0 {
```
eriksjolund, line 256:
```suggestion
	if len(doptions) != 0 {
```

cgwalters:
So I was trying this out like this:

```
$ podman --log-level debug --storage-opt use_composefs=true --root /var/tmp/containers --runroot /run/tmp/containers run --rm -ti quay.io/fedora/fedora:39 echo hello
```

I see

```
DEBU[0000] overlay: use_composefs=true 
DEBU[0000] Cached value indicated that overlay is supported 
DEBU[0000] Cached value indicated that data-only layers for overlay are supported 
DEBU[0000] Cached value indicated that overlay is supported 
DEBU[0000] Cached value indicated that metacopy is not being used 
```
Yet AFAICS `generateComposeFsBlob()` is never being invoked because `ApplyDiffFromStagingDirectory` is never being invoked?  But I don't understand the conditions under which that is invoked.

giuseppe:
> ```
> 
> ```
> 
> 
>     
>   
> 
> Yet AFAICS `generateComposeFsBlob()` is never being invoked because `ApplyDiffFromStagingDirectory` is never being invoked?  But I don't understand the conditions under which that is invoked.

it works only with the zstd:chunked code path now, you need to add this snippet to `storage.conf`:

```
[storage.options]

pull_options = {convert_images = "true", enable_partial_images = "true", use_hard_links = "false", ostree_repos=""}

```

cgwalters:
Ah, OK only operating on zstd:chunked *and* requiring an opt-in *and* previously a build tag definitely makes it hard to use :smile: 
Removing one of those things definitely helps.

I think I've done what you suggested, but I am still not managing to trigger the composefs path.  I do see

```
DEBU[0000] Found image "83a6861c2dbd64301df6dde82f9d99b3659b50478fa9cb7296e21552212a3757" as "83a6861c2dbd64301df6dde82f9d99b3659b50478fa9cb7296e21552212a3757" in local containers storage ([overlay@/var/tmp/containers/storage+/run/tmp/containers:overlay.mountopt=nodev,metacopy=on,overlay.use_composefs=true]@83a6861c2dbd64301df6dde82f9d99b3659b50478fa9cb7296e21552212a3757) 
```

which is different, but apparently `ApplyDiffFromStagingDirectory` still isn't being run.

giuseppe:
you already had the image in the storage?  Was it re-pulled once you enabled the storage options?

rhatdan:
LGTM
@cgwalters @eriksjolund @nalind @vrothberg @saschagrunert PTAL

drivers/overlay/overlay.go
```diff
@@ -387,6 +389,22 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error)
 		}
 	}
 
+	if opts.useComposefs {
+		if unshare.IsRootless() {
+			return nil, fmt.Errorf("composefs is not supported in user namespaces")
```
saschagrunert, line 394: non blocking nit:
```suggestion
			return nil, errors.New("composefs is not supported in user namespaces")
```

drivers/overlay/overlay.go
```diff
@@ -122,6 +123,7 @@ type Driver struct {
 	supportsDType    bool
 	supportsVolatile *bool
 	usingMetacopy    bool
+	usingComposefs   bool
```
saschagrunert, line 126: non-blocking nit:
```suggestion
	useComposefs   bool
```

saschagrunert (approved):
LGTM

rhatdan:
@giuseppe could you fix @saschagrunert Nit's

rhatdan:
I am going to merge to fix CI/CD system on other PRs.  @giuseppe if you care to fix the nits, please open a new PR.

PR #1770, opened 2023-12-09
pkg/unshare: return true from IsRootless if user is not root

dfr:
This helps to avoid a confusing error on FreeBSD when commands such as 'podman run' are attempted by a non-root user. Changing IsRootless to return true for these users allows types.DefaultConfigFile to return a user-accessible path for storage.conf which, in turn, allows 'podman run' to get as far as SetupRootless which will generate a more informative message than 'Error: stat
/var/db/containers/storage/libpod/bolt_state.db: permission denied'

### DIFF
--- a/pkg/unshare/unshare_unsupported.go
+++ b/pkg/unshare/unshare_unsupported.go
@@ -17,7 +17,7 @@ const (
 
 // IsRootless tells us if we are running in rootless mode
 func IsRootless() bool {
-	return false
+	return os.Getuid() != 0
 }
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@giuseppe PTAL
/approve

giuseppe (approved):
LGTM

giuseppe:
the lint failure is not related and fixed by https://github.com/containers/storage/pull/1769/commits/269cfdacd9e42b549b775e62cb3fe23377f5256a

PR #1772, opened 2023-12-12
chunked: flush the input tarball to the output

giuseppe:
Flush the entire input tarball to the output in the zstd:chunked stream writer.  This is needed to include any trailing zeros that affect the uncompressed digest.

Closes: https://github.com/containers/storage/issues/1771

### DIFF
--- a/pkg/chunked/compressor/compressor.go
+++ b/pkg/chunked/compressor/compressor.go
@@ -420,6 +420,14 @@ func writeZstdChunkedStream(destFile io.Writer, outMetadata map[string]string, r
 		zstdWriter.Close()
 		return err
 	}
+
+	// make sure the entire tarball is flushed to the output as it might contain
+	// some trailing zeros that affect the checksum.
+	if _, err := io.Copy(zstdWriter, its); err != nil {
+		zstdWriter.Close()
+		return err
+	}
+
 	if err := zstdWriter.Flush(); err != nil {
 		zstdWriter.Close()
 		return err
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
But linter is not happy.

pkg/chunked/compressor/compressor.go
```diff
@@ -420,6 +420,14 @@ func writeZstdChunkedStream(destFile io.Writer, outMetadata map[string]string, r
 		zstdWriter.Close()
 		return err
 	}
+
+	// make sure the entire tarball is flushed to the output as it might contain
+	// some trailing zeros that affect the checksum.
+	if _, err := io.Copy(zstdWriter, its); err != nil {
```
mtrmac, line 426: For the record: This looked very surprising to me (if notthing else, I doubted that after passing `its` to `tar.NewReader` there are any promises at all about the position of `its`)…

But tar-split internally does the same thing in https://github.com/vbatts/tar-split/blob/9a95e026024d3929928021244d72e59f758616b3/tar/asm/disassemble.go#L121-L151 . 

(And for the purposes of this PR, we don’t care whether the unexpected trailing padding is recorded into tar-split metadata (that _is_ important, but hopefully ensured by existing code), but whether we write it into the `zstdWriter` unmodified; so we must do that separately.)

mtrmac (commented):
LGTM, modulo the tests.

rhatdan:
I merged the other PR which I believe fixes the lint, could you rebase to make sure tests pass.

PR #1778, opened 2023-12-15
chunked: disallow both zstd:chunked and eStargz TOCs

giuseppe:
reject a layer if it contains both a zstd:chunked and an eStargz TOC since there are no guarantees that the two TOCs are consistent.


rhatdan (approved):
/lgtm

mtrmac:
Thanks; please do this in `chunked/toc/GetTOCDigest` as well, to make it unambiguous (and, given the current call stack, to catch this early).

Issue #1779, opened 2023-12-18
`podman image mount` issue when additional image stores are used


mvo5:
We ran into a quite vexing issue when building a relatively simple stage for image-builder [0]. The code below copies a container ` ghcr.io/ondrejbudai/bootc:fedora` in our tests into a `/tmp/$tmpdir` image store and then uses that image store in /tmp to mount the image using `podman --imagestore /tmp/$tmpdir image mount ...`.  The issue is that in that setup the final layer of the image (which in this case contains a modified /usr/bin/bootc file) is missing and the "old" bootc file is in the mounted path but not the expected one.

After some debugging it turns out that the way the final mount call is generated is incorrect for the case of additional storages. It will generate something like:
```
lowerdir=/var/lib/containers/storage/overlay/8d72fb04445c9b041287bf1aa467de4bb8a95fb2795bb6fbe214aee32f8f1cd0/diff:/tmp/tmpy7x_x8wm/overlay/l/EWJCNVO7BXHKURAPHEKM3HOTFC:....
```
however the `/var/lib/containers/storage/overlay/8d72fb04445c9b041287bf1aa467de4bb8a95fb2795bb6fbe214aee32f8f1cd0/diff` dir is empty because the actual data is stored in   /tmp/tmp7x... and the "normal" store does not contain any data.

Attached is a small patch that works for my test-case in image-builder (and the "normal" case seems to be still working fine when running this manully) - unfortunately I do not have a better reproducer then running it with the image builder commit below right now also sorry that e patch needs `-p4` to be applied, I tested it directly in a build podman, I hope this bugreport and the diff are still useful.

[extra-storage.diff.txt](https://github.com/containers/storage/files/13708315/extra-storage.diff.txt)

Thanks!
 Michael

[0] https://github.com/ondrejbudai/osbuild/commit/6d7d411b0c4f6d3ab5d8cd06745ff2204bdb5f0f#diff-2b1fc5cd8b2ebf1d121181a2570dff7a923f476fd8a9e6c928c9f9f7085d8f8fR34

rhatdan:
Please open a PR for us to review.
@giuseppe @nalind PTAL

mvo5:
> Please open a PR for us to review. @giuseppe @nalind PTAL

Thanks a lot for your quick reply. As suggested I opened a PR.
(fwiw I didn't right away because my change does not contain a test and opening a PR without a test makes me uneasy :)

mvo5:
[...]

PR #1781, opened 2023-12-19
overlay: use `dir` for final `diff`/`merged` dirs

mvo5:
Dan suggested to open this as a PR and I'm happy to do this. For details see my short commit message below and the issue #1779. I did not open a PR initially because I don't have a test for my change (sorry for that!). I hope it's still useful and maybe someone more familiar with the codebase can write a regression test for this change. Feel free to close/rework as needed of course.

---

This commit fixes an issue with the `overlay.go:Driver.get()` code when the requested `id` comes from an extra storage dir that got specified with e.g. `podman --imagestore`. In this scenario the final layer with the `.../diff` dir must come from the extra storage dir instead of from the main storage (that is may not contain this id).

This fixes issue #1779 for my test case and I see no other issues in my "light" testing for the "normal" cases. However I'm not familiar with this code base and this is (of course) an extremly sensitive part of the code so please have a careful look.

More details in https://github.com/containers/storage/issues/1779

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1598,7 +1598,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		return dest, nil
 	}
 
-	diffDir := path.Join(workDirBase, "diff")
+	diffDir := path.Join(dir, "diff")
 
 	if dest, err := maybeAddComposefsMount(id, 0, readWrite); err != nil {
 		return "", err
@@ -1698,7 +1698,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		return "", err
 	}
 
-	mergedDir := path.Join(workDirBase, "merged")
+	mergedDir := path.Join(dir, "merged")
 	// Create the driver merged dir
 	if err := idtools.MkdirAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
 		return "", err
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1599,6 +1599,9 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	}
 
 	diffDir := path.Join(dir, "diff")
+	if _, err := os.Stat(diffDir); os.IsNotExist(err) {
+		diffDir = path.Join(workDirBase, "diff")
+	}
 
 	if dest, err := maybeAddComposefsMount(id, 0, readWrite); err != nil {
 		return "", err
@@ -1698,9 +1701,9 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 		return "", err
 	}
 
-	mergedDir := path.Join(dir, "merged")
+	mergedDir := path.Join(workDirBase, "merged")
 	// Create the driver merged dir
-	if err := idtools.MkdirAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
+	if err := idtools.MkdirAllAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
 		return "", err
 	}
 	if count := d.ctr.Increment(mergedDir); count > 1 {


mvo5:
Fwiw, https://github.com/osbuild/osbuild/commit/60238b1fa86c59ae40793a8b091785bc1a4e901b is a testcase for our code in osbuild now to reproduce the issue. It essentially does:
1. build an image with two layers that contain both "/file1" with different content 
2. export this image to a tempdir $tmpstore container storage via skopeo
3. remove it from the podman container storage
4.  run `podman --imagestore $tmpstore image mount` and observe that `/file1` is taken from the base layer and not the expected layer

The unpatched podman fails in the expected way, the patched one works correctly in this testcase.

rhatdan:
You have to sign your commit.
git commit -a --amend -s
git push --force

rhatdan:
@giuseppe @nalind PTAL
I will open a WIP PR against Podman to test this out.

giuseppe:
Do you have a reproducer for the issue you are facing?

Should we fix this condition instead?

```
	if imageStore != "" {
		workDirBase = imageStore
	}
```

and make sure that the correct value is used when `--imagestore` is passed?

mvo5:
> Do you have a reproducer for the issue you are facing?
[..]

I do, it's part of our tests in osbuild, the easiest way is probably to do the following:
```
$ git clone https://github.com/mvo5/osbuild
$ cd osbuild
$ git checkout a7db0c0d6e85d7644c5c98d116fb4a227353ccd6   # part of the osbuild-container-deploy branch
$ sudo pytest -s -vv -k test_container_deploy
```
the relevant test is stages/test/test_container_deploy.py and the code (very small) is stages/org.osbuild.container-deploy

I'm sorry that I did not extract it for easier consumption but I hope it's useful.

mvo5:
[..]
> Should we fix this condition instead?
> 
> ```
> 	if imageStore != "" {
> 		workDirBase = imageStore
> 	}
> ```
> 
> and make sure that the correct value is used when `--imagestore` is passed?

I don't feel qualified to give a good answer here - I will ramble a bit anyway :) My understanding of this code is that `dir` is the place that the `id` comes from so AIUI it should be used for the final path of the overlay. AIUI podman always adds `--imagestore` as an optional store the `id` could come from the "normal" imageStore and the one added via `--imagestore`. But I may be totally off here so take all of this with a grain of salt.

giuseppe:
would it be possible to use just podman commands for the reproducer?

I've tried your script, but I don't understand what is going on behind and what commands are involved

mvo5:
> would it be possible to use just podman commands for the reproducer?
> 
> I've tried your script, but I don't understand what is going on behind and what commands are involved

Sure thing, I put a standalone reproducer into https://github.com/containers/storage/issues/1779#issuecomment-1863419315 - it's still slightly more complicated than I would like, I'm sorry for that, I can give it another attempt tomororw to simply or convert to shell, just let me know :)

mvo5:
Fwiw, the tests in the podman PR in https://github.com/containers/podman/pull/21058 seems to be failing so my changes are too simplistic and/or incorrect. Should we close this again? Or keep it to be able to test changes here?

giuseppe:
we can keep it open if you plan to work more on this.  Otherwise we can close it until we have a minimal reproducer that uses only Podman

mvo5:
> we can keep it open if you plan to work more on this. Otherwise we can close it until we have a minimal reproducer that uses only Podman

I tweaked the change a bit, I was able to run one of the failing tests locally. I also tweaked the reproducer in https://gist.github.com/mvo5/36c35e9d2a17ad56f73aa90970dbbc27 (still pytest though but better output and a bit more linear)

giuseppe:
could you try https://github.com/containers/storage/pull/1784 ?

mvo5:
> could you try #1784 ?

Sure, sorry that it took me so long. I applied #1784 into the vendor tree in podman and build it. The file in my test is now coming from the correct layer \o/ - thank you! However I do see a leftover mount in `/tmp/test-issue1179/overlay`:
```
$ grep overlay /proc/self/mountinfo 
13551 32 259:3 /tmp/test-issue1179/overlay /tmp/test-issue1179/overlay rw,relatime - ext4 /dev/nvme0n1p2 rw,errors=remount-ro
```
I am not sure where this one comes from, iirc my test does not create any bind mounts and I double checked that `podman image umount` is called in my script.

mvo5:
I close this one now given the very nice work in 1784. @giuseppe please let me know if I should copy my comment https://github.com/containers/storage/pull/1781#issuecomment-1866706672 into #1784

PR #1791, opened 2024-01-03
Move minimal golang to v1.20

rhatdan:


mtrmac:
@rhatdan Please document the reason — is it “just” to be fresh, or is there something that doesn’t build with 1.19 now?

mtrmac:
(The change LGTM, it’s just that I think it’s _so much_ easier to document the rationale now than to retroactively figure it out later, if the latter is even possible to do with certainty, that writing that down now is well worth it.)

rhatdan:
The entire tool chain is moving to 1.20, so I see no reason to leave this behind.  1.19 support gos away soon, I believe if it is not already gone.

mtrmac:
/lgtm

Fair enough.

Issue #1787, opened 2023-12-22
why GetConfigHome only work with a 755 config directory


sword-jin:
In https://github.com/containers/storage/blob/main/pkg/homedir/homedir_unix.go#L121, checking `~/.config` only is writeable by owner.

I found my config perm is 775, so it doesn't work.

```
drwxrwxr-x 26 sword sword    4096 Dec 22 02:38 .config
```

I have to change it to `755`, could someone explain that, why need a so strict validation.

rhatdan:
@giuseppe is there a reason that we care if the directory is only writeable by the user?  Shouldn't this check if the directory is owned by the user and is writable?  I don't know why we would care if a group can write there?

Although it is pretty strange permission to have on config in your homedir?

@sword-jin why is your homedir set like this?  Why would groups be able to write to $HOME/.config?

giuseppe:
> @giuseppe is there a reason that we care if the directory is only writeable by the user? Shouldn't this check if the directory is owned by the user and is writable? I don't know why we would care if a group can write there?

yes, we care about that because if a directory is writeable to any user, then a malicious user could write a config file and potentially exploit the target user.  I've no strong objections to relax the check and also allow write mode for the group.  

@mtrmac do you have any concerns?

rhatdan:
We probably should loadly warn, but I think we should wait for @sword-jin to respond, since this is an unusual setup, and just changing the config to chmod 755 would fix it.

Our check is a little bogus since we are not checking all the way down. to the directory.  We should be checking ~/.config/containers as well, and ~/.local/share/containers/* if we were really concerned about a hacker attacking the user via Group or Other write access.

sword-jin:
Thanks for the quick response @rhatdan 
I forgot why my config dir is 775(maybe I use a test account share my main account’s cache).
The error message is too confusing. I guess we could have two options, 1 is provide a better message with a fix command. 2 is skip this error and give user a warning

rhatdan:
Or three remove the check altogether, since it provides minimal protection to the user.

sword-jin:
I would like to remove the check, if the basic config dir is not writable, I think every developer will find it easily.

rhatdan:
SGTM

giuseppe:
Could you open a PR with the change you like to have?

mtrmac:
[...]

PR #1793, opened 2024-01-08
homedir: drop check for ConfigHome mode

giuseppe:
do not check the mode for the config home directory, as it might not be writeable to the current user.

Closes: https://github.com/containers/storage/issues/1787

### DIFF
--- a/pkg/homedir/homedir_unix.go
+++ b/pkg/homedir/homedir_unix.go
@@ -118,10 +118,13 @@ func GetConfigHome() (string, error) {
 			tmpDir := filepath.Join(resolvedHome, ".config")
 			_ = os.MkdirAll(tmpDir, 0o700)
 			st, err := os.Stat(tmpDir)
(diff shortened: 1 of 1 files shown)


pkg/homedir/homedir_unix.go
```diff
@@ -169,7 +172,7 @@ func GetRuntimeDir() (string, error) {
 			if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() && isWriteableOnlyByOwner(st.Mode().Perm()) {
```
rhatdan, line 172: Did  you forget to remove this isWritableOnlyByOwner check?  Or the error message below should not be changed.

pkg/homedir/homedir_unix.go
```diff
@@ -169,7 +172,7 @@ func GetRuntimeDir() (string, error) {
...
 				runtimeDir = tmpDir
 			} else {
-				rootlessRuntimeDirError = fmt.Errorf("path %q exists and it is not writeable only by the current user", tmpDir)
```
rhatdan, line 172: This error then is specific to the case of users homedir, not the case of /tmp.

mtrmac (commented):
/lgtm

Thanks!

PR #1794, opened 2024-01-08
chunked: validate converted images

giuseppe:
validate that the retrieved data for converted images matches the expected digest.

The related change in c/image:

```diff
diff --git a/storage/storage_dest.go b/storage/storage_dest.go
index 0cd4392a..56f16461 100644
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -279,7 +279,7 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
                blobInfo:      srcInfo,
        }
 
-       differ, err := chunked.GetDiffer(ctx, s.imageRef.transport.store, srcInfo.Size, srcInfo.Annotations, &fetcher)
+       differ, err := chunked.GetDiffer(ctx, s.imageRef.transport.store, &srcInfo.Digest, srcInfo.Size, srcInfo.Annotations, &fetcher)
        if err != nil {
                return private.UploadedBlob{}, err
        }
```

### DIFF
--- a/cmd/containers-storage/diff.go
+++ b/cmd/containers-storage/diff.go
@@ -13,6 +13,7 @@ import (
 	"github.com/containers/storage/pkg/chunked/compressor"
 	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/mflag"
+	digest "github.com/opencontainers/go-digest"
 )
 
 var (
@@ -155,7 +156,10 @@ func applyDiffUsingStagingDirectory(flags *mflag.FlagSet, action string, m stora
 		return 1, err
 	}
 
-	if _, err := io.Copy(compressor, tr); err != nil {
+	digesterCompressed := digest.Canonical.Digester()
+	r := io.TeeReader(tr, digesterCompressed.Hash())
+
+	if _, err := io.Copy(compressor, r); err != nil {
 		return 1, err
 	}
 	if err := compressor.Close(); err != nil {
@@ -171,7 +175,7 @@ func applyDiffUsingStagingDirectory(flags *mflag.FlagSet, action string, m stora
 		file: tar,
 	}
 
-	differ, err := chunked.GetDiffer(context.Background(), m, size, metadata, &fetcher)
+	differ, err := chunked.GetDiffer(context.Background(), m, digesterCompressed.Digest(), size, metadata, &fetcher)
 	if err != nil {
 		return 1, err
 	}
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -86,6 +86,11 @@ type chunkedDiffer struct {
 	// the layer are trusted and should not be validated.
 	skipValidation bool
 
+	// blobDigest is the digest of the whole compressed layer.  It is used if
+	// convertToZstdChunked to validate a layer when it is converted since there
+	// is no TOC referenced by the manifest.
+	blobDigest digest.Digest
+
 	blobSize int64
 
 	storeOpts *types.StoreOptions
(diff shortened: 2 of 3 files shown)


cmd/containers-storage/diff.go
```diff
@@ -171,7 +171,7 @@ func applyDiffUsingStagingDirectory(flags *mflag.FlagSet, action string, m stora
 		file: tar,
 	}
 
-	differ, err := chunked.GetDiffer(context.Background(), m, size, metadata, &fetcher)
+	differ, err := chunked.GetDiffer(context.Background(), m, nil, size, metadata, &fetcher)
```
mtrmac, line 174: Non-blocking? This debug-only command could take the trouble to compute the digest, and then the generic code would not have that risky extra condition for “digest is not present”.

pkg/chunked/storage_linux.go
```diff
@@ -86,6 +86,11 @@ type chunkedDiffer struct {
 	// the layer are trusted and should not be validated.
 	skipValidation bool
 
+	// blobDigest is the digest of the compressed layer.  It is required
```
mtrmac, line 89:
```suggestion
	// blobDigest is the digest of the whole compressed .stream .  It is required if convertToZstdChunked
```

pkg/chunked/storage_linux.go
```diff
@@ -1530,6 +1546,11 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 		c.fileType = fileTypeZstdChunked
 		c.manifest = manifest
 		c.tarSplit = tarSplit
+
+		if compressedDigest != *c.blobDigest {
```
mtrmac, line 1550: Check this before doing _anything_ with the file. At the very least immediately after `converTarToZstdChunked`, but … (A)

pkg/chunked/storage_linux.go
```diff
@@ -232,28 +241,30 @@ func convertTarToZstdChunked(destDirectory string, blobSize int64, iss ImageSour
...
 		f.Close()
-		return nil, "", nil, err
+		return nil, "", "", nil, err
 	}
 	is := seekableFile{
 		file: f,
 	}
-	return &is, digester.Digest(), newAnnotations, nil
+
+	compressedDigest := digesterCompressed.Digest()
```
mtrmac, line 260: (A) … I think the whole-layer digest match should be enforced here, at the latest.

---

I would actually argue that it would be preferable to enforce the digest match first, and _then_ only start parsing the tar file. This implementation exposes the tar parser and the zstd:chunked creation code to untrusted potentially-hostile data, unlike ordinary non-chunked layer pulls.

pkg/chunked/storage_linux.go
```diff
@@ -232,28 +241,30 @@ func convertTarToZstdChunked(destDirectory string, blobSize int64, iss ImageSour
...
 		f.Close()
-		return nil, "", nil, err
+		return nil, "", "", nil, err
 	}
 
-	digester := digest.Canonical.Digester()
-	hash := digester.Hash()
-
-	if _, err := io.Copy(io.MultiWriter(chunked, hash), diff); err != nil {
+	digesterUncompressed := digest.Canonical.Digester()
```
mtrmac, line 247: AFAICS now we have `digesterCompressed` (computing a digest of the pre-conversion layer) and `digesterUncompressed` (computing a digest of the converted chunked data`. The naming suggest they correspond to each other, which confuses me a lot. Please use distinct names (`originalRawDigester` / `convertedOutputDigester` or something).

mtrmac (commented):
Thanks!

I think this works; but really it seems to me that the `convert_images` boolean (… if it continues to exist in the production version of composefs; I don’t know if that’s the plan) should be handled around the ordinary `Store.PutLayer`. If nothing else, then it would work with registries which don’t support range requests.

giuseppe:
@mtrmac thanks for the review.

I've addressed your comments and pushed a new version

mtrmac (approved):
/lgtm

Thanks!

giuseppe:
thanks!   The related change in c/image:

- https://github.com/containers/image/pull/2243

PR #1799, opened 2024-01-11
overlay: use the default mode for the root directory

giuseppe:
when pulling a partial image, use the default mode specified for the traditional case too.


flouthoc (approved):
LGTM

nalind:
The regular flow should be defaulting to the creating the root directory for a diff with its permissions set to match those of its parent layer's diff directory, with 0555 being used if it has no parent.  That way, if a layer in an image includes a `./` entry, its permissions won't be overridden by mistake.

giuseppe:
> The regular flow should be defaulting to the creating the root directory for a diff with its permissions set to match those of its parent layer's diff directory, with 0555 being used if it has no parent. That way, if a layer in an image includes a `./` entry, its permissions won't be overridden by mistake.

thanks, I've implemented that behaviour and updated the PR

rhatdan:
@nalind PTANL

PR #1804, opened 2024-01-17
overlay: move the fs-verity measurement to the chunked package

giuseppe:
move the logic to calculate the fs-verity for the files payload to the chunked package, so it is calculated when the files are first created.

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -25,6 +25,7 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/chunked/compressor"
 	"github.com/containers/storage/pkg/chunked/internal"
+	"github.com/containers/storage/pkg/fsverity"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/system"
 	"github.com/containers/storage/types"
@@ -46,6 +47,7 @@ const (
 	chunkedData             = "zstd-chunked-data"
 	chunkedLayerDataKey     = "zstd-chunked-layer-data"
 	tocKey                  = "toc"
+	fsVerityDigestsKey      = "fs-verity-digests"
 
 	fileTypeZstdChunked = iota
 	fileTypeEstargz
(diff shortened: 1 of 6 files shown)


rhatdan:
@alexlarsson PTAL

pkg/chunked/storage_linux.go
```diff
@@ -994,6 +1023,35 @@ func closeDestinationFiles(files chan *destinationFile, errors chan error) {
...
+	err := fsverity.EnableVerity(path, int(roFile.Fd()))
+	if err != nil {
+		if c.useFsVerity == graphdriver.DifferFsVerityRequired {
+			return err
+		}
+
+		// If it is not required, ignore the error if the filesystem does not support it.
+		if errors.Is(err, unix.ENOTSUP) || errors.Is(err, unix.ENOTTY) {
+			return nil
+		}
```
alexlarsson, line 1040: Should we not return err here?

alexlarsson: I guess maybe we shouldn't fail if the error is "already enabled", so we'll get the error in the measure call? If so, maybe add a comment about this?

pkg/chunked/storage_linux.go
```diff
@@ -994,6 +1023,35 @@ func closeDestinationFiles(files chan *destinationFile, errors chan error) {
...
 }
 
+func (c *chunkedDiffer) recordFsVerity(path string, roFile *os.File) error {
+	if c.useFsVerity == graphdriver.DifferFsVerityDisabled {
+		return nil
+	}
+
+	err := fsverity.EnableVerity(path, int(roFile.Fd()))
+	if err != nil {
+		if c.useFsVerity == graphdriver.DifferFsVerityRequired {
```
alexlarsson, line 1033: This should handle EEXIST for the fs-verity already enabled case and make sure to then read and record the existing digest.

pkg/chunked/storage_linux.go
giuseppe: EnableVerity doesn't return an error on EEXIST, I'll add a comment to make it clearer

pkg/chunked/storage_linux.go
giuseppe: `fsverity.EnableVerity` returns nil if fs-verity was already enabled, so we should already get into the `fsverity.MeasureVerity` function even if `DifferFsVerityRequired`

rhatdan:
:GT<

PR #1807, opened 2024-01-23
Store: add ImageDirectory() and ImageRunDirectory()

nalind:
Add ImageDirectory() and ImageRunDirectory(), which return the paths of directories which the caller can use to store image-specific data which will be cleaned up automatically when the image is removed or the system is restarted, respectively.

### DIFF
--- a/store.go
+++ b/store.go
@@ -397,6 +397,18 @@ type Store interface {
 	// allow ImagesByDigest to find images by their correct digests.
 	SetImageBigData(id, key string, data []byte, digestManifest func([]byte) (digest.Digest, error)) error
 
+	// ImageDirectory returns a path of a directory which the caller can
+	// use to store data, specific to the image, which the library does not
+	// directly manage.  The directory will be deleted when the image is
+	// deleted.
+	ImageDirectory(id string) (string, error)
+
+	// ImageRunDirectory returns a path of a directory which the caller can
+	// use to store data, specific to the image, which the library does not
+	// directly manage.  The directory will be deleted when the host system
+	// is restarted.
+	ImageRunDirectory(id string) (string, error)
+
 	// ListLayerBigData retrieves a list of the (possibly large) chunks of
 	// named data associated with a layer.
 	ListLayerBigData(id string) ([]string, error)
@@ -3311,6 +3323,27 @@ func (s *store) ContainerByLayer(id string) (*Container, error) {
 	return nil, ErrContainerUnknown
 }
 
+func (s *store) ImageDirectory(id string) (string, error) {
+	foundImage := false
+	if res, done, err := readAllImageStores(s, func(store roImageStore) (string, bool, error) {
+		if store.Exists(id) {
+			foundImage = true
+		}
+		middleDir := s.graphDriverName + "-images"
+		gipath := filepath.Join(s.GraphRoot(), middleDir, id, "userdata")
+		if err := os.MkdirAll(gipath, 0o700); err != nil {
+			return "", true, err
+		}
+		return gipath, true, nil
+	}); done {
+		return res, err
+	}
+	if foundImage {
+		return "", fmt.Errorf("locating image with ID %q (consider removing the image to resolve the issue): %w", id, os.ErrNotExist)
+	}
+	return "", fmt.Errorf("locating image with ID %q: %w", id, ErrImageUnknown)
+}
+
 func (s *store) ContainerDirectory(id string) (string, error) {
 	res, _, err := readContainerStore(s, func() (string, bool, error) {
 		id, err := s.containerStore.Lookup(id)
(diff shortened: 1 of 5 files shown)


rhatdan:
/approve
LGTM
@giuseppe @saschagrunert @vrothberg @mtrmac PTAL

giuseppe (approved):
/lgtm

store.go
```diff
@@ -3311,6 +3323,27 @@ func (s *store) ContainerByLayer(id string) (*Container, error) {
...
+	foundImage := false
+	if res, done, err := readAllImageStores(s, func(store roImageStore) (string, bool, error) {
+		if store.Exists(id) {
+			foundImage = true
+		}
+		middleDir := s.graphDriverName + "-images"
+		gipath := filepath.Join(s.GraphRoot(), middleDir, id, "userdata")
+		if err := os.MkdirAll(gipath, 0o700); err != nil {
+			return "", true, err
+		}
```
mtrmac, line 3336: I’d expect the responsibility for creating these directories to be in the same object that is responsible for deleting them — so I guess inside `imageStore`, and calling `imageStore.datadir` instead of hard-coding the structure.

(Applies to both.)

store.go
```diff
@@ -397,6 +397,18 @@ type Store interface {
 	// allow ImagesByDigest to find images by their correct digests.
 	SetImageBigData(id, key string, data []byte, digestManifest func([]byte) (digest.Digest, error)) error
 
+	// ImageDirectory returns a path of a directory which the caller can
+	// use to store data, specific to the image, which the library does not
+	// directly manage.  The directory will be deleted when the image is
+	// deleted.
```
mtrmac, line 403: In both cases, a small warning that the user is responsible for dealing with concurrency (including concurrency with image deletion, deleting the stored files in unpredictable order) might be valuable.

store.go
```diff
@@ -3328,6 +3361,28 @@ func (s *store) ContainerDirectory(id string) (string, error) {
...
 
+func (s *store) ImageRunDirectory(id string) (string, error) {
+	foundImage := false
+	if res, done, err := readAllImageStores(s, func(store roImageStore) (string, bool, error) {
+		if store.Exists(id) {
+			foundImage = true
+		}
+
+		middleDir := s.graphDriverName + "-images"
+		rcpath := filepath.Join(s.RunRoot(), middleDir, id, "userdata")
```
mtrmac, line 3372: (Yes the API documents that these directories are not explicitly deleted. A small comment here might help, e.g. for refactoring efforts which don’t naturally end up reading that API documentation.)

store.go
```diff
@@ -3311,6 +3323,27 @@ func (s *store) ContainerByLayer(id string) (*Container, error) {
...
+		middleDir := s.graphDriverName + "-images"
+		gipath := filepath.Join(s.GraphRoot(), middleDir, id, "userdata")
+		if err := os.MkdirAll(gipath, 0o700); err != nil {
+			return "", true, err
+		}
+		return gipath, true, nil
+	}); done {
+		return res, err
+	}
+	if foundImage {
```
mtrmac, line 3341: AFAICS this is dead code; `done` is always set to `true` on the first iteration of `readAllImageStores`.

Also this code seems to read all image stores, but in fact only ever reads the first one and terminates; right now that loop could be dropped entirely.

(Applies to both.)

store.go
```diff
@@ -3311,6 +3323,27 @@ func (s *store) ContainerByLayer(id string) (*Container, error) {
 	return nil, ErrContainerUnknown
 }
 
+func (s *store) ImageDirectory(id string) (string, error) {
+	foundImage := false
+	if res, done, err := readAllImageStores(s, func(store roImageStore) (string, bool, error) {
+		if store.Exists(id) {
+			foundImage = true
+		}
```
mtrmac, line 3331: This is surprising, is this right? Besides `foundImage` being dead, as noted elsewhere, the effect is that for images in an additional store, we create these data directories in the _primary_ store which doesn’t store the image at all.

That _might_ be right (assuming the additional stores are read-only and we simply can’t write them to additional stores — I don’t know what the use case is and whether we need to write this data to read-only images), but then we probably need some new image cleanup code to remove this data in the otherwise-unrelated store when the image is being deleted.

If that was not the intent, I’d expect this to look more like the image part of `store.Metadata`:
```go
if res, done, err := readAllImageStores … {
    if store.Exists(id) {
        res, err := …
        return res, true, err
     }
     return N/A, false, nil
}; done {
   return res, err
}
return "", “not found”
```

(Applies to both.)

mtrmac (commented):
I know this was merged, please glance at the review anyway.

store.go
```diff
@@ -397,6 +397,18 @@ type Store interface {
...
+	// ImageDirectory returns a path of a directory which the caller can
+	// use to store data, specific to the image, which the library does not
+	// directly manage.  The directory will be deleted when the image is
+	// deleted.
+	ImageDirectory(id string) (string, error)
+
+	// ImageRunDirectory returns a path of a directory which the caller can
+	// use to store data, specific to the image, which the library does not
+	// directly manage.  The directory will be deleted when the host system
+	// is restarted.
```
mtrmac, line 409: Do we want to commit to this behavior? Or should this allow the directory to be deleted as soon as the image is deleted, even if it isn’t currently implemented?

PR #1813, opened 2024-01-30
chunked: store UncompressedDigest if validated

giuseppe:
store the UncompressedDigest when the original tarball was converted to zstd:chunked, since its diffID was computed and validated.

In this way the layer can be reused as any other layer that was fully retrieved and validated.

Before this change, a layer that was converted to zstd:chunked was always retrieved since it has not a TOC Digest.


rhatdan:
LGTM
@nalind @mtrmac PTAL

pkg/chunked/storage_linux.go
```diff
@@ -1665,11 +1666,11 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 		c.tarSplit = tarSplit
 
 		// since we retrieved the whole file and it was validated, use the diffID instead of the TOC digest.
-		c.contentDigest = diffID
+		c.contentDigest = ""
```
mtrmac, line 1669: AFAICS `contentDigest` is now only ever a TOC digest. So
- Rename it to  something like `tocDigest`, to be precise
- Update the comment documenting the field to not incorrectly talk about DiffIDs
- Remove the comment just above, AFACIS it has no relationship to the code
- Remove this line entirely? For the `…convert…` case, the field is never set. Or maybe set it explicitly to `""` already in the `makeConvertFromRawDiffer` constructor, to be explicit.

mtrmac (commented):
Makes sense to me.

pkg/chunked/storage_linux.go
giuseppe: thanks, I've amended the suggested changes.

I've added a new variable `tocDigest`.  Hopefully, it is clearer than having to change the state of `chunkedDiffer` when both write and read the value within the same function.

pkg/chunked/storage_linux.go
```diff
@@ -1663,13 +1664,13 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
...
 		c.tarSplit = tarSplit
-
-		// since we retrieved the whole file and it was validated, use the diffID instead of the TOC digest.
-		c.contentDigest = diffID
 		c.tocOffset = tocOffset
 
 		// the file was generated by us and the digest for each file was already computed, no need to validate it again.
 		c.skipValidation = true
+		// since we retrieved the whole file and it was validated, do not use the TOC digest, but set the uncompressed digest.
+		tocDigest = ""
```
mtrmac, line 1672: We actually reliably know the TOC digest, don’t we? Is it deterministic enough that we would want to record it and make it available for reuse when pulling from chunked variants?

Anyway, that’s clearly a separate PR.

mtrmac (approved):
/lgtm

Thanks!

PR #1825, opened 2024-02-09
chunked: report TOCDigest for converted layers

giuseppe:
even if we validated the full layer, report the TOC Digest as well so the upper layer can use both.


giuseppe:
@mtrmac PTAL

mtrmac (commented):
LGTM.

rhatdan:
LGTM

PR #1826, opened 2024-02-12
store: new API ApplyStagedLayer 

giuseppe:
Add a race-condition-free alternative to using CreateLayer and ApplyDiffFromStagingDirectory, ensuring the store is locked for the entire duration while the layer is being created and populated.

The relative patch for c/image is:

```diff
diff --git a/storage/storage_dest.go b/storage/storage_dest.go
index 88e492b7..51b30a64 100644
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -149,7 +149,7 @@ func (s *storageImageDestination) Close() error {
 	}
 	for _, v := range s.diffOutputs {
 		if v.Target != "" {
-			_ = s.imageRef.transport.store.CleanupStagingDirectory(v.Target)
+			_ = s.imageRef.transport.store.CleanupStagedLayer(v)
 		}
 	}
 	return os.RemoveAll(s.directory)
@@ -669,11 +669,6 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 			return false, fmt.Errorf("index %d out of range for configOCI.RootFS.DiffIDs", index)
 		}
 
-		layer, err := s.imageRef.transport.store.CreateLayer(id, parentLayer, nil, "", false, nil)
-		if err != nil {
-			return false, err
-		}
-
 		// let the storage layer know what was the original uncompressed layer.
 		flags := make(map[string]interface{})
 		flags[expectedLayerDiffIDFlag] = configOCI.RootFS.DiffIDs[index]
@@ -682,8 +677,15 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 			Flags: flags,
 		}
 
-		if err := s.imageRef.transport.store.ApplyDiffFromStagingDirectory(layer.ID, diffOutput.Target, diffOutput, options); err != nil {
-			_ = s.imageRef.transport.store.Delete(layer.ID)
+		args := storage.ApplyStagedLayerOptions{
+			ID:          id,
+			ParentLayer: parentLayer,
+
+			DiffOutput:       diffOutput,
+			DiffOptions:      options,
+		}
+		layer, err := s.imageRef.transport.store.ApplyStagedLayer(args)
+		if err != nil {
 			return false, err
 		}

```

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -181,6 +181,13 @@ type DiffOptions struct {
 	Compression *archive.Compression
 }
 
+// stagedLayerOptions are the options passed to .create to populate a staged
+// layer
+type stagedLayerOptions struct {
+	DiffOutput  *drivers.DriverWithDifferOutput
+	DiffOptions *drivers.ApplyDiffWithDifferOpts
+}
+
 // roLayerStore wraps a graph driver, adding the ability to refer to layers by
 // name, and keeping track of parent-child relationships, along with a list of
 // all known layers.
@@ -267,7 +274,7 @@ type rwLayerStore interface {
 	// underlying drivers do not themselves distinguish between writeable
 	// and read-only layers.  Returns the new layer structure and the size of the
 	// diff which was applied to its parent to initialize its contents.
-	create(id string, parent *Layer, names []string, mountLabel string, options map[string]string, moreOptions *LayerOptions, writeable bool, diff io.Reader) (*Layer, int64, error)
+	create(id string, parent *Layer, names []string, mountLabel string, options map[string]string, moreOptions *LayerOptions, writeable bool, diff io.Reader, slo *stagedLayerOptions) (*Layer, int64, error)
 
 	// updateNames modifies names associated with a layer based on (op, names).
 	updateNames(id string, names []string, op updateNameOperation) error
@@ -1232,7 +1239,7 @@ func (r *layerStore) PutAdditionalLayer(id string, parentLayer *Layer, names []s
 }
 
 // Requires startWriting.
-func (r *layerStore) create(id string, parentLayer *Layer, names []string, mountLabel string, options map[string]string, moreOptions *LayerOptions, writeable bool, diff io.Reader) (layer *Layer, size int64, err error) {
+func (r *layerStore) create(id string, parentLayer *Layer, names []string, mountLabel string, options map[string]string, moreOptions *LayerOptions, writeable bool, diff io.Reader, slo *stagedLayerOptions) (layer *Layer, size int64, err error) {
 	if moreOptions == nil {
 		moreOptions = &LayerOptions{}
 	}
@@ -1424,6 +1431,11 @@ func (r *layerStore) create(id string, parentLayer *Layer, names []string, mount
 	if diff != nil {
 		if size, err = r.applyDiffWithOptions(layer.ID, moreOptions, diff); err != nil {
 			cleanupFailureContext = "applying layer diff"
+			return nil, -1, err
+		}
+	} else if slo != nil {
+		if err := r.applyDiffFromStagingDirectory(layer.ID, slo.DiffOutput, slo.DiffOptions); err != nil {
+			cleanupFailureContext = "applying staged directory diff"
 			return nil, -1, err
 		}
 	} else {
--- a/layers.go
+++ b/layers.go
@@ -312,8 +312,8 @@ type rwLayerStore interface {
 	// CleanupStagingDirectory cleanups the staging directory.  It can be used to cleanup the staging directory on errors
 	CleanupStagingDirectory(stagingDirectory string) error
 
-	// ApplyDiffFromStagingDirectory uses stagingDirectory to create the diff.
-	ApplyDiffFromStagingDirectory(id, stagingDirectory string, diffOutput *drivers.DriverWithDifferOutput, options *drivers.ApplyDiffWithDifferOpts) error
+	// applyDiffFromStagingDirectory uses diffOutput.Target to create the diff.
+	applyDiffFromStagingDirectory(id string, diffOutput *drivers.DriverWithDifferOutput, options *drivers.ApplyDiffWithDifferOpts) error
 
 	// DifferTarget gets the location where files are stored for the layer.
 	DifferTarget(id string) (string, error)
@@ -2414,7 +2414,7 @@ func (r *layerStore) DifferTarget(id string) (string, error) {
 }
 
 // Requires startWriting.
-func (r *layerStore) ApplyDiffFromStagingDirectory(id, stagingDirectory string, diffOutput *drivers.DriverWithDifferOutput, options *drivers.ApplyDiffWithDifferOpts) error {
+func (r *layerStore) applyDiffFromStagingDirectory(id string, diffOutput *drivers.DriverWithDifferOutput, options *drivers.ApplyDiffWithDifferOpts) error {
 	ddriver, ok := r.driver.(drivers.DriverWithDiffer)
 	if !ok {
 		return ErrNotSupported
@@ -2433,7 +2433,7 @@ func (r *layerStore) ApplyDiffFromStagingDirectory(id, stagingDirectory string, 
 		}
 	}
 
-	err := ddriver.ApplyDiffFromStagingDirectory(layer.ID, layer.Parent, stagingDirectory, diffOutput, options)
+	err := ddriver.ApplyDiffFromStagingDirectory(layer.ID, layer.Parent, diffOutput, options)
 	if err != nil {
 		return err
 	}
--- a/store.go
+++ b/store.go
@@ -71,6 +71,19 @@ type metadataStore interface {
 	rwMetadataStore
 }
 
+// ApplyStagedLayerOptions contains options to pass to ApplyStagedLayer
+type ApplyStagedLayerOptions struct {
+	ID           string        // Mandatory
+	ParentLayer  string        // Optional
+	Names        []string      // Optional
+	MountLabel   string        // Optional
+	Writeable    bool          // Optional
+	LayerOptions *LayerOptions // Optional
+
+	DiffOutput  *drivers.DriverWithDifferOutput  // Mandatory
+	DiffOptions *drivers.ApplyDiffWithDifferOpts // Mandatory
+}
+
 // An roBigDataStore wraps up the read-only big-data related methods of the
 // various types of file-based lookaside stores that we implement.
 type roBigDataStore interface {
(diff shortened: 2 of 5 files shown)


store.go
```diff
@@ -71,6 +71,20 @@ type metadataStore interface {
...
+// PutLayerFromStagingDirectoryOptions contains options to pass to PutLayerFromStagingDirectory
+type PutLayerFromStagingDirectoryOptions struct {
+	ID           string
+	ParentLayer  string
+	Names        []string
+	MountLabel   string
+	Writeable    bool
+	LayerOptions *LayerOptions
+
+	StagingDirectory string
```
mtrmac, line 83: Would `StagingDirectory` ever differ from `DiffOutput.Target`? If not, can this be handled internally without callers having to care (and without callers possibly breaking this)?

OTOH `CleanupStagingDirectory` forces c/image to know about `.Target` anyway…

store.go
```diff
@@ -2953,6 +2977,32 @@ func (s *store) ApplyDiffFromStagingDirectory(to, stagingDirectory string, diffO
...
+	}
+	defer s.containerStore.stopWriting()
+
+	layer, _, err := s.putLayerLocked(rlstore, rlstores, args.ID, args.ParentLayer, args.Names, args.MountLabel, args.Writeable, args.LayerOptions, nil)
+	if err != nil {
+		return layer, err
+	}
+
+	if err := rlstore.ApplyDiffFromStagingDirectory(layer.ID, args.StagingDirectory, args.DiffOutput, args.DiffOptions); err != nil {
+		rlstore.Delete(layer.ID)
```
mtrmac, line 3000: Please either return both errors or add a warning-level log.

Otherwise, logs from production where some layers are missing for an unknown reason are very difficult to debug. (Sure, ideally the caller should provide a unique ID that is not reused in the meantime… but if we are looking at logs, there is a bug involved. And, yes, c/image didn’t have that.)

store.go
```diff
@@ -320,6 +334,11 @@ type Store interface {
 	// ApplyDiffFromStagingDirectory uses stagingDirectory to create the diff.
 	ApplyDiffFromStagingDirectory(to, stagingDirectory string, diffOutput *drivers.DriverWithDifferOutput, options *drivers.ApplyDiffWithDifferOpts) error
```
mtrmac, line 335: Should this be deprecated? I don’t have a strong opinion.

`PutLayer` exists and `CreateLayer` isn’t deprecated, that is one argument against deprecating.

store.go
```diff
@@ -71,6 +71,20 @@ type metadataStore interface {
 	rwMetadataStore
 }
 
+// PutLayerFromStagingDirectoryOptions contains options to pass to PutLayerFromStagingDirectory
```
mtrmac, line 74: (Documenting which of the options are mandatory and which are optional would be nice.)

mtrmac:
Thanks for working on this!

store.go
giuseppe: yes good idea, let's deprecate it and remove once c/image uses the new API

giuseppe:
@mtrmac thanks for the review.

I've addressed your comments, except one pending question: https://github.com/containers/storage/pull/1826#discussion_r1486741274

mtrmac (commented):
Changing the API, after looking a bit more, seems to to require a bunch of fairly intrusive changes all at once… so I don’t think it is worth blocking this PR on them.

Or maybe you can see some way to make all of that simple.

rhatdan:
LGTM

store.go
```diff
@@ -318,11 +331,21 @@ type Store interface {
 	ApplyDiffWithDiffer(to string, options *drivers.ApplyDiffWithDifferOpts, differ drivers.Differ) (*drivers.DriverWithDifferOutput, error)
 
 	// ApplyDiffFromStagingDirectory uses stagingDirectory to create the diff.
+	// Deprecated: it will be removed soon.
```
mtrmac, line 334:
```suggestion
	// Deprecated: it will be removed soon. Use CleanupStagingDirectory instead.
```
... or however it is going to be called.

store.go
```diff
@@ -318,11 +331,21 @@ type Store interface {
...
 	ApplyDiffFromStagingDirectory(to, stagingDirectory string, diffOutput *drivers.DriverWithDifferOutput, options *drivers.ApplyDiffWithDifferOpts) error
 
 	// CleanupStagingDirectory cleanups the staging directory.  It can be used to cleanup the staging directory on errors
+	// Deprecated: it will be removed soon.
 	CleanupStagingDirectory(stagingDirectory string) error
```
mtrmac, line 339: Non-blocking: Maybe `CleanupStagedLayer`?

mtrmac (changes requested):
On second thought, I’m afraid this isn’t sufficient.

It is fine for a running system: it prevents other processes from observing the WIP layer.

But it doesn’t handle crashes sufficiently. For that, the layer metadata needs to be saved to disk with `incompleteFlag` (so that if we are recovering from a crash, we delete everything), and after the contents are set up, the flag is removed again.

(Or, hypothetically, we could first write the on-disk contents and only afterwards write the layer metadata?? But that would be a new unproven code path, and we would have to worry about re-creating a layer on top of previously partially-created files. Seems risky, when the other path is well-understood.)

Very roughly speaking, I think this can be done by applying the staged data from inside `layerStore.create`, around the place where `applyDiffWithOptions` is called for non-chunked layers.

store.go
```diff
@@ -318,11 +331,21 @@ type Store interface {
 	ApplyDiffWithDiffer(to string, options *drivers.ApplyDiffWithDifferOpts, differ drivers.Differ) (*drivers.DriverWithDifferOutput, error)
 
 	// ApplyDiffFromStagingDirectory uses stagingDirectory to create the diff.
+	// Deprecated: it will be removed soon.
 	ApplyDiffFromStagingDirectory(to, stagingDirectory string, diffOutput *drivers.DriverWithDifferOutput, options *drivers.ApplyDiffWithDifferOpts) error
 
 	// CleanupStagingDirectory cleanups the staging directory.  It can be used to cleanup the staging directory on errors
+	// Deprecated: it will be removed soon.
```
mtrmac, line 338:
```suggestion
	// Deprecated: it will be removed soon. Use CleanupStagingDirectory instead.
```
... or however it is going to be called.

store.go
mtrmac: “`Use ApplyStagedLayer`…” here, my mistake.

mtrmac:
The API design LGTM.

For the record:
> The relative patch for c/image is:
> ```diff
> diff --git a/storage/storage_dest.go b/storage/storage_dest.go
> index 88e492b7..51b30a64 100644
> --- a/storage/storage_dest.go
> +++ b/storage/storage_dest.go
> 
> +		layer, err := s.imageRef.transport.store.ApplyStagedLayer(args)
> +		if err != nil {
This path in c/image also needs to check for, and succeed with, `ErrDuplicateID`. And it would be convenient, and consistent with `PutLayer`, for the new function to return a `layer` value with the duplicate object in this case.

giuseppe:
> On second thought, I’m afraid this isn’t sufficient.
> 
> It is fine for a running system: it prevents other processes from observing the WIP layer.
> 
> But it doesn’t handle crashes sufficiently. For that, the layer metadata needs to be saved to disk with `incompleteFlag` (so that if we are recovering from a crash, we delete everything), and after the contents are set up, the flag is removed again.
> 
> (Or, hypothetically, we could first write the on-disk contents and only afterwards write the layer metadata?? But that would be a new unproven code path, and we would have to worry about re-creating a layer on top of previously partially-created files. Seems risky, when the other path is well-understood.)
> 
> Very roughly speaking, I think this can be done by applying the staged data from inside `layerStore.create`, around the place where `applyDiffWithOptions` is called for non-chunked layers.

you are right, we need to use the incompleteFlag.  I've pushed a new version where I set the incompleteFlag, barely tested.  I'll work more on it through the day

giuseppe:
@mtrmac what do you think of the last version?

mtrmac (commented):
I’d rather prefer if the `incompleteFlag` remained an internal implementation detail of `layers.go`.

giuseppe:
> I’d rather prefer if the `incompleteFlag` remained an internal implementation detail of `layers.go`.

moved the `incompleteFlag` usage to `layers.go`

layers.go
```diff
@@ -1442,10 +1442,12 @@ func (r *layerStore) create(id string, parentLayer *Layer, names []string, mount
 		}
 	}
 
-	delete(layer.Flags, incompleteFlag)
-	if err = r.saveFor(layer); err != nil {
-		cleanupFailureContext = "saving finished layer metadata"
-		return nil, -1, err
+	if !partial {
```
mtrmac, line 1445: If `partial`, the code below `applyDiffWithOptions() would have updated r.bycompressedsum` modifies `layer` data and we wouldn’t save it here.

I think that might be correct enough right now (because those fields aren’t set) but it’s hard for me to track such indirect effects — the code might modify more fields in the future.

mtrmac (commented):
This is a half-way step, but `the `partial` option is still basically “leave the layer incomplete” and/or “I promise to call `applyDiffFromStagingDirectory` later” (with neither effect documented), with `store.go` being responsible for that.

It seems to me
```patch
	if diff != nil {
		if size, err = r.applyDiffWithOptions …
+	else if staged != nil {
+		if size, err = r.applyDiffFromStagingDirectory …
	else { …
```

should be possible and and not _too_ invasive a change.

giuseppe:
pushed a new version, I moved the `applyDiffFromStagingDirectory()` call inside `create()`

layers.go
```diff
@@ -2482,6 +2494,8 @@ func (r *layerStore) ApplyDiffFromStagingDirectory(id, stagingDirectory string,
 			return err
 		}
 	}
+
+	delete(layer.Flags, incompleteFlag)
```
mtrmac, line 2498: This can now be dropped.

store.go
```diff
@@ -1421,20 +1444,8 @@ func (s *store) canUseShifting(uidmap, gidmap []idtools.IDMap) bool {
 	return true
 }
 
-func (s *store) PutLayer(id, parent string, names []string, mountLabel string, writeable bool, lOptions *LayerOptions, diff io.Reader) (*Layer, int64, error) {
+func (s *store) putLayerLocked(rlstore rwLayerStore, rlstores []roLayerStore, id, parent string, names []string, mountLabel string, writeable bool, lOptions *LayerOptions, diff io.Reader, slo *stagedLayerOptions) (*Layer, int64, error) {
```
mtrmac, line 1447: Absolutely non-blocking: This could be `putLayer(… diff, slo)` which also does the locking; then `PutLayer` and `ApplyStagedLayer` would be very trivial wrappers. This works fine as it is.

mtrmac (commented):
ACK.

store.go
giuseppe: that is better, amended in the last version

mtrmac (commented):
/lgtm

Thanks!

giuseppe:
@mtrmac the related change in c/image: https://github.com/containers/image/pull/2301

Issue #1827, opened 2024-02-12
additionalimagestores Not Working as Expected


kasualkeef:
### Issue Description

I have been able to successfully follow the guide for setting up a read-only image store as described here: 
https://www.redhat.com/sysadmin/image-stores-podman

Unfortunately, the store does not seem read-only at all. If the user does not have **write** permissions to the files on the image store, he is unable to build images using a store image as the FROM base.

We are running rootless Podman on RHEL 8 with all of the latest patches from the RHEL repos.

### Steps to reproduce the issue

Steps to reproduce the issue
1. Pull down images with the `--root` flag as described [here](https://www.redhat.com/sysadmin/image-stores-podman).
    - This must be done as a rootless user or the in-container permissions will be completely broken
2. chmod the files so that other users can read the files
    - We have the `force_mask='shared'` storage.conf setting but this still doesn't seem to affect the file mode on the pulled files (they're still 0700)
3. At this stage, another rootless user on the system can point `additionalimagestores` to the image store directory and do simple things like view the images and run a container instance. However, image builds do not work correctly unless the rootless user has full write permissions to the image store.

        ## Pull Image into RO Store
        [rootless.user@rhel8dev01 ~]$ podman pull --root /data/image-cache/ harbor.domain.com/dod-ironbank/ironbank/redhat/ubi/ubi8
        Trying to pull harbor.domain.com/dod-ironbank/ironbank/redhat/ubi/ubi8:latest...
        Getting image source signatures
        Copying blob 1ea03711adea done  
        Copying blob e5dd65eaa632 done  
        Copying config b7c4db37ab done  
        Writing manifest to image destination
        b7c4db37ab5f29a75921f41120161a998149ac52f4605b3acd46325c08f903ba
        [rootless.user@rhel8dev01 ~]$ ls -l /data/image-cache/
        total 8
        drwx------. 2 rootless.user admins   27 Feb 12 07:24 libpod
        drwx------. 5 rootless.user admins  185 Feb 12 07:24 overlay
        drwx------. 2 rootless.user admins   29 Feb 12 07:24 overlay-containers
        drwx------. 3 rootless.user admins  116 Feb 12 07:24 overlay-images
        drwx------. 2 rootless.user admins 4096 Feb 12 07:24 overlay-layers
        -rw-r--r--. 1 rootless.user admins   64 Feb 12 07:24 storage.lock
        -rw-r--r--. 1 rootless.user admins    0 Feb 12 07:24 userns.lock
[...]

PR #1828, opened 2024-02-13
overlay: ignore chown errors in additionalimagestore

giuseppe:
the store might be owned by another user.

Closes: https://github.com/containers/storage/issues/1827

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1700,10 +1700,6 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	}
 
 	mergedDir := path.Join(workDirBase, "merged")
-	// Create the driver merged dir
-	if err := idtools.MkdirAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
-		return "", err
-	}
 	if count := d.ctr.Increment(mergedDir); count > 1 {
 		return mergedDir, nil
 	}
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1867,7 +1867,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 
 // Put unmounts the mount path created for the give id.
 func (d *Driver) Put(id string) error {
-	dir := d.dir(id)
+	dir, _, inAdditionalStore := d.dir2(id)
 	if _, err := os.Stat(dir); err != nil {
 		return err
 	}
@@ -1928,11 +1928,27 @@ func (d *Driver) Put(id string) error {
 		}
 	}
 
-	if err := unix.Rmdir(mountpoint); err != nil && !os.IsNotExist(err) {
-		logrus.Debugf("Failed to remove mountpoint %s overlay: %s - %v", id, mountpoint, err)
-		return fmt.Errorf("removing mount point %q: %w", mountpoint, err)
-	}
+	if !inAdditionalStore {
+		uid, gid := int(0), int(0)
+		fi, err := os.Stat(mountpoint)
+		if err != nil {
+			return err
+		}
+		if stat, ok := fi.Sys().(*syscall.Stat_t); ok {
+			uid, gid = int(stat.Uid), int(stat.Gid)
+		}
 
+		tmpMountpoint := path.Join(dir, "merged.1")
+		if err := idtools.MkdirAs(tmpMountpoint, 0o700, uid, gid); err != nil && !errors.Is(err, os.ErrExist) {
+			return err
+		}
+		// rename(2) can be used on an empty directory, as it is the mountpoint after umount, and it retains
+		// its atomic semantic.  In this way the "merged" directory is never removed.
+		if err := unix.Rename(tmpMountpoint, mountpoint); err != nil {
+			logrus.Debugf("Failed to replace mountpoint %s overlay: %s - %v", id, mountpoint, err)
+			return fmt.Errorf("replacing mount point %q: %w", mountpoint, err)
+		}
+	}
 	return nil
 }
 
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1696,7 +1696,13 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	}
 
 	if err := idtools.MkdirAllAs(diffDir, perms, rootUID, rootGID); err != nil {
-		return "", err
+		if !inAdditionalStore {
+			return "", err
+		}
+		// if it is in an additional store, do not fail if the directory already exists
+		if _, err2 := os.Stat(diffDir); err2 != nil {
+			return "", err
+		}
 	}
 
 	mergedDir := path.Join(workDirBase, "merged")


mtrmac (commented):
This would ignore all kinds of errors; is there a way to restrict this to chown errors only?

`if !inAdditionalStore { MkdirAllAs } else { MkdirAll }` seems to be an option. I don’t immediately know about the `Rmdir` one.

---

Or, really, if the store is supposed to be read-only, then why change it at all? Test that the directories exist, and fail if they don’t.

But if the `merged` directory is supposed to be created on a `Get` and always deleted on a `Put`, I can’t see how that can possibly work as truly read-only additional image store: a “legitimate” image store won’t have the `merged` directory, and we won’t be able to create it. Am I missing something?

giuseppe:
> This would ignore all kinds of errors; is there a way to restrict this to chown errors only?
> 
> `if !inAdditionalStore { MkdirAllAs } else { MkdirAll }` seems to be an option. I don’t immediately know about the `Rmdir` one.
> 
> Or, really, if the store is supposed to be read-only, then why change it at all? Test that the directories exist, and fail if they don’t.
> 
> But if the `merged` directory is supposed to be created on a `Get` and always deleted on a `Put`, I can’t see how that can possibly work as truly read-only additional image store: a “legitimate” image store won’t have the `merged` directory, and we won’t be able to create it. Am I missing something?

the correct test would be to check whether the directory has the `ContainersOverrideXattr` xattr set and assume fuse-overlayfs is used, in this case the mode specified in the xattr will be used, no matter what is the mode on the directory itself.  But do we care to make it so specific and not just try in any case?

If you prefer, I can rewrite it as:

```
if !inAdditionalStore { MkdirAllAs } else { // do nothing, just assume it exists }
```

are you ok with it?

mtrmac:
(Warning: I don’t understand the full feature, I’m going just by the bug report and the PR.)

Yes, the talk about “read-only additional image store” would lead to basically what you propose. I don’t feel too strongly about the “test that the directories exist” part. But then I saw the `Remove` of `merged` — and I don’t understand how that could work. I assume it does actually work for you — so I’m missing something trivial.

giuseppe:
> Yes, the talk about “read-only additional image store” would lead to basically what you propose. I don’t feel too strongly about the “test that the directories exist” part. But then I saw the `Remove` of `merged` — and I don’t understand how that could work. I assume it does actually work for you — so I’m missing something trivial.

I am not even sure why we need to rm the mount directory, we could just drop that code IMO, so we don't have to worry whether the store is read-only or writeable

mtrmac:
Git blame points at https://github.com/containers/storage/pull/390 .

And I’m still interested in understanding how this _works_ in the first place, in https://github.com/containers/storage/pull/1828#pullrequestreview-1878708827 . (Admittedly I didn’t try that in practice.)

giuseppe:
> And I’m still interested in understanding how this _works_ in the first place, in [#1828 (review)](https://github.com/containers/storage/pull/1828#pullrequestreview-1878708827) . (Admittedly I didn’t try that in practice.)

the question is: should we fail when the directory in the additional image store (which is supposed to be R/O) doesn't exist but the user has enough privileges to create it?  Or should we try to create it anyway?

```go
	if err := idtools.MkdirAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
		if !inAdditionalStore {
			return "", err
		}
		// if it is in an additional store, do not fail if the directory already exists
		if _, err2 := os.Stat(mergedDir); err2 != nil {
			return "", err
		}
	}
```

mtrmac:
Oh I don’t have a strong opinion on that design question.

I’m stuck at an earlier phase: If the user _doesn’t have enough privileges_, I understand `mergedDir` should not exist most of the time. So how does this work when we can’t create it, and it doesn’t exist? … and if it doesn’t, the primary goal of this PR can’t be achieved at all AFAICT.

giuseppe:
> I understand `mergedDir` should not exist most of the time.

if you just perform a "pull" then the directory exists, at least until you try to mount/unmount it.

Maybe we can replace the `rmdir` with a `rename("merged.tmp", "merged")`?  It is an atomic operation with empty directories

mtrmac:
> if you just perform a "pull" then the directory exists, at least until you try to mount/unmount it.

OK I’ll need to look into the situation in more detail, then.

---

> Maybe we can replace the `rmdir` with a `rename`

Does that retain the originally-wanted benefit of handling leaked mounts?

giuseppe:
> > ``
> 
> Does that retain the originally-wanted benefit of handling leaked mounts?

I think it does, although after started looking into it, I am not sure we need to overcomplicate the code for such a corner case where an image is mounted/unmounted on an additional image store

mtrmac:
\*shrug\* the original thing smells like a workaround, and it seems no-one cared enough to document in detail what it is working around (or maybe it is not known at all).

The trouble is that when creating the image, we don’t know that the store is going to later be used as an additional one.

---

One entirely different possibility might be to create the mounts somewhere else, not in the read-only additional store. That’s rather more disruptive.

But really I need to understand what happens on pull, first. Hopefully tomorrow. For now, I’m just confused.

giuseppe:
I've replaced the `Rmdir` with a `Rename` so we can safely assume the `merged` directory is never removed.

Differently the `diffDir` errors are ignored, it might have the wrong ownership but should not really matter with additional image stores since we are already expecting the files in the additional image store owned by a different user to have different ownership.

drivers/overlay/overlay.go
```diff
@@ -1928,11 +1930,27 @@ func (d *Driver) Put(id string) error {
...
 
+		tmpMountpoint := path.Join(dir, "merged.1")
+		if err := idtools.MkdirAs(tmpMountpoint, 0o700, uid, gid); err != nil && !os.IsExist(err) {
+			return err
+		}
+		// rename(2) can be used on an empty directory, as it is the mountpoint after umount, and it retains
+		// its atomic semantic.  In this way the "merged" directory is never removed.
+		if err := unix.Rename(tmpMountpoint, mountpoint); err != nil {
+			logrus.Debugf("Failed to replace mountpoint %s overlay: %s - %v", id, mountpoint, err)
+			return fmt.Errorf("replacingmount point %q: %w", mountpoint, err)
```
rhatdan, line 1951: nit: "replacing mount point %q: %w"

drivers/overlay/overlay.go
```diff
@@ -1928,11 +1930,27 @@ func (d *Driver) Put(id string) error {
...
+		fi, err := os.Stat(mountpoint)
+		if err != nil {
+			return err
+		}
+		if stat, ok := fi.Sys().(*syscall.Stat_t); ok {
+			uid, gid = int(stat.Uid), int(stat.Gid)
+		}
 
+		tmpMountpoint := path.Join(dir, "merged.1")
+		if err := idtools.MkdirAs(tmpMountpoint, 0o700, uid, gid); err != nil && !os.IsExist(err) {
```
rhatdan, line 1944: sed 's/os.IsExists(err)/errors.Is(err, os.ErrNotExist)/g'

giuseppe:
comments addressed :arrow_up:

rhatdan:
LGTM
@mtrmac @nalind @saschagrunert @vrothberg PTAL

TomSweeneyRedHat:
LGTM
but definitely want a @mtrmac  or @nalind head nod

giuseppe:
there are two reviews, can we get this merged?

TomSweeneyRedHat:
Addresses: https://issues.redhat.com/browse/RHEL-32589

PR #1830, opened 2024-02-13
Extend PutLayer to optimize reusing data from existing layers

mtrmac:
- If the caller does not provide digests (e.g. when reusing a layer found by `LayersByTOCDigest`), only digest uncompressed input once, not twice
- Let the caller provide a `CompressedSize` value to match `CompressedDigest`, so that we don't create layers with incorrect metadata.

Motivated by https://github.com/containers/image/pull/2288 .

Cc: @giuseppe (fairly low priority)


giuseppe (approved):
LGTM

PR #1833, opened 2024-02-15
storage: enable partial images by default

giuseppe:
by default enable pulling a partial image, it is still possible to disable the feature through the configuration file.

@rhatdan @mtrmac

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -84,7 +84,7 @@ The `storage.options` table supports the following options:
 **additionalimagestores**=[]
   Paths to additional container image stores. Usually these are read/only and stored on remote network shares.
 
-**pull_options** = {enable_partial_images = "false", use_hard_links = "false", ostree_repos=""}
+**pull_options** = {enable_partial_images = "true", use_hard_links = "false", ostree_repos=""}
 
 Allows specification of how storage is populated when pulling images. This
 option can speed the pulling process of images compressed with format zstd:chunked. Containers/storage looks
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -1701,7 +1701,7 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 		UncompressedDigest: uncompressedDigest,
 	}
 
-	if !parseBooleanPullOption(c.storeOpts, "enable_partial_images", false) {
+	if !parseBooleanPullOption(c.storeOpts, "enable_partial_images", true) {
 		return output, errors.New("enable_partial_images not configured")
 	}
 
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -241,6 +241,10 @@ func GetDiffer(ctx context.Context, store storage.Store, blobDigest digest.Diges
 		return nil, err
 	}
 
+	if !parseBooleanPullOption(&storeOpts, "enable_partial_images", true) {
+		return nil, errors.New("enable_partial_images not configured")
+	}
+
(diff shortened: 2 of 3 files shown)


mtrmac (commented):
Code LGTM; but doing this is ~blocked on #1826 and c/image using that API, otherwise power loss can lead to broken stores.

---

As discussed in chat, having the check for `enable_partial_images` at the very first entry point (`chunked.GetDiffer`) would isolate more of the risk.

mtrmac:
FWIW tests are failing

giuseppe:
fixed now

pkg/chunked/storage_linux.go
```diff
@@ -259,10 +263,6 @@ func GetDiffer(ctx context.Context, store storage.Store, blobDigest digest.Diges
 }
 
 func makeConvertFromRawDiffer(ctx context.Context, store storage.Store, blobDigest digest.Digest, blobSize int64, annotations map[string]string, iss ImageSourceSeekable, storeOpts *types.StoreOptions) (*chunkedDiffer, error) {
-	if !parseBooleanPullOption(storeOpts, "convert_images", false) {
```
mtrmac, line 262: Was this removal intentional?

giuseppe: no sorry, that was a mistake.  Re-added

mtrmac (commented):
Code LGTM again… still #1826.

docs/containers-storage.conf.5.md
```diff
@@ -96,7 +96,7 @@ storage of content and can allow the kernel to use less memory when running
 containers.
 
 containers/storage supports three keys
-  * enable_partial_images="true" | "false"
```
rhatdan, line 99: This looks wrong options is true or falsenot true or true.

docs/containers-storage.conf.5.md
```diff
@@ -96,7 +96,7 @@ storage of content and can allow the kernel to use less memory when running
...
+  * enable_partial_images="true" | "true"
```
rhatdan, line 99: This looks wrong options is true or falsenot true or true.

giuseppe: fixed

giuseppe:
> Code LGTM again… still #1826.

unblocked now

mtrmac (commented):
/lgtm

cgwalters:
Likely fallout in https://github.com/containers/bootc/issues/509#issuecomment-2156630865

cgwalters:
Thanks for following up. Still to this day, we haven't really dug into zstd:chunked images and ostree because https://github.com/containers/bootc/issues/20 means it's basically not useful to create them right now - in theory it should be pullable (and most of the time was), but without the actual delta effect.

> But the total sequence of operations is not trivially clear from that report, and I’m naively thinking that if we broke podman pull && podman push, 

Right but remember "podman pull" is exercising the *combination* of c/image and c/storage which are wired together for zstd:chunked.

What ostree is doing is more like `skopeo copy docker://quay.io/some-zstd-chunked-image oci:ocidir` - and actually I guess not even that because that path goes through the whole "copy machinery" where c/image is driving things instead of being pull based. But it's at least a lot closer to "just fetch the blobs for an zstd:chunked images".

cgwalters:
I guess we could try stress testing this path with just raw code talking to the proxy and calling GetBlob (https://github.com/containers/containers-image-proxy-rs/blob/main/examples/client.rs is a trivial client that doesn't involve anything ostree, etc.)

mtrmac:
Hum… `skopeo copy docker://… non-c/storage:…` would not trigger the chunked path at all. That _is_ less likely to be exercised in existing tests, but it’s also simple enough in implementation that I don’t see anything that could obviously be wrong when creating +pushing, or consuming zstd:chunked layers on registries.

PR #1837, opened 2024-02-19
pkg/archive: avoid user lookups when generating tarball

giuseppe:
This change introduces a new approach to generating tar headers in a way that avoids system-dependent lookups and potential calls to glibc, enhancing portability and security.

The same logic is used by Moby, and the override function is based on code from Moby.

Closes: https://github.com/containers/storage/issues/1836


rhatdan:
LGTM
although CI/CD system is broken right now.

giuseppe:
> LGTM although CI/CD system is broken right now.

green now

pkg/archive/archive.go
```diff
@@ -339,12 +339,43 @@ func (compression *Compression) Extension() string {
...
+	if sys, ok := fi.FileInfo.Sys().(*tar.Header); ok {
+		return sys
+	}
+	return nil
+}
+
+// sysStatOverride, if non-nil, populates hdr from system-dependent fields of fi.
+var sysStatOverride func(fi os.FileInfo, hdr *tar.Header) error
+
+func FileInfoHeaderNoLookups(fi os.FileInfo, link string) (*tar.Header, error) {
```
nalind, line 362: Does this need to be exported?

giuseppe: unexported and pushed a new version.

PR #1843, opened 2024-02-26
chunked: add chunk size to cache file

giuseppe:
include the chunk length in the generated  file location format, This enhancement is designed to facilitate the use of the cache by external tools which may not have knowledge of the chunk size.


rhatdan (approved):
LGTM

PR #1845, opened 2024-02-27
chunked: preserve the original value for symlinks

giuseppe:
the symlinks must preserve their original value without any alteration.


dougsland (commented):
LGTM

PR #1846, opened 2024-02-28
chunked: improve function to merge chunks

giuseppe:
improve the function that combines neighbor chunks.  Instead of using the number of parts, which also includes local files, use only the number of chunks that must be retrieved from the network.

In addition, introduce a threshold limit to merge chunks so that we further reduce the number of requested ranges.

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -41,6 +41,7 @@ import (
 
 const (
 	maxNumberMissingChunks  = 1024
+	autoMergePartsThreshold = 128 // if the gap between two ranges is below this threshold, automatically merge them.
 	newFileFlags            = (unix.O_CREAT | unix.O_TRUNC | unix.O_EXCL | unix.O_WRONLY)
 	containersOverrideXattr = "user.containers.override_stat"
 	bigDataKey              = "zstd-chunked-manifest"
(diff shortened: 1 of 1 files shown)


giuseppe:
@mtrmac @rhatdan PTAL

rhatdan:
LGTM
@alexlarsson PTAL

rhatdan:
@mtrmac @nalind @saschagrunert PTAL

saschagrunert (approved):

pkg/chunked/storage_linux.go
```diff
@@ -1184,18 +1185,8 @@ func mergeMissingChunks(missingParts []missingPart, target int) []missingPart {
 		prev := missingParts[i-1].SourceChunk.Offset + missingParts[i-1].SourceChunk.Length
 		return int(missingParts[i].SourceChunk.Offset - prev)
```
mtrmac, line 1186: Unrelated, why it OK to down-cast to int(32) instead of keeping all offsets (u)int64?

(Also there seems to be only one caller now, with a `== 0`, so maybe we don’t need the subtraction in the first place.

Or, alternatively, use this also when computing the new `cost`, as `getGap(from, to)`)

pkg/chunked/storage_linux.go
```diff
@@ -1281,7 +1295,14 @@ func (c *chunkedDiffer) retrieveMissingFiles(stream ImageSourceSeekable, dest st
 		}
 
 		if _, ok := err.(ErrBadRequest); ok {
-			requested := len(missingParts)
+			requested := 0
```
mtrmac, line 1298: Isn’t this just `len(chunksToRequest)`?

mtrmac (commented):
Locally LGTM (but I forgot all about how this works, if I ever knew, and I didn’t study what `OriginFile`/`Hole`/`Chunks`/… mean now.)

pkg/chunked/storage_linux.go
giuseppe: good observation.  It is safer to avoid the down-cast.  I'll change the type

pkg/chunked/storage_linux.go
giuseppe: yes, it is.  I'll simplify it, thanks

mtrmac (commented):
/lgtm

PR #1851, opened 2024-03-03
Draft: Bump to v1.52.1, and then to v1.52.2-dev

TomSweeneyRedHat:
In preparation of Podman v5.0, the first vendor dance step.

@nalind, @mheon, or @vrothberg. I'm on the fence with the version number.  I made a v1.52.0 release several weeks back, but not a branch, and we've had several commits since.    Since this is going into Podman v5.0, I thought we would want a release branch for it.  If that's the case, then perhaps I should bump this to v1.53 now and make a release-1.53 branch?

Thoughts?


rhatdan:
Yes bump to v1.53

vrothberg:
> Yes bump to v1.53

If new features went in, I agree.  If the commits are only bug fixes, we an bump the patch number (i.e., the Y in the X.Z.Y).

nalind:
> If new features went in, I agree. If the commits are only bug fixes, we an bump the patch number (i.e., the Y in the X.Z.Y).

Which are you saying is the case here?

vrothberg:
I see a couple of new APIs in the git-log.  So a minor bump to 1.53.0 sounds like the right thing to do.

TomSweeneyRedHat:
Just looking back, v1.52.0 was spun up on January 16, and is about 3 pages back in the PR history.  As there looks to be more than enough updates to go to 1.53, and it will be cleaner, I'll kill this and will create 1.53.0.

Thanks all

TomSweeneyRedHat:
Closing to bump to v1.53.0 instead

PR #1857, opened 2024-03-05
chunked: use mmap to load cache files

giuseppe:
reduce memory usage for the process by not loading entirely in memory any cache file for the layers.
    
The memory mapped files can be shared among multiple instances of Podman, as well as not being fully loaded in memory.

### DIFF
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
@@ -21,6 +22,7 @@ import (
 	jsoniter "github.com/json-iterator/go"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 const (
@@ -41,10 +43,19 @@ type layer struct {
 	id        string
 	cacheFile *cacheFile
 	target    string
+	// mmapBuffer is nil when the cache file is fully loaded in memory.
+	// Otherwise it points to a mmap'ed buffer that is referenced by cacheFile.vdata.
+	mmapBuffer []byte
+
+	// reloadWithMmap is set when the current process generates the cache file,
+	// and cacheFile reuses the memory buffer used by the generation function.
+	// Next time the layer cache is used, attempt to reload the file using
+	// mmap.
+	reloadWithMmap bool
 }
 
 type layersCache struct {
-	layers  []layer
+	layers  []*layer
 	refs    int
 	store   storage.Store
 	mutex   sync.RWMutex
@@ -56,14 +67,29 @@ var (
 	cache      *layersCache
 )
 
+func (c *layer) release() {
+	runtime.SetFinalizer(c, nil)
+	if c.mmapBuffer != nil {
+		unix.Munmap(c.mmapBuffer)
+	}
+}
+
+func layerFinalizer(c *layer) {
+	c.release()
+}
+
 func (c *layersCache) release() {
 	cacheMutex.Lock()
 	defer cacheMutex.Unlock()
 
 	c.refs--
-	if c.refs == 0 {
-		cache = nil
+	if c.refs != 0 {
+		return
 	}
+	for _, l := range c.layers {
+		l.release()
+	}
+	cache = nil
 }
 
 func getLayersCacheRef(store storage.Store) *layersCache {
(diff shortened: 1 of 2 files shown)


rhatdan:
Is this something that will need to be back ported into podman 5.0 or wait for 5.1?

rhatdan:
LGTM
@mtrmac PTAL

store.go
```diff
@@ -440,6 +444,10 @@ type Store interface {
 	// associated with a layer.
 	LayerBigData(id, key string) (io.ReadCloser, error)
 
+	// LayerBigDataFile retrieves the backing file for the layers big data,
+	// when used.
+	LayerBigDataFilePath(id, key string) (string, error)
```
mtrmac, line 449: This is the public API, so this is the place to document that it can return `""`.

But this is fundamentally racy. After this returns, the layer’s BigData can be added/modified/removed at any time, hypothetically changing what the path points to. Fine, we can document that this returns a non-empty string only if the path will only refer to an item for (`id, key`) and not for any other item.

But the whole layer can be removed after this returns. And recreated again. And the BigData item can be created, with different contents, afterwards. All of this could happen before “we” get to consume the returned path.

Also, some other process can _change_ the BigData item at any time, as long as “we” are not inside a `Store` method and holding a lock.

_Maybe_ one option would be to return a file descriptor, _and_ to guarantee that we use `Rename()` on BigData creation so that the file descriptor refers to immutable data.

store.go
```diff
@@ -127,6 +127,10 @@ type roLayerBigDataStore interface {
 	// with this ID.
 	BigData(id, key string) (io.ReadCloser, error)
 
+	// BigDataFilePath returns the path to the underlying file, if the big data
+	// is backed by a file.  Otherwise it returns the empty string.
+	BigDataFilePath(id, key string) (string, error)
```
mtrmac, line 132: Make this a package-private method, please.

mtrmac (commented):
For now just a note on the public API, I didn’t read the actual cache part.

It’s possible that these things are not much a concern for the cache, I didn’t check.

giuseppe:
I've pushed a new version that doesn't require any API change, and try to cast the output from `LayerBigData` to a `os.File`

mtrmac:
Oh, that’s clever. Maybe worth a comment in the `layers.go` implementation, pointing at this assumption, so that it isn’t randomly broken by some wrapping helper.

giuseppe:
@kolyshkin PTAL

giuseppe:
rebased, please take a look

kolyshkin:
I'm still looking at it, hope to finish tomorrow.

pkg/chunked/cache_linux.go
```diff
@@ -91,83 +101,135 @@ func getLayersCache(store storage.Store) (*layersCache, error) {
...
-			c.addLayer(r.ID, metadata)
+		if loaded {
+			continue
 		}
-	}
-
-	var newLayers []layer
-	for _, l := range c.layers {
-		if _, found := currentLayers[l.id]; found {
-			newLayers = append(newLayers, l)
```
kolyshkin, line 166: It's a bit suspicious that you've removed this part (which re-sets `c.layers`). My guess is, either you assume that 
- there can be no situation when some layers are gone during a subsequent call to `getLayersCache(store)`, or
- it is not important that some (stale?) layers are not removed from the cache.

Please clarify.

kolyshkin (commented):
I will continue tomorrow. For now, I've [split this into two patches](https://github.com/kolyshkin/storage/tree/improve-chunked-cache-loading.split) for easier review:
1. rename manifest to cacheFile
2. the rest of your changes

I might split it further as there are some code cleanups in here that may benefit from being in a separate commit.

pkg/chunked/cache_linux.go
giuseppe: you are right the new code didn't deal correctly with stale layers.

I've changed the code to cleanup stale layers and override `c.layers` as before

pkg/chunked/cache_linux.go
```diff
@@ -455,19 +529,18 @@ func prepareMetadata(manifest []byte, format graphdriver.DifferOutputFormat) ([]
...
-
 	l := layer{
-		id:        id,
-		cacheFile: cacheFile,
-		target:    target,
+		id:         id,
+		cacheFile:  cacheFile,
+		target:     target,
+		mmapBuffer: mmapBuffer,
 	}
```
kolyshkin, line 542: In here, we can do something like
```go
if mmapBuffer != nil {
    runtime.SetFinalizer(l, l.release)
}
```

(and amend release() to do `SetFinalizer(l, nil)`).

kolyshkin (commented):
Added some more comments.

pkg/chunked/cache_linux.go
```diff
@@ -41,6 +42,9 @@ type layer struct {
 	id        string
 	cacheFile *cacheFile
 	target    string
+	// mmapBuffer is nil when the cache file is fully loaded in memory.
+	// Otherwise it points to a mmap'ed buffer that is referenced by cacheFile.vdata.
+	mmapBuffer []byte
```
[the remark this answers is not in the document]
giuseppe: `vdata != mmapBuffer`, as it doesn't include the header of the file.  We still need a way to store the pointer to the mmap, or the offset of vdata inside the buffer

pkg/chunked/cache_linux.go
```diff
@@ -91,83 +105,144 @@ func getLayersCache(store storage.Store) (*layersCache, error) {
...
+	}
+	return c.createLayer(layerID, cacheFile, mmapBuffer)
+}
+
+func (c *layersCache) createCacheFileFromTOC(layerID string) (*layer, error) {
+	clFile, err := c.store.LayerBigData(layerID, chunkedLayerDataKey)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return nil, err
+	}
+	if clFile == nil {
```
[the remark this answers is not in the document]
giuseppe: I'll restore the previous behavior, since the change doesn't belong to this PR

pkg/chunked/cache_linux.go
```diff
@@ -91,83 +105,144 @@ func getLayersCache(store storage.Store) (*layersCache, error) {
...
+		// best effort advise to the kernel.
+		_ = unix.Madvise(buf, unix.MADV_RANDOM)
+
+		return buf, buf, nil
+	}
+	buf, err := io.ReadAll(inputFile)
+	return buf, nil, err
+}
+
+func (c *layersCache) loadLayerCache(layerID string) (l *layer, errRet error) {
```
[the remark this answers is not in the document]
giuseppe: fixed

pkg/chunked/cache_linux.go
```diff
@@ -91,83 +105,144 @@ func getLayersCache(store storage.Store) (*layersCache, error) {
...
+			return nil, nil, err
+		}
+		// best effort advise to the kernel.
+		_ = unix.Madvise(buf, unix.MADV_RANDOM)
+
+		return buf, buf, nil
+	}
+	buf, err := io.ReadAll(inputFile)
+	return buf, nil, err
+}
```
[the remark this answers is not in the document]
giuseppe: good idea, I'll do that

pkg/chunked/cache_linux.go
```diff
@@ -91,83 +105,144 @@ func getLayersCache(store storage.Store) (*layersCache, error) {
...
-		}
-
-		bigData, err := c.store.LayerBigData(r.ID, cacheKey)
-		// if the cache already exists, read and use it
-		if err == nil {
-			defer bigData.Close()
-			cacheFile, err := readCacheFileFromReader(bigData)
-			if err == nil {
-				c.addLayer(r.ID, cacheFile)
+		// the layer is present in the store and it is already loaded.  Attempt to use
```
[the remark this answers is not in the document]
giuseppe: fixed

pkg/chunked/cache_linux.go
giuseppe: changed to use `runtime.SetFinalizer`

pkg/chunked/cache_linux.go
```diff
@@ -91,83 +111,147 @@ func getLayersCache(store storage.Store) (*layersCache, error) {
...
-		// if the cache already exists, read and use it
-		if err == nil {
-			defer bigData.Close()
-			cacheFile, err := readCacheFileFromReader(bigData)
-			if err == nil {
-				c.addLayer(r.ID, cacheFile)
+		// The layer is present in the store and it is already loaded.  Attempt to
+		// re-use it if mmap'ed.
+		if l, found := loadedLayers[r.ID]; found {
+			if l.mmapBuffer != nil {
```
kolyshkin, line 225: Does that mean if a layer is loaded via io.ReadAll (rather than Mmap), we are re-reading it?

giuseppe: this was meant to be an optimization to use only when the file is first created, so we already have its content in memory and to avoid reloading it.  I've fixed it to not reload the cache when the file was initially loaded using io.ReadAll.

Applied the following fixup patch and pushed the new version:

```diff
diff --git a/pkg/chunked/cache_linux.go b/pkg/chunked/cache_linux.go
index 386bda515..01bfc8d92 100644
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -46,6 +46,12 @@ type layer struct {
 	// mmapBuffer is nil when the cache file is fully loaded in memory.
 	// Otherwise it points to a mmap'ed buffer that is referenced by cacheFile.vdata.
 	mmapBuffer []byte
+
+	// reloadWithMmap is set when the current process generates the cache file,
+	// and cacheFile reuses the memory buffer used by the generation function.
+	// Next time the layer cache is used, attempt to reload the file using
+	// mmap.
+	reloadWithMmap bool
 }
 
 type layersCache struct {
@@ -201,7 +207,12 @@ func (c *layersCache) createCacheFileFromTOC(layerID string) (*layer, error) {
 	if err != nil {
 		return nil, err
 	}
-	return c.createLayer(layerID, cacheFile, nil)
+	l, err := c.createLayer(layerID, cacheFile, nil)
+	if err != nil {
+		return nil, err
+	}
+	l.reloadWithMmap = true
+	return l, nil
 }
 
 func (c *layersCache) load() error {
@@ -222,8 +233,8 @@ func (c *layersCache) load() error {
 		// The layer is present in the store and it is already loaded.  Attempt to
 		// re-use it if mmap'ed.
 		if l, found := loadedLayers[r.ID]; found {
-			if l.mmapBuffer != nil {
-				// It is loaded through mmap.  Re-use it.
+			// If the layer is not marked for re-load, move it to newLayers.
+			if !l.reloadWithMmap {
 				delete(loadedLayers, r.ID)
 				newLayers = append(newLayers, l)
 				continue
```

giuseppe:
@kolyshkin are you fine with the last version?

giuseppe:
rebased

pkg/chunked/cache_linux.go
```diff
@@ -91,83 +117,152 @@ func getLayersCache(store storage.Store) (*layersCache, error) {
...
+	return c.createLayer(layerID, cacheFile, mmapBuffer)
+}
+
+func (c *layersCache) createCacheFileFromTOC(layerID string) (*layer, error) {
+	clFile, err := c.store.LayerBigData(layerID, chunkedLayerDataKey)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return nil, err
+	}
+	var lcd chunkedLayerData
+	if clFile != nil {
```
mtrmac, line 184: Pre-existing:
```suggestion
	if err == nil && clFile != nil {
```
for pedantic correctness, should not matter in practice

mtrmac, line 184: Pre-existing: `clFile` should be closed.

mtrmac (commented):
Please add  a comment around `layerStore.BigData` noting that `pkg/chunked` relies on the returned type being exactly `*os.File`.

LGTM otherwise.

giuseppe:
@mtrmac addressed your comments and pushed a new version

mtrmac (commented):
Thanks! LGTM. Up to @kolyshkin now.

kolyshkin (approved):
lgtm

PR #1858, opened 2024-03-11
overlay: create the merged path only if it does not exist

giuseppe:
follow-up for ccb70a79a69a1a5137ff24720520534bfbcc2316

more information here: https://github.com/containers/storage/issues/1827#issuecomment-1988332922

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1666,9 +1666,11 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	}
 
 	mergedDir := path.Join(dir, "merged")
-	// Create the driver merged dir
-	if err := idtools.MkdirAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
-		return "", err
+	// Attempt to create the merged dir only if it doesn't exist.
+	if _, err := os.Stat(mergedDir); err != nil && os.IsNotExist(err) {
+		if err := idtools.MkdirAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
+			return "", err
+		}
 	}
 	if count := d.ctr.Increment(mergedDir); count > 1 {
 		return mergedDir, nil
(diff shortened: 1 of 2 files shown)


drivers/overlay/overlay.go
```diff
@@ -1666,9 +1666,12 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	}
 
 	mergedDir := path.Join(dir, "merged")
-	// Create the driver merged dir
-	if err := idtools.MkdirAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
-		return "", err
+	// attempt to create the merged dir only if it doesn't exist.
+	if _, err := os.Stat(mergedDir); err != nil {
```
vrothberg, line 1670: I think we need to move the `os.IsExist` check here.

Any chance we can add a test?

giuseppe:
added test and amended the suggestion

drivers/overlay/overlay.go
```diff
@@ -1666,9 +1666,11 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	}
 
 	mergedDir := path.Join(dir, "merged")
-	// Create the driver merged dir
-	if err := idtools.MkdirAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
-		return "", err
+	// Attempt to create the merged dir only if it doesn't exist.
+	if _, err := os.Stat(mergedDir); err != nil && !os.IsExist(err) {
```
vrothberg, line 1670:
```suggestion
	if _, err := os.Stat(mergedDir); err != nil && os.IsExist(err) {
```

giuseppe: ah, it should really be `os.IsNotExist(err)`

rhatdan: And you are not supposted to use IsExist any longer, `errors.Is(err, fs.ErrNotExist)` But it looks like we have some cases of IsNotExist in podman.

PR #1862, opened 2024-03-14
chunked: skip file metadata for composefs-like links

giuseppe:
if the file is created using the object-store flat directory format, there is no need to set its inodes attributes, as anyway they are ignored when creating the composefs binary blob.

Similarly, we do not need to set file attributes when the deduplication happened through a hard link.

### DIFF
--- a/pkg/chunked/internal/compression.go
+++ b/pkg/chunked/internal/compression.go
@@ -48,9 +48,6 @@ type FileMetadata struct {
 	ChunkOffset int64  `json:"chunkOffset,omitempty"`
 	ChunkDigest string `json:"chunkDigest,omitempty"`
 	ChunkType   string `json:"chunkType,omitempty"`
-
-	// internal: computed by mergeTOCEntries.
-	Chunks []*FileMetadata `json:"-"`
 }
 
 const (
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -153,7 +153,8 @@ func doHardLink(srcFd int, destDirFd int, destBase string) error {
 	return err
 }
 
-func copyFileContent(srcFd int, destFile string, dirfd int, mode os.FileMode, useHardLinks bool) (*os.File, int64, error) {
+func copyFileContent(srcFd int, fileMetadata *fileMetadata, dirfd int, mode os.FileMode, useHardLinks bool) (*os.File, int64, error) {
+	destFile := fileMetadata.Name
 	src := fmt.Sprintf("/proc/self/fd/%d", srcFd)
 	st, err := os.Stat(src)
 	if err != nil {
@@ -171,6 +172,8 @@ func copyFileContent(srcFd int, destFile string, dirfd int, mode os.FileMode, us
 
 			err := doHardLink(srcFd, int(destDir.Fd()), destBase)
 			if err == nil {
+				// if the file was deduplicated with a hard link, skip overriding file metadata.
+				fileMetadata.skipSetAttrs = true
 				return nil, st.Size(), nil
 			}
 		}
@@ -382,7 +385,7 @@ func copyFileFromOtherLayer(file *fileMetadata, source string, name string, dirf
 	}
 	defer srcFile.Close()
 
-	dstFile, written, err := copyFileContent(int(srcFile.Fd()), file.Name, dirfd, 0, useHardLinks)
+	dstFile, written, err := copyFileContent(int(srcFile.Fd()), file, dirfd, 0, useHardLinks)
 	if err != nil {
 		return false, nil, 0, fmt.Errorf("copy content to %q: %w", file.Name, err)
 	}
@@ -484,7 +487,7 @@ func findFileInOSTreeRepos(file *fileMetadata, ostreeRepos []string, dirfd int, 
 			continue
 		}
 
-		dstFile, written, err := copyFileContent(fd, file.Name, dirfd, 0, useHardLinks)
+		dstFile, written, err := copyFileContent(fd, file, dirfd, 0, useHardLinks)
 		if err != nil {
 			logrus.Debugf("could not copyFileContent: %v", err)
 			return false, nil, 0, nil
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -1595,6 +1595,7 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
 		hashes[d] = d
 
 		mergedEntries[i].Name = fmt.Sprintf("%s/%s", d[0:2], d[2:])
+		mergedEntries[i].skipSetAttrs = true
 
 		new = append(new, mergedEntries[i])
 	}
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -58,6 +58,16 @@ const (
 	copyGoRoutines = 32
 )
 
+// fileMetadata is a wrapper around internal.FileMetadata with additional private fields that
+// are not part of the TOC document.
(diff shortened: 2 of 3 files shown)


rhatdan:
LGTM
@mtrmac @alexlarsson @cgwalters PTAL

pkg/chunked/internal/compression.go
```diff
@@ -51,6 +51,10 @@ type FileMetadata struct {
 
 	// internal: computed by mergeTOCEntries.
 	Chunks []*FileMetadata `json:"-"`
+	// SkipSetAttrs is set when the file attributes must not be
+	// modified, e.g. it is a hard link from a different source,
+	// or a composefs file.
+	SkipSetAttrs bool
```
mtrmac, line 57: Is this really intended to be maintained in the TOC?

If not, can this field be tracked internally inside `pkg/chunked` only?

cgwalters:
Apart from the other comment, this seems sane; are there no integration tests that run with composefs enabled on this repo?  It seems like it wouldn't be too hard to add a test for this if so.  (And if not, probably need integration coverage for chunked/composefs here)

mtrmac (commented):
BTW, the code to check whether the hardlink destination matches the expected attributes (`canDedupFileWithHardLink`?) and this code to set/not set the attributes are way too distant for me to be comfortable; I’m not smart enough to believe I would be able to keep the two in sync long-term.

pkg/chunked/internal/compression.go
giuseppe: yeah, let's clean this up and remove it from the TOC definition.

What do you think about the new version?

giuseppe:
> Apart from the other comment, this seems sane; are there no integration tests that run with composefs enabled on this repo? It seems like it wouldn't be too hard to add a test for this if so. (And if not, probably need integration coverage for chunked/composefs here)

this is still something we miss, as we don't have any test for composefs

pkg/chunked/storage_linux.go
```diff
@@ -58,6 +58,20 @@ const (
 	copyGoRoutines = 32
 )
 
+// fileMetadata is a wrapper around internal.FileMetadata with additional private fields that
+// are not part of the TOC document.
+type fileMetadata struct {
+	internal.FileMetadata
+
+	// Chunks are computed by mergeTOCEntries.
+	Chunks []*internal.FileMetadata `json:"-"`
```
mtrmac, line 67: - The internal fields can be package-private
- I read the the `json` annotation to suggest that this struct can be marshaled/unmarshaled; that’s AFAICS not the case, so dropping the annotation would make this easier to follow.

pkg/chunked/storage_linux.go
```diff
@@ -58,6 +58,20 @@ const (
 	copyGoRoutines = 32
 )
 
+// fileMetadata is a wrapper around internal.FileMetadata with additional private fields that
+// are not part of the TOC document.
+type fileMetadata struct {
+	internal.FileMetadata
+
+	// Chunks are computed by mergeTOCEntries.
```
mtrmac, line 66:
```suggestion
	// If FileMetadata.Type == TypeReg, this stores the TypeChunk entries relevant to this entry.
```
(“relevant to” can probably be improved upon)

pkg/chunked/storage_linux.go
```diff
@@ -582,7 +601,10 @@ func (o *originFile) OpenFile() (io.ReadCloser, error) {
 }
 
 // setFileAttrs sets the file attributes for file given metadata
-func setFileAttrs(dirfd int, file *os.File, mode os.FileMode, metadata *internal.FileMetadata, options *archive.TarOptions, usePath bool) error {
+func setFileAttrs(dirfd int, file *os.File, mode os.FileMode, metadata *fileMetadata, options *archive.TarOptions, usePath bool) error {
```
mtrmac, line 604: (Unrelated: if I’m not mistaken, `mode` is always `os.FileMode(metadata.Mode)`, so the parameter could be eliminated throughout the call stack. Perhaps best to do in a separate PR.)

pkg/chunked/storage_linux.go
```diff
@@ -1729,9 +1752,16 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 
 	var missingParts []missingPart
 
-	output.UIDs, output.GIDs = collectIDs(toc.Entries)
+	var entries []fileMetadata
+	for _, e := range toc.Entries {
```
mtrmac, line 1756: (This works just fine; alternatively, maybe it could be the job of `mergeTocEntries` to convert from the one type to the other.

That way, the semantics could be separated by type: `[]internal.FileMetadata` can contain top-level `TypeChunk` entries, and `[]fileMetadata` never contains top-level `TypeChunk`, they are all collected in `.Chunks`.)

pkg/chunked/storage_linux.go
```diff
@@ -58,6 +58,20 @@ const (
 	copyGoRoutines = 32
 )
 
+// fileMetadata is a wrapper around internal.FileMetadata with additional private fields that
+// are not part of the TOC document.
```
mtrmac, line 62:
```suggestion
// are not part of the TOC document.
// Type: TypeChunk entries are stored in Chunks, the primary [fileMetadata] entries never use TypeChunk
```

pkg/chunked/storage_linux.go
```diff
@@ -1987,6 +2017,10 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 
 			switch chunk.ChunkType {
 			case internal.ChunkTypeData:
+				fm := fileMetadata{
+					FileMetadata: *chunk,
+				}
+				chunk := &fm
```
mtrmac, line 2023: Following the principle that `fileMetadata` entries only contain `TypeChunk` entries in the `.Chunks` field, not at the top level, it seems to me that the type of the `findChunkInOtherLayers` parameter could remain `internal.FileMetadata`, so that we don’t need to create this usually-invalid entry.

mtrmac (commented):
This is nice.

Apart from comment around `internal.ChunkTypeData` (to consistently never use the internal `fileMetadata` for `TypeChunk` values), all others are just tentative aesthetic suggestions.

giuseppe:
thanks, took care of the comments (except https://github.com/containers/storage/pull/1862#discussion_r1530966629 ) and pushed a new version

pkg/chunked/storage_linux.go
```diff
@@ -2045,9 +2068,16 @@ func mustSkipFile(fileType compressedFileType, e internal.FileMetadata) bool {
 	return false
 }
 
-func (c *chunkedDiffer) mergeTocEntries(fileType compressedFileType, entries []internal.FileMetadata) ([]internal.FileMetadata, int64, error) {
+func (c *chunkedDiffer) mergeTocEntries(fileType compressedFileType, tocEntries []internal.FileMetadata) ([]fileMetadata, int64, error) {
 	var totalFilesSize int64
 
+	var entries []fileMetadata
+	for _, e := range tocEntries {
```
mtrmac, line 2075: AFAICS this copy/allocation would also be cheap to eliminate: We are already building `mergedEntries` as an array of the right type, so…

pkg/chunked/storage_linux.go
```diff
@@ -2069,7 +2099,7 @@ func (c *chunkedDiffer) mergeTocEntries(fileType compressedFileType, entries []i
 		}
 	}
 
-	mergedEntries := make([]internal.FileMetadata, size)
+	mergedEntries := make([]fileMetadata, size)
 	m := 0
 	for i := 0; i < len(entries); i++ {
 		e := entries[i]
```
mtrmac, line 2105: …
```go
    e := fileMetadata{ FileMetadata: entries[i] }
```
might be all that is necessary.

mtrmac (commented):
Thanks!

LGTM, feel free to merge as is or add that small cleanup.

PR #1870, opened 2024-03-27
chunked: store cache as binary and use a bloom filter

giuseppe:
The bloom filter itself is useful to reduce page faults with the mmap'ed cache files, as it reduces lookups.

Storing the file as a binary instead reduces the file size considerably, with the `quay.io/giuseppe/zstd-chunked:fedora-{38,39,40}{,-updated}` images I see:

before:
```
# find -name '=Y2h1bmtlZC1tYW5pZmVzdC1jYWNoZQ==' -exec stat -c '%s' \{\} \;2547644
2575163
2547644
2476816
2462835
2533346
```

after:
```
# find -name '=Y2h1bmtlZC1tYW5pZmVzdC1jYWNoZQ==' -exec stat -c '%s' \{\} \;
1319206
1312332
1275803
1270629
1297565
```

so it is ~50% size reduction

### DIFF
--- a/pkg/chunked/bloom_filter.go
+++ b/pkg/chunked/bloom_filter.go
@@ -1,0 +1,84 @@
+package chunked
+
+import (
+	"encoding/binary"
+	"hash/crc32"
+	"io"
+)
+
+type bloomFilter struct {
+	bitArray []uint64
+	k        uint32
+}
+
+func newBloomFilter(size int, k uint32) *bloomFilter {
+	numElements := (size + 63) / 64
+	return &bloomFilter{
+		bitArray: make([]uint64, numElements),
+		k:        k,
+	}
+}
+
+func newBloomFilterFromArray(bitArray []uint64, k uint32) *bloomFilter {
+	return &bloomFilter{
+		bitArray: bitArray,
+		k:        k,
+	}
+}
+
+func (bf *bloomFilter) hashFn(item []byte, seed uint32) (uint64, uint64) {
+	if len(item) == 0 {
+		return 0, 0
+	}
+	mod := uint32(len(bf.bitArray) * 64)
+	seedSplit := seed % uint32(len(item))
+	hash := (crc32.ChecksumIEEE(item[:seedSplit]) ^ crc32.ChecksumIEEE(item[seedSplit:])) % mod
+	return uint64(hash / 64), uint64(1 << (hash % 64))
+}
+
+func (bf *bloomFilter) add(item []byte) {
+	for i := uint32(0); i < bf.k; i++ {
+		index, mask := bf.hashFn(item, i)
+		bf.bitArray[index] |= mask
+	}
+}
+
+func (bf *bloomFilter) maybeContains(item []byte) bool {
+	for i := uint32(0); i < bf.k; i++ {
+		index, mask := bf.hashFn(item, i)
+		if bf.bitArray[index]&mask == 0 {
+			return false
+		}
+	}
+	return true
+}
+
+func (bf *bloomFilter) writeTo(writer io.Writer) error {
+	if err := binary.Write(writer, binary.LittleEndian, uint64(len(bf.bitArray))); err != nil {
+		return err
+	}
+	if err := binary.Write(writer, binary.LittleEndian, uint32(bf.k)); err != nil {
+		return err
+	}
+	if err := binary.Write(writer, binary.LittleEndian, bf.bitArray); err != nil {
+		return err
+	}
+	return nil
+}
+
+func readBloomFilter(reader io.Reader) (*bloomFilter, error) {
+	var bloomFilterLen uint64
+	var k uint32
+
+	if err := binary.Read(reader, binary.LittleEndian, &bloomFilterLen); err != nil {
+		return nil, err
+	}
+	if err := binary.Read(reader, binary.LittleEndian, &k); err != nil {
+		return nil, err
+	}
+	bloomFilterArray := make([]uint64, bloomFilterLen)
+	if err := binary.Read(reader, binary.LittleEndian, &bloomFilterArray); err != nil {
+		return nil, err
+	}
+	return newBloomFilterFromArray(bloomFilterArray, k), nil
+}
--- a/pkg/chunked/bloom_filter_test.go
+++ b/pkg/chunked/bloom_filter_test.go
@@ -29,6 +29,7 @@ var (
 func initCache(sizeCache int) (*cacheFile, string, string, *bloomFilter) {
 	var tagsBuffer bytes.Buffer
 	var vdata bytes.Buffer
+	var fnames bytes.Buffer
 	tags := [][]byte{}
 	tagLen := 0
 	digestLen := 64
@@ -59,7 +60,7 @@ func initCache(sizeCache int) (*cacheFile, string, string, *bloomFilter) {
 	hash.Write([]byte("1"))
 	notPresentDigest = digester.Digest().String()
 
-	writeCacheFileToWriter(io.Discard, bloomFilter, tags, tagLen, digestLen, vdata, &tagsBuffer)
+	writeCacheFileToWriter(io.Discard, bloomFilter, tags, tagLen, digestLen, vdata, fnames, &tagsBuffer)
 
 	cache := &cacheFile{
 		digestLen: digestLen,
--- a/pkg/chunked/bloom_filter_test.go
+++ b/pkg/chunked/bloom_filter_test.go
@@ -59,7 +59,7 @@ func initCache(sizeCache int) (*cacheFile, string, string, *bloomFilter) {
 	hash.Write([]byte("1"))
 	notPresentDigest = digester.Digest().String()
 
-	writeCacheFileToWriter(io.Discard, tags, tagLen, digestLen, &vdata, &tagsBuffer)
+	writeCacheFileToWriter(io.Discard, bloomFilter, tags, tagLen, digestLen, vdata, &tagsBuffer)
 
 	cache := &cacheFile{
 		digestLen: digestLen,
--- a/pkg/chunked/bloom_filter_test.go
+++ b/pkg/chunked/bloom_filter_test.go
@@ -1,0 +1,126 @@
+package chunked
+
+import (
+	"bytes"
+	"io"
+	"testing"
+
+	digest "github.com/opencontainers/go-digest"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	presentDigestInCache          string
+	notPresentDigestInCache       string
+	presentDigestInCacheBinary    []byte
+	notPresentDigestInCacheBinary []byte
+	preloadedCache                *cacheFile
+	preloadedbloomFilter          *bloomFilter
+	benchmarkN                    int = 100000
+)
+
+// Using 3 hashes functions and n/m = 10 gives a false positive rate of ~1.7%:
+// https://pages.cs.wisc.edu/~cao/papers/summary-cache/node8.html
+var (
+	factorNM     int    = 10
+	numberHashes uint32 = 3
+)
+
+func initCache(sizeCache int) (*cacheFile, string, string, *bloomFilter) {
+	var tagsBuffer bytes.Buffer
+	var vdata bytes.Buffer
+	tags := [][]byte{}
+	tagLen := 0
+	digestLen := 64
+	var presentDigest, notPresentDigest string
+
+	bloomFilter := newBloomFilter(sizeCache*factorNM, numberHashes)
+
+	digester := digest.Canonical.Digester()
+	hash := digester.Hash()
+	for i := 0; i < sizeCache; i++ {
+		hash.Write([]byte("1"))
+		d := digester.Digest().String()
+		digestLen = len(d)
+		presentDigest = d
+		tag, err := appendTag([]byte(d), 0, 0)
+		if err != nil {
+			panic(err)
+		}
+		tagLen = len(tag)
+		tags = append(tags, tag)
+		bd, err := makeBinaryDigest(d)
+		if err != nil {
+			panic(err)
+		}
+		bloomFilter.add(bd)
+	}
+
+	hash.Write([]byte("1"))
+	notPresentDigest = digester.Digest().String()
+
+	writeCacheFileToWriter(io.Discard, tags, tagLen, digestLen, &vdata, &tagsBuffer)
+
+	cache := &cacheFile{
+		digestLen: digestLen,
+		tagLen:    tagLen,
+		tags:      tagsBuffer.Bytes(),
+		vdata:     vdata.Bytes(),
+	}
+	return cache, presentDigest, notPresentDigest, bloomFilter
+}
+
+func init() {
+	var err error
+	preloadedCache, presentDigestInCache, notPresentDigestInCache, preloadedbloomFilter = initCache(10000)
+	presentDigestInCacheBinary, err = makeBinaryDigest(presentDigestInCache)
+	if err != nil {
+		panic(err)
+	}
+	notPresentDigestInCacheBinary, err = makeBinaryDigest(notPresentDigestInCache)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func BenchmarkLookupBloomFilter(b *testing.B) {
+	for i := 0; i < benchmarkN; i++ {
+		if preloadedbloomFilter.maybeContains(notPresentDigestInCacheBinary) {
+			findTag(notPresentDigestInCache, preloadedCache)
+		}
+		if preloadedbloomFilter.maybeContains(presentDigestInCacheBinary) {
+			findTag(presentDigestInCache, preloadedCache)
+		}
+	}
+}
+
+func BenchmarkLookupBloomRaw(b *testing.B) {
+	for i := 0; i < benchmarkN; i++ {
+		findTag(notPresentDigestInCache, preloadedCache)
+		findTag(presentDigestInCache, preloadedCache)
+	}
+}
+
+func TestBloomFilter(t *testing.T) {
+	bloomFilter := newBloomFilter(1000, 1)
+	digester := digest.Canonical.Digester()
+	hash := digester.Hash()
+	for i := 0; i < 1000; i++ {
+		hash.Write([]byte("1"))
+		d := digester.Digest().String()
+		bd, err := makeBinaryDigest(d)
+		assert.NoError(t, err)
+		contains := bloomFilter.maybeContains(bd)
+		assert.False(t, contains)
+	}
+	for i := 0; i < 1000; i++ {
+		hash.Write([]byte("1"))
+		d := digester.Digest().String()
+		bd, err := makeBinaryDigest(d)
+		assert.NoError(t, err)
+		bloomFilter.add(bd)
+
+		contains := bloomFilter.maybeContains(bd)
+		assert.True(t, contains)
+	}
+}
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	cacheKey     = "chunked-manifest-cache"
-	cacheVersion = 2
+	cacheVersion = 3
 
 	digestSha256Empty = "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -320,6 +320,52 @@ type setBigData interface {
 	SetLayerBigData(id, key string, data io.Reader) error
 }
 
+func writeCacheFileToWriter(writer io.Writer, tags []string, tagLen, digestLen int, vdata bytes.Buffer, tagsBuffer *bytes.Buffer) error {
+	sort.Strings(tags)
+	for _, t := range tags {
+		if _, err := tagsBuffer.Write([]byte(t)); err != nil {
+			return err
+		}
+	}
+
+	// version
+	if err := binary.Write(writer, binary.LittleEndian, uint64(cacheVersion)); err != nil {
+		return err
+	}
+
+	// len of a tag
+	if err := binary.Write(writer, binary.LittleEndian, uint64(tagLen)); err != nil {
+		return err
+	}
+
+	// len of a digest
+	if err := binary.Write(writer, binary.LittleEndian, uint64(digestLen)); err != nil {
+		return err
+	}
+
+	// tags length
+	if err := binary.Write(writer, binary.LittleEndian, uint64(tagsBuffer.Len())); err != nil {
+		return err
+	}
+
+	// vdata length
+	if err := binary.Write(writer, binary.LittleEndian, uint64(vdata.Len())); err != nil {
+		return err
+	}
+
+	// tags
+	if _, err := writer.Write(tagsBuffer.Bytes()); err != nil {
+		return err
+	}
+
+	// variable length data
+	if _, err := writer.Write(vdata.Bytes()); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // writeCache write a cache for the layer ID.
 // It generates a sorted list of digests with their offset to the path location and offset.
 // The same cache is used to lookup files, chunks and candidates for deduplication with hard links.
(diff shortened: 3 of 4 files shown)


giuseppe:
@kolyshkin @mtrmac @rhatdan some more improvements to the cache file

pkg/chunked/bloom_filter.go
```diff
@@ -0,0 +1,95 @@
...
+	"io"
+	"unsafe"
+)
+
+type bloomFilter struct {
+	bitArray []uint64
+	k        uint32
+}
+
+func newBloomFilter(size uint, k uint32) *bloomFilter {
```
mtrmac, line 14: (Array indices and sizes in Go are `int`; they can’t go any larger anyway, so using `uint` here only means that the caller has to cast.)

pkg/chunked/cache_linux.go
```diff
@@ -465,17 +600,20 @@ func writeCache(manifest []byte, format graphdriver.DifferOutputFormat, id strin
...
-		tagLen:    tagLen,
-		tags:      tagsBuffer.Bytes(),
-		vdata:     vdata.Bytes(),
+		digestLen:   digestLen,
+		tagLen:      tagLen,
+		tags:        tagsBuffer.Bytes(),
+		vdata:       vdata.Bytes(),
+		fnames:      fnames.Bytes(),
+		fnamesLen:   len(fnames.Bytes()),
+		bloomFilter: bloomFilterFromTags(tags, digestLen),
```
mtrmac, line 609: This was already computed inside `writeCacheFileToWriter`; couldn’t that happen only once?

pkg/chunked/cache_linux.go
```diff
@@ -488,26 +626,41 @@ func readCacheFileFromMemory(bigDataBuffer []byte) (*cacheFile, error) {
 	if err := binary.Read(bigData, binary.LittleEndian, &digestLen); err != nil {
 		return nil, err
 	}
+
+	bloomFilter, err := readBloomFilter(bigData)
```
mtrmac, line 630: What is a typical size of the filter? Would it pay off to access the mapped data directly instead of making a copy?

(I didn’t even try to estimate, for all I know this might clearly not be worth the extra complexity. If it is a few percent of the total size, let’s not bother.)

pkg/chunked/cache_linux.go
```diff
@@ -609,20 +767,32 @@ func (c *layersCache) findDigestInternal(digest string) (string, string, int64,
...
+	}
+
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
 
 	for _, layer := range c.layers {
+		if !layer.cacheFile.bloomFilter.maybeContains(binaryDigest) {
+			continue
+		}
 		digest, off, tagLen := findTag(digest, layer.cacheFile)
```
mtrmac, line 782: This could use the already-computed `binaryDigest`

pkg/chunked/cache_linux.go
```diff
@@ -488,26 +626,41 @@ func readCacheFileFromMemory(bigDataBuffer []byte) (*cacheFile, error) {
...
 
 	return &cacheFile{
-		tagLen:    int(tagLen),
-		digestLen: int(digestLen),
-		tags:      tags,
-		vdata:     vdata,
+		bloomFilter: bloomFilter,
+		digestLen:   int(digestLen),
+		fnames:      fnames,
+		fnamesLen:   int(fnamesLen),
```
mtrmac, line 660: (Nit: I’d aesthetically prefer if the type declaration, and the two constructors, were all listing the fields in the same order — and if that order were somehow consistent. Here we have (data, len) for file names, but (len, data) for tags, for example.)

pkg/chunked/cache_linux.go
```diff
@@ -609,20 +767,32 @@ func (c *layersCache) findDigestInternal(digest string) (string, string, int64,
...
+			fileLocationData := layer.cacheFile.vdata[off : off+tagLen]
+
+			fnamePosition, offFile, _, err := parseFileLocation(fileLocationData)
+			if err != nil {
+				return "", "", 0, err
 			}
-			offFile, _ := strconv.ParseInt(parts[0], 10, 64)
+
+			lenPath := int(binary.LittleEndian.Uint32(layer.cacheFile.fnames[fnamePosition : fnamePosition+4]))
+			path := string(layer.cacheFile.fnames[fnamePosition+4 : fnamePosition+lenPath+4])
```
mtrmac, line 792: Consider bounds checks to avoid a panic on invalid data

pkg/chunked/cache_linux_test.go
```diff
@@ -164,7 +167,8 @@ func TestWriteCache(t *testing.T) {
 			if digest != r.ChunkDigest {
 				t.Error("wrong digest found")
 			}
-			expectedLocation := generateFileLocation(r.Name, uint64(r.ChunkOffset), uint64(r.ChunkSize))
+			expectedLocation, err := generateFileLocation(0, uint64(r.ChunkOffset), uint64(r.ChunkSize))
```
mtrmac, line 170: Non-blocking: Testing this with at least two distinct paths, to truly exercise the indexing logic, would be nice.

pkg/chunked/cache_linux.go
```diff
@@ -303,23 +324,132 @@ func calculateHardLinkFingerprint(f *fileMetadata) (string, error) {
 	return string(digester.Digest()), nil
 }
 
-// generateFileLocation generates a file location in the form $OFFSET:$LEN:$PATH
-func generateFileLocation(path string, offset, len uint64) []byte {
-	return []byte(fmt.Sprintf("%d:%d:%s", offset, len, path))
+// generateFileLocation generates a file location in the form $OFFSET$LEN$PATH_POS
+func generateFileLocation(pathPos int, offset, len uint64) ([]byte, error) {
```
mtrmac, line 328: This can never fail, so the error paths in callers can be dropped.

pkg/chunked/cache_linux.go
```diff
@@ -609,20 +767,32 @@ func (c *layersCache) findDigestInternal(digest string) (string, string, int64,
...
+		if !layer.cacheFile.bloomFilter.maybeContains(binaryDigest) {
+			continue
+		}
 		digest, off, tagLen := findTag(digest, layer.cacheFile)
 		if digest != "" {
-			position := string(layer.cacheFile.vdata[off : off+tagLen])
-			parts := strings.SplitN(position, ":", 3)
-			if len(parts) != 3 {
-				continue
+			fileLocationData := layer.cacheFile.vdata[off : off+tagLen]
```
mtrmac, line 784: Consider bounds checks to avoid a panic on invalid data

mtrmac (commented):
Thanks _so much_ for splitting this into smaller commits. That made the review very enjoyable.

pkg/chunked/cache_linux_test.go
```diff
@@ -211,3 +215,60 @@ func TestUnmarshalToc(t *testing.T) {
...
+
+	cache := &cacheFile{
+		digestLen: digestLen,
+		tagLen:    tagLen,
+		tags:      tagsBuffer.Bytes(),
+		vdata:     vdata.Bytes(),
+	}
+
+	for i := 0; i < n; i++ {
+		hash.Write([]byte("1"))
```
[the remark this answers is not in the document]
giuseppe: shouldn't matter too much as it is a binary search, but you are right, it is better to make it exhaustive and also measure hits

pkg/chunked/cache_linux_test.go
```diff
@@ -211,3 +215,60 @@ func TestUnmarshalToc(t *testing.T) {
 	_, err = unmarshalToc([]byte(jsonTOC + "123"))
 	assert.Error(t, err)
 }
+
+func benchmarkLookup(b *testing.B, sizeCache, n int) {
```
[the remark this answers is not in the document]
giuseppe: I think I'll drop these tests (and the `chunked: add benchmark tests for lookup` commit) from the series since these tests are already covered in `bloom_filter_test.go`.

With the tests in `bloom_filter_test.go`, I've:

```
$ benchstat <(go test -bench=. -count 20 -benchtime=1x)
name                  time/op
LookupBloomFilter-12  8.24ms ±10%
LookupBloomRaw-12      196ms ± 7%
```

pkg/chunked/cache_linux.go
```diff
@@ -154,6 +162,19 @@ fallback:
...
+func getBinaryDigest(stringDigest string) ([]byte, error) {
+	d, err := digest.Parse(stringDigest)
+	if err != nil {
+		return nil, err
+	}
+	digestBytes, err := hex.DecodeString(d.Hex())
+	if err != nil {
+		return nil, err
+	}
+	digest := append([]byte(d.Algorithm()+":"), digestBytes...)
```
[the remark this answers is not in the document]
giuseppe: with that I get:

```
./cache_linux.go:174:47: too many arguments in call to append
	have ([]byte, number, []byte)
	want ([]byte, ...byte)
```

pkg/chunked/bloom_filter_test.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func BenchmarkLookupBloomRaw(b *testing.B) {
+	for i := 0; i < benchmarkN; i++ {
+		findTag(notPresentDigestInCache, preloadedCache)
+		findTag(presentDigestInCache, preloadedCache)
+	}
+}
+
+func TestBloomFilter(t *testing.T) {
```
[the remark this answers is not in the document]
giuseppe: This test is not very helpful, and I could simplify it to test only one case, but testing for elements that are not part of the filter is not useful since both answers are correct

pkg/chunked/cache_linux_test.go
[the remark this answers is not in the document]
mtrmac: Oh, I didn’t realize this is not triggering the bloom filter. I agree for benchmarking the binary search, this is representative.

pkg/chunked/bloom_filter_test.go
```diff
@@ -0,0 +1,127 @@
...
+		panic(err)
+	}
+}
+
+func BenchmarkLookupBloomFilter(b *testing.B) {
+	for i := 0; i < benchmarkN; i++ {
+		if preloadedbloomFilter.maybeContains(notPresentDigestInCacheBinary) {
+			findTag(notPresentDigestInCache, preloadedCache)
+		}
+		if preloadedbloomFilter.maybeContains(notPresentDigestInCacheBinary) {
```
mtrmac, line 92:
```suggestion
		if preloadedbloomFilter.maybeContains(presentDigestInCacheBinary) {
```

pkg/chunked/bloom_filter_test.go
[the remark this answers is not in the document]
mtrmac: I was thinking that if we expect a false positive rate of 1.7 % and the test finds > 10 % (> 50 %?), it should fail. OTOH, *shrug*, the cost vs. benefit balance of doing that work is unclear.

pkg/chunked/cache_linux.go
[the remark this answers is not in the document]
mtrmac: I’m sorry, I have misremembered the language spec. So this would have to be two separate `append` calls.

Anyway, I think this micro-optimization can wait for the updated shape, if any, of `generateTag`/`getBinaryDigest`, per the other comments.

pkg/chunked/bloom_filter.go
```diff
@@ -0,0 +1,95 @@
...
+func (bf *bloomFilter) hashFn(item []byte, seed uint32) (uint64, uint64) {
+	var hash uint64 = 0
+
+	// long a multiple of 8
+	off := len(item) % 8
+	for _, b := range item[:off] {
+		hash = 17*hash + uint64(uint32(b)^seed)
+	}
+	// handle the remaining data as uint64
+	itemsAsUint64 := byteSliceAsUint64([]byte(item[off:]))
```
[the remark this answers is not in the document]
giuseppe: unfortunately using `encoding.ReadUint64` has a big impact on performance.

giuseppe:
the PR is ready for review

rhatdan:
@mtrmac needs another review.

pkg/chunked/cache_linux.go
```diff
@@ -579,50 +724,89 @@ func byteSliceAsString(b []byte) string {
 }
 
 func findTag(digest string, cacheFile *cacheFile) (string, uint64, uint64) {
```
mtrmac, line 726: (AFAICS this function is now only called from tests; maybe move it to a test file.)

pkg/chunked/bloom_filter.go
```diff
@@ -0,0 +1,80 @@
...
+func newBloomFilterFromArray(bitArray []uint64, k uint32) *bloomFilter {
+	return &bloomFilter{
+		bitArray: bitArray,
+		k:        k,
+	}
+}
+
+func (bf *bloomFilter) hashFn(item []byte, seed uint32) (uint64, uint64) {
+	mod := uint32(len(bf.bitArray) * 64)
+	hash := (seed ^ crc32.ChecksumIEEE(item)) % mod
```
mtrmac, line 31: Is the way `seed` is integrated here sufficient? Apart from the `% mod` part, this seems to be effectively computing 3 hashes of (crc32, crc32 ^ 1, crc32 ^ 2); intuitively (I didn’t do the math) 3 hash functions like this should be about as effective as using a single hash — a collision in the first hash implies a collision in the other hashes.

For the same reason, doing a CRC32 step “appending” `seed` to `ChecksumIEEE(item)` wouldn’t work well, and AFAICS integrating the seed as a data input to CRC32 wouldn’t work well either (all inputs are of the same length, so the seed affect the same 2 bits of output).

Ugh… I didn’t expect to be evaluating independence of hash functions this week ;)

---

It’s quite possible this is a big enough improvement with CRC32… we can always bump the cache version later to change the functions.

pkg/chunked/cache_linux.go
```diff
@@ -579,50 +724,89 @@ func byteSliceAsString(b []byte) string {
...
-			parts := strings.SplitN(position, ":", 3)
-			if len(parts) != 3 {
-				continue
+		if !layer.cacheFile.bloomFilter.maybeContains(binaryDigest) {
+			continue
+		}
+		found, off, tagLen := findBinaryTag(binaryDigest, layer.cacheFile)
+		if found {
+			if uint64(len(layer.cacheFile.vdata)) <= off+tagLen {
+				return "", "", 0, err
```
mtrmac, line 789: In all the added bound checks: This returns `"", "", 0, nil`, AFAICS that's considered a “success, found”. It should either contain an error value, or a -1 offset to indicate “not found”.

(And if the intent is to return `nil`, I’d find an explicit `nil` more readable.)

pkg/chunked/cache_linux.go
```diff
@@ -579,50 +724,89 @@ func byteSliceAsString(b []byte) string {
...
-			position := string(layer.cacheFile.vdata[off : off+tagLen])
-			parts := strings.SplitN(position, ":", 3)
-			if len(parts) != 3 {
-				continue
+		if !layer.cacheFile.bloomFilter.maybeContains(binaryDigest) {
+			continue
+		}
+		found, off, tagLen := findBinaryTag(binaryDigest, layer.cacheFile)
+		if found {
+			if uint64(len(layer.cacheFile.vdata)) <= off+tagLen {
```
mtrmac, line 788: In all three added bound checks: pedantically, the condition should be `<`.

mtrmac (commented):
I’m sorry about the delayed response.

pkg/chunked/bloom_filter.go
giuseppe: good point.

I've played more with it and use the seed to split item in two slices and checksum each of them.  The benchmark tests still show a benefit for using the bloom filter (with 3 hash functions):

```
$ go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/containers/storage/pkg/chunked
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
BenchmarkLookupBloomFilter-12    	1000000000	         0.08671 ns/op
BenchmarkLookupBloomRaw-12       	1000000000	         0.1495 ns/op
PASS
ok  	github.com/containers/storage/pkg/chunked	2.628s
```

giuseppe:
I've fixed your comments, except https://github.com/containers/storage/pull/1870#discussion_r1556134916.  What would you like me to do here?

pkg/chunked/cache_linux.go
```diff
@@ -303,23 +326,124 @@ func calculateHardLinkFingerprint(f *fileMetadata) (string, error) {
...
 }
 
-// generateTag generates a tag in the form $DIGEST$OFFSET@LEN.
-// the [OFFSET; LEN] points to the variable length data where the file locations
+// appendTag appends the $OFFSET$LEN information to the provided $DIGEST.
+// The [OFFSET; LEN] points to the variable length data where the file locations
 // are stored.  $DIGEST has length digestLen stored in the cache file file header.
-func generateTag(digest string, offset, len uint64) string {
-	return fmt.Sprintf("%s%.20d@%.20d", digest, offset, len)
+func appendTag(digest []byte, offset, len uint64) ([]byte, error) {
```
mtrmac, line 365: (Absolutely non-blocking: the error value is always `nil` now.)

pkg/chunked/bloom_filter.go
```diff
@@ -0,0 +1,81 @@
...
+func newBloomFilterFromArray(bitArray []uint64, k uint32) *bloomFilter {
+	return &bloomFilter{
+		bitArray: bitArray,
+		k:        k,
+	}
+}
+
+func (bf *bloomFilter) hashFn(item []byte, seed uint32) (uint64, uint64) {
+	mod := uint32(len(bf.bitArray) * 64)
+	seedSplit := seed % mod
```
mtrmac, line 31: Shouldn’t this be
```suggestion
	seedSplit := seed % len(item)
```
?

---

(I guess we don’t need to _actually_ care about zero-length items.)

pkg/chunked/bloom_filter.go
mtrmac: OK. I can’t immediately see that this is not uniform, so I’ll go with the practical evidence :)

pkg/chunked/cache_linux.go
```diff
@@ -465,17 +590,20 @@ func writeCache(manifest []byte, format graphdriver.DifferOutputFormat, id strin
...
-		tagLen:    tagLen,
-		tags:      tagsBuffer.Bytes(),
-		vdata:     vdata.Bytes(),
+		digestLen:   digestLen,
+		tagLen:      tagLen,
+		tags:        tagsBuffer.Bytes(),
+		vdata:       vdata.Bytes(),
+		fnames:      fnames.Bytes(),
+		fnamesLen:   len(fnames.Bytes()),
+		bloomFilter: bloomFilterFromTags(tags, digestLen),
```
mtrmac, line 599:
```suggestion
		bloomFilter: bloomFilter,
```
I think here.

mtrmac:
> I've fixed your comments, except [#1870 (comment)](https://github.com/containers/storage/pull/1870#discussion_r1556134916). What would you like me to do here?

https://github.com/containers/storage/pull/1870/files#r1557856169 , or perhaps I’m missing something.

rhatdan:
@giuseppe This is waiting on you now?

pkg/chunked/bloom_filter_test.go
```diff
@@ -29,6 +29,7 @@ var (
 func initCache(sizeCache int) (*cacheFile, string, string, *bloomFilter) {
 	var tagsBuffer bytes.Buffer
 	var vdata bytes.Buffer
+	var fnames bytes.Buffer
```
kolyshkin: nit:
```go
var tagsBuffer, vdata, fnames bytes.Buffer
```

(unless you don't like that style; to me it's less repetition)

pkg/chunked/cache_linux.go
```diff
@@ -423,14 +436,37 @@ func writeCache(manifest []byte, format graphdriver.DifferOutputFormat, id strin
...
+		fnamesMap[name] = pos
+
+		nameBytes := []byte(name)
+
+		if err := binary.Write(&fnames, binary.LittleEndian, uint32(len(nameBytes))); err != nil {
+			return 0, err
+		}
+		if _, err := fnames.Write(nameBytes); err != nil {
+			return 0, err
+		}
```
kolyshkin, line 481: You can avoid `[]byte` conversion here, using `name` directly (as in `len(name)` and `fnames.WriteString(name)`)

kolyshkin (commented):
Left some very minor comments. Overall it looks like a good case for protobuf, or is that an overkill?

It looks like `appendBinaryDigest`'s first argument is always `[]byte{}`. First, it could be `nil`, second, if it's not ever used maybe drop it (and rename the function to `getBinaryDigest`).

pkg/chunked/cache_linux.go
```diff
@@ -311,19 +327,22 @@ func generateFileLocation(path string, offset, len uint64) []byte {
 // generateTag generates a tag in the form $DIGEST$OFFSET@LEN.
 // the [OFFSET; LEN] points to the variable length data where the file locations
 // are stored.  $DIGEST has length digestLen stored in the cache file file header.
-func generateTag(digest string, offset, len uint64) string {
-	return fmt.Sprintf("%s%.20d@%.20d", digest, offset, len)
+func generateTag(digest []byte, offset, len uint64) []byte {
+	tag := append(digest[:], []byte(fmt.Sprintf("%.20d@%.20d", offset, len))...)
```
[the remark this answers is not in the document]
giuseppe: this code is replaced by a next commit.  I'll squash the two patches so it will just disappear

giuseppe:
thanks @mtrmac and @kolyshkin.  I've addressed your last comments and pushed a new version

pkg/chunked/cache_linux.go
```diff
@@ -311,19 +327,22 @@ func generateFileLocation(path string, offset, len uint64) []byte {
...
 	// SetLayerBigData stores a (possibly large) chunk of named data
 	SetLayerBigData(id, key string, data io.Reader) error
 }
 
-func writeCacheFileToWriter(writer io.Writer, tags []string, tagLen, digestLen int, vdata bytes.Buffer, tagsBuffer *bytes.Buffer) error {
-	sort.Strings(tags)
+func writeCacheFileToWriter(writer io.Writer, tags [][]byte, tagLen, digestLen int, vdata, tagsBuffer *bytes.Buffer) error {
+	sort.Slice(tags, func(i, j int) bool {
+		return bytes.Compare(tags[i], tags[j]) == -1
+	})
```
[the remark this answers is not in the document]
giuseppe: had to revert this change as we are using go 1.20

pkg/chunked/cache_linux.go
```diff
@@ -154,6 +161,23 @@ fallback:
...
+	d, err := digest.Parse(stringDigest)
+	if err != nil {
+		return nil, err
+	}
+	digestBytes, err := hex.DecodeString(d.Encoded())
+	if err != nil {
+		return nil, err
+	}
+	algo := []byte(d.Algorithm())
+	buf := make([]byte, len(algo)+1+len(digestBytes))
```
mtrmac, line 174: FIXME

pkg/chunked/cache_linux.go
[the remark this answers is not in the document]
mtrmac: (BTW we are now agreed on updating to Go 1.21 (around https://github.com/containers/skopeo/pull/2297 ). That would probably be a separate PR.)

pkg/chunked/cache_linux.go
mtrmac, line 174:
```suggestion
	buf := make([]byte, 0, len(algo)+1+len(digestBytes))
```
otherwise this produces double-length output, with first half all zeroes.

mtrmac (commented):
LGTM otherwise.

pkg/chunked/cache_linux.go
mtrmac: (Maybe worth writing a tiny unit test … it’s not much more work than testing manually and throwing the test away.)

giuseppe: ouch.  This is embarrassing, thanks for catching it!

Pushed a new version, and a test

mtrmac:
/lgtm

Thanks!

PR #1873, opened 2024-04-04
store: improve error from MountImage()

giuseppe:
improve the error message returned when MountImage() tries to use a layer from a read-only store.

Previously it returned a simple "layer not known".

### DIFF
--- a/store.go
+++ b/store.go
@@ -2820,7 +2820,7 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
 	}
 	defer s.stopUsingGraphDriver()
 
-	rlstore, err := s.getLayerStoreLocked()
+	rlstore, lstores, err := s.bothLayerStoreKindsLocked()
 	if err != nil {
 		return "", err
 	}
@@ -2836,6 +2836,13 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
 	if rlstore.Exists(id) {
 		return rlstore.Mount(id, options)
 	}
+	// check if the layer is in a read-only store, and return a better error message
+	for _, store := range lstores {
+		if store.Exists(id) {
+			return "", fmt.Errorf("mounting read/only store images is not allowed: %w", ErrLayerUnknown)
+		}
+	}
+
 	return "", ErrLayerUnknown
 }
 


giuseppe:
@rhatdan @nalind PTAL

store.go
```diff
@@ -2836,6 +2836,13 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
 	if rlstore.Exists(id) {
 		return rlstore.Mount(id, options)
 	}
+	// check if the layer is in a read-only store, and return a better error message
+	for _, store := range lstores {
+		if store.Exists(id) {
+			return "", fmt.Errorf("layer present in a read-only store: %w", ErrLayerUnknown)
```
rhatdan, line 2842: If I saw this error, I would ask, but I don't want to write to the layer only examine it.  Could we return the layer but marked as read/only?

giuseppe: with MountImage we always mount in read-only mode, but to mount it we still need to modify the store and need to grab a writeable lock there.

I don't see an immediate way to "copy up" the layer in a transparent way, we would need the equivalent of `CreateContainer` and reference the read-only layer as a lower layer but we will end up with two different layers.  I think it is cleaner to handle it manually with `CreateContainer` at least it is clear what is happening in the store

rhatdan: How about "mounting read/only store images is not allowed"

giuseppe: sure, fixed and pushed a new version

rhatdan:
LGTM

rhatdan:
@mtrmac @saschagrunert @flouthoc @vrothberg PTAL

flouthoc (approved):
LGTM

store.go
```diff
@@ -2836,6 +2836,13 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
 	if rlstore.Exists(id) {
 		return rlstore.Mount(id, options)
 	}
+	// check if the layer is in a read-only store, and return a better error message
+	for _, store := range lstores {
+		if store.Exists(id) {
```
mtrmac, line 2841: This requires `store.startReading`.

(A possible micro-optimization would be to end the `stopWriting` scope before this new code, to improve chance of concurrent accesses.)

giuseppe: thanks.  Opened a follow up PR: https://github.com/containers/storage/pull/1877

PR #1875, opened 2024-04-04
fileutils: new function to check for path existence

giuseppe:
On unix it uses the access syscall which is faster than stat for simply checking the existence of a file.
    
with a file in the cache I see on my machine:
    
$ go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/containers/storage/pkg/fileutils
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
BenchmarkPathExists-12           1000000              1037 ns/op
BenchmarkStat-12                  897534              1330 ns/op
    
I'd expect even better results if the file is not already in the cache.
    
### DIFF
--- a/drivers/aufs/aufs.go
+++ b/drivers/aufs/aufs.go
@@ -41,6 +41,7 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/chrootarchive"
 	"github.com/containers/storage/pkg/directory"
+	"github.com/containers/storage/pkg/fileutils"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/locker"
 	mountpk "github.com/containers/storage/pkg/mount"
@@ -243,7 +244,7 @@ func (a *Driver) Metadata(id string) (map[string]string, error) {
 // Exists returns true if the given id is registered with
 // this driver
 func (a *Driver) Exists(id string) bool {
-	if _, err := os.Lstat(path.Join(a.rootPath(), "layers", id)); err != nil {
+	if err := fileutils.Lexists(path.Join(a.rootPath(), "layers", id)); err != nil {
 		return false
 	}
 	return true
@@ -431,7 +432,7 @@ func atomicRemove(source string) error {
 	case err == nil, os.IsNotExist(err):
 	case os.IsExist(err):
 		// Got error saying the target dir already exists, maybe the source doesn't exist due to a previous (failed) remove
-		if _, e := os.Stat(source); !os.IsNotExist(e) {
+		if e := fileutils.Exists(source); !os.IsNotExist(e) {
 			return fmt.Errorf("target rename dir '%s' exists but should not, this needs to be manually cleaned up: %w", target, err)
 		}
 	default:
--- a/drivers/btrfs/btrfs.go
+++ b/drivers/btrfs/btrfs.go
@@ -32,6 +32,7 @@ import (
 
 	graphdriver "github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/directory"
+	"github.com/containers/storage/pkg/fileutils"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/mount"
 	"github.com/containers/storage/pkg/parsers"
@@ -589,11 +590,11 @@ func (d *Driver) setStorageSize(dir string, driver *Driver) error {
 // Remove the filesystem with given id.
 func (d *Driver) Remove(id string) error {
 	dir := d.subvolumesDirID(id)
-	if _, err := os.Stat(dir); err != nil {
+	if err := fileutils.Exists(dir); err != nil {
 		return err
 	}
 	quotasDir := d.quotasDirID(id)
-	if _, err := os.Stat(quotasDir); err == nil {
+	if err := fileutils.Exists(quotasDir); err == nil {
 		if err := os.Remove(quotasDir); err != nil {
 			return err
 		}
@@ -669,7 +670,7 @@ func (d *Driver) ReadWriteDiskUsage(id string) (*directory.DiskUsage, error) {
 // Exists checks if the id exists in the filesystem.
 func (d *Driver) Exists(id string) bool {
 	dir := d.subvolumesDirID(id)
-	_, err := os.Stat(dir)
+	err := fileutils.Exists(dir)
 	return err == nil
 }
 
--- a/drivers/devmapper/deviceset.go
+++ b/drivers/devmapper/deviceset.go
@@ -22,6 +22,7 @@ import (
 	graphdriver "github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/devicemapper"
 	"github.com/containers/storage/pkg/dmesg"
+	"github.com/containers/storage/pkg/fileutils"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/loopback"
 	"github.com/containers/storage/pkg/mount"
@@ -257,7 +258,7 @@ func (devices *DeviceSet) hasImage(name string) bool {
 	dirname := devices.loopbackDir()
 	filename := path.Join(dirname, name)
 
-	_, err := os.Stat(filename)
+	err := fileutils.Exists(filename)
 	return err == nil
 }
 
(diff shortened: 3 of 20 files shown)


pkg/fileutils/fileutils_darwin.go
```diff
@@ -25,3 +25,13 @@ func GetTotalUsedFds() int {
 
 	return len(fds) - 1
 }
+
+// PathExists checks whether a file or directory exists at the given path.
+func PathExists(path string, followSymlinks bool) error {
```
rhatdan, line 30: Should the name of this function just return an err.

err:=fileutils.Exist(PATH);
err = nil -> True
err == ENOEXIST == False
err == (!nil && not ENOEXIST) == err.

giuseppe: I didn't change the type of the returned value because it is easier to replace the existing calls to `os.Stat()`.  Every existing call to `_, err := os.Stat(...)` simply becomes `err := fileutils.PathExists(..., true)` and the existing code for handling `err` is the same.

rhatdan: I don't see the difference, since my function will do the same, except there is two functions, Exists, and Lexists, which is similar to Stat and Lstat.

giuseppe: pushed a new version with Exist and Lexist

giuseppe:
I think keeping them separate helps with the review and potentially the backport.  That's what @mtrmac asked me to do last time I've sent a big patch

rhatdan:
Except that you are really only making a single change here.  Adding two functions and then using them everywhere.  I don't see any difference between the changes, so no need to split them out.

rhatdan:
@mtrmac @nalind @flouthoc @saschagrunert PTAL

pkg/fileutils/fileutils_unix.go
```diff
@@ -20,3 +21,18 @@ func GetTotalUsedFds() int {
 	}
 	return -1
 }
+
+// Exist checks whether a file or directory exists at the given path.
+// It uses unix.Access which is a faster operation compared to os.Stat for
```
nalind, line 26: This should name the function it really calls, if it mentions it at all.

pkg/fileutils/fileutils_test.go
```diff
@@ -624,3 +626,58 @@ func TestMatchesAmount(t *testing.T) {
...
+	_, err2 = os.Lstat(symlinkPath)
+	assert.Equal(t, errors.As(err, &syscallErr), errors.As(err2, &syscallErr), "same error for existing symlink (follow=true)")
+
+	err = Lexist(danglingSymlinkPath)
+	_, err2 = os.Stat(danglingSymlinkPath)
+	assert.Equal(t, errors.As(err, &syscallErr), errors.As(err2, &syscallErr), "same error for not existing symlink")
+
+	err = Exist(danglingSymlinkPath)
+	_, err2 = os.Lstat(danglingSymlinkPath)
+	assert.Equal(t, errors.As(err, &syscallErr), errors.As(err2, &syscallErr), "same error for not existing symlink (follow=true)")
```
mtrmac, line 666: - the `assert.Equal` is just comparing the bool returned form `errors.As`, i.e. wether the errors are of the same _type_, not that the values are the same. Is it intentional that this is what the test is checking?
- (The comments refer to `follow=true`, I suppose that should be updated to the new function names instead.)

pkg/fileutils/fileutils_unix.go
mtrmac: I’d lean towards not mentioning it in the API documentation here — the job is to check for existence.

That `Access` is faster is an implementation detail — a non-obvious detail, so worth a comment — but that comment can be moved inside the function.

mtrmac (commented):
- Efficiency improvements are always nice
- Looking at `git grep -C5 '_, .*os.Lstat'` and `it grep -C5 '_, .*os.Stat'`, there seem to be more places where this could be used — e.g. `drivers/aufs`, and other files in `pkg/archive`. That’s not blocking this PR, of course.
- (Aesthetically, the name feels a bit off to me. `fileutils.{Exists,Lexists}` would be a small improvement, but really I’d expect a function named like that to return a `bool`, not an `error`. OTOH the returned error is clearly useful… and I can’t think of a good name. `EnsurePathExists` is so ugly.)
- Do we want this in a ~stable public API of the module? If it is expected to be beneficial in other repos like Podman, sure, let’s make it stable. OTOH making it private would make the naming concerns much less relevant. I don’t really have an opinion here, I just want to make sure this is an intentional decision.

mtrmac:
> I think keeping them separate helps with the review and potentially the backport. That's what @mtrmac asked me to do last time I've sent a big patch

My personal view of this is that it is ideal if there is one _idea_ per patch. Introduce one new feature. Fix one bug, or a homogenous class of bugs. Do one refactor. (Ideally the refactors are so trivial that in case of a merge conflict, they could just be thrown away and made just based on the commit description.)

In that sense, the migration to the new API could well be a single commit. OTOH more commits never really hurt — and they make _allocating time_ for reviews a tiny bit easier, because I know there are natural stopping points and I know before starting to look that path the code I can make definite progress in 15 minutes, (as opposed to allocating an hour or two to stare at a huge do-12-things commit; that can happen only a few times a day).

Or course, everyone thinks and works a bit differently. Seeing small commits probably makes me more likely to take on the review, and to do it quickly — OTOH there is a definite cost to creating those commits in the first place.

pkg/fileutils/fileutils_darwin.go
```diff
@@ -25,3 +25,16 @@ func GetTotalUsedFds() int {
 
 	return len(fds) - 1
 }
+
+// Exist checks whether a file or directory exists at the given path.
+func Exist(path string) error {
+	_, err := os.Stat(path)
```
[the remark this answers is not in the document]
giuseppe: I've no way to test it (and if it is worth in terms of performance)

giuseppe:
> * 
> * Looking at `git grep -C5 '_, .*os.Lstat'` and `it grep -C5 '_, .*os.Stat'`, there seem to be more places where this could be used — e.g. `drivers/aufs`, and other files in `pkg/archive`. That’s not blocking this PR, of course.

I've fixed these callers too.

> * (Aesthetically, the name feels a bit off to me. `fileutils.{Exists,Lexists}` would be a small improvement, but really I’d expect a function named like that to return a `bool`, not an `error`. OTOH the returned error is clearly useful… and I can’t think of a good name. `EnsurePathExists` is so ugly.)

Renamed the functions to `fileutils.{Exists, Lexists}.

My first version of the API looked like `func PathExists(path string) (bool, error)` but it made more difficult and error-prone to fix the current callers (if we really wanted to use the `found` return value).  With the current API instead we can mechanically replace the current code with `_, err := os.Stat(...) => err := fileutils.Exists(...)` and `_, err := os.Lstat(...) => => err := fileutils.Lexists(...)` without changing any of the conditions in place.

> * Do we want this in a ~stable public API of the module? If it is expected to be beneficial in other repos like Podman, sure, let’s make it stable. OTOH making it private would make the naming concerns much less relevant. I don’t really have an opinion here, I just want to make sure this is an intentional decision.

Yes, let's make it stable and fix the same pattern in other repos.

pkg/fileutils/fileutils_test.go
giuseppe: thanks, fixed and applied to the last version

pkg/fileutils/fileutils_test.go
```diff
@@ -624,3 +626,71 @@ func TestMatchesAmount(t *testing.T) {
...
+		assert.Equal(t, err1 == nil, err2 == nil, "only one error is set")
+		if err1 == nil {
+			return
+		}
+		// on Linux validates that the syscall error is the same
+		if runtime.GOOS == "linux" {
+			var syscallErr1 syscall.Errno
+			var syscallErr2 syscall.Errno
+			assert.True(t, errors.As(err1, &syscallErr1), "wrong error type")
+			assert.True(t, errors.As(err2, &syscallErr2), "wrong error type")
```
mtrmac, line 652: (FWIW `assert.ErrorAs` exists. This is just fine.)

mtrmac (commented):
LGTM; let’s maybe wait a few hours to confirm consensus on the macOS situation.

pkg/fileutils/fileutils_darwin.go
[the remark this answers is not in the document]
giuseppe: thanks for testing it, I will add it to Darwin too

pkg/fileutils/fileutils_windows.go
```diff
@@ -1,5 +1,9 @@
 package fileutils
 
+import (
+	"os"
```
mtrmac, line 4: This is now unused, so I’d expect it to cause a build failure.

mtrmac (commented):
LGTM other than that one thing, feel free to merge without another review.

giuseppe:
@nalind PTAL

giuseppe:
fix for c/common: https://github.com/containers/common/pull/1953

PR #1877, opened 2024-04-08
store: lock stores before calling Exists

giuseppe:
we need to lock the stores for read before checking if a layer exists.

follow up for 466ee04ed72429867df1fea80d992722ab571415

addressing the comment here: https://github.com/containers/storage/pull/1873#discussion_r1553620140

### DIFF
--- a/store.go
+++ b/store.go
@@ -2824,21 +2824,34 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if err := rlstore.startWriting(); err != nil {
-		return "", err
-	}
-	defer rlstore.stopWriting()
-
 	if options.UidMaps != nil || options.GidMaps != nil {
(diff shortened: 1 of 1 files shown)


store.go
```diff
@@ -2824,21 +2824,28 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
...
 
-	if rlstore.Exists(id) {
+	if err := rlstore.startWriting(); err != nil {
+		return "", err
+	}
+	exists := rlstore.Exists(id)
+	rlstore.stopWriting()
+
+	if exists {
 		return rlstore.Mount(id, options)
```
mtrmac, line 2838: This also requires a write lock.

mtrmac (commented):
Thanks for the followup!

store.go
giuseppe: ouch, right.

Fixed now

mtrmac (approved):
/lgtm

PR #1879, opened 2024-04-09
chunked: fix unmarshaling of file names

giuseppe:
The getString() function was used to extract string values, but it doesn't handle escaped characters.  Replace it with iter.ReadString() that is slower but handles escaped characters correctly.

Closes: https://github.com/containers/storage/issues/1878

### DIFF
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -652,45 +652,9 @@ func (c *layersCache) findChunkInOtherLayers(chunk *internal.FileMetadata) (stri
 }
 
 func unmarshalToc(manifest []byte) (*internal.TOC, error) {
-	var buf bytes.Buffer
-	count := 0
 	var toc internal.TOC
 
 	iter := jsoniter.ParseBytes(jsoniter.ConfigFastest, manifest)
-	for field := iter.ReadObject(); field != ""; field = iter.ReadObject() {
-		if strings.ToLower(field) != "entries" {
-			iter.Skip()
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@nalind @mtrmac @saschagrunert @flouthoc PTAL

mtrmac (commented):
- Does `xattrs` need this as well?
- If this does not use `StringsBuf`, the fields should not be counted in `count`
- (I am ambivalent about the concept of `StringsBuf`, there is no documented rationale. Is it to save on per-object allocation cost?)
- That `ReadStringAsSlice` does not process escapes is understandable but not immediately obvious to me. Consider turning `getString([]byte])` into a `getStringWithoutEscapes(iter)`, to document that behavior in one place (and to save a bit on code size / repetitiveness).

giuseppe:
> * (I am ambivalent about the concept of `StringsBuf`, there is no documented rationale. Is it to save on per-object allocation cost?)

yeah it was done exactly for that reason, but now that we are not using for `name` is unclear if it is still worth having.

For now, I've simplified the implementation and dropped the StringsBuf performance hack, we can revisit later if needed

mtrmac:
(Just to be explicit, I’m completely fine with `StringsBuf` continuing to exist. I have no reason to prefer either way.)

mtrmac (commented):
LGTM for correctness, the simplification is nice.

I have no idea what’s the performance impact, I trust your judgement on this.

PR #1882, opened 2024-04-09
bring read mappings proc into c/storage

kannon92:
As I was working on https://github.com/cri-o/cri-o/pull/7990, I found that ReadMappingsProc was the only function that was missing from pkg/unshare.

@giuseppe mention that this would be a good candidate for c/storage.

This PR adds this function to pkg/unshare. Once this is released, we could drop this from podman.

### DIFF
--- a/pkg/unshare/unshare_linux.go
+++ b/pkg/unshare/unshare_linux.go
@@ -708,6 +708,37 @@ func GetSubIDMappings(user, group string) ([]specs.LinuxIDMapping, []specs.Linux
 	return uidmap, gidmap, nil
 }
 
+// ReadMappingsProc parses and returns the ID mappings at the specified path.
+func ReadMappingsProc(path string) ([]idtools.IDMap, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	mappings := []idtools.IDMap{}
+
+	buf := bufio.NewReader(file)
+	for {
+		line, _, err := buf.ReadLine()
+		if err != nil {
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

giuseppe (changes requested):
second thoughts: should we just use GetHostIDMappings in Podman?

rhatdan:
@giuseppe they are trying to stop using Podman code.

rhatdan:
We could make Podman use this interface possibly?

rhatdan:
LGTM

giuseppe:
> @giuseppe they are trying to stop using Podman code.

I understand that, but if we already have an existing function doing the same thing (as it seems to be `GetHostIDMappings`), we can just use that.

@kannon92 can you please try using `GetHostIDMappings` in CRI-O?  If that works for you, we can drop `ReadMappingsProc`

giuseppe:
I've opened a PR for Podman to drop the function and use the existing function from c/storage: https://github.com/containers/podman/pull/22328

We can do the same in CRI-O and not require `podman/pkg/rootless`

kannon92:
I see. I did use that in https://github.com/cri-o/cri-o/pull/7990.

We can drop this PR if you think the above one is the right way to go.

giuseppe:
> I see. I did use that in [cri-o/cri-o#7990](https://github.com/cri-o/cri-o/pull/7990).
> 
> We can drop this PR if you think the above one is the right way to go.

yes let's just revert this change.  I've opened a PR: https://github.com/containers/storage/pull/1884

Issue #1306, opened 2022-08-11
AdditionalImageStores configuration


meox:

rhatdan:
I would be ok with this.  BUT storage.conf right now does not handle overriding fields.  We only replace and fall back to defaults.  This is different then what we do in containers.conf and registries.conf, I believe.

Changing this to inheritance my be a breaking change.

Changing it to only inherit from storage.conf.d/* Would not be a breaking change, though.

/usr/share/containers/storage.conf -> /etc/containers/storage.conf/ -> $HOME/.config/storage.conf override completely

/etc/containers/storage.conf.d/* and $HOME/.config/storage.conf/.d/* would only override fields in the default storage.conf.

meox:
Overrides in which way?
I mean if I have multiple have multiple storage.conf.d files and then I run `podman images` I will able to see all the images from all storage?

rhatdan:
Right now if I have /usr/containers/storage.conf and /etc/containers/storage.conf, None of the fields in /usr/containers/storage.conf are used or even read.
We would need to change this.  So that fields in /etc/containers/storage.conf/.d would be merged (replacing just the specified fields) that are present in /etc/containers/storage.conf.

If two drop in files used the same fields, then the last one read would win.

meox:
I'd like to have the opportunity to add multiple storage (AdditionalImageStores) just adding multiple config files in a folder.
What @rhatdan said seems not able to cover my proposed use case.

rhatdan:
Yes you would have to add multiple images to a single file.

meox:
> Yes you would have to add multiple images to a single file.

ok but is just like now if I understand correctly

rhatdan:
Best way to make this happen is to open a PR.

jacksgt:
Alright, at least it's good to know that a PR is welcome for this :-)

rhatdan:
Yes I think this is a good idea.  If we had it to do all over again, I think we would have done storage.conf the same as containers.conf.

Since .d directories are a new feature, I see them as enhancing the storage.conf file rather then replacing.

PR #1885, opened 2024-04-12
Add support for drop-in config

harche:
Helps in use cases where a drop-in config could help by allowing users to update a subsection of the config instead of touching an entire config file. 

e.g.

1. https://github.com/containers/storage/issues/1306
2. https://github.com/openshift/enhancements/pull/1600/files#diff-10bfa811cdabc678cbfd947dbeaf04b63ebd24e8d9ab5a05c7abbb998cb72f8aR135
3. https://github.com/kubernetes/enhancements/issues/4191

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -49,6 +49,12 @@ var (
 	defaultConfigFile = SystemConfigFile
 	// DefaultStoreOptions is a reasonable default set of options.
 	defaultStoreOptions StoreOptions
+
+	// defaultOverrideConfigFile path to override the default system wide storage.conf file
+	defaultOverrideConfigFile = "/etc/containers/storage.conf"
+
+	// defaultDropInConfigDir path to the folder containing drop in config files
+	defaultDropInConfigDir = defaultOverrideConfigFile + ".d"
 )
 
 func loadDefaultStoreOptions() {
@@ -114,11 +120,53 @@ func loadDefaultStoreOptions() {
 
 // loadStoreOptions returns the default storage ops for containers
 func loadStoreOptions() (StoreOptions, error) {
-	storageConf, err := DefaultConfigFile()
+	baseConf, err := DefaultConfigFile()
 	if err != nil {
 		return defaultStoreOptions, err
 	}
-	return loadStoreOptionsFromConfFile(storageConf)
+
+	// Load the base config file
+	baseOptions, err := loadStoreOptionsFromConfFile(baseConf)
+	if err != nil {
+		return defaultStoreOptions, err
+	}
+
+	if _, err := os.Stat(defaultDropInConfigDir); err == nil {
+		// The directory exists, so merge the configuration from this directory
+		err = mergeConfigFromDirectory(&baseOptions, defaultDropInConfigDir)
+		if err != nil {
+			return defaultStoreOptions, err
+		}
+	} else if !os.IsNotExist(err) {
+		// There was an error other than the directory not existing
+		return defaultStoreOptions, err
+	}
+
+	return baseOptions, nil
+}
+
+func mergeConfigFromDirectory(baseOptions *StoreOptions, configDir string) error {
+	return filepath.Walk(configDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
+
+		// Only consider files with .conf extension
+		if filepath.Ext(path) != ".conf" {
+			return nil
+		}
+
+		// Load drop-in options from the current file
+		err = ReloadConfigurationFile(path, baseOptions, false)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
 }
 
 // usePerUserStorage returns whether the user private storage must be used.
@@ -399,7 +447,7 @@ func ReloadConfigurationFileIfNeeded(configFile string, storeOptions *StoreOptio
 		return nil
 	}
 
-	if err := ReloadConfigurationFile(configFile, storeOptions); err != nil {
+	if err := ReloadConfigurationFile(configFile, storeOptions, true); err != nil {
 		return err
 	}
 
@@ -412,7 +460,7 @@ func ReloadConfigurationFileIfNeeded(configFile string, storeOptions *StoreOptio
 
 // ReloadConfigurationFile parses the specified configuration file and overrides
 // the configuration in storeOptions.
-func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) error {
+func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions, initializeOptions bool) error {
 	config := new(TomlConfig)
 
 	meta, err := toml.DecodeFile(configFile, &config)
@@ -428,8 +476,11 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) erro
 		}
 	}
 
-	// Clear storeOptions of previous settings
-	*storeOptions = StoreOptions{}
+	if initializeOptions {
+		// Clear storeOptions of previous settings
+		*storeOptions = StoreOptions{}
+	}
+
 	if config.Storage.Driver != "" {
 		storeOptions.GraphDriverName = config.Storage.Driver
 	}
@@ -519,8 +570,13 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) erro
 		storeOptions.PullOptions = config.Storage.Options.PullOptions
 	}
 
-	storeOptions.DisableVolatile = config.Storage.Options.DisableVolatile
-	storeOptions.TransientStore = config.Storage.TransientStore
+	if config.Storage.Options.DisableVolatile {
+		storeOptions.DisableVolatile = config.Storage.Options.DisableVolatile
+	}
+
+	if config.Storage.TransientStore {
+		storeOptions.TransientStore = config.Storage.TransientStore
+	}
 
 	storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, cfg.GetGraphDriverOptions(storeOptions.GraphDriverName, config.Storage.Options)...)
 
--- a/store.go
+++ b/store.go
@@ -3710,7 +3710,7 @@ func DefaultConfigFile() (string, error) {
 // the configuration in storeOptions.
 // Deprecated: Use types.ReloadConfigurationFile, which can return an error.
 func ReloadConfigurationFile(configFile string, storeOptions *types.StoreOptions) {
-	_ = types.ReloadConfigurationFile(configFile, storeOptions)
+	_ = types.ReloadConfigurationFile(configFile, storeOptions, true)
 }
 
 // GetDefaultMountOptions returns the default mountoptions defined in container/storage
--- a/types/options_darwin.go
+++ b/types/options_darwin.go
@@ -8,8 +8,6 @@ const (
 	SystemConfigFile        = "/usr/share/containers/storage.conf"
 )
 
-var defaultOverrideConfigFile = "/etc/containers/storage.conf"
-
 // canUseRootlessOverlay returns true if the overlay driver can be used for rootless containers
 func canUseRootlessOverlay(home, runhome string) bool {
 	return false
--- a/types/options_freebsd.go
+++ b/types/options_freebsd.go
@@ -8,11 +8,6 @@ const (
 	SystemConfigFile        = "/usr/local/share/containers/storage.conf"
 )
 
-// defaultConfigFile path to the system wide storage.conf file
-var (
-	defaultOverrideConfigFile = "/usr/local/etc/containers/storage.conf"
-)
-
 // canUseRootlessOverlay returns true if the overlay driver can be used for rootless containers
 func canUseRootlessOverlay(home, runhome string) bool {
 	return false
--- a/types/options_linux.go
+++ b/types/options_linux.go
@@ -16,11 +16,6 @@ const (
 	SystemConfigFile        = "/usr/share/containers/storage.conf"
 )
 
-// defaultConfigFile path to the system wide storage.conf file
-var (
-	defaultOverrideConfigFile = "/etc/containers/storage.conf"
-)
-
 // canUseRootlessOverlay returns true if the overlay driver can be used for rootless containers
 func canUseRootlessOverlay(home, runhome string) bool {
 	// we check first for fuse-overlayfs since it is cheaper.
--- a/types/options_test.go
+++ b/types/options_test.go
@@ -143,7 +143,7 @@ func TestSetRemapUIDsGIDsOpts(t *testing.T) {
 		},
 	}
 
-	err := ReloadConfigurationFile("./storage_test.conf", &remapOpts)
+	err := ReloadConfigurationFile("./storage_test.conf", &remapOpts, true)
 	require.NoError(t, err)
 	if !reflect.DeepEqual(uidmap, remapOpts.UIDMap) {
 		t.Errorf("Failed to set UIDMap: Expected %v Actual %v", uidmap, remapOpts.UIDMap)
@@ -185,7 +185,7 @@ remap-group = "%s"
 
 	mappings, err := idtools.NewIDMappings(user, user)
 	require.NoError(t, err)
-	err = ReloadConfigurationFile(configPath, &remapOpts)
+	err = ReloadConfigurationFile(configPath, &remapOpts, true)
 	require.NoError(t, err)
 	if !reflect.DeepEqual(mappings.UIDs(), remapOpts.UIDMap) {
 		t.Errorf("Failed to set UIDMap: Expected %v Actual %v", mappings.UIDs(), remapOpts.UIDMap)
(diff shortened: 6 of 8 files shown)


types/options.go
```diff
@@ -114,11 +120,102 @@ func loadDefaultStoreOptions() {
...
+	})
+
+	if err != nil {
+		return baseOptions, err
+	}
+	return baseOptions, nil
+}
+
+func mergeStoreOptions(base, dropIn StoreOptions) StoreOptions {
+	if dropIn.RunRoot != "" {
```
haircommander, line 173: instead of individually adding fields (which won't scale well, can we just serialize into the struct: https://github.com/cri-o/cri-o/blob/256fda5ac98de1d67e26133d0b25df2a74e2ebd2/pkg/config/config.go#L705

harche: Thanks. Turns out they already have a function that does that, https://github.com/containers/storage/blob/1fd0dc1d20e523f54369e874d510bf62a23f1574/types/options.go#L415

harche: Not really, they don't serialize and automatically update.

harche: @haircommander I updated the code to reuse existing [ReloadConfigurationFile](https://github.com/containers/storage/blob/1fd0dc1d20e523f54369e874d510bf62a23f1574/types/options.go#L415C6-L415C29) (instead of writing a new merge config logic) which seem to be doing [more](https://github.com/containers/storage/blob/1fd0dc1d20e523f54369e874d510bf62a23f1574/types/options.go#L446) than just updating the options. `ReloadConfigurationFile` is in use already, so I think we can safely rely on it. Although I admit it is not as scalable as the cri-o approach is, but using `ReloadConfigurationFile` allows us to reuse their existing way of reading and parsing a config file. 

WDYT?

types/options.go
```diff
@@ -49,6 +49,12 @@ var (
 	defaultConfigFile = SystemConfigFile
 	// DefaultStoreOptions is a reasonable default set of options.
 	defaultStoreOptions StoreOptions
+
+	// defaultOverrideConfigFile path to override the default system wide storage.conf file
+	defaultOverrideConfigFile = "/etc/containers/storage.conf"
+
+	// defaultDropInConfigDir path to the folder containing drop in config files
+	defaultDropInConfigDir = "/etc/containers/storage.conf.d"
```
haircommander, line 57: these may need to remain in the platform specific paths. i'm not sure if there are windows paths that need to be supported.

harche: I moved it here because all platform specific files had same value for that variable, including windows one. Although it does look incorrect for the windows, but I just moved it from [options_windows.go](https://github.com/containers/storage/pull/1885/files#diff-fa5a62b5c68d186af67295d75ab2d2978b00b39db884f9828686e68c5ccbbc91)

types/options.go
```diff
@@ -114,11 +120,50 @@ func loadDefaultStoreOptions() {
...
+
+		// Load drop-in options from the current file
+		err = ReloadConfigurationFile(path, baseOptions)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+	if err != nil {
```
rphillips, line 163: return filepath.Walk(...)... The if statements are not necessary.

types/options.go
```diff
@@ -114,11 +120,50 @@ func loadDefaultStoreOptions() {
...
+	}
+
+	// Load the base config file
+	baseOptions, err := loadStoreOptionsFromConfFile(baseConf)
+	if err != nil {
+		return defaultStoreOptions, err
+	}
+
+	if _, err := os.Stat(defaultDropInConfigDir); err != nil && os.IsNotExist(err) {
+		return defaultStoreOptions, err
```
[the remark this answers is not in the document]
harche: yes, you are right. even if the `defaultDropInConfigDir` does not exist then there is no need to return the error. Thanks for point it out. I will update the changes.

types/options.go
harche: Thanks. updated.

types/options.go
harche: Also, if I don't add it here, then just like `defaultOverrideConfigFile`, I will have to add new `defaultDropInConfigDir` to all platform specific files, otherwise the compilation would fail on those platforms.

types/options.go
```diff
@@ -114,11 +120,48 @@ func loadDefaultStoreOptions() {
...
+
+func mergeConfigFromDirectory(baseOptions *StoreOptions, configDir string) error {
+	return filepath.Walk(configDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
+
```
rphillips, line 156: Nice update. Should check for file extension here.

harche: But a text file without a `.conf` or `.toml` can also contain a valid drop-in config. For any reason the content is not valid (irrespective of the extension) it will fail in `ReloadConfigurationFile` below while parsing the toml.

rphillips: Hmm. We support a file without an extension?

harche: as long as the content of the file is valid, it shouldn't matter, right? 

crio doesn't enforce the extension either - https://github.com/cri-o/cri-o/blob/256fda5ac98de1d67e26133d0b25df2a74e2ebd2/pkg/config/config.go#L776

rphillips: Usually there is a filter, but this is fine.

haircommander: one risk is temporary files written by editor and the like. It mostly only is an issue if we're reloading automatically. If it's manually triggered (like here) then the user usually has written and edited the editor before the reload happens...

haircommander: I think requiring a .conf or .toml extension could be a nice change both here and cri-o. @sohankunkerkar added a filter for the kubelet's drop-in work

harche: Thanks @rphillips @haircommander @rhatdan for this feedback. I have updated the code to select files with `.conf` extension only and also updated corresponding unit test to verify that.

rhatdan:
Could you open a [WIP] PR against Podman to test this feature, and make sure it does not break anything in it's ci/cd system.

types/options.go
```diff
@@ -114,11 +120,53 @@ func loadDefaultStoreOptions() {
...
+		// Only consider files with .conf extension
+		if filepath.Ext(path) != ".conf" {
+			return nil
+		}
+
+		// Load drop-in options from the current file
+		err = ReloadConfigurationFile(path, baseOptions)
```
mtrmac, line 163: Very quickly skimming the implementation of `ReloadConfigurationFile`, it is not designed to merge things at all; it starts with `*storeOptions = StoreOptions{}`.

harche:
> Could you open a [WIP] PR against Podman to test this feature, and make sure it does not break anything in it's ci/cd system.

@rhatdan I opened https://github.com/containers/podman/pull/22484. Most of the jobs were green, apart from the one for the `mac`.  I am not able to trigger the retest of that job, it needs `/ok-to-test` label.

rhatdan:
@nalind @giuseppe @saschagrunert @vrothberg PTAL

types/options.go
```diff
@@ -114,11 +120,53 @@ func loadDefaultStoreOptions() {
...
+
+	// Load the base config file
+	baseOptions, err := loadStoreOptionsFromConfFile(baseConf)
+	if err != nil {
+		return defaultStoreOptions, err
+	}
+
+	if _, err := os.Stat(defaultDropInConfigDir); err == nil {
+		// The directory exists, so merge the configuration from this directory
+		err = mergeConfigFromDirectory(&baseOptions, defaultDropInConfigDir)
```
giuseppe, line 136: I am not sure we should hardcode a specific directory, but instead try to load each file from the `$STORAGE_CONF + ".d"'` directory, where `$STORAGE_CONF` is whatever file we decided to use

types/options.go
```diff
@@ -114,11 +120,53 @@ func loadDefaultStoreOptions() {
...
+	baseConf, err := DefaultConfigFile()
 	if err != nil {
 		return defaultStoreOptions, err
 	}
-	return loadStoreOptionsFromConfFile(storageConf)
+
+	// Load the base config file
+	baseOptions, err := loadStoreOptionsFromConfFile(baseConf)
+	if err != nil {
+		return defaultStoreOptions, err
```
saschagrunert, line 131: Let's wrap those errors, otherwise they're hard to trace down the API, like:
```suggestion
		return defaultStoreOptions, fmt.Errorf("load storage options from config file: %w", err)
```

types/options.go
saschagrunert, line 57: Let's make those `const` if they're not variable.

types/options.go
harche: Thanks @giuseppe, that's a good idea. I modified the code to set the drop-in config dir to `storage conf path + .d`

types/options.go
harche: Updated, thanks.

types/options.go
harche: Because of this change https://github.com/containers/storage/pull/1885#discussion_r1581176041, we may not be able to make it a `const`.

types/options.go
mtrmac: A reminder: ^^^. I can’t see how this works at all. What am I missing?

harche: The included unit test `TestMergeConfigFromDirectory` gives me the impression that the change is working as expected, but maybe I am missing something. Can you suggest a change to that test which can bring out the failure you are anticipating? 

The way I am looking at it is, for every drop-in config file we are initializing `*storeOptions = StoreOptions{}` and then merging it with the earlier values. But again, there is a good chance I might have overlooked something. A test case would definitely help me understand it better.

mtrmac: That test is not testing _merging_ at all.

As just a starting point (I don’t understand all of the c/storage option loading code, too many functions and too many code paths and too many hooks; I’m scared of that):
```patch
diff --git a/types/options_test.go b/types/options_test.go
index bd5a70c8b..92d35dc6e 100644
--- a/types/options_test.go
+++ b/types/options_test.go
@@ -227,8 +227,7 @@ graphroot = 'temp/graph2'`,
 runroot = 'should/ignore'
 graphroot = 'should/ignore'`,
                `[storage]
-runroot = 'temp/run3'
-graphroot = 'temp/graph3'`,
+runroot = 'temp/run3'`,
                `[storage]
 runroot = 'temp/run4'
 graphroot = 'temp/graph4'`,
@@ -242,14 +241,16 @@ graphroot = 'temp/graph4'`,
 
        // Set base options
        baseOptions := StoreOptions{
-               RunRoot:   "initial/run",
-               GraphRoot: "initial/graph",
+               RunRoot:        "initial/run",
+               GraphRoot:      "initial/graph",
+               TransientStore: true,
        }
 
        // Expected results after merging configurations from only .conf files
        expectedOptions := StoreOptions{
-               RunRoot:   "temp/run3", // Last .conf file (config3.conf) read overrides earlier values
-               GraphRoot: "temp/graph3",
+               RunRoot:        "temp/run3", // Last .conf file (config3.conf) read overrides earlier values
+               GraphRoot:      "temp/graph2",
+               TransientStore: true,
        }
 
        // Run the merging function
@@ -258,7 +259,5 @@ graphroot = 'temp/graph4'`,
                t.Fatalf("Error merging config from directory: %v", err)
        }
 
-       if baseOptions.RunRoot != expectedOptions.RunRoot || baseOptions.GraphRoot != expectedOptions.GraphRoot {
-               t.Errorf("Expected RunRoot to be %q and GraphRoot to be %q, got RunRoot %q and GraphRoot %q", expectedOptions.RunRoot, expectedOptions.GraphRoot, baseOptions.RunRoot, baseOptions.GraphRoot)
-       }
+       assert.DeepEqual(t, expectedOptions, baseOptions)
 }
```

harche: Thanks @mtrmac . 

/hold until we fix it.

types/options.go
```diff
@@ -519,8 +570,13 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) erro
 		storeOptions.PullOptions = config.Storage.Options.PullOptions
 	}
 
-	storeOptions.DisableVolatile = config.Storage.Options.DisableVolatile
-	storeOptions.TransientStore = config.Storage.TransientStore
+	if config.Storage.Options.DisableVolatile {
```
mtrmac, line 573: How do I set `disableFolatile = false` in the drop-in?

It seems to me that all of this is going to require some other approach, and probably very detailed testing.

mtrmac: Or, alternatively, be _very_ strict and _very_ restrictive about which options are supported in drop-ins, support/test only those, and hard-fail if any others are present. Then we can add support for more options over time … _assuming_ it’s sufficiently certain adding the others will actually be possible in the future.

rphillips: Agreed, we need to call `meta.Keys()` to pull out the keys that are being set, then selectively set those values on a storeOptions instance.

harche: Thanks @rphillips. If I get the meta keys and update the values only if the key is defined in a drop-in config then it seem  to fix the issue. 

```golang
keySet := make(map[string]bool)
for _, key := range meta.Keys() {
	keySet[key.String()] = true
}

if _, ok := keySet["storage.transient_store"]; ok {
	storeOptions.TransientStore = config.Storage.TransientStore
}
```

cc @mtrmac @haircommander @sohankunkerkar

harche: The updated snippet of the included test looks like this, 
```golang
contents := []string{
		`[storage]
runroot = 'temp/run1'
graphroot = 'temp/graph1'
transient_store = false`,
		`[storage]
runroot = 'temp/run2'
graphroot = 'temp/graph2'`,
		`[storage]
runroot = 'should/ignore'
graphroot = 'should/ignore'`,
		`[storage]
runroot = 'temp/run3'`,
		`[storage]
runroot = 'temp/run4'
graphroot = 'temp/graph4'`,
	}
	for i, fileName := range fileNames {
		filePath := filepath.Join(tempDir, fileName)
		if err := os.WriteFile(filePath, []byte(contents[i]), 0o666); err != nil {
			t.Fatalf("Failed to write to temp file: %v", err)
		}
	}

	// Set base options
	baseOptions := StoreOptions{
		RunRoot:        "initial/run",
		GraphRoot:      "initial/graph",
		TransientStore: true,
	}

	// Expected results after merging configurations from only .conf files
	expectedOptions := StoreOptions{
		RunRoot:        "temp/run3", // Last .conf file (config3.conf) read overrides earlier values
		GraphRoot:      "temp/graph2",
		TransientStore: false,
	}
```

harche: If we think this is a good approach to take, I will update that entire `ReloadConfigurationFile` function.

harche: @haircommander pointed me to how kubelet handles drop-in configs,

https://github.com/kubernetes/kubernetes/blob/02365ecec1cb1ddf993cf4ed12407737db950cea/cmd/kubelet/app/server.go#L322-L332

harche: One more option is to use [mergo](https://github.com/kubernetes/kubernetes/blob/8bd39b8eb1fa731ce104460e0b97b9e1fff429d0/staging/src/k8s.io/client-go/tools/clientcmd/client_config.go#L510) to merge the `StorageOptions`.

rphillips: I sorta worry about mergo setting options to false. Did we get past this issue in MCO where values being set to false wouldn’t get set?

harche: Yeah the mergo option didn't work. It does set the option to false as you predicated. 

code is at, https://github.com/harche/storage/commit/f0b92dfc33667a34adc1ba74c7aaa36ad32899f4#diff-5c0b7c9ae084e6c4613c438d00cede106862b604f184a331df1c3806cfbd11d2R164

So it seems like our only option is to use `meta.Keys()`?

BTW, which MCO issue are you talking about?

rphillips: https://issues.redhat.com/browse/OCPBUGS-14399 I remember we had to hack around it.

cgwalters: systemd has definitely benefited from a single consistent config file format with consistent semantics and merge support.

cgwalters: Explicitly iterating over the keys seems like it'd work, but another thing to do is to use nullable values for all entries.  This is much more elegant in Rust of course...here's how I recently did a basic "mergeable toml" in bootc:

https://github.com/containers/bootc/blob/f152bfe8da47c3bfbafca33c2142d10611375fa3/lib/src/install/config.rs#L47 (Note each entry is `Option<T>`) and there's a `Mergeable` trait which can be called recursively e.g. https://github.com/containers/bootc/blob/f152bfe8da47c3bfbafca33c2142d10611375fa3/lib/src/install/config.rs#L106

cgwalters:
This would be really quite useful to have, it's just completely annoying right now to e.g. enable composefs because one needs to copy all of the storage config.

ktock:
@harche Hi, thanks for working on  this PR. This PR looks beneficial for several use cases so it would be great if we can move this PR forward. 

> So it seems like our only option is to use meta.Keys()?

If you have the patch for the `meta.Keys()` approach, could you update this PR with that patch? Or, if you don't have enough time for this PR, I'm willing to take over this PR.

harche:
> @harche Hi, thanks for working on this PR. This PR looks beneficial for several use cases so it would be great if we can move this PR forward.
> 
> > So it seems like our only option is to use meta.Keys()?
> 
> If you have the patch for the `meta.Keys()` approach, could you update this PR with that patch? Or, if you don't have enough time for this PR, I'm willing to take over this PR.

Thanks @ktock but this is an optional (good to have) thing for the lazy image pull. In case of Openshift, there is an operator called MCO which drops the storage configuration on the node. It is totally possible to implement the logic in MCO to drop a storage config file on the node to support lazy image pulls. MCO has been doing that with KubeletConfig for awhile now. 

Since this is an optional feature from lazy image point of view, this change is not on high priority. But you can still look into it if you feel this library should support drop-in config nevertheless.

mtrmac:
My personal opinion is that this is fairly likely to be brittle and hard-to-maintain code (though I could be well wrong about that, there might be a neat trick that makes it easy); and that the option parsing in c/storage is already convoluted enough that I’d prefer not to add any extra complexity.

I’m not a c/storage maintainer, though, so it’s not up to me.

jankaluza:

PR #1887, opened 2024-04-13
Refactor to make it clearer that we return the validated TOC value

### LINKED ISSUE #1886: zstd:chunked binary footer format is broken
Since eedd976e5b3cc4bf50c62878d8a89b85602c1f16 , AFAICS the `ReadFooterDataFromBlob` code path when parsing zstd:chunked metadata always fails, because it does not set `ChecksumAnnotation`, causing a failure when validating the manifest.

@giuseppe the binary footer seems to be unused in practice. Can we eliminate it altogether, and simplify the code?

mtrmac:
Primarily, this ensures that the value in `ApplyDiff`’s `DriverWithDifferOutput.TOCDigest` is really exactly the value we used to validate the digest, by only reading it once and passing it around.

That’s to simplify auditing and make the relationship clear, but it should not change behavior.

Incidentally, this also should fix #1886 — but I still think the binary footer code path should be removed entirely.

Cc: @giuseppe

### DIFF
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -145,7 +145,7 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, toc
 
 	if offsetMetadata := annotations[internal.ManifestInfoKey]; offsetMetadata != "" {
 		var err error
-		footerData, err = internal.ReadFooterDataFromAnnotations(tocDigest, annotations)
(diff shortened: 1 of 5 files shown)


mtrmac:
(Completely untested, to be transparent.)

giuseppe (approved):
I've tested it with some existing images and it works well.

LGTM

rhatdan:
LGTM

PR #1889, opened 2024-04-15
pkg/parsers/kernel: simplify

kolyshkin:
Since golang.org/x/sys/unix provides portable way to use uname(2) syscall, let's use it for all unix variants, removing the duplicated code from uname_*.go.

While at it:
 - remove uname_unsupported*.go (not used);
 - use unix.ByteSliceToString.

NOTE this also removes the public definition of struct Utsname. I am puzzled as to why it was exported as it is not (and was not) used by any public functions in this package, nor I was able to find any usage of kernel.Utsname (using sourcegraph).


rhatdan:
LGTM
/approve
@giuseppe @nalind @mtrmac @saschagrunert PTAL

saschagrunert (approved):

kolyshkin:
> Simplifying the Unix implementation, sure.
> 
> WRT removing the public `Utsname` symbol, c/storage is supposed to have a stable API, but with things like #1622 (and many others), that’s… not been consistently the case.

As I was not able to find any existing users of `kernel.Utsname`, this API breakage seems purely theoretical.

PR #1890, opened 2024-04-18
composefs: return mkcomposefs stderr as part of error

Luap99:
Writing the error to stderr is not helpful when using the podman API. The error would only be visable on the server which is bad and likely is not noticed at all.

Also for the regual podman cli it would be hard for a user to know if the stderr from mkcompose is related to the returned podman error.

To fix this capture the stderr and append it to the error as string, this should be fine assuming mkcomposefs doesn't print a ton on stderr which I assume is not the case.

### DIFF
--- a/drivers/overlay/composefs.go
+++ b/drivers/overlay/composefs.go
@@ -4,12 +4,14 @@
 package overlay
 
 import (
+	"bytes"
 	"encoding/binary"
 	"errors"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@giuseppe @alexlarsson PTAL

giuseppe (approved):
LGTM

PR #1893, opened 2024-04-19
chunked: store original tar size for converted layers

giuseppe:
if the layer was converted from an existing one, store the original layer size.

Closes: https://github.com/containers/storage/issues/1892

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -216,15 +216,15 @@ func (f *seekableFile) GetBlobAt(chunks []ImageSourceChunk) (chan io.ReadCloser,
 	return streams, errs, nil
 }
 
-func convertTarToZstdChunked(destDirectory string, payload *os.File) (*seekableFile, digest.Digest, map[string]string, error) {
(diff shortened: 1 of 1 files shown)


pkg/chunked/storage_linux.go
```diff
@@ -1753,13 +1756,20 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
...
 		return output, err
 	}
 
 	output.UIDs, output.GIDs = collectIDs(mergedEntries)
-	output.Size = totalSize
+	if convertedBlobSize > 0 {
+		// if the image was converted, store the original tar size, so that
+		// it can be recreated correctly.
+		output.Size = totalSizeFromTOC
+		output.Size = convertedBlobSize
```
eriksjolund, line 1769: Is the second line overwriting the value that the first line just set?
(disclaimer: I don't know golang so well)

giuseppe: yes, sorry.

Fixed now

PR #1895, opened 2024-04-19
chunked: fix escape of space

giuseppe:
the code was copied from the composefs C version:

	if (noescape_space)
		hex_escape = !isprint(c);
	else
		hex_escape = !isgraph(c);

but unicode.IsGraphic() seems to behave differently and includes the space:

isgraph(' ') -> 0
unicode.IsGraphic(' ') -> true

### DIFF
--- a/pkg/chunked/dump/dump.go
+++ b/pkg/chunked/dump/dump.go
@@ -52,7 +52,7 @@ func escaped(val string, escape int) string {
 			if noescapeSpace {
 				hexEscape = !unicode.IsPrint(rune(c))
 			} else {
-				hexEscape = !unicode.IsGraphic(rune(c))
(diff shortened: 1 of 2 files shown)


giuseppe:
@rhatdan this is the fix for the `mkcomposefs: Invalid integer 628` error we have seen

kolyshkin (approved):
Indeed, `isgraph(3) says (emphasis is mine):
>        isgraph()
>              checks for any printable character **except space**.

and [unicode.IsGraphic](https://pkg.go.dev/unicode#IsGraphic) says:

> IsGraphic reports whether the rune is defined as a Graphic by Unicode. Such characters include letters, marks, numbers, punctuation, symbols, **and spaces**, from categories L, M, N, P, S, Zs.

Somewhat ironic that the test case was obv. wrong.

PR #1899, opened 2024-04-22
overlay: support AdditionalImagesStore on vfs

giuseppe:
extend the overlay driver to use an additional image store on a vfs backend.

It works as overlay can simply use the upper most vfs layer as the only lower layer for the container.

It is useful on a system already protected by composefs, to distribute images as part of the operating system itself.  The read-only store cannot be on overlay since composefs itself is using overlay, and it causes conflicts with whiteout files

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -21,6 +21,7 @@ import (
 	graphdriver "github.com/containers/storage/drivers"
 	"github.com/containers/storage/drivers/overlayutils"
 	"github.com/containers/storage/drivers/quota"
+	"github.com/containers/storage/drivers/vfs"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/chrootarchive"
 	"github.com/containers/storage/pkg/directory"
@@ -977,7 +978,7 @@ func (d *Driver) Create(id, parent string, opts *graphdriver.CreateOpts) (retErr
 }
 
 func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, readOnly bool) (retErr error) {
-	dir, homedir, _ := d.dir2(id, readOnly)
+	dir, homedir, _, _ := d.dir2(id, false, false)
 
 	disableQuota := readOnly
 
@@ -1006,15 +1007,6 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, readOnl
 
 	if err := idtools.MkdirAllAndChownNew(path.Dir(dir), 0o755, idPair); err != nil {
 		return err
-	}
-	if parent != "" {
-		parentBase := d.dir(parent)
-		st, err := system.Stat(filepath.Join(parentBase, "diff"))
-		if err != nil {
-			return err
-		}
-		rootUID = int(st.UID())
-		rootGID = int(st.GID())
 	}
 
 	if err := fileutils.Lexists(dir); err == nil {
@@ -1066,11 +1058,19 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, readOnl
 	}
 
 	if parent != "" {
-		parentBase := d.dir(parent)
-		st, err := system.Stat(filepath.Join(parentBase, "diff"))
+		parentBase, _, singleLayer, _ := d.dir2(parent, false, true)
+		var d string
+		if singleLayer {
+			d = parentBase
+		} else {
+			d = filepath.Join(parentBase, "diff")
+		}
+		st, err := system.Stat(d)
 		if err != nil {
 			return err
 		}
+		rootUID = int(st.UID())
+		rootGID = int(st.GID())
 		perms = os.FileMode(st.Mode())
 	}
 
@@ -1142,11 +1142,15 @@ func (d *Driver) parseStorageOpt(storageOpt map[string]string, driver *Driver) e
 }
 
 func (d *Driver) getLower(parent string) (string, error) {
-	parentDir := d.dir(parent)
+	parentDir, _, singleLayer, _ := d.dir2(parent, false, true)
 
 	// Ensure parent exists
 	if err := fileutils.Lexists(parentDir); err != nil {
 		return "", err
+	}
+
+	if singleLayer {
+		return parent, nil
 	}
 
 	// Read Parent link fileA
@@ -1175,7 +1179,7 @@ func (d *Driver) getLower(parent string) (string, error) {
 }
 
 func (d *Driver) dir(id string) string {
-	p, _, _ := d.dir2(id, false)
+	p, _, _, _ := d.dir2(id, false, false)
 	return p
 }
 
@@ -1187,7 +1191,15 @@ func (d *Driver) getAllImageStores() []string {
 	return additionalImageStores
 }
 
-func (d *Driver) dir2(id string, useImageStore bool) (string, string, bool) {
+// dir2 returns the directory path and home directory path associated with the given layer ID.
+// It also returns two boolean values indicating if the store is using the VFS graph driver, and
+// whether the directory is found in an image store.
+//
+//   - string: The directory path for the layer.
+//   - string: The storage home directory path.
+//   - bool: A boolean indicating whether the store is using VFS, thus doesn't require stacking of layers.
+//   - bool: A boolean indicating whether the directory is found in an additional image store.
+func (d *Driver) dir2(id string, useImageStore, useVFS bool) (string, string, bool, bool) {
 	var homedir string
 
 	if useImageStore && d.imageStore != "" {
@@ -1196,18 +1208,27 @@ func (d *Driver) dir2(id string, useImageStore bool) (string, string, bool) {
 		homedir = d.home
 	}
 
-	newpath := path.Join(homedir, id)
+	primaryPath := path.Join(homedir, id)
+	if err := fileutils.Exists(primaryPath); err == nil {
+		return primaryPath, homedir, false, false
+	}
 
-	if err := fileutils.Exists(newpath); err != nil {
-		for _, p := range d.getAllImageStores() {
-			l := path.Join(p, d.name, id)
-			err = fileutils.Exists(l)
-			if err == nil {
-				return l, homedir, true
+	allImageStores := d.getAllImageStores()
+	for _, p := range allImageStores {
+		l := path.Join(p, d.name, id)
+		if err := fileutils.Exists(l); err == nil {
+			return l, homedir, false, true
+		}
+		if useVFS {
+			l = path.Join(p, vfs.Name, "dir", id)
+			if err := fileutils.Exists(l); err == nil {
+				return l, homedir, true, true
 			}
 		}
 	}
-	return newpath, homedir, false
+
+	// It does not exist, return where to create it.
+	return primaryPath, homedir, false, false
 }
 
 func (d *Driver) getLowerDirs(id string) ([]string, error) {
@@ -1417,7 +1438,7 @@ func (d *Driver) Get(id string, options graphdriver.MountOpts) (string, error) {
 }
 
 func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountOpts) (_ string, retErr error) {
-	dir, _, inAdditionalStore := d.dir2(id, false)
+	dir, _, _, inAdditionalStore := d.dir2(id, false, true)
 	if err := fileutils.Exists(dir); err != nil {
 		return "", err
 	}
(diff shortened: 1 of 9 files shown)


drivers/overlay/overlay.go
```diff
@@ -2283,7 +2283,7 @@ func (d *Driver) Changes(id string, idMappings *idtools.IDMappings, parent strin
 
 // AdditionalImageStores returns additional image stores supported by the driver
 func (d *Driver) AdditionalImageStores() ([]string, []string) {
-	return d.options.imageStores, []string{d.name}
+	return d.options.imageStores, []string{d.name, "vfs"}
```
alexlarsson: Maybe this part should be in a separate commit

store.go
```diff
@@ -996,33 +996,35 @@ func (s *store) load() error {
...
-			}
-			s.rwImageStores = append(s.rwImageStores, imageStore)
-			ris = imageStore
-		} else {
-			ris, err = newROImageStore(gipath)
-			if err != nil {
-				if errors.Is(err, syscall.EROFS) {
-					logrus.Debugf("Ignoring creation of lockfiles on read-only file systems %q, %v", gipath, err)
-					continue
+		for _, driver := range additionalGraphDrivers {
```
mtrmac, line 1005: This is trying a cartesian product of (store, driver) _and all combinations must exist_, otherwise this fails. 

I can’t see how that makes sense even with one additional store, let alone multiple. Am I missing something?

I would have expected `.AdditionalImageStores()` to return a `[]struct InternalAdditionalStoreSpec { storePath String, driver stringOrMaybeInterface }` or something like that. (Well, modulo the pre-existing thing that the division of responsibility for parsing/processing these options between `store.go` and the graph driver seems convoluted to me.)

store.go
```diff
@@ -1162,14 +1164,18 @@ func (s *store) getROLayerStoresLocked() ([]roLayerStore, error) {
...
+	additionalImageStores, additionalGraphDrivers := s.graphDriver.AdditionalImageStores()
 
-		rls, err := newROLayerStore(rlpath, glpath, s.graphDriver)
-		if err != nil {
-			return nil, err
+	for _, store := range additionalImageStores {
+		for _, driver := range additionalGraphDrivers {
+			glpath := filepath.Join(store, driver+"-layers")
+
+			rls, err := newROLayerStore(rlpath, glpath, s.graphDriver)
```
mtrmac, line 1173: Shouldn’t this use the right graph driver for that store?

drivers/overlay/overlay.go
```diff
@@ -1175,19 +1192,19 @@ func (d *Driver) getLower(parent string) (string, error) {
...
-	additionalImageStores := d.AdditionalImageStores()
+	additionalImageStores, _ := d.AdditionalImageStores()
 	if d.imageStore != "" {
 		additionalImageStores = append([]string{d.imageStore}, additionalImageStores...)
 	}
 	return additionalImageStores
 }
 
-func (d *Driver) dir2(id string, useImageStore bool) (string, string, bool) {
+func (d *Driver) dir2(id string, useImageStore bool) (string, string, bool, bool) {
```
mtrmac, line 1207: (I didn’t really read the overlay implementation at all.)

Please document the meaning all (parameters and) return values of this function.

drivers/overlay/overlay.go
```diff
@@ -1198,16 +1215,25 @@ func (d *Driver) dir2(id string, useImageStore bool) (string, string, bool) {
...
+		for _, p := range allImageStores {
 			l := path.Join(p, d.name, id)
 			err = fileutils.Exists(l)
 			if err == nil {
-				return l, homedir, true
+				return l, homedir, false, true
+			}
+		}
+		for _, p := range allImageStores {
+			l := path.Join(p, vfs.Name, "dir", id)
```
mtrmac, line 1229: Hard-coding VFS feels a bit surprising.

If the same layer exists if multiple image stores, this will probably not find them in the same order `store.go` will. Does that matter?

mtrmac (commented):
(Just a very minimal skim of what this is about.)

> It is useful on a system already protected by composefs, to distribute images as part of the operating system itself. The read-only store cannot be on overlay since composefs itself is using overlay, and it causes conflicts with whiteout files

At a very high level, I understand how this works fine for running containers.

But what about access to the parent layers, e.g. due to layer deduplication? Would we be shipping OS images with a baked-in VFS store, where a 10-layer image contains 10 copies of the same file? (10 real copies? 10 hard-linked copies? 10 ref-linked copies?)

The alternative of not shipping the parent layers at all seems impractical to me, that would break quite a lot of assumptions. Or would we restrict this feature to squashed images where the difference doesn’t matter?

giuseppe:
> But what about access to the parent layers, e.g. due to layer deduplication? Would we be shipping OS images with a baked-in VFS store, where a 10-layer image contains 10 copies of the same file? (10 real copies? 10 hard-linked copies? 10 ref-linked copies?)

with composefs it won't matter, because these files will be deduplicated anyway.  So having multiple copies of the same file is fine, as the underlying storage will keep only one copy of that file.

This feature is only useful on a composefs system, where vfs doesn't have any additional cost compared to overlay since the deduplication is performed in the layer below.

alexlarsson:
> But what about access to the parent layers, e.g. due to layer deduplication? Would we be shipping OS images with a baked-in VFS store, where a 10-layer image contains 10 copies of the same file? (10 real copies? 10 hard-linked copies? 10 ref-linked copies?)
> 
> The alternative of not shipping the parent layers at all seems impractical to me, that would break quite a lot of assumptions. Or would we restrict this feature to squashed images where the difference doesn’t matter?

I think there is maybe something to this worry, but its not quite what you seem to say here.

First, lets start with the goal of this: We want to enable a bootc image that contains other containers, such that when you boot the container as a fullblown OS it can run containers. In such a deployment, the outer image would be using ostree with composefs, and the vfs storage directory would be part of this composefs image.

The composefs image will naturally and automatically de-duplicate any identical files. So, shipping all the layers as vfs stores will not be a problem in the final image.

That said, wouldn't we still de duplicating the files in the OCI layer tarballs, making e.g. downloads larger?

alexlarsson:
Or would zstd-chunked help here?

store.go
giuseppe: `newImageStore` automatically creates the directory.  This is a pre-existing behavior that I've kept in the current version.

I agree it is not ideal, I will add a patch to just ignore stores that do not exist instead of silently creating them

store.go
giuseppe: the only case we enabled now is overlay on top of vfs, and the overlay graph driver works well with layers stored as vfs.  I'll add a comment, not even sure how we would deal with multiple graphdrivers otherwise

drivers/overlay/overlay.go
giuseppe: it should not really matter, but I'll change the lookup code in overlay to behave in the same way as in `store.go`

giuseppe:
> That said, wouldn't we still de duplicating the files in the OCI layer tarballs, making e.g. downloads larger?

the downside is when you pull an image to vfs, it first needs to copy all the underlying layers, and then applies the current one.  zstd:chunked is not implemented for vfs (not sure it is worth the effort), but once you've the store and you've added it to a composefs image, then there are no additional costs of vfs (except the # of inodes)

rhatdan:
But running this image as a container could be huge.   As long as users of this feature understand the downsides, it should not be a problem.

Doing a `bootc upgrade` with one of these images would be able to take advantage of zstd:chunked though, correct?

alexlarsson:
> > That said, wouldn't we still de duplicating the files in the OCI layer tarballs, making e.g. downloads larger?
> 
> the downside is when you pull an image to vfs, it first needs to copy all the underlying layers, and then applies the current one. zstd:chunked is not implemented for vfs (not sure it is worth the effort), but once you've the store and you've added it to a composefs image, then there are no additional costs of vfs (except the # of inodes)

That is not exactly what I mean though. Say you have a Dockerfile that creates a bootc container, and it does `RUN podman pull --vfs --ais=/usr/lib/my-images some-image`. Now, supposed `some-image` has several layers, and these are put into the container image. Wouldn't the contents of these layers be duplicated in the resulting bootc OCI image tarball, for example when you pull it to you bootc host.

alexlarsson:
> Doing a `bootc upgrade` with one of these images would be able to take advantage of zstd:chunked though, correct?

Technically it should be possible. For example, if a chunked image contains 10 copies of the same data in different files, then we should be able to only download it once. I don't know if the current implementation works this way though, as it is probably focusing on the slightly different case where a single file in the image is already locally available.

mtrmac:
> > That said, wouldn't we still de duplicating the files in the OCI layer tarballs, making e.g. downloads larger?
> 
> the downside is when you pull an image to vfs, it first needs to copy all the underlying layers, and then applies the current one.

I’m not quite clear about the overall picture, still: Under $assumptions, once layers 1…M exist in all stores backing a single c/storage.Store, a pull of a child image with layers 1…M…N should see that all the parent layers are already available locally, not pull them, and then tell the top-level graph driver to only create the new layers.

If the primary graph driver is overlay, that should only create overlay layers.

($assumptions are that either no layers are zstd:chunked, or that they are _the same exact zstd:chunked_ TOC in the two images.)

store.go
mtrmac: That means silent failures on incorrect paths in the config. (We do validate their existence.)

The previous behavior of automatically creating a directory seems undesirable to me; but if we find _no_ matching driver, shouldn’t we fail outright instead of silently ignoring the store?

Also the logic in `load()` and `getROLayerStoresLocked()` should be consistent; that’s also easier if the graph driver detection is moved behind `driver.AdditionalImageStores` (I suppose into overlay’s `parseOptions` or even earlier).

store.go
mtrmac: Right now, the overlay driver is, via `AdditionalImageStores()`, in charge; so I imagine it could
- detect the driver / format of the additional store
- fail on unexpected ones
- return the driver ID so that this line can use the right one.

See elsewhere.

I suppose I can live with the added comment…

mtrmac (commented):
Looking a bit more — the diffs look plausible to me; I didn’t delve deep enough to understand the full impact (but at least looking at `.get` there seems to be more to be found).

giuseppe:
ping

rhatdan:
@mtrmac PTAL

rhatdan:
@alexlarsson PTAL

drivers/overlay/overlay.go
```diff
@@ -1175,19 +1179,27 @@ func (d *Driver) getLower(parent string) (string, error) {
 }
 
 func (d *Driver) dir(id string) string {
-	p, _, _ := d.dir2(id, false)
+	p, _, _, _ := d.dir2(id, false, false)
```
mtrmac, line 1182: So it seems that https://github.com/containers/storage/pull/1899/files#r1576924079 was resolved by just not _finding_ a VFS layer when `dir` is called, and instead returning an otherwise-uninitialized `primaryPath`. So some callers looking for a layer with the same ID will see a VFS layer, and some will see a different, empty directory?

Why is that safe and consistent?

After looking a bit, it seems that only parent layers can be VFS layers (because non-parent layers would presumably be handled by the VFS driver directly, not overlay?) … but then `isParent` is calling `.dir` and not `dir2`, and it will not find VFS layers?

---

I’m sorry, I just can’t deal with the complexity. Too many options, too many interactions, too many things that _seem_ like they might be possible even if something elsewhere in the call stack probably prevents them. **Most of that with documentation of neither intent nor maintained invariants**.

I straight can’t approve any PR adding a feature adding any more indirections or alternatives to the overlay driver any more. I’m not smart enough to keep up.

It seems to me that most non-trivial PRs here would require me to spend something like 2 days just graphing out and understanding the interactions and possibilities. At some point, it must pay off to instead spend 2-3 weeks to restructure and explicitly document, so that PRs are again a matter of a small number of hours, and I’m now past the point where I’m willing to enable the current state to persist, and to keep paying for the complexity those 2 days at a time, every time.

I don’t know what is the path forward. Probably easiest, find some other reviewer who is up to the task.

Or, maybe, aggressively simplify. Build abstractions where they can be built and don’t leak. Find identifiable an isolated utilities that can be cut out of the >400-line functions. Rename things to be clear (`dir` vs. `dir2` could never be clear), maybe `findParentLayerPossiblyInOtherDriver` vs. `findLayerWeOwn` vs … (yes, that is surely unproductive overkill, but the naming needs to _first_ be clear and only _afterwards_ succinct [I have no idea whether that’s the distinction that matters here, those names are not intended to be just used because I had to come up with an example]).

---

*shrug* it’s also possible that I’m overreacting and that a few sentences can make the picture clear. I’m not smart enough to figure that out myself.

alexlarsson:
I would love to have this feature, but I currently don't have time to review this.

giuseppe:
I won't deal with rebases trying to keep this PR alive for now.  I'll reopen it once it is clear we want it and can get reviewed

Issue #1888, opened 2024-04-13
zstd:chunked blocker: TarSplitChecksumKey not used in a layer ID


mtrmac:
@giuseppe Zstd:chunked has uses a `TarSplitChecksumKey` annotation, separate from the TOC digest.

That’s fine as far as per-image individual trust goes, because the annotation is authenticated by the manifest digest.

But we deduplicate individual layers by the TOC digest alone; so it is possible for two images to use a layer with the same TOC digest, but different tar-split locations and contents (e.g. one with an added set-UID bit somewhere).

Locally, it seems by far easiest to move the tar-split digest into the TOC. That way the by-now-widespread assumption that the TOC digest is a sufficient identifier can be maintained.

For other purposes, like the BlobInfoCache and metadata-only reuse, the value returned from  `toc.GetTOCDigest` is used both for reuse/match lookups, so whatever this function returns must be a _complete_ identifier; it should be unambiguous, and we can never add more annotations that change the semantics of a layer with a matching `GetTOCDigest` value. Even if we replaced `GetTOCDigest` _now_ with something that accounts for tar-split, we need this guarantee to be forward-compatible (consider pulling an image just before a new annotation type is added, upgrading, and then pulling another just after). I think that also argues for a long-term design of keeping everything anchored to a single digest value.

---

A bit more generally, I’ve been wondering whether the TOC digest is a good identifier long-term.

The way things are now, we basically can’t significantly change the format of the TOC, otherwise we risk the possibility of the same TOC blob being valid for both the old and new formats, allowing an attacker to trigger deduplication between layers with the same TOC digest but different TOC formats and differently-understood contents.

For example, we just can’t support a new partial-pull format with some _other_ JSON-based TOC format if it could be constructed ambiguously with the zstd:chunked TOC format. I’m not 100% sure this is worth worrying about — but if it were ever to happen, we would need a fairly large restructuring, or at least renaming/re-documenting. So I’m raising it right now, in case such things were expected. (E.g. will this be required for composeFS?)

mtrmac:
Maybe one way to think about this is that I think it would be clean if #1887 were extended further, so that
- (some future variant of) `toc.GetTOCDigest` returns a “complete” format+identity identifier (of a kind which is long-term stable across new added feature and new format variants)
- that complete identifier were passed to `chunked.GetDiffer`, which just “executes” it without making any decision what format/variant to consume
[...]

PR #1902, opened 2024-04-22
Move the tar-split digest into the TOC

mtrmac:
A minimal prototype for a fix of #1888 .

Note that this would immediately break pulling all currently-existing layers.

Also this probably increases memory requirements because the TOC exists as individual Go values longer.

Cc: @giuseppe

### DIFF
--- a/pkg/chunked/internal/compression.go
+++ b/pkg/chunked/internal/compression.go
@@ -85,8 +85,9 @@ func GetType(t byte) (string, error) {
 const (
 	ManifestChecksumKey = "io.github.containers.zstd-chunked.manifest-checksum"
 	ManifestInfoKey     = "io.github.containers.zstd-chunked.manifest-position"
-	TarSplitChecksumKey = "io.github.containers.zstd-chunked.tarsplit-checksum" // Deprecated: Use the TOC.TarSplitDigest field instead, this one is not authenticated by ManifestChecksumKey.
 	TarSplitInfoKey     = "io.github.containers.zstd-chunked.tarsplit-position"
+
+	TarSplitChecksumKey = "io.github.containers.zstd-chunked.tarsplit-checksum" // Deprecated: Use the TOC.TarSplitDigest field instead, this annotation is no longer read nor written.
 
 	// ManifestTypeCRFS is a manifest file compatible with the CRFS TOC file.
 	ManifestTypeCRFS = 1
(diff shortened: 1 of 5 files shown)


pkg/chunked/internal/compression.go
```diff
@@ -85,7 +86,7 @@ func GetType(t byte) (string, error) {
 const (
 	ManifestChecksumKey = "io.github.containers.zstd-chunked.manifest-checksum"
 	ManifestInfoKey     = "io.github.containers.zstd-chunked.manifest-position"
-	TarSplitChecksumKey = "io.github.containers.zstd-chunked.tarsplit-checksum"
+	TarSplitChecksumKey = "io.github.containers.zstd-chunked.tarsplit-checksum" // Deprecated: Use the TOC.TarSplitDigest field instead, this one is not authenticated by ManifestChecksumKey.
```
giuseppe, line 89: I think it is fine if we just break now, and not have to worry about it

rhatdan:
@giuseppe @mtrmac What should we do with this one?

mtrmac:
#1888 _must_ have a solution. This is one possible solution; it also breaks all currently-existing zstd:chunked images, which might be fine or very bad, I don’t know which one it is.

giuseppe:
> #1888 _must_ have a solution. This is one possible solution; it also breaks all currently-existing zstd:chunked images, which might be fine or very bad, I don’t know which one it is.

It is fine to break these images now.

mtrmac:
With this PR:
- If there is no tar-split offset, the image is accepted
- If there is a tar-split offset and the _new_ tar-split digest in the TOC, the image is accepted
- If there is a tar-split offset and the TOC does not have the new digest, the image is rejected (most recently-created images are like this)

Alternatively, in the last case, we could accept the layer and ignore the tar-split data. If we do that, the images will continue to pull fine; but previously they also pushed fine and that would now break.

Marking ready for review now.

giuseppe:
> * If there is a tar-split offset and the TOC does not have the new digest, the image is rejected (most recently-created images are like this)

could we just drop it from the image generation and always ignore it?  So we don't have to worry about it in future

giuseppe (approved):
code LGTM

mtrmac:
> could we just drop it from the image generation and always ignore it? So we don't have to worry about it in future

I’m not sure which part you mean should be dropped. We need:
- the offset to be provided somehow. Currently it is an annotation. (It could be in the TOC instead.)
- a digest, in the TOC. (That’s being added in this PR. The previous digest, in the annotation, is still being added, but it is now ignored).

We can’t _drop_ the tar-split offset. Do you mean to stop adding the `TarSplitChecksumKey` annotation? Something else?

giuseppe:
> We can’t _drop_ the tar-split offset. Do you mean to stop adding the `TarSplitChecksumKey` annotation? Something else?

yes, to stop doing that and use only the information in the TOC

giuseppe (approved):
Thanks!

LGTM

PR #1904, opened 2024-04-23
chunked: fix divide by zero in bloom filter

giuseppe:
if the bloom filter size is zero, the "% size" operation fails with a divide by zero.

Closes: https://github.com/containers/storage/issues/1903

### DIFF
--- a/pkg/chunked/bloom_filter.go
+++ b/pkg/chunked/bloom_filter.go
@@ -13,6 +13,9 @@ type bloomFilter struct {
 
 func newBloomFilter(size int, k uint32) *bloomFilter {
 	numElements := (size + 63) / 64
+	if numElements == 0 {
+		numElements = 1
+	}
 	return &bloomFilter{
 		bitArray: make([]uint64, numElements),
 		k:        k,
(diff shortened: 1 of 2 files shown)


rhatdan:
lgtm
@mtrmac @nalind @saschagrunert @flouthoc PTAL

pkg/chunked/bloom_filter.go
```diff
@@ -28,7 +31,7 @@ func newBloomFilterFromArray(bitArray []uint64, k uint32) *bloomFilter {
 
 func (bf *bloomFilter) hashFn(item []byte, seed uint32) (uint64, uint64) {
 	if len(item) == 0 {
-		return 0, 0
+		return 0, 1
```
mtrmac, line 34: Is this correct? If we have a 1-bit filter, it’s a bit `hash == 0`, i.e. `(0 / 64, 0 % 64) == (0, 0)`, isn’t it?

mtrmac: … never mind, the value returned is a mask, not a bit index.

mtrmac (commented):
/lgtm

PR #1906, opened 2024-04-23
chunked: downgrade loading cache file msg to info

giuseppe:
it can happen for any reason, like for example using a new cache file format, in this case the file is recreated with the last version. This is internal only and should not be displayed by default.

Closes: https://github.com/containers/storage/issues/1905

### DIFF
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -268,7 +268,7 @@ func (c *layersCache) load() error {
 		// try to read the existing cache file.
 		l, err := c.loadLayerCache(r.ID)
 		if err != nil {
-			logrus.Warningf("Error loading cache file for layer %q: %v", r.ID, err)
+			logrus.Infof("Error loading cache file for layer %q: %v", r.ID, err)
 		}
 		if l != nil {
 			newLayers = append(newLayers, l)


mtrmac (commented):
My first impression is that logging `io.EOF` (if that’s what it is) as an error, as reported in #1905 ,  indicates a programming bug; that bug should be fixed, not papered over. (`io.EOF` is a non-failure value that should be handled, perhaps by converting into `UnexpectedEOF`).

E.g. there already is
```go
	if version != cacheVersion {
		return nil, nil //nolint: nilnil
	}
```
so we shouldn’t be failing on a version change, in principle.

I guess `cacheVersion` was not bumped recently and that’s the _root cause_. Oh well, not something we can fix retroactively.

---

The API of `binary.Read` is inconvenient for us, in that it returns `io.EOF` if the EOF falls on a value boundary. I suppose it would be easiest to wrap it into something that converts `io.EOF` into `UnexpectedEOF`.

Alternatively, the use of the `io.Reader` abstraction over `bigDataBuffer` could be eliminated entirely, but dealing with byte indices manually would be tedious.

---

Decreasing the severity of the warning LGTM either way — still, typically we shouldn’t be triggering that log line at all.

PR #1907, opened 2024-04-23
Update to Go 1.21

mtrmac:
We are already doing that for downstream consumers, so let’s do it here as well, so that we not surprised by a dependency requiring Go 1.21 suddenly forcing an upgrade.

This is a _very minimal_ update; the code could benefit from some of the new features, e.g. quite a few of the global `sync.Once` variables could be replaced by `sync.OnceValue`/`sync.OnceValues`.

Cc: @nalind @cevich

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -171,7 +171,7 @@ vendor_task:
 cross_task:
     alias: cross
     container:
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM

PR #1908, opened 2024-04-24
Ignore failure to ignore thinpool keys

rhatdan:
Fixes: https://github.com/containers/podman/issues/22473

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -115,6 +115,9 @@ type OptionsConfig struct {
 	// Btrfs container options to be handed to btrfs drivers
 	Btrfs struct{ BtrfsOptionsConfig } `toml:"btrfs,omitempty"`
 
+	// Thinpool container options to be handed to thinpool drivers (NOP)
+	Thinpool struct{} `toml:"thinpool,omitempty"`
+
 	// Overlay container options to be handed to overlay drivers
 	Overlay struct{ OverlayOptionsConfig } `toml:"overlay,omitempty"`
 


rhatdan:
@edsantiago @Luap99 PTAL

types/options.go
```diff
@@ -418,8 +418,11 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) erro
 	meta, err := toml.DecodeFile(configFile, &config)
 	if err == nil {
 		keys := meta.Undecoded()
-		if len(keys) > 0 {
-			logrus.Warningf("Failed to decode the keys %q from %q", keys, configFile)
+		for _, key := range keys {
+			if key.String() == "storage.options.thinpool" {
```
Luap99, line 422: Seems like a good compromise to me, but can you add a comment why this is added here, i.e.
`Do not cause warnings for users with old storage.conf files`.

An alternative approach could be adding the struct definition back so toml will not complain in the first place.

types/options.go
```diff
@@ -418,8 +418,11 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) erro
...
+				continue
+			}
+			logrus.Warningf("Failed to decode the keys %q from %q", key, configFile)
```
Luap99, line 425: This seems to be a change in behavior, previously it would have logged all errors in one line now it will log each key individually. This is fine for me but you should `/s/keys/key` in the message then.

rhatdan:
@edsantiago PTAL

types/options.go
mtrmac: I mean, are the warnings _bogus_ in the first place? The users presumably put the options in there for a reason, and ignoring them is exactly why we have the warning.

If this is intended to specifically silence the case where the default config contains an _empty_ thinpool section, it would be more appropriate to only silence the case where that section really _is_ empty.

mtrmac: Oh I’m not disagreeing at that high level. It’s the details of the implementation that I think might ignore too much.

edsantiago: My initial ask was to ignore `thinpool` iff it comes from `/usr/share/conf`; this is known at the time the error message is issued. My thinking was that this allows mixed-sequence updating of podman and c-common. Ignoring `thinpool` if the block is empty is also a great idea, but seems much harder.

Either way, I do think that if a `thinpool` block is found anywhere outside `/usr/share`, it should trigger a warning.

I am fully aware that this is asking for a big ugly special case conditional and that we will never find out how much (if at all) it helps anyone in the field.

rhatdan: Some distros ship storage.conf in /etc, so I think it make sense to just ignore it.

Luap99: I think the simple fix is to restore the old type definition for the thinpool table
```
diff --git a/pkg/config/config.go b/pkg/config/config.go
index dc70531de..7f49d029d 100644
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -115,6 +115,9 @@ type OptionsConfig struct {
        // Btrfs container options to be handed to btrfs drivers
        Btrfs struct{ BtrfsOptionsConfig } `toml:"btrfs,omitempty"`
 
+       // Thinpool container options to be handed to thinpool drivers (NOP)
+       Thinpool struct{} `toml:"thinpool,omitempty"`
+
        // Overlay container options to be handed to overlay drivers
        Overlay struct{ OverlayOptionsConfig } `toml:"overlay,omitempty"`
 
```

untested but I would assume that makes toml happy only if the table is empty because all other keys would be still unknown

mtrmac: On second thought, if the
```go
if key.String() == "storage.options.thinpool" {
```
is an exact match, then this already does what we want (does it?); and I might have been complicating a simple thing.

mtrmac: In particular, does
```toml
[storage.options.thinpool]
blocksize=".." # or some other previously-valid option
```
still trigger a warning?

Luap99 (approved):
LGTM, let's get this merged before we cut a release for 5.1.
I checked that this silences the empty config but still errors for any keys were set under the thinpool section.

edsantiago:
LGTM

Issue #1909, opened 2024-04-23
composefs: Error creating cache file for layer "sha": blah blah zstd blah blah


edsantiago:
Seeing this when running integration tests, podman + c-c + c-s @ latest.
```
time="2024-04-23T14:46:27-04:00" level=warning msg="Error creating cache file for layer \"03901b4a2ea88eeaad62dbe59b072b28b6efa00491962b8741081c5df50c65e0\": locating item named \"zstd-chunked-manifest\" for layer with ID \"03901b4a2ea88eeaad62dbe59b072b28b6efa00491962b8741081c5df50c65e0\": file does not exist"
```
I can't figure out a reproducer and have wasted waaaaay too much energy trying. Sorry.

giuseppe:
thanks, PR here: https://github.com/containers/storage/pull/1910

PR #1910, opened 2024-04-24
chunked: skip cache file for non-partial layers

giuseppe:
if the layer does not have a manifest TOC, just ignore it instead of raising a warning.  There is no need to create a cache file since there is no manifest file to parse.

Closes: https://github.com/containers/storage/issues/1909

### DIFF
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -199,6 +199,8 @@ func (c *layersCache) loadLayerCache(layerID string) (_ *layer, errRet error) {
 	return c.createLayer(layerID, cacheFile, mmapBuffer)
 }
 
+// createCacheFileFromTOC attempts to create a cache file for the specified layer.
+// If a TOC is not available, the cache won't be created and nil is returned.
 func (c *layersCache) createCacheFileFromTOC(layerID string) (*layer, error) {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@mtrmac @Luap99 PTAL

pkg/chunked/cache_linux.go
```diff
@@ -219,6 +219,10 @@ func (c *layersCache) createCacheFileFromTOC(layerID string) (*layer, error) {
 	}
 	manifestReader, err := c.store.LayerBigData(layerID, bigDataKey)
 	if err != nil {
+		// the cache file is not needed since there is no manifest file.
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, nil
```
Luap99, line 224: I see the only caller of this function handles a nil return correctly but I think most go devs do only check err != nil and then assume they got a valid pointer back. So at the very least I would expect the function commend to say that this may return nil,nil and the caller has to check layer != nil before using it.

giuseppe: sure, added a comment

mtrmac (commented):
Technically this works — but I’d really expect the early exit to be, well, more early.

Historically `bigDataKey` was used before we added `chunkedLayerDataKey`, so maybe start reading `bigDataKey`, and immediately cleanly exit if that one is missing; then we can optionally load the other one.

PR #1914, opened 2024-04-24
Remove dependabot, we use renovate now

rhatdan:


rhatdan:
@cevich Is there anything else I need to do to remove dependabot?

PR #1921, opened 2024-05-13
Fix assignment of quota IDs for XFS quotas

mheon:
It appears that setting the FS_XFLAG_PROJINHERIT flag when assigning a quota to a directory overrides any requested project ID set (even in the same ioctl!) if a parent directory already has a project ID. And our parent directories always do, as we set the top-level directory to the base project ID for all volumes.

The result of this is that the inherit flag forces every subdirectory to share the same project ID, and thus every time we set a quota, it's set for everything using quotas - not just the single directory we wanted to set it on.

Lacking the inherit flag made me a bit worried about subdirectories, but I can confirm that the quotas themselves are still function (creating a subdirectory in a dir with a quota set and creating files in that directory still applies their size towards the quota). There's a chance it would impact doing a separate project ID in a subdirectory of a subdirectory but from my read of the code that isn't supported at all right now.

Fixes RH Jira RHEL-18038

### DIFF
--- a/drivers/quota/projectquota_supported.go
+++ b/drivers/quota/projectquota_supported.go
@@ -19,16 +19,6 @@ package quota
 #include <linux/quota.h>
 #include <linux/dqblk_xfs.h>
 
-#ifndef FS_XFLAG_PROJINHERIT
-struct fsxattr {
-	__u32		fsx_xflags;
-	__u32		fsx_extsize;
-	__u32		fsx_nextents;
-	__u32		fsx_projid;
-	unsigned char	fsx_pad[12];
-};
-#define FS_XFLAG_PROJINHERIT	0x00000200
-#endif
 #ifndef FS_IOC_FSGETXATTR
 #define FS_IOC_FSGETXATTR		_IOR ('X', 31, struct fsxattr)
 #endif
@@ -357,7 +347,6 @@ func setProjectID(targetPath string, projectID uint32) error {
 		return fmt.Errorf("failed to get projid for %s: %w", targetPath, errno)
 	}
 	fsx.fsx_projid = C.__u32(projectID)
-	fsx.fsx_xflags |= C.FS_XFLAG_PROJINHERIT
 	_, _, errno = unix.Syscall(unix.SYS_IOCTL, getDirFd(dir), C.FS_IOC_FSSETXATTR,
 		uintptr(unsafe.Pointer(&fsx)))
 	if errno != 0 {


mheon:
@nalind @rhatdan PTAL

rhatdan:
LGTM
@giuseppe PTAL

rhatdan:
@giuseppe @mtrmac PTAL

giuseppe:
we took this file from Moby where the INHERIT flag is still used: 

https://github.com/moby/moby/blob/master/quota/projectquota.go#L312

I think we should keep the current behavior for overlay.

Could we add a new function if libpod volumes require the quota without INHERIT set?

> Fixes RH Jira RHEL-18038

It seems I've no permissions to access the Jira card

mheon:
@giuseppe Is there a guide somewhere on how to work with the overlay driver quotas? I suspect from what I'm seeing they may also be broken, so I'd like to check.

giuseppe:
> @giuseppe Is there a guide somewhere on how to work with the overlay driver quotas? I suspect from what I'm seeing they may also be broken, so I'd like to check.

I am aware only of the man page.

Looking into the code more, it seems the quota is applied only to empty directories (through `create`), so for overlay it should not matter either

mtrmac:
I can’t find _any_ authoritative-looking documenting the `FS_XFLAG_PROJINHERIT` flag.

Looking at the code, at least https://github.com/torvalds/linux/blob/ea5f6ad9ad9645733b72ab53a98e719b460d36a6/fs/xfs/xfs_inode.h#L274 suggests that we do need that flag set on the directories we create.

Also https://man7.org/linux/man-pages/man8/xfs_quota.8.html#DIRECTORY_TREE_QUOTA and https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/quota/project.c#n117 seem to confirm that the _expected_ shape of a project-quota directory tree is that all directories have the flag set.

I know nothing about this problem space, I didn’t try anything in practice, and I’m perfectly happy to stay out of this PR.

If I _should_ get involved, my starting hypothesis would be that we need that flag set, and then either look to invalidate that hypothesis, or to investigate why the code setting the flag + project ID is failing, and how to change it so that it succeeds (Compare https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/quota/project.c#n209 ).

One intriguing note is https://github.com/torvalds/linux/blob/ea5f6ad9ad9645733b72ab53a98e719b460d36a6/fs/ioctl.c#L603-L607 , but if this were the cause we should have seen a failure.

rhatdan:
Could @haircommander @saschagrunert Take a look, I think CRI-O uses some of this code, most likely we can just merge, but CRI-O should know this change is coming.

saschagrunert (approved):
LGTM

PR #1923, opened 2024-05-14
Add missing #include for basename

WhyNotHugo:
POSIX mandates that the `basename` function be defined in libgen.h. GNU libc defines it in strings.h, and musl used to do the same to allow programs that target glibc to compile with musl.

As of musl 1.2.5, this compatibility quirk has been removed, and the `basename` function must be imported from the correct header file.

See: https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7

### DIFF
--- a/pkg/unshare/unshare.c
+++ b/pkg/unshare/unshare.c
@@ -8,6 +8,7 @@
 #include <sys/mman.h>
 #include <fcntl.h>
 #include <grp.h>
+#include <libgen.h>
 #include <sched.h>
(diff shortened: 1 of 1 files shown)


rhatdan:
Certainly looks like a dup.

WhyNotHugo:
It is, sorry for the noise.

nalind:
No worries, it was a timing thing.

PR #1924, opened 2024-05-14
[Additional Layer Store] Use TOCDigest as ID of each layer (patch for c/storage)

ktock:
Needs: https://github.com/containers/image/pull/2416

This commit changes the store implementation to use TOCDigset as the ID of each layer. This ensures the runtime to pass TOCDigest to the FUSE backend everytime requiring a layer.

This commit modifies c/storage's `LookupAdditionalLayer` to receive a TOCDigest as one of the arguments. Then that TOCDigest is passed to Additional Layer Store as one of the path elements of the FUSE fs. For example, the extracted view of a layer diff contents is provided in the following path:

```
 <mountpoint>/base64(imageref)/<TOCDigest>/diff
```

Additional Layer Store implementation attempts lazy pulling of the layer that has the specified TOCDigest in the image.
If that image doesn't contain the layer matching to the TOCDigest, accessing to the layer directly results in an error.

Example draft implementation of Additional Layer Store is at https://github.com/containerd/stargz-snapshotter/pull/1673

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -885,11 +885,11 @@ func (d *Driver) Cleanup() error {
 }
 
 // LookupAdditionalLayer looks up additional layer store by the specified
-// digest and ref and returns an object representing that layer.
+// TOC digest and ref and returns an object representing that layer.
 // This API is experimental and can be changed without bumping the major version number.
 // TODO: to remove the comment once it's no longer experimental.
-func (d *Driver) LookupAdditionalLayer(dgst digest.Digest, ref string) (graphdriver.AdditionalLayer, error) {
-	l, err := d.getAdditionalLayerPath(dgst, ref)
+func (d *Driver) LookupAdditionalLayer(tocDigest digest.Digest, ref string) (graphdriver.AdditionalLayer, error) {
+	l, err := d.getAdditionalLayerPath(tocDigest, ref)
 	if err != nil {
 		return nil, err
 	}
@@ -2405,14 +2405,14 @@ func nameWithSuffix(name string, number int) string {
 	return fmt.Sprintf("%s%d", name, number)
 }
 
-func (d *Driver) getAdditionalLayerPath(dgst digest.Digest, ref string) (string, error) {
+func (d *Driver) getAdditionalLayerPath(tocDigest digest.Digest, ref string) (string, error) {
 	refElem := base64.StdEncoding.EncodeToString([]byte(ref))
 	for _, ls := range d.options.layerStores {
 		ref := ""
 		if ls.withReference {
 			ref = refElem
 		}
-		target := path.Join(ls.path, ref, dgst.String())
+		target := path.Join(ls.path, ref, tocDigest.String())
 		// Check if all necessary files exist
 		for _, p := range []string{
 			filepath.Join(target, "diff"),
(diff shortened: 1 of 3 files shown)


drivers/overlay/overlay.go
```diff
@@ -885,11 +885,11 @@ func (d *Driver) Cleanup() error {
...
 
 // LookupAdditionalLayer looks up additional layer store by the specified
-// digest and ref and returns an object representing that layer.
+// TOC digest and ref and returns an object representing that layer.
 // This API is experimental and can be changed without bumping the major version number.
 // TODO: to remove the comment once it's no longer experimental.
-func (d *Driver) LookupAdditionalLayer(dgst digest.Digest, ref string) (graphdriver.AdditionalLayer, error) {
-	l, err := d.getAdditionalLayerPath(dgst, ref)
+func (d *Driver) LookupAdditionalLayer(tocDigest digest.Digest, ref string) (graphdriver.AdditionalLayer, error) {
+	l, err := d.getAdditionalLayerPath(tocDigest, ref)
```
mtrmac, line 892: Update the parameter names of this function as well?

store.go
```diff
@@ -549,14 +549,14 @@ type Store interface {
...
-	// which can create a layer with the specified digest associated with the specified image
+	// which can create a layer with the specified TOC digest associated with the specified image
 	// reference. Note that this hasn't been stored to this store yet: the actual creation of
 	// a usable layer is done by calling the returned object's PutAs() method.  After creating
 	// a layer, the caller must then call the object's Release() method to free any temporary
 	// resources which were allocated for the object by this method or the object's PutAs()
 	// method.
 	// This API is experimental and can be changed without bumping the major version number.
-	LookupAdditionalLayer(d digest.Digest, imageref string) (AdditionalLayer, error)
+	LookupAdditionalLayer(tocDigest digest.Digest, imageref string) (AdditionalLayer, error)
```
mtrmac, line 559: In general, changing the semantics of a security-critical parameter in a caller-invisible way like this feels rather risky to me.

In this case, assuming there’s exactly the one caller in c/image, _and_ because this breaks the API of `AdditionalLayer`, I suppose I can live with it.

mtrmac (commented):
(I didn’t read the caller https://github.com/containers/storage/pull/1924 yet.)

store.go
```diff
@@ -578,8 +578,8 @@ type AdditionalLayer interface {
 	// layer store.
 	PutAs(id, parent string, names []string) (*Layer, error)
 
-	// UncompressedDigest returns the uncompressed digest of this layer
-	UncompressedDigest() digest.Digest
+	// TOCDigest returns the digest of TOC of this layer.
```
mtrmac, line 581: https://github.com/containers/image/pull/2416 suggests that this can return `""`. If so, that should be documented here.

Or, alternatively, the code should promise (and check?) that that’s not the case.

drivers/overlay/overlay.go
ktock: Thanks for the comment. Fixed.

store.go
ktock: Added the comment.

ktock:
After this PR is merged, we'll merge https://github.com/containerd/stargz-snapshotter/pull/1673 and release a new version as soon as possible.

giuseppe (approved):
LGTM

store.go
```diff
@@ -578,8 +578,8 @@ type AdditionalLayer interface {
 	// layer store.
 	PutAs(id, parent string, names []string) (*Layer, error)
 
-	// UncompressedDigest returns the uncompressed digest of this layer
-	UncompressedDigest() digest.Digest
+	// TOCDigest returns the digest of TOC of this layer. Returns "" if unknown.
+	TOCDigest() digest.Digest
```
mtrmac, line 582: Compare https://github.com/containers/image/pull/2416#discussion_r1606886998 / https://github.com/containerd/stargz-snapshotter/pull/1673#discussion_r1607020173 ; do we need this provided by the ALS backend at all?

mtrmac: … we actually, sort of, do: if this returns `""`, we know the FUSE server is older than this PR expects, and interprets the digest as non-TOC compressed digest.

AFAICS the proposed `stargz-snapshotter` implementation never returns a _different_ digest than the one we looked up, but at least the “supported/unsupported” bit seems valuable.

PR #1926, opened 2024-05-21
follow-up for "overlay: lock staging directories"

giuseppe:
fix a couple of issues reported here: https://github.com/containers/storage/pull/1916


mtrmac (changes requested):
AFAICS that’s the same bug the other way.

See https://github.com/containers/storage/pull/1927 .

PR #1927, opened 2024-05-21
Fix locking bugs from #1916, and one more

mtrmac:
… and add some comments.

Cc: @giuseppe


giuseppe:
LGTM, just a nit, should "Fix the documentation of store.putLayer" and "Move the containerStore locking inside putLayer" be squashed since the latter changes the comment added by the former?

mtrmac:
Thanks!

You’re right that future maintainers won’t care that the changes originally happened for unrelated reasons, and a simpler history helps. Squashed.

PR #1929, opened 2024-05-23
composefs: add parent directory if missing

giuseppe:
it solves this error with mkcomposefs when the TOC doesn't specify the parent directory for an entry:

Error: committing the finished image: failed to put layer using a partial pull: failed to convert json to erofs: exit status 1: mkcomposefs: Parent directory missing for /usr/share/locale/ca/LC_MESSAGES/libc.mo

### DIFF
--- a/pkg/chunked/dump/dump.go
+++ b/pkg/chunked/dump/dump.go
@@ -104,8 +104,22 @@ func sanitizeName(name string) string {
 	return path
 }
 
-func dumpNode(out io.Writer, links map[string]int, verityDigests map[string]string, entry *internal.FileMetadata) error {
+func dumpNode(out io.Writer, added map[string]struct{}, links map[string]int, verityDigests map[string]string, entry *internal.FileMetadata) error {
 	path := sanitizeName(entry.Name)
+
+	parent := filepath.Dir(path)
+	if _, found := added[parent]; !found && entry.Name != "/" {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@mtrmac PTAL

pkg/chunked/dump/dump.go
```diff
@@ -218,7 +233,7 @@ func GenerateDump(tocI interface{}, verityDigests map[string]string) (io.Reader,
 				Mode: 0o755,
```
mtrmac, line 233: Isn’t this special case for `/` now redundant with the new logic?

mtrmac (commented):
LGTM otherwise.

pkg/chunked/dump/dump.go
giuseppe: I could drop the condition `|| (sanitizeName(toc.Entries[0].Name) != "/")` but we still need it for `len(toc.Entries) == 0` as we never call `dumpNode()`

mtrmac: Oh, I didn’t realize that was also a practical possibility.

I do think removing the second condition would be valuable in itself — and it would make the “empty entries” special case a bit more noticeable.

mtrmac (commented):
/lgtm

PR #1931, opened 2024-05-29
archive: always fix mode for root dir with ForceMask

akihikodaki:
We used to override the root directory permission only when the record for the root directory exists in the archive. However, docker.io/ubuntu:jammy-20240427 lacks such a record and not overriding the root directory permission makes all other permission overrides invalid.

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -1113,8 +1113,11 @@ loop:
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe PTAL

rhatdan:
LGTM

giuseppe (approved):
/lgtm

PR #1933, opened 2024-05-30
docs: Add `containers-storage-architecture.md`

cgwalters:
I'd find it really useful to have a document with the implementation details of containers-storage (really just the overlay driver).

I started on this, using some of the source code.  Is there already


cgwalters:
BTW, feel free to (please do) just force push updates to this PR to add/fix things.

docs/containers-storage-architecture.md
```diff
@@ -0,0 +1,76 @@
+## containers-storage 1 "May 2024"
```
mtrmac, line 1: Nit: section 5 or 7

docs/containers-storage-architecture.md
```diff
@@ -0,0 +1,76 @@
...
+
+NOTE: These are implementation details subject to change. This document
+is intended to be used only as an aid for debugging and development
+of the containers-storage system. Do *not* directly access the filesystem
+components below from external code, only from the Go API.
+
+## DESCRIPTION
+The *containers/storage* library manages the following object types:
+
+- layer: A filesystem tree, along with metadata
```
mtrmac, line 14: … organized in, and optimized for, a tree of parent/child links, where a single “layer” may either serve as a complete filesystem tree, or a difference from the parent.

docs/containers-storage-architecture.md
```diff
@@ -0,0 +1,76 @@
...
+- image: A reference to a layer, along with metadata
+- container: A layer whose parent is an image
```
mtrmac, line 16: These feel a bit too reductionist to me.

The important thing about images is that, unlike, layers, users frequently with deal with them directly; that they have human-usable names; and that they are nevertheless deduplicated.

The important thing about containers is… ? That the top layer is read-write? That a lot of the metadata is actually not in c/storage but in Podman’s / CRI-O’s data? Also, technically the parent of the container’s layer is still a layer.

docs/containers-storage-architecture.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Layer storage
```
mtrmac, line 18: Discusssion of graph drivers?

docs/containers-storage-architecture.md
```diff
@@ -0,0 +1,76 @@
...
+
+### `overlay/`
+
+The `overlay/` directory contains a set of layers, named
+by their digest (note, this is the "diffid" or *uncompressed*
+digest, not the compressed one).
```
mtrmac, line 24: No, this is a “chain ID” identifying the precise sequence of parent1→parent2→…→child layer IDs; and with zstd:chunked, also identifying the contents of the layer depending on how we actually created it.

I.e. a layer with the same DiffID can have _several_ physical objects in this directory.

docs/containers-storage-architecture.md
```diff
@@ -0,0 +1,76 @@
...
+
+### layer:
+
+A layer has the following components:
+
+- `diff/`: The expanded filesystem tree
+- `work/`: A directory used by `overlayfs` mounts
+- `merged/`: This will be a mount point when a layer is mounted.
+- `link`: A regular file; this contains a unique string for the layer; see `l/` below.
+- `lower`: Contains all the lower layer mounts separated by ":" and ordered from uppermost to lowermost layers.
```
mtrmac, line 34: There _can_ be layers with other layout / features nowadays, but the combinatorics of `drivers/overlay` scares me.

docs/containers-storage-architecture.md
```diff
@@ -0,0 +1,76 @@
...
+Global list of metadata associated with layers.
+
+#### `volatile-layers.json`
+
+Like `layers.json`, but used for layers which are defined not to
+persist across reboots.
+
+### storage.lock
+
+A cross-process lockfile, protecting overall layer state.
```
mtrmac, line 76: I’d much prefer the lock semantic documentation to be centralized in one place, near the code, throughout. Having these short paragraphs duplicating/paraphrasing the actual rules seems to me more risky than helpful.

E.g., the semantics of `graphLock` is much more narrow (but, also, under documented).

docs/containers-storage-architecture.md
```diff
@@ -0,0 +1,76 @@
...
+## DESCRIPTION
+The *containers/storage* library manages the following object types:
+
+- layer: A filesystem tree, along with metadata
+- image: A reference to a layer, along with metadata
+- container: A layer whose parent is an image
+
+## Layer storage
+
+### `overlay/`
```
mtrmac, line 20: I, very weakly, think that pointing at, and improving, the documentation in `drivers/overlay/overlay.go` might be more helpful than duplicating it.

mtrmac (commented):
Just some quick comments because I glanced at the PR list…

One vague thought: that there are two very different audiences who might read this: support investigators who need to understand some specific broken filesystem state, and developers who need to understand all possible combinations and rules and invariants.

docs/containers-storage-architecture.md
cgwalters: Yes, this makes total sense.  I've moved this text into `overlay.go.`

docs/containers-storage-architecture.md
cgwalters: Would make sense to add in theory but I personally don't care much since I think we're in a place where there is only the `overlay` driver that matters, right?

rhatdan:
@giuseppe PTAL

rhatdan:
LGTM

cgwalters:
This one isn't urgent, I think we can just leave it as draft and I may circle back to it later. No need for others to dive in for review now, we have more important things.

PR #1935, opened 2024-06-01
Trivial patches to minimize patches needed by NetBSD

coypoop:
Hi folks, I've been patching your repo to build podman on NetBSD.
The resulting podman seems to work well enough.

I'd like to minimize the amount of patches needed, a lot of them are about list of supported OSes.
(There are some remaining patches that I still have to think about).


rhatdan:
/approve
LGTM

rhatdan:
@giuseppe @saschagrunert @nalind @mheon @vrothberg @flouthoc @TomSweeneyRedHat PTAL

saschagrunert (approved):

nalind:
Thanks!  A few notes.
Some of these could use the "unix" build constraint that was added in Go 1.19 instead of "!windows", though I don't think we've done that before.
Removing `homedir_others.go` looks fine, since I don't think we care about plan9 and javascript and wasi platforms.
This should probably add one or more netbsd targets to the list of targets that the "local-cross" target in the top-level `Makefile` attempts to compile for, to prevent regressions on this by catching them in CI.

dfr:
Looks good, thanks!

PR #1936, opened 2024-06-03
chunked: ignore the tar-split data if digest is empty

giuseppe:
if a digest was not specified in the TOC, ignore completely the tar-split data.

Otherwise the clients fail to pull images created before commit b5413c2bd64742a08384300f09616a66231b21d1.

### DIFF
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -210,14 +210,18 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
 
 	decodedTarSplit := []byte{}
 	if tarSplitChunk.Offset > 0 {
-		tarSplit, err := readBlob(tarSplitChunk.Length)
-		if err != nil {
-			return nil, nil, nil, 0, err
-		}
+		tarSplitDigest := toc.TarSplitDigest.String()
+		// ignore the tar-split data if the digest was not specified
+		if tarSplitDigest != "" {
+			tarSplit, err := readBlob(tarSplitChunk.Length)
+			if err != nil {
+				return nil, nil, nil, 0, err
+			}
 
-		decodedTarSplit, err = decodeAndValidateBlob(tarSplit, tarSplitLengthUncompressed, toc.TarSplitDigest.String())
-		if err != nil {
-			return nil, nil, nil, 0, fmt.Errorf("validating and decompressing tar-split: %w", err)
+			decodedTarSplit, err = decodeAndValidateBlob(tarSplit, tarSplitLengthUncompressed, toc.TarSplitDigest.String())
+			if err != nil {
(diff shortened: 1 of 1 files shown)


giuseppe:
@mtrmac PTAL

I wonder if it would be cleaner to also move the tar-split offsets into the TOC, instead of having half the information in the TOC and half in the annotations.

cgwalters (approved):

pkg/chunked/compression_linux.go
```diff
@@ -210,14 +210,18 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
...
-		tarSplit, err := readBlob(tarSplitChunk.Length)
-		if err != nil {
-			return nil, nil, nil, 0, err
-		}
-
-		decodedTarSplit, err = decodeAndValidateBlob(tarSplit, tarSplitLengthUncompressed, toc.TarSplitDigest.String())
-		if err != nil {
-			return nil, nil, nil, 0, fmt.Errorf("validating and decompressing tar-split: %w", err)
+		tarSplitDigest := toc.TarSplitDigest.String()
+		// ignore the tar-split data if the digest was not specified
```
mtrmac, line 214: We must actually _consume_ the data, otherwise the producer of `(parts, errs)` is going to block forever (and keep the HTTP connection occupied).

pkg/chunked/compression_linux.go
```diff
@@ -210,14 +210,18 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
...
+		if tarSplitDigest != "" {
```
mtrmac, line 215: (Absolutely non-blocking: I’d just do `if tarSplitChunk.Offset > 0 && toc.TarSplitDigest != ""`, directly comparing `digest.Digest` to `""` is fairly frequent, and one less nesting level is nice.)

mtrmac (commented):
Actually … with the design of #1902, the fix on #1888 wants the layer contents to be _fully determined_ by the TOC.

So:
- if `toc.TarSplitDigest == ""`, the tar-split must be ignored (what this PR does)
- if `toc.TarSplitDigest != ""`, we _must_ have, validate, and consume the tar-split; if `tarSplitChunk.Offset <= 0`, the layer is inconsistent and using it (at least by TOC) must fail.

pkg/chunked/compression_linux.go
giuseppe: thanks!

Opened a PR: https://github.com/containers/storage/pull/1942

mtrmac:
> Actually … with the design of #1902, the fix on #1888 wants the layer contents to be _fully determined_ by the TOC.

https://github.com/containers/storage/pull/1967 , for the record.

Issue #1928, opened 2024-05-22
Partial Images Feature is slower than not using it


antheas:
### Issue Description

Hi, since we're hearing a lot about the `zstd:chunked` encoding, I ran some tests today, and across the board it appears that the partial images feature is 2-3x slower, regardless of the cached amount (although the tests do not show partially cached layers; it was tested).

When the partial images feature is on, under a registry and client that have more than ~100mbps of bandwidth, image pulling with the partial mages feature is 2-3x slower. Its closer to 2x slower when compared to `zstd:chunked` with the feature turned off and 3x slower compared to `gzip`'d layers.

### Steps to reproduce the issue

I ran the following with a 1gbps connection (as confirmed by speedtest.net ~900mbps) from github with a ping of 19ms, saved to a gen 4 nvme ssd, backed with a 7840U class AMD CPU (8c/16t/16GB RAM). 

```bash
# Without partial images
# enable_partial_images="false"
podman system prune -a
time podman run -it --rm ghcr.io/castrojo/bluefin:40-20240522 ls
# 54.45s user 36.50s system 193% cpu 47.015 total
time podman run -it --rm ghcr.io/castrojo/bluefin-dx:40-20240522 ls
# 32.62s user 18.76s system 69% cpu 1:14.38 total

# Downloading the large image on its own (!!!)
podman system prune -a
podman run -it --rm ghcr.io/castrojo/bluefin-dx:40-20240522 ls 
# 74.91s user 46.67s system 159% cpu 1:16.46 total

# With partial pulls
# enable_partial_images="true"
podman system prune -a
time podman run -it --rm ghcr.io/castrojo/bluefin:40-20240522 ls 
# 34.64s user 30.79s system 59% cpu 1:50.81 total
time podman run -it --rm ghcr.io/castrojo/bluefin-dx:40-20240522 ls 
# 19.51s user 17.06s system 22% cpu 2:40.20 total
```

For context, the image `bluefin-dx` extends `bluefin` with 2 layers, each 1GB (total 2GB).
```
REPOSITORY                   TAG          IMAGE ID      CREATED      SIZE
ghcr.io/castrojo/bluefin-dx  40-20240522  2ed1c6dd4200  4 hours ago  10.6 GB
ghcr.io/castrojo/bluefin     40-20240522  f7d19966c0e1  4 hours ago  7.09 GB
```

### Describe the results you received

[...]

PR #1937, opened 2024-06-03
chunked: change auto merge threshold to 1024

giuseppe:
Increase the threshold for auto-merging parts from 128 to 1024. This change aims to reduce the number of parts in an HTTP multi-range request, thus increasing the likelihood that the server will accept the request.

The previous threshold of 128 often resulted in a large number of small ranges, which could lead to HTTP multi-range requests being rejected by servers due to the excessive number of parts.

It partially addresses the reported issue.

Reported-by: https://github.com/containers/storage/issues/1928

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -42,7 +42,7 @@ import (
 
 const (
 	maxNumberMissingChunks  = 1024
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
Should this be configurable?

rhatdan:
@saschagrunert @mheon @vrothberg @mtrmac @nalind PTAL

mtrmac (commented):
LGTM

PR #1942, opened 2024-06-03
chunked: fix deadlock by always consuming tar-split

giuseppe:
always consume the tar-split data when present to avoid blocking the producer. Previously, the tar-split data was only read when the digest was specified.

commit 6875c9fbcf1cb6e06ee5e363982a059adec9693b introduced the regression.

follow-up for: https://github.com/containers/storage/pull/1936#discussion_r1624617123

### DIFF
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -210,14 +210,13 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
 
 	decodedTarSplit := []byte{}
 	if tarSplitChunk.Offset > 0 {
-		tarSplitDigest := toc.TarSplitDigest.String()
-		// ignore the tar-split data if the digest was not specified
-		if tarSplitDigest != "" {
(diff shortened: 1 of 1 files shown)


pkg/chunked/compression_linux.go
```diff
@@ -210,14 +210,13 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
...
-			if err != nil {
-				return nil, nil, nil, 0, err
-			}
-
+		// we must consume the data to not block the producer
+		tarSplit, err := readBlob(tarSplitChunk.Length)
+		if err != nil {
+			return nil, nil, nil, 0, err
+		}
+		// but ignore it when the digest is not present
```
mtrmac, line 218:
```suggestion
		// but ignore it when the digest is not present , because we can’t authenticate it against tocDigest
```
or ”… when we can’t authenticate it”, to say not just what, but also why.

giuseppe: fixed

PR #1944, opened 2024-06-04
layers: allow mount in ro stores

giuseppe:
There is nothing conceptually wrong in this as the Mount itself does not modify the store.

newFileGetter() works today with overlay only because it uses the drivers.DiffGetterDriver shortcut that reads files directly from their storage.

graph drivers that do not allow that, need to mount the layer instead. composefs could be changed to use the same code as overlay, but it requires more refactoring to propagate the TOC information down to the overlay layer.  For now just fix mounting from a ro store, as newFileGetter() already assumes anyway.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -408,7 +408,7 @@ type layerStore struct {
 	// The following fields can only be read/written with read/write ownership of inProcessLock, respectively.
 	// Almost all users should use startReading() or startWriting().
 	lastWrite           lockfile.LastWrite
-	mountsLastWrite     lockfile.LastWrite // Only valid if lockfile.IsReadWrite()
+	mountsLastWrite     lockfile.LastWrite
 	layers              []*Layer
 	idindex             *truncindex.TruncIndex
 	byid                map[string]*Layer
@@ -616,7 +616,7 @@ func (r *layerStore) modified() (bool, error) {
 	if m {
 		return true, nil
 	}
-	if r.lockfile.IsReadWrite() {
+	if r.mountsLockfile != nil {
 		// This means we get, release, and re-obtain, r.mountsLockfile if we actually need to do any kind of reload.
 		// That’s a bit expensive, but hopefully most callers will be read-only and see no changes.
 		// We can’t eliminate these mountsLockfile accesses given the current assumption that Layer objects have _some_ not-very-obsolete
@@ -687,7 +687,7 @@ func (r *layerStore) reloadIfChanged(lockedForWriting bool) (bool, error) {
 		r.lastWrite = lastWrite
 		return false, nil
 	}
-	if r.lockfile.IsReadWrite() {
+	if r.mountsLockfile != nil {
 		r.mountsLockfile.RLock()
 		defer r.mountsLockfile.Unlock()
 		if err := r.reloadMountsIfChanged(); err != nil {
@@ -880,7 +880,7 @@ func (r *layerStore) load(lockedForWriting bool) (bool, error) {
 	r.bytocsum = tocsums
 
 	// Load and merge information about which layers are mounted, and where.
-	if r.lockfile.IsReadWrite() {
+	if r.mountsLockfile != nil {
 		r.mountsLockfile.RLock()
 		defer r.mountsLockfile.Unlock()
 		// We need to reload mounts unconditionally, becuause by creating r.layers from scratch, we have discarded the previous
@@ -1047,9 +1047,6 @@ func (r *layerStore) saveLayers(saveLocations layerLocations) error {
 // The caller must hold r.mountsLockfile for writing.
 // The caller must hold r.inProcessLock for WRITING.
 func (r *layerStore) saveMounts() error {
-	if !r.lockfile.IsReadWrite() {
-		return fmt.Errorf("not allowed to modify the layer store at %q: %w", r.layerdir, ErrStoreIsReadOnly)
-	}
 	r.mountsLockfile.AssertLockedForWriting()
 	mpath := r.mountspath()
 	if err := os.MkdirAll(filepath.Dir(mpath), 0o700); err != nil {
@@ -1156,23 +1153,24 @@ func (s *store) newLayerStore(rundir, layerdir, imagedir string, driver drivers.
 }
 
 func newROLayerStore(rundir string, layerdir string, driver drivers.Driver) (roLayerStore, error) {
-	lockfile, err := lockfile.GetROLockFile(filepath.Join(layerdir, "layers.lock"))
+	roLockfile, err := lockfile.GetROLockFile(filepath.Join(layerdir, "layers.lock"))
 	if err != nil {
 		return nil, err
 	}
+	// Ignore errors if the lock file could not be opened
+	mountsLockfile, _ := lockfile.GetLockFile(filepath.Join(rundir, "mountpoints.lock"))
 	rlstore := layerStore{
-		lockfile:       newMultipleLockFile(lockfile),
-		mountsLockfile: nil,
+		lockfile:       newMultipleLockFile(roLockfile),
+		mountsLockfile: mountsLockfile,
 		rundir:         rundir,
 		jsonPath: [numLayerLocationIndex]string{
 			filepath.Join(layerdir, "layers.json"),
 			filepath.Join(layerdir, "volatile-layers.json"),
 		},
 		layerdir: layerdir,
-
-		byid:    make(map[string]*Layer),
-		byname:  make(map[string]*Layer),
-		bymount: make(map[string]*Layer),
+		byid:     make(map[string]*Layer),
+		byname:   make(map[string]*Layer),
+		bymount:  make(map[string]*Layer),
 
 		driver: driver,
 	}
@@ -1579,6 +1577,9 @@ func (r *layerStore) Mount(id string, options drivers.MountOpts) (string, error)
 	if !r.lockfile.IsReadWrite() && !hasReadOnlyOpt(options.Options) {
 		return "", fmt.Errorf("not allowed to update mount locations for layers at %q: %w", r.mountspath(), ErrStoreIsReadOnly)
 	}
+	if r.mountsLockfile == nil {
+		return "", fmt.Errorf("cannot mount layers at %q: %w", r.mountspath(), ErrStoreIsReadOnly)
+	}
 	r.mountsLockfile.Lock()
 	defer r.mountsLockfile.Unlock()
 	if err := r.reloadMountsIfChanged(); err != nil {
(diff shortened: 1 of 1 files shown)


layers.go
```diff
@@ -1156,23 +1153,24 @@ func (s *store) newLayerStore(rundir, layerdir, imagedir string, driver drivers.
 }
 
 func newROLayerStore(rundir string, layerdir string, driver drivers.Driver) (roLayerStore, error) {
-	lockfile, err := lockfile.GetROLockFile(filepath.Join(layerdir, "layers.lock"))
+	roLockfile, err := lockfile.GetROLockFile(filepath.Join(layerdir, "layers.lock"))
 	if err != nil {
 		return nil, err
 	}
+	// Ignore errors if the lock file could not be opened
```
mtrmac, line 1160: Why is this safe? What are callers going to do with this data?

layers.go
```diff
@@ -1579,6 +1577,9 @@ func (r *layerStore) Mount(id string, options drivers.MountOpts) (string, error)
 	if !r.lockfile.IsReadWrite() && !hasReadOnlyOpt(options.Options) {
 		return "", fmt.Errorf("not allowed to update mount locations for layers at %q: %w", r.mountspath(), ErrStoreIsReadOnly)
 	}
+	if r.mountsLockfile == nil {
+		return "", fmt.Errorf("cannot mount layers at %q: %w", r.mountspath(), ErrStoreIsReadOnly)
```
mtrmac, line 1581: `unmount` must be symmetrical.

mtrmac (changes requested):
What about the prominent `LOCKING BUG` comments? Historically they didn’t very much matter to us because they were limited to Btrfs and ZFS, which we weren’t using.

If this is proposing to make the `newFileGetter` mount/unmount code used in production, that locking bug AFAICS must be resolved _first_.

---

At a fairly cursory skim, this is missing quite a few other updates:
- `.modified()`, `.reloadIfChanged()`
- `.Mounted()`, `ParentOwners()`?
- The documentation, and uses? of `mountsLastWrite`, need updating

I didn’t look any further, there may be much more.

giuseppe:
the current code just crashes at the moment, because `mountsLockfile` is `nil` in an additional layer store.

I think the long term fix is to teach overlay how to look up files in the composefs store without having to mount the layer ( we need to propagate somehow the TOC to the overlay layer so it can lookup the file digest).

The current PR is just a band-aid to get composefs tests to pass in the Podman CI: https://github.com/containers/podman/pull/22888

mtrmac:
I’m rather opposed to band-aids that introduce locking problems, and/or incomplete/inconsistent behavior involving locks. I don’t think there will ever be a more convenient time to get that fixed; so if we are serious about locking, we must be serious now. And I think we must be serious about locking.

---

If the code currently crashes, replace that with cleanly returning an error; on all similarly relevant code paths. That’s a clear improvement. Then we can see about any other improvement.

giuseppe:
looking at the issue:

```
	// LOCKING BUG: This is reachable via store.Diff → layerStore.Diff → layerStore.newFileGetter
	// (with btrfs and zfs graph drivers) holding layerStore only locked for reading, while it modifies
	// - r.layers[].MountCount (directly and via loadMounts / saveMounts)
	// - r.layers[].MountPoint (directly and via loadMounts / saveMounts)
	// - r.bymount (via loadMounts / saveMounts)
```

shouldn't these just be protected by `mountsLockfile` as they are?

mtrmac:
- It can’t _currently_ be `mountsLockFile` because (related to this PR) read-only stores might not have that lock at all, and/or might not be able to lock it for writing.
- “load” is a _write_ to the in-memory state. I.e. holding a cross-process _read_ lock, we need an in-process _write_ lock. I don’t immediately see that using the same lock for both is desirable.

There’s a kernel of an idea here — maybe we _could_ make the in-memory storage more granular, with some fields covered a generic layer store lock, and some other fields only by a mount lock. (I would worry about the two getting completely out of sync, e.g. WRT layer existence.)

But, fundamentally, `layerStore.Diff` is currently documented to only require `startReading`, which means it is potentially concurrent with other operations (including concurrent `Diff` calls), and _outright not allowed to modify any layer-store state at all_. It’s “easy” to change that, but I vaguely remember holding a shared-lock here being important for performance.

giuseppe:
> * It can’t _currently_ be `mountsLockFile` because (related to this PR) read-only stores might not have that lock at all, and/or might not be able to lock it for writing.

I don't understand why it would be a problem.  If the `mountsLockFile` file is not usable, then it must not be possible to mount an image (the code in the PR returns an error now: https://github.com/containers/storage/pull/1944/files#diff-04662d0ee53d0732912a23987c7631b5d07aa429ea69c18e1da00903a6362904R1580-R1581.

We can improve it further by teaching `DiffGetter` how to use composefs.  Once that happens, there won't be need to use `mountsLockFile`, as we can access the files in a layer without having to mount it, as it happens now with overlayfs.  But until then, I think it is fine to use the naive approach and mount the image to access its files and require a writeable `mountsLockFile` to do so.

I think we already have a granular enough locking mechanism now, where `mountsLockFile` protects only the mount information.  Anyway the `mount.Mounted()` function is the ultimate truth whether a mount exists or not, since the process could have been killed between performing `mount/unmount` and updating the mounts info.  This was the cause of a long term Podman race condition, where the podman cleanup process is terminated before it could update the mounts info.

layers.go
```diff
@@ -1156,23 +1153,24 @@ func (s *store) newLayerStore(rundir, layerdir, imagedir string, driver drivers.
...
-
-		byid:    make(map[string]*Layer),
-		byname:  make(map[string]*Layer),
-		bymount: make(map[string]*Layer),
+		byid:     make(map[string]*Layer),
+		byname:   make(map[string]*Layer),
+		bymount:  make(map[string]*Layer),
 
 		driver: driver,
 	}
```
mtrmac, line 1176: Also add the `rlstore.mountsLastWrite is initialized inside rlstore.load().` comment below, matching the other constructor.

layers.go
```diff
@@ -1156,23 +1153,24 @@ func (s *store) newLayerStore(rundir, layerdir, imagedir string, driver drivers.
 }
 
 func newROLayerStore(rundir string, layerdir string, driver drivers.Driver) (roLayerStore, error) {
-	lockfile, err := lockfile.GetROLockFile(filepath.Join(layerdir, "layers.lock"))
+	roLockfile, err := lockfile.GetROLockFile(filepath.Join(layerdir, "layers.lock"))
```
mtrmac, line 1156: Reading `getROLayerStoresLocked`, `rundir` is the same for _all_ RO layer stores, so this doesn’t work; they are going to stomp over each others’ data.

And if we assume that two concurrent processes could share the same layer store on the same system, one process as primary, the second as a read-only additional … this just doesn’t work. _And_ I can’t see a upgrade/migration path for processes currently using the `rundir`-based locations.

mtrmac (commented):
Regardless of incremental improvements, building composefs on top of the `newFileGetter` race condition continues to be a hard blocker for me.

mtrmac:
> > * It can’t _currently_ be `mountsLockFile` because (related to this PR) read-only stores might not have that lock at all, and/or might not be able to lock it for writing.
> 
> I don't understand why it would be a problem. If the `mountsLockFile` file is not usable, then it must not be possible to mount an image

An example of the relevant race is between `layerStore.Mount` changing `layer.MountPoint` vs. `layerStore.Get` making a copy of the string. A race here (e.g. a read in the middle of a torn write) results in undefined behavior. I think you’re right that those races are not reachable if `mountsLockFile` is not usable (assuming we rigorously document that the fields driven by `mountpoints.json` are never modified if the lock is `nil`).

But, also…

---

The ABI is such that a `store.Layer.Mount{Point,Count}` _always exist_, we don’t get to declare a value “unavailable”. Sure, that’s a mistake, and any out-of-c/storage consumers of those values are unfixably racy, but there are such consumers in Podman, and we need to somehow deal with them.

So far, the design has been that _only_ layers in the primary store get mounted; for the other stores, the layers would have no mount point and zero mount count, and that’s AFAICT consistent and correct (unless a primary layer store in one process is a secondary layer store in another process — the latter one wouldn’t see counts).

If, after this PR, the RO layer stores can be mounted, but that data is _maybe_, _without any reporting_ ( https://github.com/containers/storage/pull/1944#discussion_r1625927036 ) available or not … what does that mean in consumers?

And ”`mountsLockFile` is not usable” is not necessarily a system-wide consistent value: the file might exist, but _some_ processes might fail to use it, e.g. due to SELinux.

---

> Anyway the `mount.Mounted()` function is the ultimate truth whether a mount exists or not, since the process could have been killed between performing `mount/unmount` and updating the mounts info.

Because we reference-count unmounts, `mount.Mounted()` is not _sufficient_, but, plausibly, we could do the reference counting without involving `inProcessLock` — if we didn’t expose that data in `Layer`.

Now, I agree that the data in `Layer.Mount*` should, in principle, not exist, but I also don’t see a short-term path for avoiding that. The “easy” start for that change would be to remove all consumers of those fields, and that already looked non-obvious IIRC. I don’t know how we would handle correctness across multiple versions of c/storage e.g. during an upgrade or if Podman/Buildah/CRI-O get out of sync. (Also https://github.com/containers/storage/pull/1944#discussion_r1626442118 )

layers.go
giuseppe: I've not touched the existing logic here.  I've only renamed the variable since I need to refer to the `lockfile` package later

mtrmac: I’m sorry, the comment was actually directed at `mountsLockFile` (which is based on `rundir`)

layers.go
```diff
@@ -1156,23 +1153,24 @@ func (s *store) newLayerStore(rundir, layerdir, imagedir string, driver drivers.
...
 	if err != nil {
 		return nil, err
 	}
+	// Ignore errors if the lock file could not be opened
+	mountsLockfile, _ := lockfile.GetLockFile(filepath.Join(rundir, "mountpoints.lock"))
```
mtrmac, line 1161: Reading `getROLayerStoresLocked`, `rundir` is the same for _all_ RO layer stores, so this doesn’t work; they are going to stomp over each others’ data.

And if we assume that two concurrent processes could share the same layer store on the same system, one process as primary, the second as a read-only additional … this just doesn’t work. _And_ I can’t see a upgrade/migration path for processes currently using the `rundir`-based locations.

layers.go
mtrmac: I’ve moved this thread to https://github.com/containers/storage/pull/1944/files#r1626500559 for clarity.

PR #1947, opened 2024-06-04
chunked: Set O_CLOEXEC

cgwalters:
I was just reading the code and I have a mental checklist item for "invoking open without O_CLOEXEC" that triggered here. (See also e.g.
https://github.com/containers/composefs/pull/185#discussion_r1322925050 )

It has security-relevant properties for us, xref
CVE-2024-21626 for example.

This isn't the only missing variant of this in this codebase, just using this targeted PR to test the waters for more PRs.


giuseppe (approved):
LGTM

PR #1949, opened 2024-06-05
graphdriver: Clarify name of fsverity const, tweak docs

cgwalters:
- Add comments that make clear these are internal values
- Rename "Enabled" to "IfAvailable" to clarify what it does
- Also note that `DifferFsVerityRequired` is not actually set by anything
  in this codebase.

Signed-off-by: Colin Walters <walters@verbum.org>

---

### DIFF
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -215,20 +215,25 @@ const (
 	DifferOutputFormatFlat
 )
 
+// DifferFsVerity is a part of the experimental Differ interface and should not be used from outside of c/storage.
+// It configures the fsverity requirement.
 type DifferFsVerity int
 
 const (
 	// DifferFsVerityDisabled means no fs-verity is used
 	DifferFsVerityDisabled = iota
 
-	// DifferFsVerityEnabled means fs-verity is used when supported
-	DifferFsVerityEnabled
+	// DifferFsVerityIfAvailable means fs-verity is used when supported by
+	// the underlying kernel and filesystem.
+	DifferFsVerityIfAvailable
 
-	// DifferFsVerityRequired means fs-verity is required
+	// DifferFsVerityRequired means fs-verity is required.  Note this is not
+	// currently set or exposed by the overlay driver.
 	DifferFsVerityRequired
 )
 
-// DifferOptions overrides how the differ work
+// DifferOptions is a part of the experimental Differ interface and should not be used from outside of c/storage.
+// It overrides how the differ works.
 type DifferOptions struct {
 	// Format defines the destination directory layout format
 	Format DifferOutputFormat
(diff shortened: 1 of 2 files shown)


drivers/driver.go
```diff
@@ -221,10 +221,12 @@ const (
 	// DifferFsVerityDisabled means no fs-verity is used
 	DifferFsVerityDisabled = iota
```
mtrmac, line 222: Drive-by: these are public symbols.

Yes, the documentation of `Differ` opts out of the stability promise … but if we are going to keep changing this, could we actually hide the details?

Make `graphdriver.Differ` a public alias to a private struct containing a private field of an interface type, with all the private fields in an internal subpackage so that the `Differ` object can be only created, and consumed, by users who can access that internal subpackage. And then move all the helper optiosn/types into that same internal subpackage. Compare `c/image/signature/signer`.

---

That’s all probably a separate PR (Cc: @Honny1 ); in this PR, maybe just add
```go
// This is a part of the experimental Differ interface and should not be used from outside of c/storage
```
to the `DifferFsVerity` and `DifferOptions` symbols.

mtrmac (commented):
ACK in general.

rhatdan:
/approve
LGTM

drivers/driver.go
cgwalters: > Make graphdriver.Differ a public alias to a private struct containing a private field of an interface type, with all the private fields in an internal subpackage so that the Differ object can be only created, and consumed, by users who can access that internal subpackage.

Sounds like Go really needs the equivalent of Rust's `pub(crate)`...

I think there's pkg/internal as a somewhat idiomatic approach for stuff like this though? https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface

cgwalters: Briefly looked, it seems like `mkdir pkg/internal && git mv drivers pkg/internal` would make sense as a starting point?  There's some imports that would need to be fixed, e.g. `cmd` but ISTM that those should be using a higher level API anyways?

mtrmac: In principle, I’d very much like moving graph drivers to `internal` (maybe without the `pkg`, I don’t think that helps), but the graph driver API is public via `Store.GraphDriver()`, and actually used in Podman and CRI-O a bit.

Those should almost certainly be migrated to some other API, `GraphDriver` callers have no access to c/storage locks — but we’d actually need to do that in c/storage, migrate callers, and only afterwards remove this API surface. [Or, maybe, keep a stub “implementing” the currently-visible public API, but that is probably too much work for no known benefit.]

mtrmac: (For the record, `internal` is not just an idiom, it is tool-enforced. But, like some other things related to subpackages, it’s not documented in the language spec. There’s https://go.dev/doc/go1.4#internalpackages .)

PR #1950, opened 2024-06-05
overlay: extend overlayFileGetter to understand composefs

giuseppe:
extend the overlayFileGetter interface to correctly read files from a composefs layer.

The overlayFileGetter creates a temporary mount of the composefs blob, grab a directory to its root and immediately unmount it so the mount is visible (and could be leaked) only for a short period.

Alternative to #1944 

Marked as a Draft until Podman tests pass

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -9,6 +9,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
@@ -2020,11 +2022,27 @@ func (d *Driver) getWhiteoutFormat() archive.WhiteoutFormat {
 }
 
 type overlayFileGetter struct {
-	diffDirs []string
+	diffDirs        []string
+	composefsMounts map[string]*os.File // map from diff dir to the directory with the composefs blob mounted
 }
 
 func (g *overlayFileGetter) Get(path string) (io.ReadCloser, error) {
+	buf := make([]byte, unix.PathMax)
 	for _, d := range g.diffDirs {
+		if f, found := g.composefsMounts[d]; found {
+			// there is no *at equivalent for getxattr, but it can be emulated by opening the file under /proc/self/fd/$FD/$PATH
+			len, err := unix.Getxattr(fmt.Sprintf("/proc/self/fd/%d/%s", int(f.Fd()), path), "trusted.overlay.redirect", buf)
+			if err != nil {
+				if errors.Is(err, unix.ENODATA) {
+					continue
+				}
+				return nil, &fs.PathError{Op: "getxattr", Path: path, Err: err}
+			}
+
+			// the xattr value is the path to the file in the composefs layer diff directory
+			return os.Open(filepath.Join(d, string(buf[:len])))
+		}
+
 		f, err := os.Open(filepath.Join(d, path))
 		if err == nil {
 			return f, nil
@@ -2037,7 +2055,16 @@ func (g *overlayFileGetter) Get(path string) (io.ReadCloser, error) {
 }
 
 func (g *overlayFileGetter) Close() error {
-	return nil
+	var errs *multierror.Error
+	for _, f := range g.composefsMounts {
+		if err := f.Close(); err != nil {
+			errs = multierror.Append(errs, err)
+		}
+		if err := unix.Rmdir(f.Name()); err != nil {
+			errs = multierror.Append(errs, err)
+		}
+	}
+	return errs.ErrorOrNil()
 }
 
 func (d *Driver) getStagingDir(id string) string {
@@ -2048,10 +2075,7 @@ func (d *Driver) getStagingDir(id string) string {
 // DiffGetter returns a FileGetCloser that can read files from the directory that
 // contains files for the layer differences, either for this layer, or one of our
 // lowers if we're just a template directory. Used for direct access for tar-split.
-func (d *Driver) DiffGetter(id string) (graphdriver.FileGetCloser, error) {
-	if d.usingComposefs {
-		return nil, nil
-	}
+func (d *Driver) DiffGetter(id string) (_ graphdriver.FileGetCloser, Err error) {
 	p, err := d.getDiffPath(id)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


drivers/overlay/overlay.go
```diff
@@ -2060,7 +2084,40 @@ func (d *Driver) DiffGetter(id string) (graphdriver.FileGetCloser, error) {
...
+	diffDirs := append([]string{p}, paths...)
+	for _, diffDir := range diffDirs {
+		// diffDir has the form $GRAPH_ROOT/overlay/$ID/diff, so grab the $ID from the parent directory
+		id := path.Base(path.Dir(diffDir))
+		composefsBlob := d.getComposefsData(id)
+		if fileutils.Exists(composefsBlob) != nil {
+			// not a composefs layer, ignore it
+			continue
+		}
+		dir, err := ioutil.TempDir(d.runhome, "composefs-mnt")
```
cgwalters, line 2106: Does anything GC this directory?

giuseppe: `overlayFileGetter.Close()` removes it, but it won't GC if the process is killed (although it is on tmpfs)

mtrmac:
I’m afraid this is not helpful, but just do avoid later surprises: Per https://github.com/containers/storage/pull/1899#discussion_r1608590912, I will not be approving PRs adding any more alternatives or indirections to the overlay driver.

That doesn’t at all mean that I think no-one should be working on those things, just that I can’t be helping.

cgwalters:
Seems fair; at a procedural level here: we're looking at using c/storage underneath bootc and I'm trying to spin up on the codebase more as I'm sure you've noticed.  Let me know what would be the most helpful/useful way to do that.  It'd be good to try to get to a place where have rough consensus and working code.

giuseppe:
> I’m afraid this is not helpful, but just to avoid later surprises: Per [#1899 (comment)](https://github.com/containers/storage/pull/1899#discussion_r1608590912), I will not be approving PRs adding any more alternatives or indirections to the overlay driver.

so where should it be done if the composefs implementation is in the overlay driver?

mtrmac:
I’m not saying this is the wrong place; just that I can’t keep up with the complexity of this subpackage as it is, and as it continues to grow.

mtrmac:
… thats “a me problem”, as the young kids say.

drivers/overlay/overlay.go
```diff
@@ -2036,8 +2053,18 @@ func (g *overlayFileGetter) Get(path string) (io.ReadCloser, error) {
...
 
-func (g *overlayFileGetter) Close() error {
-	return nil
+func (g *overlayFileGetter) Close() (Err error) {
+	for _, f := range g.composefsMounts {
+		err := f.Close()
+		if Err == nil {
+			Err = err
+		}
+		err = unix.Rmdir(f.Name())
```
rhatdan, line 2062: Should we logrus.Error(err) on else path, IE If Err != nil, then we should output all or errors, or just wrap all errors?

drivers/overlay/overlay.go
```diff
@@ -2036,8 +2053,18 @@ func (g *overlayFileGetter) Get(path string) (io.ReadCloser, error) {
...
+		if Err == nil {
+			Err = err
+		}
```
rhatdan, line 2065: Ditto?

drivers/overlay/overlay.go
rhatdan: Should we remove on errors below?

drivers/overlay/overlay.go
giuseppe: yes that is better.  I've changed it to use `go-multierror` now

rhatdan:
LGTM
@cgwalters @saschagrunert @nalind @alexlarsson PTAL

drivers/overlay/overlay.go
giuseppe: amended in the last version

giuseppe:
we need this for the composefs tests in the Podman CI

@edsantiago PTAL

drivers/overlay/overlay.go
```diff
@@ -2060,7 +2083,41 @@ func (d *Driver) DiffGetter(id string) (graphdriver.FileGetCloser, error) {
...
+				unix.Rmdir(f.Name())
+			}
+		}
+	}()
+	diffDirs := append([]string{p}, paths...)
+	for _, diffDir := range diffDirs {
+		// diffDir has the form $GRAPH_ROOT/overlay/$ID/diff, so grab the $ID from the parent directory
+		id := path.Base(path.Dir(diffDir))
+		composefsBlob := d.getComposefsData(id)
+		if fileutils.Exists(composefsBlob) != nil {
```
cgwalters, line 2102: This is only very tangentially related but I find the fact that someone went out of their way to create `fileutils.Exists` but made it return an *error* type instead of a `bool` confusing.  Reading `if fileutils.Exists(composefsBlob) != nil {` I have to stop and think for a second, but `if fileutils.Exists(composefsBlob) {` just would read more nicely...

Of course in Rust, there's https://doc.rust-lang.org/std/path/struct.Path.html#method.try_exists which is *really* what we want here; we should still throw an error IMO if we get e.g. `EIO` or something else.  Anyways.

giuseppe: we still need the error in case `Exists` fails, so it would be `exists, err := fileutils.Exists()` and `exists` is true only when `err == nil`

The `fileutils.Exists()` function is a replacement for the `_, err := os.Stat(...)` idiom that becomes `err := fileutils.Exists(...)`, so it is also easier to replace in an existing code base

cgwalters: Yeah sorry this is a minor irrelevant tangent, just marking as resolved

drivers/overlay/overlay.go
```diff
@@ -2020,11 +2021,27 @@ func (d *Driver) getWhiteoutFormat() archive.WhiteoutFormat {
...
+	diffDirs        []string
+	composefsMounts map[string]*os.File // map from diff dir to the directory with the composefs blob mounted
 }
 
 func (g *overlayFileGetter) Get(path string) (io.ReadCloser, error) {
+	buf := make([]byte, unix.PathMax)
 	for _, d := range g.diffDirs {
+		if f, found := g.composefsMounts[d]; found {
+			// there is no *at equivalent for getxattr, but it can be emulated by opening the file under /proc/self/fd/$FD/$PATH
+			len, err := unix.Getxattr(fmt.Sprintf("/proc/self/fd/%d/%s", int(f.Fd()), path), "trusted.overlay.redirect", buf)
```
cgwalters, line 2033: This `Getxattr` probably wants at some point to be factored out to a helper function, we have a *lot* of copies of the `/proc/self/fd/` trick in this codebase, and there is actually finally some movement on a proper syscall for `lgetxattrat` that we could change in one place instead of many.

---

On a different topic it feels a bit like we're poking into composefs internals here?  One thing I've thought about in the past that would be nice perhaps is something like an `ioctl()` usable on an overlayfs that would reopen the file descriptor with the real lower file.  Specifically, what I hit is I wanted to be able to reflink across an overlay view and a target in the real underlying filesystem.  That's a case that it seems to me we could make safe and viable, and in that model the overlayfs kernel implementation would handle "chasing" this `trusted.overlay.redirect`.

drivers/overlay/overlay.go
```diff
@@ -2020,11 +2021,27 @@ func (d *Driver) getWhiteoutFormat() archive.WhiteoutFormat {
...
+			if err != nil {
+				continue
+			}
+
+			// the xattr value is the path to the file in the composefs layer diff directory
+			f, err := os.Open(filepath.Join(d, string(buf[:len])))
+			if err == nil {
```
cgwalters, line 2040: I wonder if we should really fall through here if we hit an error?  Under what scenario would that happen, and it would be the right thing to just do the non-composefs overlay path?

cgwalters (commented):
I don't know this code well at all, but this seems overall sane to me, FWIW.

drivers/overlay/overlay.go
giuseppe: yes that is a mistake, thanks for catching it.  There is no point in the fall through when it is a composefs layer.

I've changed it and pushed a new version

drivers/overlay/overlay.go
```diff
@@ -2020,11 +2021,24 @@ func (d *Driver) getWhiteoutFormat() archive.WhiteoutFormat {
...
 }
 
 func (g *overlayFileGetter) Get(path string) (io.ReadCloser, error) {
+	buf := make([]byte, unix.PathMax)
 	for _, d := range g.diffDirs {
+		if f, found := g.composefsMounts[d]; found {
+			// there is no *at equivalent for getxattr, but it can be emulated by opening the file under /proc/self/fd/$FD/$PATH
+			len, err := unix.Getxattr(fmt.Sprintf("/proc/self/fd/%d/%s", int(f.Fd()), path), "trusted.overlay.redirect", buf)
+			if err != nil {
+				continue
```
rhatdan, line 2035: Should this be logrus.Error if not ENOENT?

giuseppe: I've not investigated it further, but from the man page it seems `ENODATA` is returned:

```
       ENODATA
              The  named  attribute  does not exist, or the process has no access to this attribute.
```

other failures seem more difficult to hit, but if you prefer I can return the error to the caller if it is not `ENODATA`

giuseppe: changed now to ignore only `ENODATA`

rhatdan: sgtm

rhatdan:
/lgtm
/hold
Wait for Podman tests to pass.

giuseppe:
tests pass: https://github.com/containers/podman/pull/22888/checks?check_run_id=25957981302

/hold cancel

PR #1951, opened 2024-06-05
dump: handle duplicates

giuseppe:
Add validation for duplicate entries.  Duplicates are ignored, unless there is a mismatch in the values.  In that case, an error is returned.

Also refactor the tests to avoid duplication.
    
### DIFF
--- a/pkg/chunked/dump/dump.go
+++ b/pkg/chunked/dump/dump.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cgwalters:
This makes sense to me, but can you explain briefly what motivated the change just for reference?  Did you come across an image in the wild with duplicates?  Were you just thinking through the problem domain?  Is this split out of a different larger pending work PR?

rhatdan:
LGTM

giuseppe:
> This makes sense to me, but can you explain briefly what motivated the change just for reference? Did you come across an image in the wild with duplicates? Were you just thinking through the problem domain? Is this split out of a different larger pending work PR?

no, I've not encountered yet such case.  I was just looking at ways to make the `dump` code more robust, since most of the issues we had with composefs were caused by the conversion to the dump format

PR #1952, opened 2024-06-06
chunked: refactor file system ops in a new file and add some tests

giuseppe:
more details in each commit


giuseppe:
@cgwalters @rhatdan PTAL

pkg/chunked/filesystem_linux.go
```diff
@@ -304,7 +304,7 @@ func openFileUnderRoot(dirfd int, name string, flags uint64, mode os.FileMode) (
 	if errors.Is(err, unix.ENOENT) && hasCreate {
 		parent := filepath.Dir(name)
 		if parent != "" {
-			newDirfd, err2 := openOrCreateDirUnderRoot(parent, dirfd, 0)
+			newDirfd, err2 := openOrCreateDirUnderRoot(dirfd, parent, 0)
```
cgwalters, line 307: On a related topic, wouldn't it make sense to create a struct wrapper for the dirfd, and have things be methods on it?

In my Rust code, we *extensively* use https://docs.rs/cap-std/latest/cap_std/ which is related, and has a `Dir` struct with a lot of methods, but in implementation it's just:

https://github.com/bytecodealliance/cap-std/blob/1dc17d327bdcd44a0f5ec53d9fc671285be55b34/cap-std/src/fs/dir.rs#L48-L50

pkg/chunked/filesystem_linux_test.go
```diff
@@ -50,3 +56,305 @@ func TestSeekableFileGetBlobAt(t *testing.T) {
...
+	tmpDir := createTempDir(t)
+	defer os.RemoveAll(tmpDir)
+
+	srcFile := createTempFile(t, tmpDir, "source")
+	defer srcFile.Close()
+	srcFd := int(srcFile.Fd())
+
+	destDir := createTempDir(t)
+	defer os.RemoveAll(destDir)
+	destDirFd, err := syscall.Open(destDir, syscall.O_RDONLY, 0)
```
cgwalters, line 70: Hello, your friendly `O_CLOEXEC` linter reporting in here :smile:

pkg/chunked/filesystem_linux.go
```diff
@@ -0,0 +1,551 @@
...
+	doChmod := func() error {
+		if usePath {
+			return unix.Fchmodat(dirfd, baseName, uint32(mode), unix.AT_SYMLINK_NOFOLLOW)
+		}
+		return unix.Fchmod(fd, uint32(mode))
+	}
+
+	if err := doChown(); err != nil {
+		if !options.IgnoreChownErrors {
+			return &fs.PathError{Op: "chown", Path: metadata.Name, Err: err}
```
mtrmac, line 170: … and, symmetrically to how `os.Stat` returns a `PathError`, I’d expect the `PathError` creation to happen at a fairly low level.

---

I don’t know how I feel about a `PathError` reporting a `Path` which is _not exactly what we used_. 

---

Also, the commit says “improve error messages”. but it removes information about non-path parameters to the operation. Is that intentional?

(I didn’t review more parts of the “improve error messages” commit in detail.)

pkg/chunked/filesystem_linux_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	err = newFile.Close()
+	assert.NoError(t, err)
+
+	// the file was copied with hard links, the inodes must be equal
+	assert.Equal(t, st.Dev, st2.Dev)
+	assert.Equal(t, st.Ino, st2.Ino)
+}
+
+func createTempDir(t *testing.T) string {
```
mtrmac, line 350: Just use `t.TempDir()` in callers, and there’s no need to worry about cleanup.

pkg/chunked/filesystem_linux_test.go
```diff
@@ -0,0 +1,360 @@
...
+	tmpDir, err := os.MkdirTemp("", "testDir")
+	assert.NoError(t, err)
+	return tmpDir
+}
+
+func createTempFile(t *testing.T, dir, name string) *os.File {
+	tmpFile, err := os.CreateTemp(dir, name)
+	assert.NoError(t, err)
```
mtrmac, line 358: (Test case setup should typically use `require.NoError` when there’s no point in proceeding. But that can always be fixed later, if the test ever actually fail.)

mtrmac (commented):
A quick skim:

- I like the consistent (parentFD, name) parameter ordering
- More tests are always good

(I didn’t carefully read the tests, and didn’t analyze the error reporting change impact in  detail.)

pkg/chunked/filesystem_linux.go
giuseppe: we need an explicit `mode` to support `ForceMask`

giuseppe:
thanks for the comments and suggestions.

Addressed them and pushed a new version

pkg/chunked/filesystem_linux.go
```diff
@@ -58,7 +63,7 @@ func copyFileContent(srcFd int, fileMetadata *fileMetadata, dirfd int, mode os.F
 	src := fmt.Sprintf("/proc/self/fd/%d", srcFd)
 	st, err := os.Stat(src)
 	if err != nil {
-		return nil, -1, fmt.Errorf("copy file content for %q: %w", destFile, err)
+		return nil, -1, err
```
rhatdan, line 66: Not sure this is an improvement on the error. 
Now we will get "failed to stat /proc/self/fd/%d" Rather then destFile?

giuseppe: I've dropped the patch from this PR.

I will rework errors later

pkg/chunked/filesystem_linux_test.go
```diff
@@ -0,0 +1,339 @@
...
+		[]byte("World"),
+	}
+
+	i := 0
+	for stream := range streams {
+		buf := new(bytes.Buffer)
+		_, err := buf.ReadFrom(stream)
+		require.NoError(t, err)
+
+		require.Equal(t, buf.Bytes(), expectedContents[i])
```
mtrmac, line 51: (FWIW, order is `Equal(t, expected, actual)`. Not generally worth worrying about, that that can be cleaned up if the test ever fails.)

pkg/chunked/filesystem_linux.go
```diff
@@ -0,0 +1,550 @@
...
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		destDirFd = int(f.Fd())
+	}
+
+	err = doHardLink(int(sourceFile.Fd()), destDirFd, destBase)
+	if err != nil {
+		return err
```
mtrmac, line 415: Was the removal of error wrapping here intentional?

mtrmac (commented):
LGTM

pkg/chunked/filesystem_linux_test.go
giuseppe: fixed now

pkg/chunked/filesystem_linux.go
giuseppe: no, it was not.  Thanks for catching it.  I'll deal with errors in a separate PR.  I've re-added it

mtrmac (commented):
/lgtm

PR #1953, opened 2024-06-07
overlay: Fix root directoy state with extended attributes

akihikodaki:
The upper layer inherits the root directory state (UID, GID, and mode) from the lower layer. The current code does not respect extended attributes in the process so this pull request implements necessary procedures.

Concretely, we must respect the lower layer's extended attribute if present. If "force mask" option is set, the upper layer should set the state to the extended attribute instead of modifying the real directory state.


pkg/idtools/idtools.go
```diff
@@ -38,9 +38,10 @@ func (e ranges) Swap(i, j int)      { e[i], e[j] = e[j], e[i] }
 func (e ranges) Less(i, j int) bool { return e[i].Start < e[j].Start }
 
 const (
-	subuidFileName          string = "/etc/subuid"
-	subgidFileName          string = "/etc/subgid"
-	ContainersOverrideXattr        = "user.containers.override_stat"
+	subuidFileName             string = "/etc/subuid"
+	subgidFileName             string = "/etc/subgid"
+	FuseoverlayfsOverrideXattr        = "user.fuseoverlayfs.override_stat"
```
giuseppe, line 43: we added the `user.containers.override_stat` xattr to not leak fuse-overlayfs specific xattrs into c/storage.

Why can't we always use `user.containers.override_stat`?

giuseppe (commented):
do you have any tests case where the current code currently fails?

Could we add a test?

pkg/idtools/idtools.go
akihikodaki: fuse-overlayfs uses `user.fuseoverlayfs.override_stat` for the upper layer and setting `user.containers.override_stat` for the upper layer does not work.

Personally, I don't see a reason to have a distinct extended attribute for each of c/storage and fuse-overlayfs. We can just have a unified extended attribute `user.override_stat` and use it both in c/storage, fuse-overlayfs, and potentially other components (kernel overlayfs, composefs, etc).

akihikodaki:
> do you have any tests case where the current code currently fails?
> 
> Could we add a test?

Added a test with commit 399c5194ee465994315c985ca5bd0a23e9f497fe.

pkg/idtools/idtools.go
giuseppe: so let's standardize on `user.containers.override_stat` and if necessary fix fuse-overlayfs, since `user.containers.override_stat` is already understood to fuse-overlayfs.

akihikodaki: The problem is fuse-overlayfs still uses `user.fuseoverlayfs.override_stat` for the upper layer though it can consume `user.containers.override_stat` for lower layers.
We can use `user.containers.override_stat` we change fuse-overlayfs to use `user.containers.override_stat` for the upper layer. Is that what you suggest?

giuseppe: yes, I suggest to change fuse-overlayfs to use `user.containers.override_stat` so we don't leak fuse-overlayfs specific stuff into c/storage.

akihikodaki: Opened https://github.com/containers/fuse-overlayfs/pull/422 and removed references to `user.fuseoverlayfs.override_stat` from this change.

giuseppe (approved):
LGTM

rhatdan:
@akihikodaki Sorry to say this needs a rebase.

akihikodaki:
> @akihikodaki Sorry to say this needs a rebase.

Done.

giuseppe (approved):
LGTM

PR #1954, opened 2024-06-07
Fix build failures

mtrmac:
… so that `golangci-lint run ./...` on macOS at least compiles everything.


giuseppe (approved):
LGTM

PR #1956, opened 2024-06-07
chunked: Add helpers for /proc/self/fd accesses

cgwalters:
It's not hard to grep for /proc/self/fd, but this way is a bit cleaner and avoids typos, etc.


giuseppe (approved):
LGTM

mtrmac (commented):
/lgtm

PR #1939, opened 2024-06-03
compression: Add some doc comments

cgwalters:
I'm digging in more to zstd:chunked and I hope these comments are useful.

PR #1958, opened 2024-06-07
compression: Fix Deprecated: and actually comment out unused bits

cgwalters:
- Address https://github.com/containers/storage/pull/1939#pullrequestreview-2104615910
- Also just go ahead and comment this out because it's not used, but I think it makes sense to retain as a historical reference


mtrmac (commented):
Thanks for the follow-up!

Supposedly, all of c/storage has a stable API. That’s… not the case in practice.

cgwalters:
> Supposedly, all of c/storage has a stable API. That’s… not the case in practice.

This is under `internal/` right?

rhatdan:
Why not delete this rather then comment it out?

PR #1959, opened 2024-06-07
chunked: Fix two minor linter issues

cgwalters:
My IDE runs a linter by default, and these two show up. For the file one, it's because `Fd()` returns `uintptr` which is unsigned and can't be negative.  IOW, a `File` object should always be a valid opened fd.


mtrmac (commented):
/lgtm

giuseppe (approved):
LGTM

Issue #1955, opened 2024-06-07
chunked format string issues?


mtrmac:
Per `GOOS=linux golangci-lint run --no-config ./...`:
```
pkg/chunked/dump/dump.go:154:15: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
	if _, err := fmt.Fprintf(out, escapedOptional(payload, ESCAPE_LONE_DASH)); err != nil {
	             ^
pkg/chunked/dump/dump.go:168:15: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
	if _, err := fmt.Fprintf(out, escapedOptional(digest, ESCAPE_LONE_DASH)); err != nil {
```
and there may be more. Shouldn’t this use `Fprintf("%s")`, or some kind of `Write`, so that `%` characters are not interpreted?

Cc: @giuseppe 

[One of these days I’m going to make a non-consensus decision on #1579 …]

giuseppe:
it was just a too aggressive copy&paste.  Thanks for reporting it:

- https://github.com/containers/storage/pull/1960

I think it is a good idea to enable more linters

PR #1960, opened 2024-06-07
chunked: use Fprint instead of Fprintf

giuseppe:
replace Fprintf(file, string) with Fprint(file, string) since the argument is a raw string and it should not be treated as a formatting directive.

Closes: https://github.com/containers/storage/issues/1955


flouthoc (approved):
LGTM

mtrmac (commented):
/lgtm

Thanks!

PR #1962, opened 2024-06-07
pkg/archive.statDifferent(): care about mtime for directories

nalind:
When considering changes, don't ignore differences between directories when their mtimes are the only thing we check that's changed.  This creates a difference in behavior between the vfs and overlay drivers that shows up in build conformance tests, such as in https://github.com/containers/buildah/pull/5492.

### DIFF
--- a/pkg/archive/changes_unix.go
+++ b/pkg/archive/changes_unix.go
@@ -31,9 +31,9 @@ func statDifferent(oldStat *system.StatT, oldInfo *FileInfo, newStat *system.Sta
 		ownerChanged ||
 		oldStat.Rdev() != newStat.Rdev() ||
 		oldStat.Flags() != newStat.Flags() ||
+		!sameFsTimeSpec(oldStat.Mtim(), newStat.Mtim()) ||
 		// Don't look at size for dirs, its not a good measure of change
-		(oldStat.Mode()&unix.S_IFDIR != unix.S_IFDIR &&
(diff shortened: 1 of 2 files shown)


giuseppe (approved):
LGTM

pkg/archive/changes_unix.go
```diff
@@ -31,9 +31,9 @@ func statDifferent(oldStat *system.StatT, oldInfo *FileInfo, newStat *system.Sta
 		ownerChanged ||
 		oldStat.Rdev() != newStat.Rdev() ||
 		oldStat.Flags() != newStat.Flags() ||
+		!sameFsTimeSpec(oldStat.Mtim(), newStat.Mtim()) ||
 		// Don't look at size for dirs, its not a good measure of change
-		(oldStat.Mode()&unix.S_IFDIR != unix.S_IFDIR &&
-			(!sameFsTimeSpec(oldStat.Mtim(), newStat.Mtim()) || (oldStat.Size() != newStat.Size()))) {
+		(!sameFsTimeSpec(oldStat.Mtim(), newStat.Mtim()) || (oldStat.Size() != newStat.Size())) {
```
mtrmac, line 36: The code is now not making a special case for directories, while the comment still says it does.

Also the `sameFsTimeSpec` is now AFAICS happening redundantly twice.

A typo?

nalind: Good catch!  I kept the wrong check there.  PR incoming.

PR #1963, opened 2024-06-09
chunked: add fuzzing tests for the cache

giuseppe:
and fix some issues found with the new tests.

### DIFF
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -642,6 +642,9 @@ func readCacheFileFromMemory(bigDataBuffer []byte) (*cacheFile, error) {
 	if tagsLen > maxTagsLen {
 		return nil, fmt.Errorf("tags len %d exceeds the maximum allowed size %d", tagsLen, maxTagsLen)
 	}
+	if digestLen > tagLen {
+		return nil, fmt.Errorf("digest len %d exceeds the tag len %d", digestLen, tagLen)
+	}
 
 	tags := make([]byte, tagsLen)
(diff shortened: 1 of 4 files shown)


cgwalters:
I was unaware of this whole subsystem until now.  This is the thing keeping indicies of chunks that are used for inside-a-file partial pulls?

The changes seem sane superficially.

giuseppe:
> I was unaware of this whole subsystem until now. This is the thing keeping indicies of chunks that are used for inside-a-file partial pulls?

yes, this is needed to avoid parsing the json metadata files each time we lookup for a digest.  In the same cache file we store the files payload digest, the chunks digest and the hard link fingerprint.  The hard link fingerprint a digest that is used to find files that can be used for hard links deduplication (payload digest + UID + GID + mode + xattrs).

pkg/chunked/cache_linux.go
```diff
@@ -34,6 +34,8 @@ const (
 	// https://pages.cs.wisc.edu/~cao/papers/summary-cache/node8.html
 	bloomFilterScale  = 10 // how much bigger is the bloom filter than the number of entries
 	bloomFilterHashes = 3  // number of hash functions for the bloom filter
+
+	maxTagsLen = 100 * 1024 * 1024 // 100MB max size for tags len
```
rhatdan, line 38: Why not build a Const for this magic number since the two PRs both create it.

giuseppe: I've changed it to use the go-units package.

PR #1964, opened 2024-06-10
archive: use pigz|zstd if available

giuseppe:
use the command line tools when available as they are much faster than the Go libraries.

Especially for gzip, I've registered a 50% improvement

use zstd if available

the performance improvement is not as clear as with pigz, but it is
still measurable difference.

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -220,6 +220,12 @@ func DecompressStream(archive io.Reader) (_ io.ReadCloser, Err error) {
 		readBufWrapper := p.NewReadCloserWrapper(buf, xzReader)
 		return readBufWrapper, nil
 	case Zstd:
+		cleanup := func() {
+			p.Put(buf)
+		}
+		if rc, canUse := tryProcFilter([]string{"zstd", "-d"}, buf, cleanup); canUse {
+			return rc, nil
+		}
 		return zstdReader(buf)
 	default:
 		return nil, fmt.Errorf("unsupported compression format %s", (&compression).Extension())
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -169,10 +169,17 @@ func DetectCompression(source []byte) Compression {
 }
 
 // DecompressStream decompresses the archive and returns a ReaderCloser with the decompressed archive.
-func DecompressStream(archive io.Reader) (io.ReadCloser, error) {
+func DecompressStream(archive io.Reader) (_ io.ReadCloser, Err error) {
 	p := pools.BufioReader32KPool
 	buf := p.Get(archive)
 	bs, err := buf.Peek(10)
+
+	defer func() {
+		if Err != nil {
+			p.Put(buf)
+		}
+	}()
+
 	if err != nil && err != io.EOF {
 		// Note: we'll ignore any io.EOF error because there are some odd
 		// cases where the layer.tar file will be empty (zero bytes) and
@@ -214,9 +221,16 @@ func DecompressStream(archive io.Reader) (io.ReadCloser, error) {
 }
 
 // CompressStream compresses the dest with specified compression algorithm.
-func CompressStream(dest io.Writer, compression Compression) (io.WriteCloser, error) {
+func CompressStream(dest io.Writer, compression Compression) (_ io.WriteCloser, Err error) {
 	p := pools.BufioWriter32KPool
 	buf := p.Get(dest)
+
+	defer func() {
+		if Err != nil {
+			p.Put(buf)
+		}
+	}()
+
 	switch compression {
 	case Uncompressed:
 		writeBufWrapper := p.NewWriteCloserWrapper(buf, buf)
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -196,6 +196,12 @@ func DecompressStream(archive io.Reader) (_ io.ReadCloser, Err error) {
 		readBufWrapper := p.NewReadCloserWrapper(buf, buf)
 		return readBufWrapper, nil
 	case Gzip:
+		cleanup := func() {
+			p.Put(buf)
+		}
+		if rc, canUse := tryProcFilter([]string{"pigz", "-d"}, buf, cleanup); canUse {
+			return rc, nil
+		}
 		gzReader, err := gzip.NewReader(buf)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
Very cool
Would need to put in a Recommends for Podman and Buildah

rhatdan:
@mtrmac @vrothberg @cgwalters @baude @nalind @alexlarsson @saschagrunert PTAL

pkg/archive/filter.go
```diff
@@ -0,0 +1,78 @@
...
+	"io"
+	"os/exec"
+	"sync"
+
+	"github.com/containers/storage/pkg/pools"
+)
+
+var (
+	procFilterLock sync.Mutex
+	filterPath     map[string]string
```
mtrmac, line 14: Consider using a `sync.Map`; it is supposedly optimized for this kind of use.

pkg/archive/filter.go
```diff
@@ -0,0 +1,78 @@
...
+	r, w := io.Pipe()
+	cmd := exec.Command(path, args[1:]...)
+	cmd.Stdin = input
+	cmd.Stdout = w
+
+	f := &filterReader{reader: r}
+	go func() {
+		defer w.Close()
+		f.runErr = cmd.Run()
+		pool.Put(input)
```
mtrmac, line 75: I don’t think this function should be in the business of managing `pool`; let the caller worry about that, probably using `NewReadCloserWrapper` like on all the other paths.

(Uh… is the existing code in the caller safe? Shouldn’t it trigger on close of the _input_ to the decompression, rather than the _output_? It’s not obvious to me that they always happen in the expected order.

If it is not safe, that’s _also_ a reason for the caller to worry about that … and at least a separate commit.)

pkg/archive/filter.go
```diff
@@ -0,0 +1,78 @@
...
+
+	r, w := io.Pipe()
+	cmd := exec.Command(path, args[1:]...)
+	cmd.Stdin = input
+	cmd.Stdout = w
+
+	f := &filterReader{reader: r}
+	go func() {
+		defer w.Close()
+		f.runErr = cmd.Run()
```
mtrmac, line 74: - This is a data race vs the consumer goroutine
- I don’t immediately see that this `filterReader` is necessary: Isn’t it enough to do

```go
var err = errors.New("internal error: should never be seen")
defer w.CloseWithError(err) // CloseWithErr(nil) == Close()
err = cmd.Run()
```

mtrmac, line 74: (AFAICT that this captures and records stderr correctly, but I’d welcome a confirmation that you tested it — or a unit test.)

mtrmac (commented):
Taking the reported speedup on faith, sure, why not.

Tests are unhappy, I didn’t check the details of the failure.

cgwalters:
One side effect of using `pigz` is definitely that for consumers this will result in unreproducible gzip output compared to the previous Go gzip.  And pigz in particular almost always for not-small input produces different output.  Not fatal, but something to be aware of, and I suspect there are some consumers that are partially relying on the gzip "semi-reproducible" status quo, where the build context isn't caching the previous state via diffids.

giuseppe:
> One side effect of using `pigz` is definitely that for consumers this will result in unreproducible gzip output compared to the previous Go gzip. And pigz in particular almost always for not-small input produces different output. Not fatal, but something to be aware of, and I suspect there are some consumers that are partially relying on the gzip "semi-reproducible" status quo, where the build context isn't caching the previous state via diffids.

Thanks for the information, I didn't think of that problem and something to consider if we plan to use pigz also for the compressor.  The current PR changes only the decompression side

pkg/archive/filter.go
giuseppe: under what conditions will `err` ever be `errors.New("internal error: should never be seen")`?

rhatdan:
I think the pull side is much more important then the push side.  You usually push once and pull many times.  Stick to only the uncompress.

giuseppe:
ready for review, CI is green

rhatdan:
LGTM

mtrmac:
> One side effect of using `pigz` is definitely that for consumers this will result in unreproducible gzip output compared to the previous Go gzip.

FWIW we never promised the compressed blobs to be reproducible (and we can’t, compression implementations make no such promises — and e.g. https://github.com/klauspost/compress is changing too often for me to think that the output never changes).

And yes, from time to time we see users who say “it’s been working for us fine for a year”, and it’s a struggle to convince them them that is it might very well break the next week and they need to change their design.

But it does break from time to time, and yes they do need to change their design; and preventing any single output change would not help avoid that need to change the design.

pkg/archive/filter.go
mtrmac: Panic before `Run` returns.

mtrmac: … and it’s a belt-and-suspenders thing: this makes sure to avoid a hypothetical `CloseWithError(nil)` on the `panic` path — whether or not `panic` is relevant, the point is that the claim “the pipe is successfully closed _only_ on success” can be easily verified.

giuseppe:
I've added a new patch to the PR.  I've realized the existing implementation was leaking the buffer on errors

pkg/archive/archive.go
```diff
@@ -189,8 +189,15 @@ func DecompressStream(archive io.Reader) (io.ReadCloser, error) {
...
 	case Gzip:
+		cleanup := func() {
+			p.Put(buf)
+		}
+		if rc, canUse := tryProcFilter([]string{"pigz", "-d"}, buf, cleanup); canUse {
+			return rc, nil
+		}
 		gzReader, err := gzip.NewReader(buf)
 		if err != nil {
+			p.Put(buf)
```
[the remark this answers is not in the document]
giuseppe: thanks, fixed now

mtrmac (commented):
/lgtm

(Purely personal aesthetic preference: differentiate the `err` and `Err` variables more. Also, named return values are a bit dangerous, so emphasizing that it is a return value might help. But that’s fully up to the maintainers of c/storage.)

vrothberg:
@edsantiago, did you see performance improvements after vendoring this change into Podman CI?

Nice work, @giuseppe !

edsantiago:
@vrothberg I'm not ignoring your question. (1) this is not yet vendored into podman, and (2) our PR merge rate is so low these days that it's hard to get performance numbers. I am keeping this on my TODO list and will report back once there's something to learn.

PR #1965, opened 2024-06-11
chunked: use existing buffer for io.Copy

giuseppe:
utilize the existing pre-allocated buffer instead of allocating a new one in io.Copy().


rhatdan:
LGTM

mtrmac (commented):
/lgtm

PR #1966, opened 2024-06-12
chunked: honor store configuration

giuseppe:
honor the pull_options configuration set for the current store instead of using the default configuration.
    
Needed by Podman to override the pull options from the command line for a single command.
    

mtrmac (commented):
LGTM

@Honny1 PTAL

PR #1967, opened 2024-06-12
Decide on tar-split usage based on trusted data in TOC

mtrmac:
Don't ignore the tar-split when the TOC requires one, otherwise we could deduplicate a layer without tar-split with a layer with tar-split.

Fixes a bug discovered during review in https://github.com/containers/storage/pull/1936#pullrequestreview-2094139878 .

Cc: @giuseppe


rhatdan:
LGTM

giuseppe (approved):
/lgtm

PR #1968, opened 2024-06-12
pkg/archive.statDifferent(): fix the previous change

nalind:
Make the logic match the comment for real this time: instead of comparing the datestamps a second time, check the ModeType bits.  h/t to @mtrmac for spotting it at https://github.com/containers/storage/pull/1962/files#r1636517199.


mtrmac (commented):
LGTM. Thanks!

PR #1970, opened 2024-06-13
overlay: drop check for mount_program AND force_mask

giuseppe:
Using a mount_program is not a necessary requirement for users creating a shared store, as the store can be consumed by other users.

Stop enforcing this rule.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -375,9 +375,6 @@ func Init(home string, options graphdriver.Options) (graphdriver.Driver, error) 
 			return nil, err
 		}
 	} else {
-		if opts.forceMask != nil {
-			return nil, errors.New("'force_mask' is supported only with 'mount_program'")
-		}
 		// check if they are running over btrfs, aufs, overlay, or ecryptfs
 		switch fsMagic {
 		case graphdriver.FsMagicAufs, graphdriver.FsMagicOverlay, graphdriver.FsMagicEcryptfs:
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@saschagrunert @mtrmac @nalind PTAL

mtrmac:
Probably best if someone else reviews this…

---

> Using a mount_program is not a necessary requirement for users creating a shared store, as the store can be consumed by other users.

I don’t understand how “shared store” and “mount_program” interact.

It seems to me that when creating a layer applying a diff, we can set the permissions ourselves while untarring. Fine.

What happens with read-write layers? AFAICS in that case we really need the filesystem driver (in-kernel or FUSE) to enforce the mask … but I also see _nothing_ that passes the value to the mount operation.

So I don’t understand how any of this works even before this PR, and I’m just confused.

giuseppe:
> What happens with read-write layers? AFAICS in that case we really need the filesystem driver (in-kernel or FUSE) to enforce the mask … but I also see _nothing_ that passes the value to the mount operation.

yes a mount_program is still needed if you plan to use that same store, the issue though is that we require a mount_program even if the root user is not going to use that store to run containers, but only to share images to different users.  The store can be made available also on a network file system, so users are not necessarily on the same machine.  So we either require `fuse-overlayfs` to be installed, or provide a workaround with a dummy `mount_program`, since it doesn't matter when we pull an image.

mtrmac:
Would it make sense to refuse to create RW layers if `force_mask` is used without `mount_program`, then? (I don’t see how that actually works, I don’t know what I’m saying…)

mtrmac (commented):
LGTM?

I don’t see how RW layers work with `force_mask` _at all_, but this is is at least not inconsistent with how I imagine it might be working :)

(I plan to leave this for 2 days, to give knowledgeable reviewers an opportunity, and then I default to merging.)

mtrmac:
/lgtm

as I threatened.

PR #1971, opened 2024-06-13
chunked: honor the ForceMask setting

giuseppe:
make sure the ForceMask setting is honored for partial pulls.  More details in each commit.


rhatdan:
LGTM

pkg/chunked/storage_linux.go
```diff
@@ -1339,11 +1339,14 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
...
+	for i := range mergedEntries {
+		r := &mergedEntries[i]
 		if options.ForceMask != nil {
 			value := idtools.FormatOverrideXattr(r.UID, r.GID, int(r.Mode))
-			mergedEntries[i].Xattrs[idtools.ContainersOverrideXattr] = base64.StdEncoding.EncodeToString([]byte(value))
-			r.Mode = int64(*options.ForceMask)
+			if r.Xattrs == nil {
+				r.Xattrs = make(map[string]string)
+			}
+			r.Xattrs[idtools.ContainersOverrideXattr] = base64.StdEncoding.EncodeToString([]byte(value))
```
cgwalters, line 1349: It feels a bit unclean to mutate the input here; I'd trust you if you said it was safe, but still.

In ostree we have something very similar to the `ContainersOverrideXattr` in the "bare-user" mode vs "bare" and the code paths in the backend are pretty distinct. Doing something similar here would want instead for e.g. `setFileAttrs` to take an override parameter or so perhaps.

cgwalters (commented):
Looks sane to me, just one comment/question.

drivers/overlay/overlay.go
```diff
@@ -2220,11 +2223,12 @@ func (d *Driver) ApplyDiffWithDiffer(id, parent string, options *graphdriver.App
 		differOptions.UseFsVerity = graphdriver.DifferFsVerityIfAvailable
 	}
 	out, err := differ.ApplyDiff(applyDir, &archive.TarOptions{
-		UIDMaps:           idMappings.UIDs(),
+		ForceMask:         forceMask,
 		GIDMaps:           idMappings.GIDs(),
 		IgnoreChownErrors: d.options.ignoreChownErrors,
-		WhiteoutFormat:    d.getWhiteoutFormat(),
 		InUserNS:          unshare.IsRootless(),
+		UIDMaps:           idMappings.UIDs(),
```
mtrmac, line 2230: Absolutely non-blocking: `UIDMaps` and `GIDMaps` are now far from each other, I’m not sure why this change was made.

(Personally I’d aesthetically prefer the fields to be in the same order as in the structure declaration, to make it easier to see which were set / omitted.)

drivers/overlay/overlay.go
```diff
@@ -2162,11 +2162,14 @@ func supportsDataOnlyLayersCached(home, runhome string) (bool, error) {
...
 	var idMappings *idtools.IDMappings
+	forceMask := d.options.forceMask
+
 	if options != nil {
 		idMappings = options.Mappings
+		forceMask = options.ForceMask
 	}
 	if idMappings == nil {
-		idMappings = &idtools.IDMappings{}
+		idMappings = idtools.NewIDMappingsFromMaps(d.uidMaps, d.gidMaps)
```
mtrmac, line 2172: Warning: I don’t understand the whole picture at all.

Shouldn’t this happen at a (much?) higher level? Looking at `store.putLayer`, it is done there — and `canUseShifting` is involved in the decision.

I have no idea which of the two is correct, to be clear.

mtrmac (commented):
The `ForceMask` parts LGTM.

pkg/chunked/storage_linux.go
mtrmac: This should be safe because `mergedEntries` is an array of values (i.e. copies), created from scratch for this function. … ugh, except that a hash is a reference type, so AFAICS this _might_ be affecting the `c.toc` field.

Ultimately, that is _still_ safe because `c.toc` is written at most once, and read exactly once; but the number of assumptions _is_ getting a bit large.

drivers/overlay/overlay.go
giuseppe: I've sorted them alphabetically.  I'll revert this part and only add `ForceMask`

drivers/overlay/overlay.go
giuseppe: I never understood this part as well :-)  I took the chance now to dig deeper into it...

I think we are carrying the feature from Docker (e5f58bf9c25af5d7a49adbe1b597bcd1e3f33772) where it was possible to configure a single uid/gid mapping for all the containers, so the mappings are configured only once for the driver.

In facts this is the setting we get when we configure `remap-uids` and `remap-gids` in the storage.conf file.

I'll drop this patch from the current PR and remove the `remap-*ids` logic completely in a different PR, as it has been broken forever

giuseppe: opened a test PR to see how far the CI goes: https://github.com/containers/storage/pull/1976

pkg/chunked/storage_linux.go
cgwalters: In Rust:

- xattrs would likely be an `Option<T>` instead of a pointer, so the compiler would have made us check for its presence
- We likely would have been passing a `&TableOfContents` or so, and the compiler would have errored out on an attempt to mutate it
- And if we chose to pass instead `&mut TableOfContents` the compiler would verify that there weren't other threads (goroutines) that were concurrently reading it, etc.
- But we can often do something better than mutation for something like this; Rust has efficient iterators over collections, and in this case instead of mutation it'd likely be better to leave the input read-only and instead pass `impl Iterator<Item=Xattr>` to the function that sets xattrs (instead of an array). It's efficient to do e.g. `toc.xattrs.iter().chain(Xattr::new(idtools.CONTAINER_OVERRIDE_XATTR, value))` (although this would blend borrowed and owned values so we may need a `Cow` or so)

drivers/overlay/overlay.go
```diff
@@ -2162,8 +2162,11 @@ func supportsDataOnlyLayersCached(home, runhome string) (bool, error) {
 // ApplyDiffWithDiffer applies the changes in the new layer using the specified function
 func (d *Driver) ApplyDiffWithDiffer(id, parent string, options *graphdriver.ApplyDiffWithDifferOpts, differ graphdriver.Differ) (output graphdriver.DriverWithDifferOutput, errRet error) {
 	var idMappings *idtools.IDMappings
+	forceMask := d.options.forceMask
+
 	if options != nil {
 		idMappings = options.Mappings
+		forceMask = options.ForceMask
```
mtrmac, line 2169: Should this be conditioned on `options.ForceMask != nil`?

(One point is that the semantics is not documented. Another is that this field is never set when called by c/image, anyway. And yet again, if it is not used, why is it a field of `ApplyDiffWithDifferOpts` at all…)

---

AFAICS `ApplyDiffOpts.ForceMask` is never set when creating non-staged/chunked layers, so maybe we can just remove the filed and simplify?

pkg/chunked/storage_linux.go
mtrmac: I’d sign up to be a platinum member of the Go haters club, but that’s not really under discussion in this PR. (If you wanted to make a case for rewriting the stack in Rust, I’d not oppose that at all.)

---

I do agree that _this_ would be safer with a deep copy, but then … Go doesn’t have a language-native deep copy operation _either_. So it’s either risk of bugs due to overwriting data (if it ever turned from single use into multiple-use) or risk of bugs due to missing a field copy (if we ever added a field), and to me that feels like a wash, up to @giuseppe .

drivers/overlay/overlay.go
mtrmac: (Running on macOS, apparently I can’t trust the IDE:) `ApplyDiff` does
```go
	if d.options.forceMask != nil {
			options.ForceMask = d.options.forceMask
		}
```
the opposite of this logic… but, also, the field is not used on that path at all, either.

Removing `ApplyDiffOpts.ForceMask` looks increasingly attractive to me.

drivers/overlay/overlay.go
mtrmac: If I read the code correctly, the `d.[ug]idMaps` fields are never used when applying tarballs; and the generic code which triggers shifting or per-container remapping does not know about this option; that makes it really hard to see a use case.

Dealing with that in a separate PR LGTM.

drivers/overlay/overlay.go
giuseppe: thanks for spotting it.  I'd say the options should have a higher precedence, but I've adapted it to follow what `ApplyDiff` does.

Maybe something more to simplify along the lines of https://github.com/containers/storage/pull/1976

rhatdan:
LGTM

drivers/overlay/overlay.go
```diff
@@ -2162,9 +2162,15 @@ func supportsDataOnlyLayersCached(home, runhome string) (bool, error) {
 // ApplyDiffWithDiffer applies the changes in the new layer using the specified function
 func (d *Driver) ApplyDiffWithDiffer(id, parent string, options *graphdriver.ApplyDiffWithDifferOpts, differ graphdriver.Differ) (output graphdriver.DriverWithDifferOutput, errRet error) {
 	var idMappings *idtools.IDMappings
+	forceMask := options.ForceMask
+
 	if options != nil {
```
mtrmac, line 2167: Either the first line can crash, or the second one is unnecessary.

giuseppe: thanks, opened a PR: https://github.com/containers/storage/pull/1984

PR #1976, opened 2024-06-19
[TEST] [DO NOT MERGE] store: drop remap-[ug]ids

giuseppe:


cgwalters:
Right, there's no good use case for setting this globally now AFAIK.

While this obviously isn't exhaustive, at least a [Github code search](https://github.com/search?q=%22remap-uids+%3D%22+-is%3Afork&type=code) doesn't turn up any obvious uses of this in the wild.

(There are a lot of copies of the default...including some from people who are obviously committing the defaults that we write into their home directory into git...which wouldn't exist if we didn't write the defaults into a file at all)

rhatdan:
Would be good to fire off a podman test suite against this change.

giuseppe:
opened two tests PRs:

- https://github.com/containers/buildah/pull/5606
- https://github.com/containers/podman/pull/23070

giuseppe:
tests passed.  Ready for review

types/options.go
```diff
@@ -481,33 +481,6 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) erro
...
-		mappings, err := idtools.NewIDMappings(config.Storage.Options.RemapUser, config.Storage.Options.RemapGroup)
-		if err != nil {
-			logrus.Warningf("Error initializing ID mappings for %s:%s %v\n", config.Storage.Options.RemapUser, config.Storage.Options.RemapGroup, err)
-			return err
-		}
-		uidmap = mappings.UIDs()
-		gidmap = mappings.GIDs()
-	}
-	storeOptions.UIDMap = uidmap
-	storeOptions.GIDMap = gidmap
```
mtrmac, line 510: So the `StoreOptions.[UG]IDMap` fields still exist; they still affect `CreateContainer` and `putLayer`; and they, AFAICS, are potentially passed by Podman and potentially? user-configurable in `podman/pkg/domain/infra`. Yet the c/storage-level option has been removed.

What’s the full story here? Are the fields _also_ intended to be irrelevant / removed, and that just didn’t happen in order not to break Podman builds? Have the fields been used for two different purposes, so that we are now using only one? Are we changing what the fields mean now?

drivers/vfs/driver.go
```diff
@@ -152,14 +149,8 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, ro bool
 		return fmt.Errorf("--storage-opt is not supported for vfs")
 	}
 
-	idMappings := d.idMappings
-	if opts != nil && opts.IDMappings != nil {
-		idMappings = opts.IDMappings
```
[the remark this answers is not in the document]
mtrmac: “drop it from overlay as well”, I’m sorry.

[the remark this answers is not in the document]
giuseppe: you are right, this must not be deleted, opened a follow-up PR: https://github.com/containers/storage/pull/1988

types/options.go
giuseppe: The StoreOptions.[UGMap] are still honored.  I've changed them to not be read from the `storage.conf` file and to not affect the `Init()` call for each driver.   I don't think the internal directories used by each driver must be affected by the mappings.  I don't see a valid use case for that, unless we are planning to run without being root (either on the host, or in a user namespace).

```
# bin/podman run --rm -it --subuidname containers alpine sh -c 'grep . /proc/self/*id_map; ls -l /'
/proc/self/gid_map:         0     200000  268435456
/proc/self/uid_map:         0     200000  268435456
total 12
drwxr-xr-x    2 root     root          4096 Jun 18 14:16 bin
drwxr-xr-x    5 root     root           360 Jun 25 09:48 dev
drwxr-xr-x    1 root     root            25 Jun 25 09:48 etc
drwxr-xr-x    2 root     root             6 Jun 18 14:16 home
drwxr-xr-x    6 root     root          4096 Jun 18 14:16 lib
drwxr-xr-x    5 root     root            44 Jun 18 14:16 media
drwxr-xr-x    2 root     root             6 Jun 18 14:16 mnt
drwxr-xr-x    2 root     root             6 Jun 18 14:16 opt
dr-xr-xr-x  577 nobody   nobody           0 Jun 25 09:48 proc
drwx------    2 root     root             6 Jun 18 14:16 root
drwxr-xr-x    1 root     root            42 Jun 25 09:48 run
drwxr-xr-x    2 root     root          4096 Jun 18 14:16 sbin
drwxr-xr-x    2 root     root             6 Jun 18 14:16 srv
dr-xr-xr-x   13 nobody   nobody           0 Jun 23 19:41 sys
drwxrwxrwt    2 root     root             6 Jun 18 14:16 tmp
drwxr-xr-x    7 root     root            66 Jun 18 14:16 usr
drwxr-xr-x   12 root     root           137 Jun 18 14:16 var
```

mtrmac: ACK. Reading the documentation of the fields, they do seem to have been documented for two different purposes.

> I don't think the internal directories used by each driver must be affected by the mappings.

I agree.

PR #1978, opened 2024-06-20
Fix the issue where compiling containers-storage with exclude_disk_qu…

xuegege5290:
Fix the issue where compiling containers-storage with exclude_disk_quota enabled on the Linux CGO platform results in an error indicating that d.quotaCtl.GetDiskUsage is not defined.

```bash
root@syx:/storage# make binary
go build -compiler gc -tags " libdm_no_deferred_remove   exclude_disk_quota"  ./cmd/containers-storage
# github.com/containers/storage/drivers/overlay
drivers/overlay/overlay.go:1930:35: undefined: quota.BackingFsBlockDeviceLink
drivers/overlay/overlay_cgo.go:18:21: d.quotaCtl.GetDiskUsage undefined (type *quota.Control has no field or me
thod GetDiskUsage)
make: *** [Makefile:41: containers-storage] Error 1
```

Certainly, the reason I compiled it this way was an attempt to disable the xfs_quota limit feature during the compilation of CRI-O, which led to this issue. When I tried to use the exclude_disk_quota build argument, I should have chosen to compile the overlay_nocgo.go file.
```bash
root@syx:/storage# git status
HEAD detached at origin/release-1.51
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   Makefile
        modified:   drivers/overlay/overlay_cgo.go
        modified:   drivers/overlay/overlay_nocgo.go
        modified:   drivers/quota/projectquota.go
        modified:   drivers/quota/projectquota_unsupported.go

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        drivers/quota/projectquota_supported.go

no changes added to commit (use "git add" and/or "git commit -a")
root@syx:/storage# make binary
go build -compiler gc -tags " libdm_no_deferred_remove   exclude_disk_quota"  ./cmd/containers-storage
root@syx:/storage# 
```


mtrmac (changes requested):
Thanks!

I don’t think the quota build tag should affect whether `getComposeFsHelper` and others are included; this should probably be split into more files or something.

xuegege5290:
> Thanks!
> 
> I don’t think the quota build tag should affect whether `getComposeFsHelper` and others are included; this should probably be split into more files or something.

I don't understand what you mean by getComposeFsHelper. I only know that when I use the XFS file system, the overlay driver, and enable the quota, I need to use exclude_disk_quota to disable the quota feature, and then overlay_cgo.go will be compiled, and d.quotaCtl.GetDiskUsage needs to be referenced. You said 'this should probably be split into more files or something,' what exactly needs to be done? I am very willing to listen.

rhatdan:
You need to create two new files for disk quota. and not just tread no_disk_quota as no nocgo

xuegege5290:
> You need to create two new files for disk quota. and not just tread no_disk_quota as no nocgo

I've got it., thanks. I will give it a try.

Issue #1579, opened 2023-04-19
Consider allowing more linters


mtrmac:
<s>https://github.com/containers/storage/pull/1522</s>  has disabled 5 out of the 7 default golangci-lint linters. That seems excessive.

For the record, currently:
```console
% golangci-lint version              
golangci-lint has version 1.52.2 built with go1.20.2 from da04413a on 2023-03-25T18:11:28Z
% golangci-lint run --no-config ./...
pkg/chunked/compressor/compressor.go:331:21: SA1019: hdr.Xattrs is deprecated: Use PAXRecords instead. (staticcheck)
		for k, v := range hdr.Xattrs {
		                  ^
check.go:289:14: Error return value of `io.Copy` is not checked (errcheck)
						io.Copy(io.Discard, diffReader)
						       ^
check.go:883:54: SA1019: tar.TypeRegA has been deprecated since Go 1.11 and an alternative has been available since Go 1.1: Use TypeReg instead. (staticcheck)
		if hdr.Typeflag == tar.TypeLink || hdr.Typeflag == tar.TypeRegA {
		                                                   ^
check.go:902:45: SA1019: tar.TypeRegA has been deprecated since Go 1.11 and an alternative has been available since Go 1.1: Use TypeReg instead. (staticcheck)
	if typeflag == tar.TypeLink || typeflag == tar.TypeRegA {
	                                           ^
lockfile_compat.go:8:15: SA1019: lockfile.Locker is deprecated: Refer directly to *LockFile, the provided implementation, instead. (staticcheck)
type Locker = lockfile.Locker //lint:ignore SA1019 // lockfile.Locker is deprecated
              ^
pkg/regexp/regexp.go:26:9: copylocks: return copies lock value: github.com/containers/storage/pkg/regexp.Regexp contains sync.Once contains sync.Mutex (govet)
	return re
	       ^
pkg/ioutils/fswriters.go:153:18: Error return value of `w.closeTempFile` is not checked (errcheck)
		w.closeTempFile()
		               ^
pkg/ioutils/readers_test.go:82:7: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (govet)
	ctx, _ := context.WithTimeout(context.Background(), 100*time.Millisecond)
	     ^
pkg/archive/archive.go:402:5: SA1019: hdr.Xattrs has been deprecated since Go 1.10: Use PAXRecords instead. (staticcheck)
	if hdr.Xattrs == nil {
	   ^
[...]

PR #1980, opened 2024-06-20
Fix deprecated use of `tar.TypeRegA` (SA1019)

Honny1:
This PR fixes warning deprecated use of `tar.TypeRegA` (SA1019) found by `golangci` when the `staticcheck` linter is enabled. 

Partially fixes:
- #1579

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -668,7 +668,7 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 			}
 		}
 
-	case tar.TypeReg, tar.TypeRegA:
+	case tar.TypeReg:
 		// Source is regular file. We use system.OpenFileSequential to use sequential
 		// file access to avoid depleting the standby list on Windows.
 		// On Linux, this equates to a regular os.OpenFile
--- a/pkg/chunked/internal/compression.go
+++ b/pkg/chunked/internal/compression.go
@@ -89,7 +89,6 @@ const (
 
 var TarTypes = map[byte]string{
 	tar.TypeReg:     TypeReg,
-	tar.TypeRegA:    TypeReg,
 	tar.TypeLink:    TypeLink,
 	tar.TypeChar:    TypeChar,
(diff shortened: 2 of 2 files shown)


mtrmac (commented):
It’s not _immediately obvious_ to me that we can just stop processing these values.

Is there a reason why this is safe? E.g. maybe because the inputs are being created by the same process and we can see the code that generated it? Or can these values never show up nowadays?

If this is processing externally-processed tarballs built by other implementations, or 7-year-old images, and those _could actually generate those values,_ we should not be breaking that.

---

(I’m not saying that this is the wrong approach; from a _very quick_ skim, it might be the right one — but I assume you have checked, so it would be nice to have a written record of the situation.)

mtrmac:
For the record, in both affected places, the headers come from `tar.Reader.Next`, and that function never produces `TypeRegA`.

/lgtm

Honny1:
I have the same thoughts. I checked the `archive/tar` library implementation for Go Lang version 1.21+.  If the old version of the tarball header in the `Typeflag` contains the value `TypeRegA`, it is changed to `TypeReg` or `TypeDir`. [Code reference](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/archive/tar/reader.go;l=145). I think removing `TypeRegA` from the switch case and const shouldn't break other things. But I just did a static check. Any suggestions on how to check this more deeply and not only tests?

mtrmac:
In principle, generate a tar file like that, and see that it processed correctly.

https://github.com/golang/go/commit/e4bde0510465eecd4c8a8293418b1cbed1e0e623 did add a test for the read code; we don’t have a test that we are using that source of the data.

On balance, I think adding tests for this situation is overkill. It wouldn’t _hurt_ but there are whole subpackages in this project with no unit tests at all, that would be a better use of time.

(Alternatively, we could just silence the warning, but when the raw `0` byte can mean either a file and a directory, that would _still_ be making an assumption that `archive/tar.Reader` is being used to convert the value — so we might just as well remove the cases and follow the documented API.)

PR #1984, opened 2024-06-21
overlay: check for options != nil

giuseppe:
check that options != nil before deferencing it.


cgwalters:
Needs gofmt
/approve

rhatdan:
LGTM

PR #1985, opened 2024-06-24
Fix deprecated use of `hdr.Xattrs` (SA1019)

Honny1:
This PR fixes warning deprecated use of `hdr.Xattrs` (SA1019) found by `golangci` when the `staticcheck` linter is enabled. 

Partially fixes:
- #1579

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -70,6 +70,8 @@ type (
 	}
 )
 
+const PaxSchilyXattr = "SCHILY.xattr."
+
 const (
 	tarExt  = "tar"
 	solaris = "solaris"
@@ -417,11 +419,11 @@ func FileInfoHeader(name string, fi os.FileInfo, link string) (*tar.Header, erro
 	return hdr, nil
 }
 
-// ReadSecurityXattrToTarHeader reads security.capability, security,image
+// readSecurityXattrToTarHeader reads security.capability, security,image
 // xattrs from filesystem to a tar header
-func ReadSecurityXattrToTarHeader(path string, hdr *tar.Header) error {
-	if hdr.Xattrs == nil {
-		hdr.Xattrs = make(map[string]string)
+func readSecurityXattrToTarHeader(path string, hdr *tar.Header) error {
+	if hdr.PAXRecords == nil {
+		hdr.PAXRecords = make(map[string]string)
 	}
 	for _, xattr := range []string{"security.capability", "security.ima"} {
 		capability, err := system.Lgetxattr(path, xattr)
@@ -429,14 +431,14 @@ func ReadSecurityXattrToTarHeader(path string, hdr *tar.Header) error {
 			return fmt.Errorf("failed to read %q attribute from %q: %w", xattr, path, err)
 		}
 		if capability != nil {
-			hdr.Xattrs[xattr] = string(capability)
+			hdr.PAXRecords[PaxSchilyXattr+xattr] = string(capability)
 		}
 	}
 	return nil
 }
 
-// ReadUserXattrToTarHeader reads user.* xattr from filesystem to a tar header
-func ReadUserXattrToTarHeader(path string, hdr *tar.Header) error {
+// readUserXattrToTarHeader reads user.* xattr from filesystem to a tar header
+func readUserXattrToTarHeader(path string, hdr *tar.Header) error {
 	xattrs, err := system.Llistxattr(path)
 	if err != nil && !errors.Is(err, system.EOPNOTSUPP) && err != system.ErrNotSupportedPlatform {
 		return err
@@ -451,10 +453,10 @@ func ReadUserXattrToTarHeader(path string, hdr *tar.Header) error {
 				}
 				return err
 			}
-			if hdr.Xattrs == nil {
-				hdr.Xattrs = make(map[string]string)
+			if hdr.PAXRecords == nil {
+				hdr.PAXRecords = make(map[string]string)
 			}
-			hdr.Xattrs[key] = string(value)
+			hdr.PAXRecords[PaxSchilyXattr+key] = string(value)
 		}
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


rhatdan:
/approve
LGTM
@giuseppe @nalind @mtrmac @saschagrunert PTAL

pkg/archive/archive.go
```diff
@@ -420,16 +422,16 @@ func FileInfoHeader(name string, fi os.FileInfo, link string) (*tar.Header, erro
 // ReadSecurityXattrToTarHeader reads security.capability, security,image
 // xattrs from filesystem to a tar header
 func ReadSecurityXattrToTarHeader(path string, hdr *tar.Header) error {
```
mtrmac, line 424: This, and `ReadUserXattrToTarHeader`, now change how they output data. That _should_ be fine locally, but it might trip up callers that make assumptions (like `overlayWhiteoutConverter` does).

I’d prefer those two functions to be made private when we are changing them; I don’t expect there to be external users.

mtrmac (commented):
Thanks!

This LGTM locally, but this _is_ fairly critical low-level code. Could you check that the packages do have some tests that cover the changed code, please? (One possible approach would be, for each instance, to just delete the Xattr handling code, and ensure that at least one c/storage test fails. Ideally the tests that fail should go all the way between a local file and a tar stream, not just a local unit test on `ReadSecurityXattrToTarHeader` or so.)

Also Cc: @nalind for expert insights.

rhatdan:
@mtrmac PTANL

mtrmac:
I’m fine with the code, but I’d like to hear an assurance that this is covered by tests. (Yes, I could go verify that myself…)

mtrmac:
I’d _guess_ that if c/storage does not test these somewhat obscure aspects of layer extraction itself, Podman is unlikely to; but triggering a Podman CI run with a modified c/storage is probably much cheaper than manual analysis, so, sure, go for it.

pkg/archive/archive.go
```diff
@@ -782,11 +784,15 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
...
+	for key, value := range hdr.PAXRecords {
+		xattr_key, ok := strings.CutPrefix(key, PaxSchilyXattr)
+		if !ok {
+			continue
+		}
+		if _, found := xattrsToIgnore[xattr_key]; found {
 			continue
 		}
-		if err := system.Lsetxattr(path, key, []byte(value), 0); err != nil {
+		if err := system.Lsetxattr(path, xattr_key, []byte(value), 0); err != nil {
```
giuseppe, line 795: please use `xattrKey` if you really need to change it

Honny1:
@mtrmac I ran podman CI with imported c/storage with these changes. CI passes see the [PR](https://github.com/containers/podman/pull/23104) that runs the tests.

Honny1:
> I checked the unit tests by commenting out the parts of the code that were changed if the tests fail. The tests did not fail for the `readUserXattrToTarHeader`, `(o overlayWhiteoutConverter) ConvertWrite`, `writeZstdChunkedStream` functions. I haven't tried code changes that are specific to the `darwin` os. Should I also try to run unit tests of padman with the imported local storage?

I added an xattr to `TestTarUntarWithXattr` which is a user xattr. So the test fails for the commented change in `readUserXattrToTarHeader`.

mtrmac (commented):
If I understand it correctly, with this PR we have xattr test coverage for `archive.go`. Great!

---

I can’t see any c/storage test at all exercising `compressor.ZstdCompressor`; I can’t see any for the whiteout converter either, but there it’s quite possible I have missed one.

\*shrug\* if that’s the case, I think it can be reasonably argued it’s not the responsibility of this PR to add missing tests to entire large features.

pkg/archive/archive_unix_test.go
```diff
@@ -248,9 +250,13 @@ func TestTarUntarWithXattr(t *testing.T) {
...
-			t.Fatalf("Untar should have kept the 'security.capability' xattr.")
-		}
+
+		capability, err := system.Lgetxattr(filepath.Join(origin, "2"), "security.capability")
+		require.NoError(t, err)
+		assert.EqualValues(t, capability, encoded)
+
+		test, err := system.Lgetxattr(filepath.Join(origin, "1"), "user.test")
+		require.NoError(t, err)
+		assert.EqualValues(t, test, []byte("helloWord"))
```
mtrmac, line 260: (For the record, the parameter order in `assert` is (expected, actual). That only matters if the test fails, and it can be trivially fixed afterwards — no need to update the PR now.)

mtrmac (commented):
/lgtm

Thanks!

mtrmac:
@Honny1 the tests are failing

pkg/archive/archive_unix_test.go
```diff
@@ -248,9 +250,16 @@ func TestTarUntarWithXattr(t *testing.T) {
 		if len(changes) != 1 || changes[0].Path != "/3" {
 			t.Fatalf("Unexpected differences after tarUntar: %v", changes)
 		}
-		capability, _ := system.Lgetxattr(filepath.Join(origin, "2"), "security.capability")
-		if capability == nil && capability[0] != 0x00 {
-			t.Fatalf("Untar should have kept the 'security.capability' xattr.")
+
+		capability, err := system.Lgetxattr(filepath.Join(origin, "2"), "security.capability")
+		require.NoError(t, err)
+		assert.Equal(t, len(encoded), len(capability))
```
mtrmac, line 256: `assert.Equal(t, encoded[:], capability)` maybe?

Or, to avoid duplication, `encoded := []byte…`.

(I didn’t test either, I might be missing something.)

Honny1: I didn't know I could do it that way. Thanks

Honny1:
@mtrmac The tests are passing.

PR #1986, opened 2024-06-24
Fix the surrounding loop is unconditionally terminated (SA4004)

Honny1:
This PR fixes `the surrounding loop is unconditionally terminated` warning (SA4004) found by `golangci` when the `staticcheck` linter is enabled. 

Partially fixes:
- #1579

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -2532,6 +2532,19 @@ func nameWithSuffix(name string, number int) string {
 	return fmt.Sprintf("%s%d", name, number)
 }
 
+func validateOneAdditionalLayerPath(target string) error {
+	for _, p := range []string{
+		filepath.Join(target, "diff"),
+		filepath.Join(target, "info"),
+		filepath.Join(target, "blob"),
+	} {
+		if err := fileutils.Exists(p); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 func (d *Driver) getAdditionalLayerPath(tocDigest digest.Digest, ref string) (string, error) {
 	refElem := base64.StdEncoding.EncodeToString([]byte(ref))
 	for _, ls := range d.options.layerStores {
@@ -2540,18 +2553,11 @@ func (d *Driver) getAdditionalLayerPath(tocDigest digest.Digest, ref string) (st
 			ref = refElem
 		}
 		target := path.Join(ls.path, ref, tocDigest.String())
-		// Check if all necessary files exist
-		for _, p := range []string{
-			filepath.Join(target, "diff"),
-			filepath.Join(target, "info"),
-			filepath.Join(target, "blob"),
-		} {
-			if err := fileutils.Exists(p); err != nil {
-				wrapped := fmt.Errorf("failed to stat additional layer %q: %w", p, err)
-				return "", fmt.Errorf("%v: %w", wrapped, graphdriver.ErrLayerUnknown)
-			}
+		err := validateOneAdditionalLayerPath(target)
+		if err == nil {
+			return target, nil
 		}
-		return target, nil
+		logrus.Debugf("additional Layer Store %v failed to stat additional layer: %v", ls, err)
 	}
 
 	return "", fmt.Errorf("additional layer (%q, %q) not found: %w", tocDigest, ref, graphdriver.ErrLayerUnknown)


drivers/overlay/overlay.go
```diff
@@ -2554,18 +2568,11 @@ func (d *Driver) getAdditionalLayerPath(tocDigest digest.Digest, ref string) (st
...
-		} {
-			if err := fileutils.Exists(p); err != nil {
-				wrapped := fmt.Errorf("failed to stat additional layer %q: %w", p, err)
-				return "", fmt.Errorf("%v: %w", wrapped, graphdriver.ErrLayerUnknown)
-			}
+		err := checkIfAllNesseseryFilesExistsInLayerStore(target)
+		if err == nil {
+			return target, nil
 		}
-		return target, nil
```
mtrmac, line 2568: @giuseppe @ktock “Additional layer stores” support multiple stores, but apparently the code was only able to work with the first one, since the very start of the feature ( https://github.com/containers/storage/commit/64f018103e1dba3de80e93a820b4ca50d9926ef6 ).

What do you think about simplifying to only support one store? Admittedly that would not help simplify the codebase all that much, ALS is AFAICS _only_ a feature within this driver and ~this one loop — and instead, we would need to add even more code to reject multiple stores when parsing options.

Either way, we seem not to have tests sufficiently covering the feature.

drivers/overlay/overlay.go
```diff
@@ -2546,6 +2546,20 @@ func nameWithSuffix(name string, number int) string {
 	return fmt.Sprintf("%s%d", name, number)
 }
 
+func checkIfAllNesseseryFilesExistsInLayerStore(target string) error {
```
mtrmac, line 2549: - Typo in “necessary”
- The graph driver is all about layers; to be specific for this feature, “in additional layer store”.
- (absolutely non-blocking: consider naming this in a way that shows the relationship to `getAdditionalLayerPath`; maybe `validateOneAdditionalLayerPath` or something like that.)

drivers/overlay/overlay.go
```diff
@@ -2554,18 +2568,11 @@ func (d *Driver) getAdditionalLayerPath(tocDigest digest.Digest, ref string) (st
...
-			if err := fileutils.Exists(p); err != nil {
-				wrapped := fmt.Errorf("failed to stat additional layer %q: %w", p, err)
-				return "", fmt.Errorf("%v: %w", wrapped, graphdriver.ErrLayerUnknown)
-			}
+		err := checkIfAllNesseseryFilesExistsInLayerStore(target)
+		if err == nil {
+			return target, nil
 		}
-		return target, nil
+		logrus.Warn(err)
```
mtrmac, line 2575: _If_ there are going to be multiple ALS stores, it’s nothing unexpected if a layer does not exist in the first one, but exits in the second one. So this should not be a warning — at most a debug log.

drivers/overlay/overlay.go
```diff
@@ -2546,6 +2546,20 @@ func nameWithSuffix(name string, number int) string {
...
 
+func checkIfAllNesseseryFilesExistsInLayerStore(target string) error {
+	for _, p := range []string{
+		filepath.Join(target, "diff"),
+		filepath.Join(target, "info"),
+		filepath.Join(target, "blob"),
+	} {
+		if err := fileutils.Exists(p); err != nil {
+			wrapped := fmt.Errorf("failed to stat additional layer %q: %w", p, err)
+			return fmt.Errorf("%v: %w", wrapped, graphdriver.ErrLayerUnknown)
```
mtrmac, line 2557: _If_ there will be a loop and we don’t decide to just support one value:

This error now exists only to be (not?) logged in the loop; the ultimate `ErrLayerUnknown` error returned to callers of `getAdditionalLayerPath` is created elsewhere.

So I think we can avoid the `ErrLayerUnknown` use, and just format an error string that makes sense in the logs.

drivers/overlay/overlay.go
```diff
@@ -2532,6 +2532,20 @@ func nameWithSuffix(name string, number int) string {
...
 
+func validateOneAdditionalLayerPath(target string) error {
+	for _, p := range []string{
+		filepath.Join(target, "diff"),
+		filepath.Join(target, "info"),
+		filepath.Join(target, "blob"),
+	} {
+		if err := fileutils.Exists(p); err != nil {
+			wrapped := fmt.Errorf("failed to stat additional layer %q: %w", p, err)
+			return fmt.Errorf("Additional Layer Store %v: %w", target, wrapped)
```
mtrmac, line 2543: AFAICS the log entry will now be something like `Additional Layer Store /var/foo/docker.io/library/sha256:…: failed to stat additional layer /var/foo/docker.io/library/sha256:…/diff: faccessat /var/foo/docker.io/library/sha256:…/diff: no such file or directory`

That still seems excessive. Also the `wrapped` step is unnecessary, use one `fmt.Errorf` to create the whole message.

mtrmac (commented):
LGTM, just a non-blocking stylistic nit.

Please squash follow-on fixes instead of leaving the original versions of commits in the PR in the history.

rhatdan:
/approve
/lgtm
Thanks @Honny1

PR #1987, opened 2024-06-24
Add comented out example definition of driver_priority to storage.conf

dcermak:


storage.conf
```diff
@@ -19,6 +19,10 @@ driver = "overlay"
 # Temporary storage location
 runroot = "/run/containers/storage"
 
+# Priority list for the storage drivers that will be tested one
```
rhatdan, line 22: I don't believe this is a hard coded list and true for all systems.

rhatdan: Usually the examples display the defaults, but I am fine with this.

LGTM

giuseppe (approved):
LGTM

rhatdan:
/lgtm
THanks @dcermak

PR #1995, opened 2024-07-02
Move `junkfile` to `${TESTDIR}` directory

Honny1:
This PR moves `junkfile` to the `${TESTDIR}` directory. The `junkfile` is generated during `bats` tests in the `tests` directory. Moving the `junkfile` to the `${TESTDIR}` directory avoids unintended writing this file to the git history.


giuseppe (approved):
LGTM

PR #1996, opened 2024-07-02
chunked: skip setting time if empty

giuseppe:
do not specify the timestamps fields in the TOC if they are empty.

### DIFF
--- a/pkg/chunked/compressor/compressor.go
+++ b/pkg/chunked/compressor/compressor.go
@@ -10,6 +10,7 @@ import (
 	"encoding/base64"
 	"io"
 	"strings"
+	"time"
 
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/chunked/internal"
@@ -233,6 +234,14 @@ func newTarSplitData(level int) (*tarSplitData, error) {
 	}, nil
 }
 
+// timeIfNotZero returns a pointer to the time.Time if it is not zero, otherwise it returns nil.
+func timeIfNotZero(t *time.Time) *time.Time {
+	if t == nil || t.IsZero() {
+		return nil
+	}
+	return t
+}
+
 func writeZstdChunkedStream(destFile io.Writer, outMetadata map[string]string, reader io.Reader, level int) error {
 	// total written so far.  Used to retrieve partial offsets in the file
 	dest := ioutils.NewWriteCounter(destFile)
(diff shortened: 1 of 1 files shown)


cgwalters:
/approve
Looks sane to me, but a single sentence for why would be useful to me. What motivated the change? I can guess: is this an optimization to reduce the size of the ToC JSON? Or something else?

giuseppe:
> Looks sane to me, but a single sentence for why would be useful to me. What motivated the change? I can guess: is this an optimization to reduce the size of the ToC JSON? Or something else?

it improves readability if anyone inspects the TOC file itself and it slightly reduces the TOC size, e.g. with Fedora I get:

- compressed TOC before: 86643308
- compressed TOC after:    86642710

so it is just 598 bytes after compression, not really worth mentioning as an improvement :)

cgwalters:
Hum, the Go docs say:

> The zero value of type Time is January 1, year 1, 00:00:00.000000000 UTC. As this time is unlikely to come up in practice

Which is actually different than what I thought (I thought it'd be the unix epoch).

So...apparently this actually did come up in practice, but...why? 

> with Fedora

Which Fedora? quay.io/fedora/fedora:40 ?

giuseppe:
> > The zero value of type Time is January 1, year 1, 00:00:00.000000000 UTC. As this time is unlikely to come up in practice
> 
> Which is actually different than what I thought (I thought it'd be the unix epoch).
> 
> So...apparently this actually did come up in practice, but...why?

It could be that the tar stream lacks this value and it automatically gets assigned the "zero" value for time.Time

> Which Fedora? quay.io/fedora/fedora:40 ?

I've built a local image with:

```Dockerfile
FROM registry.fedoraproject.org/fedora:latest
RUN touch /timestamp
```

and squashed it with `podman build --squash-all`

cgwalters:
Sorry I know this is a trivial PR and that makes it more likely to be bike-shedded but...this just keeps generating more questions for me.

Looking at e.g. the [PAX format](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html), mtime is a mandatory field.

Presumably it's actually the access time or ctime which are zero?

Actually I start to wonder - why the heck did the people creating eStargz add explicit support for atime and ctime? Even mtime is questionable...but, it's hard to escape the concept of files having a time (e.g. Python famously depends on it). But nothing should care if we always made atime == ctime == mtime... (xref https://github.com/containers/composefs/pull/163 too here, ctime is simply not representable in composefs).

Anyways,
/lgtm

cgwalters:
(As far as atime...well, I agree with a lot of the sentiment behind https://lwn.net/Articles/244829/ ...it should be an opt-in thing for `/tmp` at most. At least on an image-based system like bootc with `/` being a read-only mount by default, simply *running binaries* doesn't result in writes to the underlying filesystem due to atime...but, that is what happens on a traditional apt/dnf system...)

cgwalters:
And for containers, we're kind of [saved by overlayfs](https://docs.kernel.org/filesystems/overlayfs.html):

> POSIX mandates updating st_atime for reads. This is currently not done in the case when the file resides on a lower layer.

cgwalters:
Right, clicking on the composefs link, we already covered this topic earlier in https://github.com/containers/storage/pull/1646#issuecomment-1597646576 

So omitting the times here makes sense.

PR #1997, opened 2024-07-03
check: mount layer as read-only

giuseppe:
mount the image as read-only so it can work also for composefs images.


nalind:
LGTM

mtrmac (commented):
/lgtm

PR #2001, opened 2024-07-05
chunked: store compressed digest if validated

giuseppe:
if the compressed digest was validated, as it happens when 'pull_options = {convert_images = "true"}' is set, then store it as well so that reusing the blob by its compressed digest works.

Previously, when an image converted to zstd:chunked was pulled a second time, it would not be recognized by its compressed digest, resulting in the need to re-pull the image again.

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -2529,7 +2529,9 @@ func (r *layerStore) applyDiffFromStagingDirectory(id string, diffOutput *driver
 	layer.GIDs = diffOutput.GIDs
 	updateDigestMap(&r.byuncompressedsum, layer.UncompressedDigest, diffOutput.UncompressedDigest, layer.ID)
 	layer.UncompressedDigest = diffOutput.UncompressedDigest
-	updateDigestMap(&r.bytocsum, diffOutput.TOCDigest, diffOutput.TOCDigest, layer.ID)
+	updateDigestMap(&r.bycompressedsum, layer.CompressedDigest, diffOutput.CompressedDigest, layer.ID)
+	layer.CompressedDigest = diffOutput.CompressedDigest
+	updateDigestMap(&r.bytocsum, layer.TOCDigest, diffOutput.TOCDigest, layer.ID)
 	layer.TOCDigest = diffOutput.TOCDigest
 	layer.UncompressedSize = diffOutput.Size
 	layer.Metadata = diffOutput.Metadata
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -193,6 +193,7 @@ type DriverWithDifferOutput struct {
 	UIDs               []uint32
 	GIDs               []uint32
 	UncompressedDigest digest.Digest
+	CompressedDigest   digest.Digest
 	Metadata           string
 	BigData            map[string][]byte
 	TarSplit           []byte
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -1126,6 +1126,7 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 	// stream to use for reading the zstd:chunked or Estargz file.
 	stream := c.stream
 
+	var compressedDigest digest.Digest
 	var uncompressedDigest digest.Digest
 	var convertedBlobSize int64
 
@@ -1142,7 +1143,7 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 		}()
 
 		// calculate the checksum before accessing the file.
-		compressedDigest, err := c.copyAllBlobToFile(blobFile)
+		compressedDigest, err = c.copyAllBlobToFile(blobFile)
 		if err != nil {
 			return graphdriver.DriverWithDifferOutput{}, err
 		}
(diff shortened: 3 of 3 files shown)


mtrmac (commented):
I think this works for simple cases, but it is not optimal in general: A single `Layer` can have multiple compressed representations, and this only stores one.

For ordinary uncompressed layers, that is handled by `putBlobToPendingFile` (computing both digests and) calling `RecordDigestUncompressedPair`; that can fully record a M-to-1 mapping. And that mapping is then consulted in `tryReusingBlobAsPending`.

So, it seems to me that `PutBlobPartial` should call `RecordDigestUncompressedPair` if the uncompressed digest is available. It’s already making an assumption that the compressed digest has been validated … though having `pkg/chunked` explicitly return the value in `DriverWithDifferOutput`, as this PR does, and explicitly validating that `CompressedDigest == blobDigest` in `TryReusingBlobAsPending`, would be more obviously correct.

layers.go
```diff
@@ -2529,6 +2529,7 @@ func (r *layerStore) applyDiffFromStagingDirectory(id string, diffOutput *driver
 	layer.GIDs = diffOutput.GIDs
 	updateDigestMap(&r.byuncompressedsum, layer.UncompressedDigest, diffOutput.UncompressedDigest, layer.ID)
 	layer.UncompressedDigest = diffOutput.UncompressedDigest
+	layer.CompressedDigest = diffOutput.CompressedDigest
```
mtrmac, line 2532: - This might overwrite a valid `CompressedDigest` value (hypothetically, from a template layer) with `""`
- If this field is modified, a corresponding `updateDigestMap` call needs to happen as well.

I think the other parts of this PR are clearly valuable; this one doesn’t really hurt but seems unnecessary for the purpose. OTOH handling the `DriverWithDifferOutput` fields consistently _is_ somewhat valuable.

giuseppe:
> I think this works for simple cases, but it is not optimal in general: A single `Layer` can have multiple compressed representations, and this only stores one.

The Layer object already knows about the `CompressedDigest`, so I think we can use for the case where we convert an image and validate the compressed digest.

when we pull a layer without any conversion happening, we store the following information in the `overlay-layers/layers.json` file:

```json
[{"id":"94e5f06ff8e3d4441dc3cd8b090ff38dc911bfa8ebdb0dc28395bc98f82f983f","created":"2024-07-08T09:28:18.17830826Z","compressed-diff-digest":"sha256:ec99f8b99825a742d50fb3ce173d291378a46ab54b8ef7dd75e5654e2a296e99","compressed-size":3623844,"diff-digest":"sha256:94e5f06ff8e3d4441dc3cd8b090ff38dc911bfa8ebdb0dc28395bc98f82f983f","diff-size":8082944,"compression":2,"uidset":[0],"gidset":[0,42]}]
```

previously, without the current change we weren't storing the `compressed-diff-digest`:

```json
[{"id":"d51af96cf93e225825efd484ea457f867cb2b967f7415b9a3b7e65a2f803838a","created":"2024-07-08T09:49:07.867152639Z","diff-digest":"sha256:d51af96cf93e225825efd484ea457f867cb2b967f7415b9a3b7e65a2f803838a","diff-size":4495360,"uidset":[65534,1,8,0],"gidset":[8,0,65534,1],"big-data-names":["zstd-chunked-manifest","zstd-chunked-layer-data"]}]
```

layers.go
giuseppe: thanks, I'll fix it and call `updateDigestMap`

mtrmac:
> > I think this works for simple cases, but it is not optimal in general: A single `Layer` can have multiple compressed representations, and this only stores one.
> 
> The Layer object already knows about the `CompressedDigest`

It knows _one_ per deduplicated layer. That loses information if the same layer is compressed multiple ways. (And, also, the information is lost when the layer is deleted.)

The way this is expected to work is that is _primarily_ recorded in the c/image BlobInfoCache.

mtrmac:
@giuseppe https://github.com/containers/image/pull/2478 is what I think code based on this PR can look like; that should work correctly even without the `layerStoreapplyDiffFromStagingDirectory` part of this PR.

Alternatively, WIP https://github.com/containers/image/pull/2321 among various other changes, also updates BIC, without needing a new c/storage API (but hard-coding the same assumption about what c/storage validates).

giuseppe:
> It knows _one_ per deduplicated layer. That loses information if the same layer is compressed multiple ways. (And, also, the information is lost when the layer is deleted.)
> 
> The way this is expected to work is that is _primarily_ recorded in the c/image BlobInfoCache.

I am not opposing the change in c/image.  I think we need that as well to deal with multiple compressed digests for the same uncompressed diff.  I agree your change in https://github.com/containers/image/pull/2478 is correct and we should have it.  I still think though we need also the current change in c/storage so it can be used by APIs like `LayersByCompressedDigest`

mtrmac:
`LayersByCompressedDigest` is fundamentally lossy … but, sure, as long as we have the API, it’s valuable to try to keep it as good as possible.

mtrmac (commented):
I’m sorry, I screwed something up and this remained “pending” .

LGTM otherwise.

layers.go
```diff
@@ -2529,6 +2529,8 @@ func (r *layerStore) applyDiffFromStagingDirectory(id string, diffOutput *driver
 	layer.GIDs = diffOutput.GIDs
 	updateDigestMap(&r.byuncompressedsum, layer.UncompressedDigest, diffOutput.UncompressedDigest, layer.ID)
 	layer.UncompressedDigest = diffOutput.UncompressedDigest
+	updateDigestMap(&r.bycompressedsum, diffOutput.CompressedDigest, diffOutput.CompressedDigest, layer.ID)
```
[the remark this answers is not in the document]
giuseppe: thanks for catching this one.  I've copied the same mistake from `updateDigestMap(TOCDigest)`.  Fixed the other occurrence too

mtrmac:
LGTM, for the record.

PR #2002, opened 2024-07-08
Solve `errcheck` warnings (part 3)

Honny1:
This PR is one of several fixes of warnings found by `golangci` when the `errcheck` linter is enabled. 

Partially fixes:
- #1579

### DIFF
--- a/drivers/aufs/aufs.go
+++ b/drivers/aufs/aufs.go
@@ -349,7 +349,9 @@ func (a *Driver) createDirsFor(id, parent string) error {
 // Remove will unmount and remove the given id.
 func (a *Driver) Remove(id string) error {
 	a.locker.Lock(id)
-	defer a.locker.Unlock(id)
+	defer func() {
+		_ = a.locker.Unlock(id)
+	}()
 	a.pathCacheLock.Lock()
 	mountpoint, exists := a.pathCache[id]
 	a.pathCacheLock.Unlock()
@@ -440,7 +442,10 @@ func atomicRemove(source string) error {
 // This will mount the dir at its given path
 func (a *Driver) Get(id string, options graphdriver.MountOpts) (string, error) {
 	a.locker.Lock(id)
-	defer a.locker.Unlock(id)
+	defer func() {
+		_ = a.locker.Unlock(id)
+	}()
+
 	parents, err := a.getParentLayerPaths(id)
 	if err != nil && !os.IsNotExist(err) {
 		return "", err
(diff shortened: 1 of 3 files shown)


drivers/aufs/aufs.go
```diff
@@ -347,7 +347,9 @@ func (a *Driver) createDirsFor(id, parent string) error {
 // Remove will unmount and remove the given id.
 func (a *Driver) Remove(id string) error {
 	a.locker.Lock(id)
-	defer a.locker.Unlock(id)
+	defer func(id string) {
```
mtrmac, line 350: (All of these can be done without the explicit `id` parameter, although this works fine as is.)

pkg/chrootarchive/archive_unix.go
```diff
@@ -110,7 +110,9 @@ func invokeUnpack(decompressedArchive io.Reader, dest string, options *archive.T
 		// when `xz -d -c -q | storage-untar ...` failed on storage-untar side,
 		// we need to exhaust `xz`'s output, otherwise the `xz` side will be
 		// pending on write pipe forever
-		io.Copy(io.Discard, decompressedArchive)
+		if _, err := io.Copy(io.Discard, decompressedArchive); err != nil {
+			fatal(err)
```
mtrmac, line 114: AFAICS `fatal` exists to be used in child processes; this is running in the parent and it would terminate the whole process.

pkg/chrootarchive/archive_unix.go
```diff
@@ -107,12 +107,15 @@ func invokeUnpack(decompressedArchive io.Reader, dest string, options *archive.T
 	w.Close()
 
 	if err := cmd.Wait(); err != nil {
+		errorOut := fmt.Errorf("processing tar file(%s): %w", output, err)
 		// when `xz -d -c -q | storage-untar ...` failed on storage-untar side,
 		// we need to exhaust `xz`'s output, otherwise the `xz` side will be
 		// pending on write pipe forever
-		io.Copy(io.Discard, decompressedArchive)
+		if _, err := io.Copy(io.Discard, decompressedArchive); err != nil {
+			return fmt.Errorf("%w and exhaust `xz`'s output failed: %w", errorOut, err)
```
mtrmac, line 115: I think this would be rather a confusing output. e.g. `processing tar file(invalid header): exit status 1 and exhaust xz's output failed: I/O error`, it does not express the relationship of the two errors:
- “unpacking failed (error: …; output: …)”; and
- “exhausting input failed (error: …)”

---

Also note that `xz` in the above is an example, not an inherent component of the situation. (I’m not even sure that we _should_ have to exhaust the input, it seems to me that that’s what SIGPIPE / EPIPE is for, but I didn’t look into the details and that’s probably another PR.)

mtrmac (commented):
/lgtm

Thanks!

PR #2006, opened 2024-07-08
Solve `errcheck` warnings (part 5)

Honny1:
This PR is one of several fixes of warnings found by `golangci` when the `errcheck` linter is enabled. 

Partially fixes:
- #1579


cmd/containers-storage/diff.go
```diff
@@ -192,7 +193,9 @@ func applyDiffUsingStagingDirectory(flags *mflag.FlagSet, action string, m stora
 		DiffOptions: &options,
 	}
 	if _, err := m.ApplyStagedLayer(applyStagedLayerArgs); err != nil {
-		m.CleanupStagedLayer(out)
+		if err := m.CleanupStagedLayer(out); err != nil {
+			logrus.Warn(err)
```
mtrmac, line 197: Please add some context

drivers/overlay/overlay.go
```diff
@@ -2085,7 +2087,9 @@ func (d *Driver) DiffGetter(id string) (_ graphdriver.FileGetCloser, Err error)
 		if Err != nil {
 			for _, f := range composefsMounts {
 				f.Close()
-				unix.Rmdir(f.Name())
+				if err := unix.Rmdir(f.Name()); err != nil && !os.IsNotExist(err) {
+					logrus.Debugf("Failed to remove %s: %v", f.Name(), err)
```
mtrmac, line 2091: `Warnf`, AFAICS the left-over directory could affect other operations later.

drivers/vfs/driver.go
```diff
@@ -193,7 +193,9 @@ func (d *Driver) create(id, parent string, opts *graphdriver.CreateOpts, ro bool
 	}
 	labelOpts := []string{"level:s0"}
 	if _, mountLabel, err := label.InitLabels(labelOpts); err == nil {
-		label.SetFileLabel(dir, mountLabel)
+		if err := label.SetFileLabel(dir, mountLabel); err != nil {
```
mtrmac, line 196: Can we fail the entire operation if this fails? Wouldn’t it break on non-SELinux systems? Maybe a debug log is the best we can do here.

Cc: @rhatdan , history is 73617e5e18159e5f791d2860c7857cd5dea31be4 .

mtrmac (commented):
LGTM I think, I’ll give this a few days to allow others review the `SetFileLabel` part at least.

drivers/vfs/driver.go
rhatdan: mountlabel should be "" on a non SELinux system.

SetFileLabel becomes a no-op.

PR #2008, opened 2024-07-10
Don't discard fields of `LayerOptions`

mtrmac:
Instead of making a copy with some fields, pass the ~deep copy of options we are already using with all of them.

Originally motivated by a linter pointing out that the copy of `options.BigData` has no readers.

In some cases it's not entirely clear that we intended the fields to be a public API; at least I can't see any external caller of any of these fields. But now that they _are_ a public API, we can either implement it, or mark the fields as deprecated ... and then we would need to introduce an internal option structure to carry the fields for the call paths where we do want these fields. That's possible but more work, so I'm doing the lazy thing.

@giuseppe @nalind PTAL

### DIFF
--- a/store.go
+++ b/store.go
@@ -1509,23 +1509,17 @@ func (s *store) putLayer(rlstore rwLayerStore, rlstores []roLayerStore, id, pare
 			gidMap = s.gidMap
 		}
 	}
-	layerOptions := LayerOptions{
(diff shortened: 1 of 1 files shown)


nalind:
LGTM

giuseppe (approved):
/lgtm

PR #2010, opened 2024-07-10
Clean up some warnings

mtrmac:
from what `golangci-lint` and `gopls check` show. This is not comprehensive, primarily to minimize distractions when working on pkg/chunked.

Cc: @giuseppe primarily for a review of the `context.Context` removals.


giuseppe (approved):
LGTM

PR #2019, opened 2024-07-12
drivers: change verbosity to debug for modprobe failures

giuseppe:
commit 48e010f485ffe0dabc8a4e6092dfba2fc74f59c3 caused the regression with rootless since an unprivileged user cannot use `modprobe`.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -693,7 +693,7 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 	selinuxLabelTest := selinux.PrivContainerMountLabel()
 
 	if err := exec.Command("modprobe", "overlay").Run(); err != nil {
-		logrus.Warnf("Execution of `modprobe overlay` ended with error: %v", err)
+		logrus.Debugf("Execution of `modprobe overlay` ended with error: %v", err)
 	}
 
 	logLevel := logrus.ErrorLevel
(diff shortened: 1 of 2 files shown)


giuseppe:
added a small optimization to avoid running `modprobe` if the module is already loaded

drivers/overlay/overlay.go
```diff
@@ -692,8 +692,11 @@ func supportsOverlay(home string, homeMagic graphdriver.FsMagic, rootUID, rootGI
 
 	selinuxLabelTest := selinux.PrivContainerMountLabel()
 
-	if err := exec.Command("modprobe", "overlay").Run(); err != nil {
-		logrus.Warnf("Execution of `modprobe overlay` ended with error: %v", err)
+	// don't try to modprobe overlay if it's already loaded
+	if err := fileutils.Exists("/sys/module/overlay"); err != nil {
+		if err := exec.Command("modprobe", "overlay").Run(); err != nil {
```
Luap99, line 697: Do we actually have to load the module via modprobe? Is this not auto loaded when mounting the fs?

giuseppe: the auto loading in the kernel seems to use a hardcoded path for modprobe:

```
$ grep CONFIG_MODPROBE_PATH /boot/config-$(uname -r)
CONFIG_MODPROBE_PATH="/usr/sbin/modprobe"
```

so if modprobe is installed at a different directory it would not be found.  I am fine with such a change because anyway the user can still load the module manually without us having to run modprobe each time we initialize the storage

giuseppe:
opened an alternative fix, as suggested by @Luap99: https://github.com/containers/storage/pull/2021

Pick the one you prefer :-)

Honny1:
I like more an alternative fix, as suggested by @Luap99: #2021

PR #2021, opened 2024-07-12
drivers: drop modprobe calls

giuseppe:
drop an expensive call to "modprobe" each time we initialize the storage.  The kernel already tries to automatically load the module if needed, and still the user has the possibility to load the module manually if required.

Alternative fix for: https://github.com/containers/storage/pull/2019


Luap99 (approved):
LGTM

Honny1 (approved):
LGTM

Issue #2023, opened 2024-07-12
chunked layersCache seems to never be actually reused


mtrmac:
A side discovery I can’t delve into now:
https://github.com/containers/storage/blob/98ad80d6d16593c959d77b03b33af555eb620891/pkg/chunked/cache_linux.go#L114 defines a new local variable, it does not change the value of the global singleton: https://github.com/containers/storage/blob/98ad80d6d16593c959d77b03b33af555eb620891/pkg/chunked/cache_linux.go#L77

AFAICS that means every cache user ( = every chunked differ = every layer extraction) builds the cache from scratch. The fix is trivial in itself, but it would expose currently-impossible concurrency, so it might need specific testing. 

Cc: @giuseppe

giuseppe:
tentative fix: https://github.com/containers/storage/pull/2024

still marked as Draft as I need to test it better and validate that the locking is still correct

PR #2024, opened 2024-07-12
[WIP] chunked: fix reuse of the layers cache

giuseppe:
the global singleton was never updated, causing the cache to be always recreated for each layer.

It is not possible to keep the layersCache mutex for the entire load() since it calls into some store APIs causing a deadlock since findDigestInternal() is already called while some store locks are held.

Closes: https://github.com/containers/storage/issues/2023

### DIFF
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -111,7 +111,7 @@ func getLayersCacheRef(store storage.Store) *layersCache {
 		cache.refs++
 		return cache
 	}
-	cache := &layersCache{
+	cache = &layersCache{
 		store:   store,
 		refs:    1,
 		created: time.Now(),
@@ -254,13 +254,13 @@ func (c *layersCache) createCacheFileFromTOC(layerID string) (*layer, error) {
 }
 
 func (c *layersCache) load() error {
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
-
 	loadedLayers := make(map[string]*layer)
+	c.mutex.RLock()
 	for _, r := range c.layers {
 		loadedLayers[r.id] = r
 	}
+	c.mutex.RUnlock()
+
 	allLayers, err := c.store.Layers()
 	if err != nil {
 		return err
@@ -300,7 +300,9 @@ func (c *layersCache) load() error {
 	for _, l := range loadedLayers {
 		l.release()
 	}
+	c.mutex.Lock()
 	c.layers = newLayers
+	c.mutex.Unlock()
 	return nil
 }
 


rhatdan:
LGTM
@mtrmac PTAL

pkg/chunked/cache_linux.go
```diff
@@ -111,7 +111,7 @@ func getLayersCacheRef(store storage.Store) *layersCache {
 		cache.refs++
 		return cache
 	}
-	cache := &layersCache{
+	cache = &layersCache{
```
mtrmac, line 114: (I wonder about making this kind of mistake harder to do … naming the global `cacheSingleton`, `globalCache`, or something, might make it less likely to conflict with a local. OTOH that’s very likely an overreaction.)

pkg/chunked/cache_linux.go
```diff
@@ -300,7 +300,9 @@ func (c *layersCache) load() error {
 	for _, l := range loadedLayers {
 		l.release()
```
mtrmac, line 301: AFAICS multiple `load` operations can now be happening in parallel; in that case this might release the same layer multiple times, concurrently.

(I have no opinion on an alternative design at this point, I didn’t look for one yet.)

mtrmac (commented):
Thanks for handling this!

A fairly brief skim for now…

---

(It would be also convenient to have the locking rules documented in detail — e.g. which fields of `layersCache` and `layers` are immutable; which are protected by a lock (and which one).

Especially the lock hierarchy of `layersCache.mutex` vs. the locks of `store`/`layerStore` seems fairly non-obvious to me, given how far the code locations are from each other and how dissimilar the situation seems.

But all of that is blocked on having settled on design in the first place.)

pkg/chunked/cache_linux.go
giuseppe: could we just drop the release() here and let the GC deal with it?  We are already using `runtime.SetFinalizer()`

TomSweeneyRedHat:
@giuseppe @mtrmac this too looks like a good one to get in for the vendor dance.  Would it be possible to wrap this up in the next day or so, or should we consider it for the 1.55.1 release (1.55.0 is the vendor)?

pkg/chunked/cache_linux.go
mtrmac: That would probably _work_… intuitively I don’t like relying on the GC that way, but I don’t have _much_ of an objective reason (at best, that the time of release, and total memory consumption, are hard to manage).

---

It’s not quite sufficient, though, especially for a single-operation command like `podman pull`. Normally, c/image would trigger a pull of up to 6 layers in parallel, all of those would ~concurrently trigger `layersCache.load`, and at least _read_/map all those cache files into memory, or in the worst case create them from scratch, up to 6 times in parallel.

So before this PR we build/load the cache for every pulled layer; with this PR as is, we build/load the cache for up to 6 concurrent layers (and after the first batch of layers, all future cache uses are only the metadata check, and/or update for the just-added layers). This PR _is_ an improvement but not fully where we want to be.

Ideally, I think, the cache loading / update, at least in the case where we do some I/O, should be serialized.

Would two locks work? Nested “exclusive layer loading lock” → c/storage store+layer locks → “the current `layersCache.mutex”? It seems to me that that should be possible, but I didn’t try and I didn’t check very hard.

---

Very vaguely, the current reference count vs. per-cache mutex vs. loading state seems badly structured, or at least it’s unclear to me why it is so. `getLayersCacheRef` is already serialized and it _knows_ that it needs the cache is old. But then `getLayersCache` triggers a `load` (and a reconciliation against `store.Layers`) _every time_. Do we even need the timer there? Do we need the concurrent existence of _multiple_ `layersCache` objects, with each `chunkedDiffer` keeping a reference to a particular cache version?

I’m not saying it’s _obviously wrong_, and for all I know it might be the best design or the only possible design, but nothing in the code writes down why it is so, and it just … feels … off.

TomSweeneyRedHat:
If this is merged on or before August 12, 2024, please cherry-pick this to the release-1.55 branch

cgwalters:
> @giuseppe @mtrmac this too looks like a good one to get in for the vendor dance. 

This is just my opinion but basically this is attempting to optimize zstd:chunked some, and introduces some open questions around concurrency. It doesn't make sense to rush it, and if it merged it's the type of change that should only be cherry picked if there was a known reason to do so.

layers.go
```diff
@@ -2567,14 +2566,22 @@ func (r *layerStore) ApplyDiffWithDiffer(to string, options *drivers.ApplyDiffWi
 			},
 		}
 	}
-	output, err := ddriver.ApplyDiffWithDiffer(layer.ID, layer.Parent, options, differ)
+	out, err := ddriver.ApplyDiffWithDiffer(layer.ID, layer.Parent, options, differ)
```
mtrmac, line 2569: ⚠️ 

I don’t think this works for pre-existing layers (`layer != nil`); if the layer is unlocked:
- Two callers can get here simultaneously, and write over each others’ `getStagingDir`
- The layer can be deleted concurrently to this operation!

---

c/image _only ever_ calls `ApplyDiffWithDiffer` without the layer ID, that’s the only code path I _fully_ care about. (The other code path is used in `cmd/containers-storage` and in tests, per comments in `store.go`.)

 It is very tempting to me to have a new specialized `PrepareStagedLayer` (to match the names of `{Apply,Cleanup}StagedLayer`) that only supports this simpler care; then several of the operations here could just disappear.

But `ApplyDiffWithDiffer` is now a stable API; I don’t immediately know how to deal with that. The concurrent layer deletion _might_ be possible to handle by holding a read lock on the layer store only (assuming all operations in `pkg/chunked` lock the layer store _only_ for reading, **and that’s not currently the case**), but the exclusion of two concurrent callers, and of any possible concurrent readers, is fundamentally the existing exclusive layer store lock.

Outright refusing to do the operation on existing layers … is code that could be written… but probably shouldn’t be merged. I don’t know.

layers.go
```diff
@@ -314,9 +314,12 @@ type rwLayerStore interface {
 	// applies its changes to a specified layer.
 	ApplyDiff(to string, diff io.Reader) (int64, error)
 
-	// ApplyDiffWithDiffer applies the changes through the differ callback function.
+	// ApplyDiffWithDifferNoLock applies the changes through the differ callback function.
 	// If to is the empty string, then a staging directory is created by the driver.
```
mtrmac, line 318: (this needs updating)

layers.go
```diff
@@ -314,9 +314,12 @@ type rwLayerStore interface {
...
-	ApplyDiffWithDiffer(to string, options *drivers.ApplyDiffWithDifferOpts, differ drivers.Differ) (*drivers.DriverWithDifferOutput, error)
+	ApplyDiffWithDifferNoLock(layer *Layer, options *drivers.ApplyDiffWithDifferOpts, differ drivers.Differ) (*drivers.DriverWithDifferOutput, error)
+
+	// ApplyDiffWithDifferFinalize stores the changes performed by ApplyDiffWithDifferNoLock
```
mtrmac, line 321: Documenting the relationship with `…NoLock` explicitly would be useful.

layers.go
mtrmac, line 318: Documenting the relationship with `…Finalize` explicitly would be useful.

layers.go
```diff
@@ -2543,22 +2546,18 @@ func (r *layerStore) applyDiffFromStagingDirectory(id string, diffOutput *driver
 	return err
 }
 
-// Requires startWriting.
-func (r *layerStore) ApplyDiffWithDiffer(to string, options *drivers.ApplyDiffWithDifferOpts, differ drivers.Differ) (*drivers.DriverWithDifferOutput, error) {
+// It must be called without calling startReading or startWriting first.
```
mtrmac, line 2549:
```suggestion
// It must be called without any c/storage locks held to allow differ to make c/storage calls.
```

mtrmac (commented):
I like the approach but I don’t think it works with the current API.

(Start with ⚠️ , the other review comments are ~irrelevant at this point.)

layers.go
giuseppe: > But `ApplyDiffWithDiffer` is now a stable API; I don’t immediately know how to deal with that. The concurrent layer deletion _might_ be possible to handle by holding a read lock on the layer store only (assuming all operations in `pkg/chunked` lock the layer store _only_ for reading, **and that’s not currently the case**), but the exclusion of two concurrent callers, and of any possible concurrent readers, is fundamentally the existing exclusive layer store lock.
> 
> Outright refusing to do the operation on existing layers … is code that could be written… but probably shouldn’t be merged. I don’t know.

I don't know either how to solve it, unless introducing another per-layer lock file.  As you said, this code path is used only by `cmd/containers-storage`.

Could we just document this behavior, and warn any potential user that using `to != ""` is not safe as there can be multiple writers, and the layer could potentially be deleted while the operation runs?

mtrmac: To maybe overstate the case a bit: either we are keeping a stable API, and this operation, which used to work correctly, needs to continue to work; or we are fine with imposing changes on callers, and in that case we can force the callers to stop using `to != ""`.

It’s probably not _so_ black/white, but I don’t think documenting the problem fundamentally changes the situation. Pragmatically, if there _were_ callers (and I don’t know about any), they wouldn’t notice the warning. Unless, I suppose, we hard-coded a `logrus.Warn*` on every call?

mtrmac: Introducing new locks on a code path _with no known users_ seems like a waste and unnecessary risk to me… and I guess runs into my policy of “not reviewing things that add complexity to drivers/overlay”.

I guess this is where @ rhatdan swoops in and merges a PR that outright breaks the API, and I pretend to grumble a bit but I’m secretly relieved 😈 So I suppose we should just do that? At least GitHub shows `ApplyDiffWithDiffer` references only in forks of c/storage and c/image.

giuseppe: I'd just break the only caller, but there as well, I am not sure how to keep the CLI :-) working as there is a clear split between creating a layer and populating it

mtrmac: Is anyone using the CLI, apart from tests?

giuseppe: no

mtrmac: … answering myself, CRI-O is now telling users to run the CLI’s `check` command, or rather providing one-off binaries to do so. But I also don’t know about any users using the CLI to trigger partial pulls… and its’ implementation based on a `fileFetcher` is not “partial” unless backed by some hypothetical on-demand-fetch filesystem.

rhatdan: `To maybe overstate the case a bit: either we are keeping a stable API, and this operation, which used to work correctly, needs to continue to work; or we are fine with imposing changes on callers, and in that case we can force the callers to stop using to != ""`

Force users to stop using !="". I will merge and @mtrmac can grumble.  Lets move on.

giuseppe: I've added a new function `PrepareStagedLayer` that doesn't have the `to` argument.  We can drop the old function once we update c/image

mtrmac: Thanks.

Now that the `to != ""` path outright fails, can you follow that to all the consequences, please, so that we don’t leave known-deadlocking code around?
- We won’t need the …`NoLock`/`Finalize` dichotomy (and `Finalize` has no callers)
- The always-`""` parameter of `layerStore.ApplyDiffWithDiffer` can be removed, as well as most of the code
- The always-`""` parameter of `overlay.Driver.ApplyDiffWithDiffer` can be removed, along with users
- (I didn’t realize that the CLI has already been migrated, so that’s simpler.)

store.go
```diff
@@ -341,11 +341,17 @@ type Store interface {
 	//   }
 	ApplyDiff(to string, diff io.Reader) (int64, error)
 
-	// ApplyDiffer applies a diff to a layer.
+	// ApplyDiffWithDiffer applies a diff to a layer.
 	// It is the caller responsibility to clean the staging directory if it is not
-	// successfully applied with ApplyDiffFromStagingDirectory.
+	// successfully applied with ApplyStagedLayer.
+	// Deprecated: Use PrepareStagedLayer instead.  It is going to be removed in a future release
```
mtrmac, line 347:
```suggestion
	// Deprecated: Use PrepareStagedLayer instead.  ApplyDiffWithDiffer is going to be removed in a future release
```

layers.go
```diff
@@ -2543,22 +2549,18 @@ func (r *layerStore) applyDiffFromStagingDirectory(id string, diffOutput *driver
 	return err
 }
 
-// Requires startWriting.
-func (r *layerStore) ApplyDiffWithDiffer(to string, options *drivers.ApplyDiffWithDifferOpts, differ drivers.Differ) (*drivers.DriverWithDifferOutput, error) {
+// It must be called without any c/storage locks held to allow differ to make c/storage calls.
+func (r *layerStore) ApplyDiffWithDifferNoLock(layer *Layer, options *drivers.ApplyDiffWithDifferOpts, differ drivers.Differ) (*drivers.DriverWithDifferOutput, error) {
```
mtrmac, line 2553: (If we are changing the interface of this method, we might just as well make it package-private at the same time, to make it clearer there are no external callers.)

giuseppe:
@mtrmac thanks for the review.  Addressed the comments in the last version

store.go
```diff
@@ -3103,13 +3109,21 @@ func (s *store) CleanupStagedLayer(diffOutput *drivers.DriverWithDifferOutput) e
...
+	rlstore, err := s.getLayerStore()
+	if err != nil {
+		return nil, err
+	}
+
+	if to != "" {
+		return nil, fmt.Errorf("ApplyDiffWithDiffer does not support non-empty 'layer' parameter")
+	}
+
+	return rlstore.applyDiffWithDifferNoLock(nil, options, differ)
```
mtrmac, line 3126: The first parameter can be entirely removed, as well as the relevant code paths down the call stack.

giuseppe: thanks, fixed.

I'll work on a follow up to simplify the `ApplyDiffWithDiffer` in the graph driver too

giuseppe: > I'll work on a follow up to simplify the `ApplyDiffWithDiffer` in the graph driver too

added a patch to this PR

store.go
```diff
@@ -3103,13 +3109,21 @@ func (s *store) CleanupStagedLayer(diffOutput *drivers.DriverWithDifferOutput) e
 	return err
 }
 
+func (s *store) PrepareStagedLayer(options *drivers.ApplyDiffWithDifferOpts, differ drivers.Differ) (*drivers.DriverWithDifferOutput, error) {
+	return s.ApplyDiffWithDiffer("", options, differ)
```
mtrmac, line 3113: (Non-blocking: Consider switching the direction of the indirection, so that the deprecated `ApplyDiffWithDiffer` calls the new `PrepareStagedLayer`. That way, there won’t be any callers of ADWD remaining, and removing it later will be a smaller change.

LGTM either way.)

mtrmac (commented):
LGTM. Thanks!

store.go
giuseppe: the check for `rlstore.Exists(to)` requires to take a lock.

If you prefer I could duplicate the code:

```go
func (s *store) PrepareStagedLayer(options *drivers.ApplyDiffWithDifferOpts, differ drivers.Differ) (*drivers.DriverWithDifferOutput, error) {
	return writeToLayerStore(s, func(rlstore rwLayerStore) (*drivers.DriverWithDifferOutput, error) {
		return rlstore.ApplyDiffWithDiffer(options, differ)
	})
}
```

giuseppe: yeah sorry, I was looking at doing this change in earlier commits from the PR, but I can apply it directly to the "chunked: fix reuse of the layers cache" patch.

Applied now, and pushed a new version

mtrmac (commented):
/lgtm

Thanks again!

Issue #2022, opened 2024-07-12
Improve on top of #2012


mtrmac:
… per https://github.com/containers/storage/pull/2012/files#r1676275842

PR #2025, opened 2024-07-12
archive: improve filter error reporting

giuseppe:
follow-up for https://github.com/containers/storage/pull/2012

report the error as seen by the input stream, instead of attempting another read.

Closes: https://github.com/containers/storage/issues/2022

### DIFF
--- a/pkg/archive/filter.go
+++ b/pkg/archive/filter.go
@@ -26,6 +26,19 @@ func getFilterPath(name string) string {
 	return path.(string)
 }
 
+type errorRecordingReader struct {
+	r   io.Reader
+	err error
+}
+
+func (r *errorRecordingReader) Read(p []byte) (int, error) {
+	n, err := r.r.Read(p)
+	if r.err == nil && err != io.EOF {
+		r.err = err
+	}
+	return n, err
+}
+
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@mtrmac PTAL

pkg/archive/filter.go
```diff
@@ -26,6 +26,19 @@ func getFilterPath(name string) string {
...
 
+type errorRecordingReader struct {
+	r   io.Reader
+	err error
+}
+
+func (r *errorRecordingReader) Read(p []byte) (int, error) {
+	n, err := r.r.Read(p)
+	if err != nil && err != io.EOF {
+		r.err = err
```
mtrmac, line 37: I think this should prefer the first error, not the last one.

mtrmac (commented):
Thanks!

LGTM otherwise.

pkg/archive/filter.go
giuseppe: yes that is better, changed now

mtrmac: … but still not store `err == nil`?

mtrmac: … never mind, this works without an explicit check for `err != nil` correctly.

PR #2031, opened 2024-07-16
chunked: do not write cache file to RO store

giuseppe:
if the layer is R/O, do not write a cache file.

### DIFF
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -287,6 +287,13 @@ func (c *layersCache) load() error {
 			newLayers = append(newLayers, l)
 			continue
 		}
+
+		if r.ReadOnly {
+			// if the layer is coming from a read-only store, do not attempt
+			// to write to it.
+			continue
+		}
+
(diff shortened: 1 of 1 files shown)


giuseppe:
@rhatdan @mtrmac an easy one, but it solves an annoying warning when using an additional image store

mtrmac (commented):
Doesn’t this effectively _completely stop_ looking for files in the read-only layers? It doesn’t stop _writing_ the cache, it stops also _creating_ it. And without an entry in `c.layers`, there can be no matches found in that layer.

\*shrug\* I don’t have a strong opinion on impact of that — avoiding cross-store linking might well be a good idea — but I’d like at least the comment in the code to be correct about the impact.

(And, ideally, the various implications of ALS/AIS, especially WRT cross-store sharing and dependencies, would be documented in detail somewhere, but I can see an argument that that’s not this PR.)

rhatdan:
LGTM

mtrmac:
> I’d like at least the comment in the code to be correct about the impact.

#2079 .

PR #2035, opened 2024-07-16
Ensure chunked TOC and tar-split metadata are consistent

### LINKED ISSUE #2014: zstd:chunked metadata ambiguity
Filing separately, earlier discussion around https://github.com/containers/storage/issues/1888#issuecomment-2145636434 :

zstd:chunked layers, when pulling, contain metadata in _three_ places:
- The ordinary uncompressed tar format
- TOC metadata
- tar-split metadata

When pulling, we ignore the uncompressed tar, build files from TOC metadata, and record tar-split 

When pushing, we ignore the metadata of individual files, and use the tar-split metadata.

The net outcome is that a user can “pull; inspect; push”, and the pushed metadata will be different from what the user saw.

I think this_must_ be addressed.

---

Vaguely, I think that could either happen by having the tar-split “drive” the chunked pull, using the TOC only to look up data; or by overwriting the TOC metadata by the tar-split data.

The latter seems a bit easier because the “push” compression code already has a tar → TOC conversion code, so we would “only” need to read through tar-split; the former is _conceptually_ nicer because, hypothetically, we could eventually only have a _single_ “apply tar header → filesystem metadata” code for all of c/storage, instead of now having “tar → metadata” for ordinary overlay, “TOC → metadata” for chunked, and things like https://github.com/containers/storage/pull/1653#discussion_r1256423401 — but it would probably be more disruptive and not practical within current short time limits.

Cc: @giuseppe — I’ll be looking at this over the next few days, but I’d very much appreciate any insight, advice, or help.

mtrmac:
This should resolve #2014. Basically untested, filing now for early design review.

We simply enforce that the TOC and tar-split have exactly the same contents. To the extent the zstd:chunked format is only being produced by this package, we can expect an equal match (right now? ), but that may become harder as the format evolves (e.g. the recent `timeIfNotZero` change).

@giuseppe PTAL. Cc: @cgwalters .


pkg/chunked/tar_split_linux.go
```diff
@@ -0,0 +1,68 @@
...
+			if err == io.EOF {
+				return nil
+			}
+			return fmt.Errorf("reading tar-split entries: %w", err)
+		}
+		switch tsEntry.Type {
+		case storage.SegmentType:
+			payload := tsEntry.Payload
+			// This is horrible, but we don’t know how much padding to skip. (It can be computed from the previous hdr.Size for non-sparse
+			// files, but for sparse files that is set to the logical size.)
```
mtrmac, line 37: This is a pretty horrible hack.

It should be possible for `tar-split/archive/tar` to expose the expected padding size in `Header`.

(Alternatively, the OCI spec suggests that sparse files should not be used, but I don’t expect that can be relied upon.)

mtrmac:
> `tar-split and TOC data is inconsistent: reading tar-split entries: invalid character 'T' looking for beginning of value`

The test failure is on unrealistic data in a different test in this package, I’ll update that tomorrow.

pkg/chunked/compression_linux.go
```diff
@@ -234,6 +243,121 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
 	return decodedBlob, toc, decodedTarSplit, int64(manifestChunk.Offset), err
 }
 
+// ensureTOCMatchesTarSplit validates that toc and tarSplit contain _exactly_ the same entries.
+func ensureTOCMatchesTarSplit(toc *internal.TOC, tarSplit []byte) error {
+	pendingFiles := map[string]*internal.FileMetadata{} // Name -> an entry in toc.Entries
+	for i := range toc.Entries {
+		e := &toc.Entries[i]
+		if e.Type != internal.TypeChunk {
```
cgwalters, line 251: Very minor but I paused on this for a bit, uncertain; perhaps:

```
// Chunks are just part of files, they won't appear explicitly
// in the tar stream, so we don't validate them.
if e.Type == internal.TypeChunk {
    continue
}
```
?

pkg/chunked/tar_split_linux.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"io"
+
+	"github.com/vbatts/tar-split/archive/tar"
+	"github.com/vbatts/tar-split/tar/storage"
+)
+
+// iterateTarSplit calls handler for each tar header in tarSplit
+func iterateTarSplit(tarSplit []byte, handler func(hdr *tar.Header) error) error {
+	// This, strictly speaking, hard-codes undocumented assumptions about how github.com/vbatts/tar-split/tar/asm.NewInputTarStream
```
cgwalters, line 14: In theory this API could land as a PR to that repo?

pkg/chunked/tar_split_linux_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/containers/storage/pkg/chunked/internal"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/vbatts/tar-split/archive/tar"
+	"github.com/vbatts/tar-split/tar/asm"
+	"github.com/vbatts/tar-split/tar/storage"
+)
+
+func testTarheader(index int, typeFlag byte, size int64) tar.Header {
```
cgwalters, line 18: The name made me thing this is itself a test, how about `createTestTarheader`?

cgwalters (approved):
Only a relatively superficial skim but generally LGTM.

pkg/chunked/compression_linux.go
mtrmac: I think this would be best documented somewhere around `pkg/chunked/internal/compression.go`, this is “just” straightforwardly consuming the structure as designed.

https://github.com/containers/storage/pull/1939 did add the basic documentation of `TypeChunk`, although the full semantics of `Offset`/`ChunkOffset` etc. is not written down anywhere.

pkg/chunked/tar_split_linux.go
mtrmac: That would certainly be better. The timing will probably force us to carry this in c/storage at for a few days/weeks, but I should definitely prepare a tar-split PR.

pkg/chunked/tar_split_linux_test.go
mtrmac: Thanks, fixed.

giuseppe (approved):
LGTM

mtrmac:
Podman tests passed in https://github.com/containers/podman/pull/23307 .

pkg/chunked/compression_linux.go
mtrmac: Just to be a good citizen, I have added a commit documenting the format (to the extend I can reverse-engineer it from the code) to the documentation of `internal.FileMetadata`.

giuseppe (approved):
/lgtm

pkg/chunked/tar_split_linux.go
mtrmac: I have recorded the task to file a tar-split PR as an item on https://github.com/containers/image/issues/2189 .

PR #2039, opened 2024-07-23
loopback: fix race condition opening loopback device

giuseppe:
the loopback device file could be already used/removed by another process.  Since the process is inherently racy, just grab the next available index and try again until it succeeds.

Closes: https://github.com/containers/storage/issues/2038

### DIFF
--- a/pkg/loopback/attach_loopback.go
+++ b/pkg/loopback/attach_loopback.go
@@ -6,6 +6,7 @@ package loopback
 import (
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"syscall"
 
@@ -53,24 +54,26 @@ func openNextAvailableLoopback(index int, sparseName string, sparseFile *os.File
 		target := fmt.Sprintf("/dev/loop%d", index)
 		index++
 
-		fi, err := os.Stat(target)
-		if err != nil {
-			if os.IsNotExist(err) {
-				logrus.Error("There are no more loopback devices available.")
-			}
-			return nil, ErrAttachLoopbackDevice
-		}
-
-		if fi.Mode()&os.ModeDevice != os.ModeDevice {
-			logrus.Errorf("Loopback device %s is not a block device.", target)
-			continue
-		}
-
 		// OpenFile adds O_CLOEXEC
 		loopFile, err = os.OpenFile(target, os.O_RDWR, 0o644)
 		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				continue
+			}
 			logrus.Errorf("Opening loopback device: %s", err)
 			return nil, ErrAttachLoopbackDevice
+		}
+
+		fi, err := loopFile.Stat()
+		if err != nil {
+			loopFile.Close()
+			logrus.Errorf("Stat loopback device: %s", err)
+			return nil, ErrAttachLoopbackDevice
+		}
+		if fi.Mode()&os.ModeDevice != os.ModeDevice {
+			loopFile.Close()
+			logrus.Errorf("Loopback device %s is not a block device.", target)
+			continue
 		}
 
 		// Try to attach to the loop file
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@nalind @mtrmac @saschagrunert PTAL

rhatdan:
@edsantiago PTAL

cgwalters:
A few notes:

- If we do this then it's likely that composefs (which also does the same thing as we do here, but we don't link to libcomposefs I guess because linking Go and C is painful?) needs the same changes
- systemd has a lot of similar logic https://github.com/systemd/systemd/blob/ae2bf016bc946f47dd8c027ccf76d19143e2e9c5/src/shared/loop-util.c#L528 - one thing it does (and documents well) is keeping a flock() around the loop control device
- I think we should keep retrying with `getNextFreeLoopbackIndex()` and not just calling that once and thereafter incrementing the index, right?

cgwalters:
(Also xref https://github.com/containers/composefs/issues/144 for the larger better fix)

edsantiago:
>I think we should keep retrying with getNextFreeLoopbackIndex() and not just calling that once and thereafter incrementing the index, right?

The `index++` is now a NOP, and super misleading, hence my earlier comment about a nice fresh cleanup

giuseppe:
I agree the module requires a refactoring, the error handling is quite unusual: everything is logged to stderr and it just reports `ErrAttachLoopbackDevice`, but I don't think it must be done as part of this PR.

The goal of this PR is to address a specific bug and the diff (except the new test) is very small so it can be easily backported it if needed.


> Also... again, the commit sequence is hard to follow

what exactly should be changed?

> The `index++` is now a NOP, and super misleading, hence my earlier comment about a nice fresh cleanup

how is that a NOP?  It is part of a for loop.  The `index, err = getNextFreeLoopbackIndex()` is done only on the error path.

> * I think we should keep retrying with `getNextFreeLoopbackIndex()` and not just calling that once and thereafter incrementing the index, right?

I am fine to change that, but it is still racy (another process can use that device) and it is slightly slower as we need an additional syscall

edsantiago:
`index++` is a NOP because `index` is never again referenced after that point. It is redefined on error, but never read.

As for what should be changed, `openNextAvailableLoopback()` should be rewritten as
```
func openNextAvailableLoopback(sparseName string, sparseFile ...) {
    index := getNextFreeLoopbackIndex()
```
There is, AFAICT, no purpose whatsoever to `attachLoopDevice()` calling `getNext...`. It does not do anything with `startIndex` other than pass it as an arg to the function that needs it. As a reader, I find it unnecessary and confusing.

Perhaps I've missed something, though?

edsantiago:
Never mind my `index++` comment, I missed the `continue`s in various other cases. I agree with @cgwalters, the code should probably `getNext` on each iteration even if it's slow.

giuseppe:
changed to always use `getNextFreeLoopbackIndex()`

cgwalters:
/approve

I like this one better, thanks. I do think there's more cleanup to do here but it can come as a followup.

pkg/loopback/attach_loopback.go
```diff
@@ -50,29 +52,40 @@ func openNextAvailableLoopback(index int, sparseName string, sparseFile *os.File
 
 	// Start looking for a free /dev/loop
 	for {
```
mtrmac, line 54: Shouldn’t there be a limit against infinite loops? It can be a ridiculous number — 100 attempts, 1000 attempts — but infinitely looping (e.g. if some weird security software prevented usage of `/dev/loop*`), with no output even, would be bad.

mtrmac (commented):
ACK overall.

pkg/loopback/attach_loopback.go
cgwalters: Probably yes; looks like systemd does 64 attempts with a random sleep delay.

giuseppe: I've added an upper bound with 1000 attempts

PR #2041, opened 2024-07-24
chunked: do not use a temporary file

giuseppe:
and use directly the stream to create the temporary zstd:chunked file.

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -26,6 +26,7 @@ import (
 	"github.com/containers/storage/pkg/chunked/toc"
 	"github.com/containers/storage/pkg/fsverity"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/system"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/klauspost/compress/zstd"
@@ -107,7 +108,7 @@ type chunkedLayerData struct {
 	Format graphdriver.DifferOutputFormat `json:"format"`
 }
 
-func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload *os.File) (int64, *seekableFile, digest.Digest, map[string]string, error) {
+func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload io.Reader) (int64, *seekableFile, digest.Digest, map[string]string, error) {
 	diff, err := archive.DecompressStream(payload)
 	if err != nil {
 		return 0, nil, "", nil, err
@@ -1076,7 +1077,7 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
 	return new, nil
 }
 
-func (c *chunkedDiffer) copyAllBlobToFile(destination *os.File) (digest.Digest, error) {
+func (c *chunkedDiffer) requestWholeBlob() (io.ReadCloser, digest.Digester, error) {
 	var payload io.ReadCloser
 	var streams chan io.ReadCloser
 	var errs chan error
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
and I'd like to get this in before starting the vendor dance.  @nalind @mtrmac PTAL

pkg/chunked/storage_linux.go
```diff
@@ -1131,32 +1133,17 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
...
-		blobFile := os.NewFile(uintptr(fd), "blob-file")
 		defer func() {
-			if blobFile != nil {
-				blobFile.Close()
+			if payload != nil {
+				payload.Close()
 			}
 		}()
 
-		// calculate the checksum before accessing the file.
```
cgwalters, line 1145: Presumably we were doing this for a reason (security?) before...is it just that we think `convertTarToZstdChunked` can now safely operate on untrusted input?

cgwalters (commented):
The goal is improved performance, right? Worth stating in the commit message if so.

pkg/chunked/storage_linux.go
```diff
@@ -1165,9 +1152,15 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 		// need to keep it open until the entire file is processed.
 		defer fileSource.Close()
 
-		// Close the file so that the file descriptor is released and the file is deleted.
-		blobFile.Close()
-		blobFile = nil
+		// Make sure the entire payload is consumed.
+		_, _ = io.Copy(io.Discard, payload)
```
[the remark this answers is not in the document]
giuseppe: I've not measured performance, this is just a preparatory work for extending `ApplyDiff()` to use this codepath when we convert images, so we can support pulling from other sources too, not only registries.

I've not yet found a nice way to extend the API though.

pkg/chunked/storage_linux.go
giuseppe: hmm I don't remember the security implications, @mtrmac do you think this approach is fine or should I just close this PR?

pkg/chunked/storage_linux.go
```diff
@@ -1091,26 +1092,27 @@ func (c *chunkedDiffer) copyAllBlobToFile(destination *os.File) (digest.Digest,
...
-	originalRawDigester := digest.Canonical.Digester()
+	digester := digest.Canonical.Digester()
 
-	r := io.TeeReader(payload, originalRawDigester.Hash())
+	r := io.TeeReader(payload, digester.Hash())
 
-	// copy the entire tarball and compute its digest
-	_, err = io.CopyBuffer(destination, r, c.copyBuffer)
+	rc := ioutils.NewReadCloserWrapper(r, func() error {
+		return payload.Close()
```
mtrmac, line 1112: We were not closing `payload` before? That fix should not be forgotten.

pkg/chunked/storage_linux.go
mtrmac: On balance, I’d prefer this not to be merged as is.

---

This exposes the conversion code and decompression, to malicious input in more situations.

In principle, the conversion code and decompression must be able to handle malicious input _anyway_, because an attacker could, in many cases, refer to the malicious input in a valid manifest, without triggering this check.

But _some_ users might have a policy which only accepts signed images, i.e. the malicious input would only be digested, not processed otherwise.

I’m really more worried about the decompression than the chunked conversion. That’s a lot of bit manipulation, potentially in an unsafe language “for performance”, with a fair history of vulnerabilities.

… and for ordinary image pulls, we currently also stream the input through decompression, only validating the digest concurrently; we don’t validate the digest before starting to decompress. So, in that sense, we are already accepting a larger part of the risk.

So I don’t feel that strongly about it here.

---

More importantly, if this is aiming at the c/storage `ApplyDiff` = c/image `PutBlob` (not `PutBlobPartial`) path, c/image is currently creating a temporary file, _and verifying the digest_, on that path, without c/storage having any way to prevent it; and c/storage wouldn’t need to do this. (Also, c/image streams the data to the file, and digests it, concurrently for several layers, without holding any storage locks, which seems valuable.)

So, if the goal is _code structure_, not _performance or `PutBlobPartial`_, I don’t see that this makes any difference for the `ApplyDiff` path, and it is just a performance/risk trade-off for the `PutBlobPartial` path.

If you don’t actually _care_ about the performance at this point, we get the increased risk but not any benefit we value — so I’d prefer to close the PR and not merge this; we can always resurrect it later.

Except for the one line that allows `convertTarToZstdChunked` to accept an arbitrary `io.Reader`, and the missing `payload.Close`.

pkg/chunked/storage_linux.go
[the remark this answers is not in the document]
mtrmac: Yes (if this goes in) I think it would be better to report a read error (the cause) instead of a digest mismatch that is hard to diagnose.

mtrmac (commented):
See elsewhere for the higher-level evaluation.

cgwalters:
> I’m really more worried about the decompression than the chunked conversion. That’s a lot of bit manipulation, potentially in an unsafe language “for performance”, with a fair history of vulnerabilities.

Though now that we're forking a separate process we could isolate those quite aggressively.

mtrmac:
Which forking does that refer to?

cgwalters:
> Which forking does that refer to?

https://github.com/containers/storage/pull/1964

cgwalters:
> … and for ordinary image pulls, we currently also stream the input through decompression, only validating the digest concurrently; we don’t validate the digest before starting to decompress. So, in that sense, we are already accepting a larger part of the risk.

Right there's that and also ultimately we still need to be safe against untrusted/malicious zstd:chunked inputs *even if* the checksum matches.

So I'm going to proactively reopen this PR since I think it makes sense.

TomSweeneyRedHat:
If this is merged on or before August 12, 2024, please cherry-pick this to the release-1.55 branch

Issue #2042, opened 2024-07-24
composefs: random EINVALs


edsantiago:
Sometimes on commit, sometimes on diff
```
# podman-remote [options] commit -q testcon newimage
Error: copying layers and metadata for container "cecc9c90f4107cf89808b8234df98e173e1bf052a006665f460a1e86f8d070b8": initializing source containers-storage:testcon: extracting layer "1745071114b3933b11b73a9f099d2b82d4e80262c1211b1b6e7f388466ebdf3c": invalid argument
```
```
# podman [options] --pull-option=enable_partial_images=true --pull-option=convert_images=true container checkpoint --create-image alpine-checkpoint-bmjmtt --keep e3793c9b35baf7d6a62a976fb4da4e4fc139c535edd159cd6edcf96a8b4e1d02
Error: exporting root file-system diff for "e3793c9b35baf7d6a62a976fb4da4e4fc139c535edd159cd6edcf96a8b4e1d02": invalid argument
```
* rawhide : int podman rawhide root host sqlite
    * [07-24 07:14](https://api.cirrus-ci.com/v1/artifact/task/5036223750733824/html/int-podman-rawhide-root-host-sqlite.log.html#t--Podman-checkpoint-podman-checkpoint-create-image-with-running-container--1) in Podman checkpoint podman checkpoint --create-image with running container
* rawhide : int remote rawhide root host sqlite [remote]
    * [07-22 11:56](https://api.cirrus-ci.com/v1/artifact/task/6664205737132032/html/int-remote-rawhide-root-host-sqlite.log.html#t--Podman-commit-podman-commit-adds-exposed-ports--1) in Podman commit podman commit adds exposed ports

| x | x | x | x | x | x |
| ---: | ---: | ---: | ---: | ---: | ---: |
| int(2) | podman(1) | rawhide(2) | root(2) | host(2) | sqlite(2) |
|  | remote(1) |  |  |  |  |

edsantiago:
[Another one on commit](https://api.cirrus-ci.com/v1/artifact/task/4890561826521088/html/int-remote-rawhide-root-host-sqlite.log.html#t--Podman-commit-podman-commit-container-with-message--1):
```
# podman-remote [options] commit -q -f docker --message testing-commit test1 foobar.com/test1-image:latest
[...]

PR #2043, opened 2024-07-24
Add some error context in Changes codepaths

cgwalters:
Motivated by https://github.com/containers/storage/issues/2042 where we just get a bare `invalid argument` out of the entire storage stack.

My offhand guess skimming some of the code is by far the most likely thing here is the raw `lgetxattr` call. It'd be useful to know that for sure.


giuseppe (approved):
LGTM

giuseppe:
I was working on something similar: https://github.com/containers/storage/pull/2044 :-)

cgwalters:
> I was working on something similar: https://github.com/containers/storage/pull/2044 :-)

And they're complementary which is nice!

PR #2048, opened 2024-07-29
overlay,chunked: Add some more O_CLOEXEC

cgwalters:
Just some more missing `O_CLOEXEC` I noticed while reading the code for unrelated reasons.


vrothberg (commented):
LGTM

PR #2049, opened 2024-07-29
chrootarchive: Pass root via fd

cgwalters:
I'd like to support passing a file descriptor root for the container storage, and not an absolute path.

In the bootc codebase (partially a philosophy inherited from ostree) we've heavily invested in fd-relative accesses, primarily because it's common for us to operate in different namespaces/roots, and fd-relative access avoids a lot of possible footguns when dealing with absolute paths. It's also more efficient, avoiding the need for the kernel to traverse full paths a lot.

This is just one of a few preparatory changes necessary in making it work to do:

`podman --root=/proc/self/fd/3 --runroot=... pull busybox`

Note that as part of doing this, I refactored code here to move the "split root and dest" logic into the caller... there was some logically dead code here because

### DIFF
--- a/pkg/chrootarchive/archive.go
+++ b/pkg/chrootarchive/archive.go
@@ -83,6 +83,12 @@ func untarHandler(tarArchive io.Reader, dest string, options *archive.TarOptions
 		}
 	}
 
+	destVal, err := newUnpackDestination(root, dest)
+	if err != nil {
+		return err
+	}
+	defer destVal.Close()
+
 	r := tarArchive
 	if decompress {
 		decompressedArchive, err := archive.DecompressStream(tarArchive)
@@ -93,7 +99,7 @@ func untarHandler(tarArchive io.Reader, dest string, options *archive.TarOptions
 		r = decompressedArchive
 	}
 
-	return invokeUnpack(r, dest, options, root)
+	return invokeUnpack(r, destVal, options)
 }
 
 // Tar tars the requested path while chrooted to the specified root.
--- a/pkg/chrootarchive/archive_unix.go
+++ b/pkg/chrootarchive/archive_unix.go
@@ -30,6 +30,9 @@ func (dst *unpackDestination) Close() error {
 	return dst.root.Close()
 }
 
+// tarOptionsDescriptor is passed as an extra file
+const tarOptionsDescriptor = 3
+
 // rootFileDescriptor is passed as an extra file
 const rootFileDescriptor = 4
 
@@ -51,7 +54,7 @@ func untar() {
 	var options archive.TarOptions
 
 	// read the options from the pipe "ExtraFiles"
-	if err := json.NewDecoder(os.NewFile(3, "options")).Decode(&options); err != nil {
+	if err := json.NewDecoder(os.NewFile(tarOptionsDescriptor, "options")).Decode(&options); err != nil {
 		fatal(err)
 	}
 
@@ -132,6 +135,7 @@ func invokeUnpack(decompressedArchive io.Reader, dest *unpackDestination, option
 	cmd := reexec.Command("storage-untar", dest.dest, procPathForFd(rootFileDescriptor))
 	cmd.Stdin = decompressedArchive
 
+	// If you change this, change tarOptionsDescriptor above
 	cmd.ExtraFiles = append(cmd.ExtraFiles, r) // fd 3
 	// If you change this, change rootFileDescriptor above too
 	cmd.ExtraFiles = append(cmd.ExtraFiles, dest.root) // fd 4
--- a/pkg/chrootarchive/archive_unix.go
+++ b/pkg/chrootarchive/archive_unix.go
@@ -9,14 +9,37 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/reexec"
 )
+
+type unpackDestination struct {
+	root *os.File
+	dest string
+}
+
+func (dst *unpackDestination) Close() error {
+	return dst.root.Close()
+}
+
+// rootFileDescriptor is passed as an extra file
+const rootFileDescriptor = 4
+
+// procPathForFd gives us a string for a descriptor.
+// Note that while Linux supports actually *reading* this
+// path, FreeBSD and other platforms don't; but in this codebase
+// we only compare strings.
+func procPathForFd(fd int) string {
+	return fmt.Sprintf("/proc/self/fd/%d", fd)
+}
 
 // untar is the entry-point for storage-untar on re-exec. This is not used on
 // Windows as it does not support chroot, hence no point sandboxing through
@@ -38,7 +61,17 @@ func untar() {
 		root = flag.Arg(1)
 	}
 
-	if root == "" {
+	// FreeBSD doesn't have proc/self, but we can handle it here
+	if root == procPathForFd(rootFileDescriptor) {
+		// Take ownership to ensure it's closed; no need to leak
+		// this afterwards.
+		rootFd := os.NewFile(rootFileDescriptor, "tar-root")
+		defer rootFd.Close()
+		if err := unix.Fchdir(int(rootFd.Fd())); err != nil {
+			fatal(err)
+		}
+		root = "."
+	} else if root == "" {
 		root = dst
 	}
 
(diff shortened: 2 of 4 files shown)


cgwalters:
See https://github.com/containers/bootc/pull/731 particularly https://github.com/containers/bootc/pull/724/files#diff-30bb05ef4e35724331b84e50edacf12aaf7eb21e3a5576a341d556accb0d1400R75 which is doing a complex dance right now to bridge from our internal fds to the absolute paths this package currently requires.

cgwalters:
Fixed up the cross-compilation issues.

cgwalters:
Actually...is this macos/windows code ever used?

pkg/chrootarchive/archive.go
```diff
@@ -82,6 +82,17 @@ func untarHandler(tarArchive io.Reader, dest string, options *archive.TarOptions
...
+	if err != nil {
+		return err
+	}
+	if relDest == "." {
+		relDest = "/"
+	}
+	if relDest[0] != '/' {
+		relDest = "/" + relDest
+	}
+	dest = relDest
```
mtrmac, line 95: I didn’t try this and I might well be missing something, but how does this work on non-linux where `root`/`rootfd` are completely ignored?

rhatdan: I don't think we should continue the charade of supporting containers/storage on non linux/freeBSD machines.

mtrmac: I vaguely remember the tar/untar functions being used for non-graph-driver content as well. E.g. Podman’s `pkg/machine/ocipull/oci.go`.

mtrmac: … and if we _were_ to drop support for other systems, I think the code would _have_ to be modified to fail to compile, rather than being allowed to write to unexpected destinations by neglect.

cgwalters: > I didn’t try this and I might well be missing something, but how does this work on non-Linux where root/rootfd are completely ignored?

It doesn't.

> I vaguely remember the tar/untar functions being used for non-graph-driver content as well. E.g. Podman’s pkg/machine/ocipull/oci.go.

Hmm, good point, but in glancing slightly more deeply it seems like that...ah yes, notice the `archive` vs `chrootarchive`, so it's not using this code I think.

Of course, crossing with other threads you know of, I don't see why the podman-machine disk images are actually stored as container images (including being wrapped in a tarball) that we manually unpack vs directly as an OCI artifact which we fetch and stream directly to disk without an extra copy, which would be inherently more secure too. (Of course, if the podman machine archive today is malicious, you're kind of boned anyways because that machine ends up with write access to your home directory)

cgwalters: > on non linux/freeBSD

Of course already needing to abstract over those two is pretty nontrivial...Linux has `openat2` but FreeBSD ended up doing things differently with `O_RESOLVE_BENEATH` etc. The current zstd:chunked code is hardcoded to Linux.

Since it's been at least a whole day since I've mentioned Rust here...in bootc/ostree we've been using https://github.com/bytecodealliance/cap-std/ and it's quite nice and elegant and well designed, and has [backends for both linux and freebsd](https://github.com/bytecodealliance/cap-std/tree/main/cap-primitives/src/rustix). In theory we could fork off a helper process using that to handle untar...it even has a whole "userspace fs resolver fallback" path that is what's used today on MacOS.

cgwalters:
This triggered a memory for me that we also ship `oc image extract`...I went to look at what it does, and [it forked from moby https://github.com/openshift/oc/blob/bd7e59668c050e59b14e81cf5be00fad17f2f669/pkg/cli/image/archive/archive.go#L93].

I am not sure if `oc image extract` is currently load bearing on windows/macos? [Not much happening there](https://github.com/openshift/oc/commits/bd7e59668c050e59b14e81cf5be00fad17f2f669/pkg/cli/image/archive) but it looks like it compiles...

cgwalters:
Tossed up https://github.com/containers/storage/pull/2051

pkg/chrootarchive/archive.go
mtrmac: > > I vaguely remember the tar/untar functions being used for non-graph-driver content as well. E.g. Podman’s pkg/machine/ocipull/oci.go.
> 
> Hmm, good point, but in glancing slightly more deeply it seems like that...ah yes, notice the `archive` vs `chrootarchive`, so it's not using this code I think.

Thanks, that’s an important nuance (while we do need to _partially_ continue to support c/storage on the other systems).

---

A longer comment in https://github.com/containers/storage/pull/2051#issuecomment-2260940574 , but basically I’m leaning towards the though that a Linux-only code path is not an improvement.

The untag code is _supposed_ to be secure in any environment; AFAICT the chroot is an _additional_ isolation, not something the callers specifically require because they depend on some chroot-only property.

> In theory we could fork off a helper process using that to handle untar...

The way `pkg/reexec` works here, there are no external helper binaries. In the post-RPM world of people just copying binaries around, that makes deployment easier — probably not for Podman and all its dependencies, but for some of the other lighter-weight callers.

mtrmac: > I don't see why the podman-machine disk images are actually stored as container images

At least for the `pkg/machine/compression` code, that’s not necessarily the case — it’s a very generic decompressor handling things like ZIP, not compatible with traditional container images anyway.

mtrmac (changes requested):
(Just to have a tool-visible marker that the `dest = relDest` move is breaking non-Linux systems; let’s continue the discussion in the original thread.)

cgwalters:
OK I reworked this in a way which works on Linux (tested) and FreeBSD (not tested, but in theory), and leaves other platforms unchanged (not tested).

pkg/chrootarchive/archive_unix.go
```diff
@@ -72,24 +108,12 @@ func invokeUnpack(decompressedArchive io.Reader, dest string, options *archive.T
...
-		dest = relDest
-	}
-
-	cmd := reexec.Command("storage-untar", dest, root)
+	cmd := reexec.Command("storage-untar", dest, "/proc/self/fd/4")
 	cmd.Stdin = decompressedArchive
 
-	cmd.ExtraFiles = append(cmd.ExtraFiles, r)
+	cmd.ExtraFiles = append(cmd.ExtraFiles, r) // fd 3
+	// If you change this, change rootFileDescriptor above too
```
mtrmac, line 115: This is nice, but then `"/proc/self/fd/4"` is independently hard-coded in two places. Make that a constant as well

(Alternatively, we could turn `procPathForFd` into a repo-wide internal helper somewhere. That seems out of scope for this PR. Or, maybe, we could just drop the parameter and hard-code the behavior, but that assumes the caller and the reexec destination can _never_ get out of sync… probably a bit risky for CRI-O.)

pkg/chrootarchive/archive.go
```diff
@@ -83,6 +83,11 @@ func untarHandler(tarArchive io.Reader, dest string, options *archive.TarOptions
 		}
 	}
 
+	rootFd, dest, err := splitPathWithRoot(root, dest)
```
mtrmac, line 86: rootFD is leaked on error below. (Maybe hide that in a per-platform struct, see elsewhere; and `defer dest.close()` immediately after this succeeds?)

pkg/chrootarchive/archive_unix.go
```diff
@@ -38,7 +43,17 @@ func untar() {
 		root = flag.Arg(1)
 	}
 
-	if root == "" {
+	// FreeBSD doesn't have proc/self, but we can handle it here
+	if root == "/proc/self/fd/4" {
+		// Take ownership to ensure it's closed; no need to leak
+		// this afterwards.
```
mtrmac, line 49: (FWIW this is running a a dedicated process anyway, but I do prefer closing things explicitly, in case the code were refactored / reused in other contexts.)

mtrmac (commented):
ACK to the principle of the thing.

Consider structuring this as a per-platform data type:
```go
type unpackDestination struct { /* platform-specific */ }
func newUnpackDestination(root, dest string) unpackDestination { … }
func (dest unpackDestination) invokeUnpack(decompressedArchive …, options …)
```

That would hide the platform-specific `rootFd` in the Linux code, entirely avoid the unused parameter on non-Linux, and make it easier to close the file descriptor in the caller.

(Arguably, when I added the `_ = ` warning silencers, I probably should have done that already; I was too focused on the warnings to actually think about the code. The addition of a file descriptor makes the case even stronger.)

pkg/chrootarchive/archive.go
cgwalters: > rootFD is leaked on error 

Yes :cry: Well spotted...I try to remember this but [it's hard to context switch between Rust and Go (and C)...](https://blog.skylight.io/rust-means-never-having-to-close-a-socket/)

cgwalters:
A note: If we take this out farther, what would make sense is for c/storage to hold a fd (not an absolute path) for its storage root, then this code would need more refactoring as `dest` wouldn't necessarily be prepended with the `root` by default. That'd probably be a distinct new API calling into the shared bits, we just wouldn't need to do the "strip off the root" on the Unix paths in that case.

pkg/chrootarchive/archive.go
```diff
@@ -83,6 +83,18 @@ func untarHandler(tarArchive io.Reader, dest string, options *archive.TarOptions
 		}
 	}
 
+	destVal, err := newUnpackDestination(root, dest)
+	if err != nil {
+		return err
+	}
+	closeDestVal := true
+	defer func() {
```
mtrmac, line 91: Passing the FD to `Exec` does not close it AFAICS, so this should be closed unconditionally.

mtrmac (commented):
LGTM otherwise.

pkg/chrootarchive/archive.go
cgwalters: Right, thanks!

rhatdan:
@giuseppe @nalind PTAL

pkg/chrootarchive/archive_unix.go
```diff
@@ -57,11 +92,35 @@ func untar() {
 	os.Exit(0)
 }
 
-func invokeUnpack(decompressedArchive io.Reader, dest string, options *archive.TarOptions, root string) error {
+// splitPathWithRoot takes a root directory and a destination, and
+// returns an open file descriptor for the given root and the new
+// destination path.
```
mtrmac, line 97: (This comment is now out of date.)

pkg/chrootarchive/archive_unix.go
```diff
@@ -57,11 +92,35 @@ func untar() {
...
+	if relDest == "." {
+		relDest = "/"
+	}
+	if relDest[0] != '/' {
+		relDest = "/" + relDest
+	}
+
+	rootfdRaw, err := unix.Open(root, unix.O_RDONLY|unix.O_DIRECTORY|unix.O_CLOEXEC, 0)
+	if err != nil {
+		return nil, err
```
giuseppe, line 115: can we wrap this in a `fs.PathError`?  `&fs.PathError{Op: "open", Path: name, Err: err}`

giuseppe (approved):
a nit, otherwise LGTM

PR #2050, opened 2024-07-30
composefs: prefer std library functions

giuseppe:
a couple of cleanups for the composefs backend

### DIFF
--- a/drivers/overlay/composefs.go
+++ b/drivers/overlay/composefs.go
@@ -146,9 +146,9 @@ func mountComposefsBlob(dataDir, mountPoint string) error {
 	if err != nil {
 		return err
 	}
-	mountOpts := "ro"
+	mountOpts := ""
 	if !hasACL {
-		mountOpts += ",noacl"
+		mountOpts += "noacl"
 	}
 
 	if err := unix.Mount(loop.Name(), mountPoint, "erofs", unix.MS_RDONLY, mountOpts); err != nil {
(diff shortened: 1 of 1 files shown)


giuseppe:
we use "ro" in composefs as well: https://github.com/containers/composefs/blob/main/libcomposefs/lcfs-mount.c#L582-L585, but MS_RDONLY should be enough

giuseppe:
opened a PR for composefs as well: https://github.com/containers/composefs/pull/309

@cgwalters PTAL

drivers/overlay/composefs.go
```diff
@@ -114,23 +110,30 @@ struct lcfs_erofs_header_s {
...
-		return false, &fs.PathError{Op: "openat", Path: path, Err: err}
+		return false, err
 	}
-	defer unix.Close(fd)
+	defer file.Close()
+
 	// do not worry about checking the magic number, if the file is invalid
 	// we will fail to mount it anyway
-	flags := make([]byte, 4)
-	nread, err := unix.Pread(fd, flags, 8)
```
cgwalters, line 127: I don't really care too much, but I would note that I think we could use https://pkg.go.dev/os#File.ReadAt here instead

cgwalters (approved):

mtrmac (commented):
LGTM (I didn’t look into the `MS_RDONLY` vs `"ro"` difference, if any).

PR #2054, opened 2024-08-04
docs: Add (not installed) man pages for composefs and zstd:chunked

cgwalters:
These aren't installed yet so they'll just live in the git repository but hopefully we can expand them and use them as a central reference point for "how do I do (zstd:chunked)|(composefs)". We have copies of that kind of instruction in other downstream places but I think here makes the most sense.


rhatdan:
LGTM
@giuseppe PTAL

giuseppe (approved):
LGTM

PR #2056, opened 2024-08-06
loopback: always return error if no loopback device

giuseppe:
simplify the error handling for openNextAvailableLoopback and avoid named return values

### DIFF
--- a/pkg/loopback/attach_loopback.go
+++ b/pkg/loopback/attach_loopback.go
@@ -41,11 +41,10 @@ func getNextFreeLoopbackIndex() (int, error) {
 	return index, err
 }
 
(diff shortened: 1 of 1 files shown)


rhatdan (approved):
LGTM

rhatdan:
@mtrmac @nalind @saschagrunert PTAL

mtrmac (commented):
LGTM, I have learned to strongly distrust named return values and I like this.

Absolutely non-blocking: consider turning the top-level loop into a canonical `for remaining := …; remaining > 0; remaining--` instead of a notionally-infinite one.

saschagrunert (approved):

PR #2057, opened 2024-08-06
composefs use new mount api

giuseppe:
the main advantage in this case is that it splits the mount operation in multiple steps, so it is easier (hopefully) to find out what goes wrong.

### DIFF
--- a/drivers/overlay/composefs.go
+++ b/drivers/overlay/composefs.go
@@ -149,13 +149,39 @@ func mountComposefsBlob(dataDir, mountPoint string) error {
 	if err != nil {
 		return err
 	}
-	mountOpts := ""
-	if !hasACL {
-		mountOpts += "noacl"
+
+	fsfd, err := unix.Fsopen("erofs", 0)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


drivers/overlay/composefs.go
```diff
@@ -154,8 +154,38 @@ func mountComposefsBlob(dataDir, mountPoint string) error {
 		mountOpts += "noacl"
 	}
 
-	if err := unix.Mount(loop.Name(), mountPoint, "erofs", unix.MS_RDONLY, mountOpts); err != nil {
-		return fmt.Errorf("failed to mount erofs image at %q: %w", mountPoint, err)
+	fsfd, err := unix.Fsopen("erofs", 0)
```
alexlarsson, line 157: How old is the new mount api? Don't we want to fall back to mount on ENOSYS?

giuseppe: I didn't care about the fallback as old kernels do not support anyway the features we need in overlay (and possibly erofs itself)

alexlarsson (approved):

PR #2061, opened 2024-08-14
Create `Store.List` method for listing layers, images, containers

Honny1:
This PR implements a method for listing layers, images and containers. Method `Store.List` requires parameter type `ListOptions` that specifies the content of the return value. The return value is the structure `ListResult` that contains slices of layers, images or containers according to given options. The `List` method uses a locking mechanism so that the result is the current snapshot of the items listed. 

Fixes: 
- https://github.com/containers/podman/issues/23331

### DIFF
--- a/store.go
+++ b/store.go
@@ -84,6 +84,20 @@ type ApplyStagedLayerOptions struct {
 	DiffOptions *drivers.ApplyDiffWithDifferOpts // Mandatory
 }
 
+// MultiListOptions contains options to pass to MultiList
+type MultiListOptions struct {
+	Images     bool // if true, Images will be listed in the result
+	Layers     bool // if true, layers will be listed in the result
+	Containers bool // if true, containers will be listed in the result
+}
+
+// MultiListResult contains slices of Images, Layers or Containers listed by MultiList method
+type MultiListResult struct {
+	Images     []Image
+	Layers     []Layer
+	Containers []Container
+}
+
 // An roBigDataStore wraps up the read-only big-data related methods of the
 // various types of file-based lookaside stores that we implement.
 type roBigDataStore interface {
@@ -561,6 +575,12 @@ type Store interface {
 	// usually by deleting layers and images which are damaged.  If the
 	// right options are set, it will remove containers as well.
 	Repair(report CheckReport, options *RepairOptions) []error
+
+	// MultiList returns a MultiListResult structure that contains layer, image, or container
+	// extracts according to the values in MultiListOptions.
+	// MultiList returns consistent values as of a single point in time.
+	// WARNING: The values may already be out of date by the time they are returned to the caller.
+	MultiList(MultiListOptions) (MultiListResult, error)
 }
 
 // AdditionalLayer represents a layer that is contained in the additional layer store
@@ -3814,4 +3834,56 @@ func (s *store) GarbageCollect() error {
 	}
 
 	return firstErr
+}
+
+// List returns a MultiListResult structure that contains layer, image, or container
+// extracts according to the values in MultiListOptions.
+func (s *store) MultiList(options MultiListOptions) (MultiListResult, error) {
+	// TODO: Possible optimization: Deduplicate content from multiple stores.
+	out := MultiListResult{}
+
+	if options.Layers {
+		layerStores, err := s.allLayerStores()
+		if err != nil {
+			return MultiListResult{}, err
+		}
+		for _, roStore := range layerStores {
+			if err := roStore.startReading(); err != nil {
+				return MultiListResult{}, err
+			}
+			defer roStore.stopReading()
+			layers, err := roStore.Layers()
+			if err != nil {
+				return MultiListResult{}, err
+			}
+			out.Layers = append(out.Layers, layers...)
+		}
+	}
+
+	if options.Images {
+		for _, roStore := range s.allImageStores() {
+			if err := roStore.startReading(); err != nil {
+				return MultiListResult{}, err
+			}
+			defer roStore.stopReading()
+
+			images, err := roStore.Images()
+			if err != nil {
+				return MultiListResult{}, err
+			}
+			out.Images = append(out.Images, images...)
+		}
+	}
+
+	if options.Containers {
+		containers, _, err := readContainerStore(s, func() ([]Container, bool, error) {
+			res, err := s.containerStore.Containers()
+			return res, true, err
+		})
+		if err != nil {
+			return MultiListResult{}, err
+		}
+		out.Containers = append(out.Containers, containers...)
+	}
+	return out, nil
 }
--- a/store_test.go
+++ b/store_test.go
@@ -12,28 +12,47 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestStore(t *testing.T) {
+func newTestStore(t *testing.T, testOptions StoreOptions) Store {
 	wd := t.TempDir()
 
-	pullOpts := map[string]string{"Test1": "test1", "Test2": "test2"}
-	store, err := GetStore(StoreOptions{
-		RunRoot:            filepath.Join(wd, "run"),
-		GraphRoot:          filepath.Join(wd, "root"),
-		GraphDriverName:    "vfs",
-		GraphDriverOptions: []string{},
-		UIDMap: []idtools.IDMap{{
+	options := testOptions
+	if testOptions.RunRoot == "" {
+		options.RunRoot = filepath.Join(wd, "run")
+	}
+	if testOptions.GraphRoot == "" {
+		options.GraphRoot = filepath.Join(wd, "root")
+	}
+	if testOptions.GraphDriverName == "" {
+		options.GraphDriverName = "vfs"
+	}
+	if testOptions.GraphDriverOptions == nil {
+		options.GraphDriverOptions = []string{}
+	}
+	if len(testOptions.UIDMap) == 0 {
+		options.UIDMap = []idtools.IDMap{{
 			ContainerID: 0,
 			HostID:      os.Getuid(),
 			Size:        1,
-		}},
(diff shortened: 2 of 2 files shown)


store.go
```diff
@@ -561,6 +575,12 @@ type Store interface {
 	// usually by deleting layers and images which are damaged.  If the
 	// right options are set, it will remove containers as well.
 	Repair(report CheckReport, options *RepairOptions) []error
+
+	// List returns a ListResult structure that contains layer, image, or container
+	// extracts according to the values in ListOptions.
+	// List method uses a locking mechanism so that the result is the current snapshot
+	// of the items listed.
```
mtrmac, line 582: Callers don’t care that this is implemented with locks (and we don’t want to promise that), just that the values are consistent.

store.go
```diff
@@ -3815,3 +3835,55 @@ func (s *store) GarbageCollect() error {
 
 	return firstErr
 }
+
+// List returns a ListResult structure that contains layer, image, or container
+// extracts according to the values in ListOptions.
+func (s *store) List(options ListOptions) (ListResult, error) {
```
mtrmac, line 3841: Throughout: If the relevant option is not set, it’s not necessary to lock the store.

That’s particularly important for the container store if we are not going to use it at all in the initial caller.

store_test.go
```diff
@@ -485,3 +486,107 @@ func TestWithSplitStore(t *testing.T) {
 	store.Free()
 	store.Free()
 }
+
+func TestStoreList(t *testing.T) {
+	reexec.Init()
+	wd := t.TempDir()
+
+	pullOpts := map[string]string{"Test1": "test1", "Test2": "test2"}
+	store, err := GetStore(StoreOptions{
```
mtrmac, line 495: To the extent this is non-trivial, it should be a shared utility, not copied in each test.

Maybe a `newTestStore(*testing.T, StoreOptions)` where `StoreOptions` contains only the non-obvious specific items (like `PullOptions` or `ImageStore` in this file), and the `newTestStore` creates options with the other relevant values.

store_test.go
```diff
@@ -485,3 +486,107 @@ func TestWithSplitStore(t *testing.T) {
 	store.Free()
 	store.Free()
 }
+
+func TestStoreList(t *testing.T) {
+	reexec.Init()
+	wd := t.TempDir()
+
+	pullOpts := map[string]string{"Test1": "test1", "Test2": "test2"}
```
mtrmac, line 494: This is unnecessary for this test.

store_test.go
```diff
@@ -485,3 +486,107 @@ func TestWithSplitStore(t *testing.T) {
...
+		PullOptions: pullOpts,
+	})
+	require.NoError(t, err)
+
+	root := store.RunRoot()
+	require.NotNil(t, root)
+
+	_, err = store.GraphDriver()
+	require.Nil(t, err)
+
```
mtrmac, line 519: This is unnecessary for this test.

store_test.go
```diff
@@ -485,3 +486,107 @@ func TestWithSplitStore(t *testing.T) {
...
+			containerCounts: 1,
+		},
+	}
+
+	for _, test := range tests {
+		listResults, err := store.List(test.options)
+		require.NoError(t, err)
+		require.Equal(t, test.layerCounts, len(listResults.Layers))
+		require.Equal(t, test.imageCounts, len(listResults.Images))
+		require.Equal(t, test.containerCounts, len(listResults.Containers))
```
mtrmac, line 585: Non-blocking, this works fine:
- `require` for things that should fail the entire execution; `assert` for failures that allow continuing.
- `assert.Len()` exists

mtrmac (commented):
Thanks!

The approach looks good overall.

store.go
```diff
@@ -84,6 +84,20 @@ type ApplyStagedLayerOptions struct {
 	DiffOptions *drivers.ApplyDiffWithDifferOpts // Mandatory
 }
 
+// StoreListOptions contains options to pass to MultiList
+type StoreListOptions struct {
```
mtrmac, line 88: If it is `MultiList`, then `MultiListOptions` (or `StoreMultiListOptions`, but that’s getting unergonomic).

Similarly for the result type — and that one is typically not going to be named by callers, so a long name doesn’t hurt.

store.go
```diff
@@ -561,6 +575,11 @@ type Store interface {
 	// usually by deleting layers and images which are damaged.  If the
 	// right options are set, it will remove containers as well.
 	Repair(report CheckReport, options *RepairOptions) []error
+
+	// MultiList returns a StoreListResult structure that contains layer, image, or container
+	// extracts according to the values in StoreListOptions.
+	// MultiList returns consistet values.
```
mtrmac, line 581: On another look,
```suggestion
	// MultiList returns consistent values as of a single point in time.
	// WARNING: The values may already be out of date by the time they are returned to the caller.
```

store.go
```diff
@@ -3815,3 +3834,56 @@ func (s *store) GarbageCollect() error {
...
 	return firstErr
 }
+
+// List returns a ListResult structure that contains layer, image, or container
+// extracts according to the values in ListOptions.
+func (s *store) MultiList(options StoreListOptions) (StoreListResult, error) {
+	// TODO: Possible optimization: Deduplicate content from multiple stores.
+	out := StoreListResult{}
+
+	layerStores, err := s.allLayerStores()
```
mtrmac, line 3844: This can also be moved inside `if options.Layers`.

store.go
```diff
@@ -3815,3 +3834,56 @@ func (s *store) GarbageCollect() error {
...
+		}
+	}
+
+	if options.Containers {
+		containers, _, err := readContainerStore(s, func() ([]Container, bool, error) {
+			res, err := s.containerStore.Containers()
+			return res, true, err
+		})
+		if err != nil {
+			return out, err
```
mtrmac, line 3884: Throughout: On error paths, it’s better to explicitly return a `…Result{}`, so that callers don’t get any ideas about using the value anyway.

store_test.go
```diff
@@ -6,16 +6,23 @@ import (
...
 
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/reexec"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+func newTestStore(t *testing.T, options StoreOptions) Store {
+	store, err := GetStore(options)
```
mtrmac, line 16: Creating this helper would primarily be beneficial if most of the `StoreOptions` (set up to make the tests runnable unprivileged on macOS, and the like) were set here. Deduplicating 2 lines but still having 15 lines copy&pasted is not that much of an improvement; removing 15 lines, and having the individual tests ~not care about how to deal with VFS and ID mapping, would be more useful.

I.e.
```go
   store := newTestStore(t, StoreOptions{PullOptions: pullOpts}) // in the caller which tests pullOpts

func newTestStore(… testOptions) {
   options := testOptions // A shallow copy
   options.RunRoot = …
   … // 15 more lines
   … GetStore…
}
```

store_test.go
```diff
@@ -6,16 +6,23 @@ import (
...
+	require.NoError(t, err)
+	return store
+}
+
 func TestStore(t *testing.T) {
 	wd := t.TempDir()
 
 	pullOpts := map[string]string{"Test1": "test1", "Test2": "test2"}
-	store, err := GetStore(StoreOptions{
+	storeOptions := StoreOptions{
```
mtrmac, line 25: (A personal style nit: Inlining a single-purpose options structure directly seems a tiny bit cleaner to me than adding a new single-use variable. It’s more similar to keyword parameters in languages like Python, and the reader does not have to look for any other uses of the variable. Opinions do differ on this.)

store_test.go
```diff
@@ -485,3 +492,99 @@ func TestWithSplitStore(t *testing.T) {
 	store.Free()
 	store.Free()
 }
+
+func TestStoreList(t *testing.T) {
```
mtrmac, line 496: `TestStoreMultiList` now.

store.go
```diff
@@ -3815,3 +3835,55 @@ func (s *store) GarbageCollect() error {
 
 	return firstErr
 }
+
+// List returns a MultiListResult structure that contains layer, image, or container
+// extracts according to the values in ListOptions.
```
mtrmac, line 3840: `… in options`, or `… in MultiListOptions` now.

store_test.go
```diff
@@ -6,33 +6,51 @@ import (
...
 
-	pullOpts := map[string]string{"Test1": "test1", "Test2": "test2"}
-	store, err := GetStore(StoreOptions{
-		RunRoot:            filepath.Join(wd, "run"),
-		GraphRoot:          filepath.Join(wd, "root"),
-		GraphDriverName:    "vfs",
-		GraphDriverOptions: []string{},
-		UIDMap: []idtools.IDMap{{
+	options := testOptions
+	if testOptions.RunRoot == "" {
```
mtrmac, line 19: (I’m not sure the conditionals are necessary but they don’t hurt.)

store_test.go
```diff
@@ -485,3 +489,84 @@ func TestWithSplitStore(t *testing.T) {
 	store.Free()
 	store.Free()
 }
+
+func TestStoreMultiList(t *testing.T) {
+	reexec.Init()
+
+	store := newTestStore(t, StoreOptions{
+		GraphDriverName: "vfs",
```
mtrmac, line 497: Using the `vfs` driver is also common to the configuration used to run tests, and should be shared.

mtrmac (commented):
LGTM.

I suppose we should wait with merging this until we know that this is sufficient to fix the user; that’s https://github.com/containers/common/pull/2125 I think.

rhatdan:
/approve
/lgtm
Seems like decent functions to have.

PR #2065, opened 2024-08-17
Add doc comments around "string bool" usage

cgwalters:
I've been poking at composefs related things here and stumbled across the very confusing fact that some values in the storage config are booleans - but must be
*strings* even though TOML supports native boolean values.

Further, digging in more we actually use two different parsers for boolean values; one silently accepts anything that is not the string `"true"` to mean `false`, but the other uses the Go strconv function and does return an error.

I'm not trying to fix any of this yet; the clear fix is to widen what we accept into a native TOML boolean type *in addition* which would require a bit of custom deserialization. Let's just document this to start.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,6 +20,8 @@ type BtrfsOptionsConfig struct {
 type OverlayOptionsConfig struct {
 	// IgnoreChownErrors is a flag for whether chown errors should be
 	// ignored when building an image.
+	// This can take any value accepted by Go's https://pkg.go.dev/strconv#ParseBool
+	// Search the source code for the term "string bool" to find other places which behave similarly.
 	IgnoreChownErrors string `toml:"ignore_chown_errors,omitempty"`
 	// MountOpt specifies extra mount options used when mounting
 	MountOpt string `toml:"mountopt,omitempty"`
@@ -30,8 +32,12 @@ type OverlayOptionsConfig struct {
 	// Inodes is used to set a maximum inodes of the container image.
 	Inodes string `toml:"inodes,omitempty"`
 	// Do not create a bind mount on the storage home
+	// This can take any value accepted by Go's https://pkg.go.dev/strconv#ParseBool
+	// Search the source code for the term "string bool" to find other places which behave similarly.
 	SkipMountHome string `toml:"skip_mount_home,omitempty"`
-	// Specify whether composefs must be used to mount the data layers
+	// Specify whether composefs must be used to mount the data layers.
+	// This can take any value accepted by Go's https://pkg.go.dev/strconv#ParseBool
+	// Search the source code for the term "string bool" to find other places which behave similarly.
 	UseComposefs string `toml:"use_composefs,omitempty"`
 	// ForceMask indicates the permissions mask (e.g. "0755") to use for new
 	// files and directories
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -973,6 +973,14 @@ type hardLinkToCreate struct {
 	metadata *fileMetadata
 }
 
+// parseBooleanPullOption coerces a string value we deserialized from toml into a bool value.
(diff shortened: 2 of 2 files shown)


rhatdan:
/approve
LGTM

rhatdan:
@giuseppe @nalind @mtrmac @saschagrunert @Honny1 PTAL

Honny1 (commented):
LGTM, I don't know how important it is to fix this, or if the comments are enough. At least I would create the issue.

pkg/config/config.go
```diff
@@ -20,6 +20,8 @@ type BtrfsOptionsConfig struct {
 type OverlayOptionsConfig struct {
 	// IgnoreChownErrors is a flag for whether chown errors should be
 	// ignored when building an image.
+	// This can take any value accepted by Go's https://pkg.go.dev/strconv#ParseBool
+	// Search the source code for the term "string bool" to find other places which behave similarly.
```
mtrmac, line 24: For all three comments in this file… Referring to `ParseBool` is a correct description of the current readers. But, well, do we _want_ to commit to accepting things like `[01tTfF]`? (Think possible Rust consumers of the config.)

I’d rather document just `"true"`/`"false"`, and treat the behavior of `drivers/overlay`, accepting the other values, as an undocumented… (non?)bug when processing invalid input.

Especially when, right now, the `docs/containers-storage.conf.5.md` man page does not document the values _at all_, adding the `ParseBool` references here could be construed as a commitment to support the values in the config file. I’d rather not do that.

---

(Ceterum autem censeo, the `c/storage/drivers` subpackage should almost entirely become private; and the option parsing should pass these Go structures down to `drivers/overlay` directly instead of round-tripping through `[]string`. But also, I’m scared of the c/storage option parsing / loading code.)

mtrmac (commented):
There’s a fine line between warning readers about unusual situations, and reaffirming known-incorrect behavior as desirable … I think this PR is closer to the latter than I’d like. 

If this is worth commenting, I think it’s worth commenting what we, as of today, think about the situation — not just commenting what the code does (because that can be read in the code).

---

The most important thing is _users_. For that purpose, I think the “string bool” options should be explicitly documented as such in `docs/containers-storage.conf.5.md` . If we document that only `"true"`/`"false"` are valid values, that implicitly constraints the Go implementation:
- The values directly in `OverlayOptionsConfig` don’t _need_ comments added; they can have a comment pointing out the unusual semantics in the man page, and/or a "to do: also accept native TOML booleans".
- The implementation in drivers/overlay might get a comment that we accept undocumented values, and that should ideally be fixed.

cgwalters:
> For that purpose, I think the “string bool” options should be explicitly documented as such in docs/containers-storage.conf.5.md

That was already done in https://github.com/containers/storage/pull/2066

mtrmac:
> > For that purpose, I think the “string bool” options should be explicitly documented as such in docs/containers-storage.conf.5.md
> 
> That was already done in #2066

That PR does not touch the authoritative-named containers-storage.conf(5) man page  AFAICS.

cgwalters:
> That PR does not touch the authoritative-named containers-storage.conf(5) man page AFAICS.

You're right, sorry. Hmm...probably we should try to *remove* some of the text from the config file in favor of referencing the man page, but for now I'll just keep them in sync.

https://github.com/containers/storage/pull/2076

> But, well, do we want to commit to accepting things like [01tTfF]? (Think possible Rust consumers of the config.)

My worry is that there are some people that accidentally set values to one of those. It doesn't seem impossible at all. But, I'm only "kind of" a developer here so if others feel we should pull off the bandaid and just e.g. error out on non-booleans I'm good with that too!

> There’s a fine line between warning readers about unusual situations, and reaffirming known-incorrect behavior as desirable … I think this PR is closer to the latter than I’d like.

OK, I understand. Closing then in favor of just updating the man page and then we can circle back later and actually a fix to at least support native TOML booleans.

PR #2066, opened 2024-08-17
storage.conf: Various cleanups

cgwalters:
The default storage.conf we ship is inconsistent in a few ways; there's a lot of fixes rolled up into this single commit.

First: we were using a toml "inline table" for the pull options, and then documenting each key in that table in one blob above. It simply looks much nicer to use a non-inline table - then we can move the docs next to each individual value. This is also more consistent with other sections of the config.

I also thinned out a bit the doc comments; I think instead of trying to have a longer explanation of zstd:chunked in the comments here we should refer to the man page, which is a better place to have details (and that we should fill out more).

Per another PR, I also stumbled across the fact that we have a lot of "string bool" values and cannot be native TOML booleans. Document that clearly next to each type.

We already have default values in the *code* for all of these, so comment them all out to be consistent with other values. (We're then getting closer to having the config file be entirely comments, but that's a distinct project)

Finally, update the recent man pages I added to match these changes.

### DIFF
--- a/docs/containers-storage-composefs.md
+++ b/docs/containers-storage-composefs.md
@@ -3,7 +3,6 @@
 ## NAME
 containers-storage-composefs - Information about composefs and containers/storage
 
-
 ## DESCRIPTION
 
 To enable composefs at a baseline requires the following configuration in `containers-storage.conf`:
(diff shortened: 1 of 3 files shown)


rhatdan:
Can we support both native and string at the same time?

LGTM

@giuseppe @nalind @mtrmac @flouthoc @saschagrunert PTAL

flouthoc (approved):
LGTM

flouthoc:
I agree that maybe both `native` and string format must be supported to make things easier.

cgwalters:
> Can we support both native and string at the same time?

Yes, I believe so, but we should look at that after this.

rhatdan:
Agreed.

PR #2069, opened 2024-08-20
composefs fixes

giuseppe:
some improvements for composefs, more details in each patch

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -2128,24 +2128,16 @@ func (d *Driver) DiffGetter(id string) (_ graphdriver.FileGetCloser, Err error) 
 	for _, diffDir := range diffDirs {
 		// diffDir has the form $GRAPH_ROOT/overlay/$ID/diff, so grab the $ID from the parent directory
 		id := path.Base(path.Dir(diffDir))
-		composefsBlob := d.getComposefsData(id)
-		if fileutils.Exists(composefsBlob) != nil {
+		composefsData := d.getComposefsData(id)
+		if fileutils.Exists(composefsData) != nil {
 			// not a composefs layer, ignore it
 			continue
 		}
-		dir, err := os.MkdirTemp(d.runhome, "composefs-mnt")
+		fd, err := openComposefsMount(composefsData)
 		if err != nil {
 			return nil, err
 		}
-		if err := mountComposefsBlob(composefsBlob, dir); err != nil {
-			return nil, err
-		}
-		fd, err := os.Open(dir)
-		if err != nil {
-			return nil, err
-		}
-		composefsMounts[diffDir] = fd
-		_ = unix.Unmount(dir, unix.MNT_DETACH)
+		composefsMounts[diffDir] = os.NewFile(uintptr(fd), composefsData)
 	}
 	return &overlayFileGetter{diffDirs: diffDirs, composefsMounts: composefsMounts}, nil
 }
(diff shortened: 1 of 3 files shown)


cgwalters:
Is this aiming to fix  https://github.com/containers/storage/issues/2042 ?

drivers/overlay/overlay.go
```diff
@@ -2128,24 +2128,16 @@ func (d *Driver) DiffGetter(id string) (_ graphdriver.FileGetCloser, Err error)
...
 		id := path.Base(path.Dir(diffDir))
-		composefsBlob := d.getComposefsData(id)
-		if fileutils.Exists(composefsBlob) != nil {
+		composefsData := d.getComposefsData(id)
+		if fileutils.Exists(composefsData) != nil {
 			// not a composefs layer, ignore it
 			continue
 		}
-		dir, err := os.MkdirTemp(d.runhome, "composefs-mnt")
+		fd, err := openComposefsMount(composefsData)
```
cgwalters, line 2136: > but access directly the files through the mount fd.

Huh...I didn't know one could do that...interesting and quite useful!

cgwalters (commented):
This is only a slightly-past-superficial level of review, but LGTM

drivers/overlay/overlay.go
cgwalters: > Huh...I didn't know one could do that...interesting and quite useful!

Reminder to myself and anyone interested in more: https://brauner.io/2023/02/28/mounting-into-mount-namespaces.html notes this:

> ... and with the added benefit that the mount never actually had to appear anywhere in the filesystem hierarchy and thus never had to belong to any mount namespace. This alone is already a very powerful tool but we won’t go into depth today.

giuseppe:
> Is this aiming to fix #2042 ?

Since it fails on `move_mount`, I feel it has something to do with mount propagations, so this PR might have an impact as we need less mounts, but I am not sure as I was never able to reproduce it locally yet.

drivers/overlay/overlay.go
```diff
@@ -1456,6 +1456,32 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
...
+	// user namespace requires this to move a directory from lower to upper.
+	rootUID, rootGID, err := idtools.GetRootUIDGID(options.UidMaps, options.GidMaps)
+	if err != nil {
+		return "", err
+	}
+
+	mergedDir := d.getMergedDir(id, dir, inAdditionalStore)
+	// Attempt to create the merged dir only if it doesn't exist.
+	if err := fileutils.Exists(mergedDir); err != nil && os.IsNotExist(err) {
+		if err := idtools.MkdirAllAs(mergedDir, 0o700, rootUID, rootGID); err != nil && !os.IsExist(err) {
```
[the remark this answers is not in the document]
giuseppe: I think we added the outer `fileutils.Exists` because `MkdirAllAs` chowns an existing directory, while in this case we don't want to do it when the merged directory exists (it might be in an additional store).

I'll change it to use `MkdirAllAndChownNew`

cgwalters:
> Since it fails on move_mount, I feel it has something to do with mount propagations, so this PR might have an impact as we need less mounts, but I am not sure as I was never able to reproduce it locally yet.

OK, makes sense - that's the kind of thing that I think would be good in one of the commit messages just for future reference for others to avoid needing to backref to the PR: say in https://github.com/containers/storage/pull/2069/commits/f6ca3171c1b7e0b9b34c6d6c005e8d185a5b7f44 or so.

But that's just something for future changes
/lgtm

PR #2076, opened 2024-08-27
Sync man page with storage.conf

cgwalters:
xref https://github.com/containers/storage/pull/2065#issuecomment-2313002081

This does the same for the man page that I did for the default storage.conf previously; instead of trying to squash all the `pull_options` into a single option that we document "ad-hoc", make it its own table just like the other sections.

- Add notes about "string bools" around relevant values.
- Avoid redundantly specifying the default; the default is the first thing we show after the `=`.

Synchronize the text between the man page and the default `storage.conf`.


mtrmac (commented):
/lgtm

Thanks!

PR #2078, opened 2024-08-30
Ignore rootless EPERM failures setting security xattrs

mheon:
This is specifically for the IMA xattrs, which cannot be set except as root; as rootless, they fail with EPERM. We never noticed this before (likely because IMA xattrs seem uncommon in the wild) but if there is a file in an image with an IMA xattr rootless Podman becomes completely unable to use the image. This is particularly relevant because the catatonit binary Podman uses for building its pause image has started to include an IMA xattr on Fedora Rawhide, which is breaking rootless Podman there rather badly.

Since this cannot work as rootless, it seems simplest to try to set the xattr, but tolerate failure iff the error is EPERM and we are not run as root.

Fixes: https://github.com/containers/podman/issues/18543

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -427,7 +427,7 @@ func readSecurityXattrToTarHeader(path string, hdr *tar.Header) error {
 	}
 	for _, xattr := range []string{"security.capability", "security.ima"} {
 		capability, err := system.Lgetxattr(path, xattr)
-		if err != nil && !errors.Is(err, system.EOPNOTSUPP) && err != system.ErrNotSupportedPlatform {
+		if err != nil && !errors.Is(err, system.EOPNOTSUPP) && err != system.ErrNotSupportedPlatform && !(unshare.IsRootless() && errors.Is(err, syscall.EPERM)) {
 			return fmt.Errorf("failed to read %q attribute from %q: %w", xattr, path, err)
 		}
 		if capability != nil {


mheon:
I don't know if we can reasonably test this in CI as any test run without root will be unable to create a file with the right xattr. I suppose upgrading the CI VMs to F41 once that comes out (or ensuring rpm-plugin-ima is installed in them now?) would allow testing.

pkg/archive/archive.go
```diff
@@ -427,7 +427,7 @@ func readSecurityXattrToTarHeader(path string, hdr *tar.Header) error {
 	}
 	for _, xattr := range []string{"security.capability", "security.ima"} {
 		capability, err := system.Lgetxattr(path, xattr)
-		if err != nil && !errors.Is(err, system.EOPNOTSUPP) && err != system.ErrNotSupportedPlatform {
+		if err != nil && !errors.Is(err, system.EOPNOTSUPP) && err != system.ErrNotSupportedPlatform && !(unshare.IsRootless() && errors.Is(err, syscall.EPERM)) {
```
mtrmac, line 430: If we should have the exception _here_, I think, it would be better to scope it to `security.ima` only; if we are asked to represent a set-capability executable and we can’t do that, I think the operation should fail.

mtrmac (commented):
Does this work end-to-end?

I didn’t deeply investigate the failure code path, and this _does_ match with the claim that https://github.com/containers/storage/pull/657 introduced the problem, still:

- The PR description clearly doesn’t match the actual operation: this ignores EPERM _reading_, not _setting_ attributes
- Reading https://github.com/containers/podman/issues/18543 , in there it seems that we _did_ succeed reading the attribute, at least enough to want to set it on a file, but failed setting it (`setting value of extended attribute` )

rhatdan:
@mheon Still working on this one?

mheon:
Yes, this is the wrong implementation, rewriting.

PR #2080, opened 2024-08-30
Be clearer about the layer store locking rules

mtrmac:
... as modified by https://github.com/containers/storage/pull/2036 .

Also document a LOCKING BUG I don't now care to fix.

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1878,10 +1878,21 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 
 // getMergedDir returns the directory path that should be used as the mount point for the overlayfs.
 func (d *Driver) getMergedDir(id, dir string, inAdditionalStore bool) string {
-	// If the layer is in an additional store, the lock we might hold only a reading lock.  To prevent
-	// races with other processes, use a private directory under the main store rundir.  At this point, the
-	// current process is holding an exclusive lock on the store, and since the rundir cannot be shared for
-	// different stores, it is safe to assume the current process has exclusive access to it.
+	// Ordinarily, .Get() (layer mounting) callers are supposed to guarantee exclusion.
+	//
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@nalind @saschagrunert @giuseppe @flouthoc PTAL

giuseppe (approved):
/lgtm

drivers/overlay/overlay.go
```diff
@@ -1878,10 +1878,21 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
...
+	// To prevent races with other processes mounting or unmounting the layer,
+	// use a private directory under the main store rundir, not the "merged" directory inside the
+	// original layer store holding the layer data.
+	//
+	// To support this, contrary to the _general_ locking rules for .Diff / .Changes (which allow a RO lock),
+	// the top-level Store implementation uses an exclusive lock for the primary layer store;
+	// and since the rundir cannot be shared for different stores, it is safe to assume the
+	// current process has exclusive access to it.
+	//
+	// LOCKING BUG? the .DiffSize operation does not currently hold an exclusive lock on the primary store.
```
flouthoc, line 1895: Is this yet to be verified ? I mean about the comment `LOCKING BUG?` if yes then maybe adding `TODO:` here would help someone else to revisit this in spare time.

mtrmac: I do think it is a locking bug.

It’s not 100% clear to me that it needs to be fixed _in entirety_ — c/image’s “image size” operation uses layer metadata instead of `store.ImageSize`. There are some callers that use this `.DiffSize` path, but maybe they could be migrated to the metadata-only implementation. (We would still need to either reimplement, or deprecate, this operation as is…)

But `.store.ContainerSize` does seem to be impacted, with no obvious alternative implementation.

mtrmac: Addressed in #2088. Thanks!

PR #2083, opened 2024-09-02
overlay: ignore EINVAL for cleanup unmount

giuseppe:
ignore EINVAL for unmount in the cleanup function as it means the directory is not a mount point.  It can happen if the cleanup happens before the mount point is created.


flouthoc (approved):
LGTM

PR #2084, opened 2024-09-04
Treat XDG_CONFIG_HOME and $HOME/.config the same way

rhatdan:
Currently we handle XDG_CONFIG_HOME and the fallback $HOME/.config differently. We create the later if it does not exist and verify that it is owned by the current user. This patch makes XDG_CONFIG_HOME follow the same pattern.

### DIFF
--- a/pkg/homedir/homedir.go
+++ b/pkg/homedir/homedir.go
@@ -6,11 +6,16 @@ import (
 	"path/filepath"
 )
 
-// GetDataHome returns XDG_DATA_HOME.
-// GetDataHome returns $HOME/.local/share and nil error if XDG_DATA_HOME is not set.
+// DataHome (deprecated)
+func GetDataHome() (string, error) {
+	return DataHome()
+}
+
+// DataHome returns XDG_DATA_HOME.
+// DataHome returns $HOME/.local/share and nil error if XDG_DATA_HOME is not set.
 //
 // See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
-func GetDataHome() (string, error) {
+func DataHome() (string, error) {
 	if xdgDataHome := os.Getenv("XDG_DATA_HOME"); xdgDataHome != "" {
 		return xdgDataHome, nil
 	}
@@ -21,11 +26,16 @@ func GetDataHome() (string, error) {
 	return filepath.Join(home, ".local", "share"), nil
 }
 
-// GetCacheHome returns XDG_CACHE_HOME.
-// GetCacheHome returns $HOME/.cache and nil error if XDG_CACHE_HOME is not set.
+// GetCacheHome (deprecated)
+func GetCacheHome() (string, error) {
+	return CacheHome()
+}
+
+// CacheHome returns XDG_CACHE_HOME.
+// CacheHome returns $HOME/.cache and nil error if XDG_CACHE_HOME is not set.
 //
 // See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
-func GetCacheHome() (string, error) {
+func CacheHome() (string, error) {
 	if xdgCacheHome := os.Getenv("XDG_CACHE_HOME"); xdgCacheHome != "" {
 		return xdgCacheHome, nil
 	}
@@ -34,4 +44,14 @@ func GetCacheHome() (string, error) {
 		return "", errors.New("could not get either XDG_CACHE_HOME or HOME")
 	}
 	return filepath.Join(home, ".cache"), nil
+}
+
+// GetRuntimeDir (deprecated)
+func GetRuntimeDir() (string, error) {
+	return RuntimeDir()
+}
+
+// GetShortcutString (deprecated)
+func GetShortcutString() string {
+	return ShortcutString()
 }
--- a/pkg/homedir/homedir_unix.go
+++ b/pkg/homedir/homedir_unix.go
@@ -38,9 +38,9 @@ func Get() string {
 	return homedir
 }
 
-// GetShortcutString returns the string that is shortcut to user's home directory
+// ShortcutString returns the string that is shortcut to user's home directory
 // in the native shell of the platform running on.
-func GetShortcutString() string {
+func ShortcutString() string {
 	return "~"
 }
 
@@ -52,7 +52,7 @@ func GetShortcutString() string {
 //
 // See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
 func StickRuntimeDirContents(files []string) ([]string, error) {
-	runtimeDir, err := GetRuntimeDir()
+	runtimeDir, err := RuntimeDir()
 	if err != nil {
 		// ignore error if runtimeDir is empty
 		return nil, nil //nolint: nilerr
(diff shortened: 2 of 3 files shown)


Luap99:
Can we first get an agreement what this should do? I still like to push hard that this function should not try to create and state the directory at all.

rhatdan:
I have no problem with that other then it is potentially a breaking change.

pkg/homedir/homedir_unix.go
```diff
@@ -101,9 +101,39 @@ func isWriteableOnlyByOwner(perm os.FileMode) bool {
 	return (perm & 0o722) == 0o700
 }
 
-// GetConfigHome returns XDG_CONFIG_HOME.
+// ConfigHome returns XDG_CONFIG_HOME.
+// ConfigHome returns $HOME/.config and nil error if XDG_CONFIG_HOME is not set.
+// Verifies ownership of config directory matches the current process or
+// returns error.
+// See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
```
mtrmac, line 108: (This link is now broken. Maybe https://specifications.freedesktop.org/basedir-spec/latest/index.html#variables ?)

pkg/homedir/homedir_unix.go
```diff
@@ -101,9 +101,39 @@ func isWriteableOnlyByOwner(perm os.FileMode) bool {
...
+func GetConfigHome() (string, error) {
```
mtrmac, line 109: This does not compile; was this intended to be `ConfigHome`?

pkg/homedir/homedir_unix.go
```diff
@@ -101,9 +101,39 @@ func isWriteableOnlyByOwner(perm os.FileMode) bool {
...
+	rootlessConfigHomeDirOnce.Do(func() {
```
mtrmac, line 110: If this is a new function with a new semantics, the `…{DirOnce,Dir,DirError}` variables must not be shared with the other function.

---

(`sync.OnceValues` now exists. Aesthetically I don’t think it’s all that great, but it does have one advantage of allowing the nested function to use an ordinary `return value, err` control flow.)

pkg/homedir/homedir_unix.go
```diff
@@ -101,9 +101,39 @@ func isWriteableOnlyByOwner(perm os.FileMode) bool {
...
+		tmpDir := os.Getenv("XDG_CONFIG_HOME")
```
mtrmac, line 111: (`tmpDir` is a surprising name for this value to me in this context, I’d read it as imposing a “temporary directory” semantics.

I’d expect “configDir” or something like “candidate”.)

pkg/homedir/homedir_unix.go
```diff
@@ -101,9 +101,39 @@ func isWriteableOnlyByOwner(perm os.FileMode) bool {
...
+			home := Get()
+			resolvedHome, err := filepath.EvalSymlinks(home)
+			if err != nil {
+				rootlessConfigHomeDirError = fmt.Errorf("cannot resolve %s: %w", home, err)
+				return
+			}
+			tmpDir = filepath.Join(resolvedHome, ".config")
+		}
+		st, err := os.Stat(tmpDir)
+		if err == nil {
```
mtrmac, line 122: If the dir is missing, this implementation sets (…HomeDir, HomeDirError) to (`""`, `nil`), which seems undesirable to me.

mtrmac (commented):
This is _presuming_ a consensus on something like https://github.com/containers/podman/issues/23818#issuecomment-2329502508 ; I’m not sure we have that yet.

pkg/homedir/homedir_unix.go
```diff
@@ -101,9 +101,39 @@ func isWriteableOnlyByOwner(perm os.FileMode) bool {
...
+				rootlessConfigHomeDirError = fmt.Errorf("path %q exists and it is not owned by the current user", tmpDir)
+				return
+			}
+		}
+	})
+
+	return rootlessConfigHomeDir, rootlessConfigHomeDirError
+}
+
+// GetConfigHome returns XDG_CONFIG_HOME. (Deprecated)
```
mtrmac, line 135: If we go with the plan of gradual migration per https://github.com/containers/podman/issues/23818#issuecomment-2329502508 , please
- point at the alternative
- explain why it is deprecated / what needs to change in callers
- consider using the standard `Deprecated: ` syntax that linters understand. OTOH that would force us to migrate immediately, I’m not 100% sure we want to sign up for doing that

pkg/homedir/homedir_unix.go
mtrmac: Still outstanding.

pkg/homedir/homedir_unix.go
```diff
@@ -101,11 +101,20 @@ func isWriteableOnlyByOwner(perm os.FileMode) bool {
...
 
-// GetConfigHome returns XDG_CONFIG_HOME.
-// GetConfigHome returns $HOME/.config and nil error if XDG_CONFIG_HOME is not set.
-//
-// See also https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html
+// GetConfigHome deprecated
 func GetConfigHome() (string, error) {
+	rootlessConfigHomeDir, rootlessConfigHomeDirError = ConfigHome()
+	if rootlessConfigHomeDirError == nil {
+		_ = os.MkdirAll(rootlessConfigHomeDir, 0o700)
```
mtrmac, line 108: If this function is deprecated and we intend to migrate all users, I don’t see that _changing_ the existing one to not report failures is useful: that forces us to immediately review/migrate all users, negating the benefit of the deprecation and introduction of a new name for the new version.

pkg/homedir/homedir_unix.go
```diff
@@ -101,11 +101,20 @@ func isWriteableOnlyByOwner(perm os.FileMode) bool {
...
+	}
+
+	return rootlessConfigHomeDir, rootlessConfigHomeDirError
+}
+
+// ConfigHome returns XDG_CONFIG_HOME. (Deprecated)
```
mtrmac, line 114: Presumably this one is not deprecated.

pkg/homedir/homedir.go
```diff
@@ -6,11 +6,16 @@ import (
 	"path/filepath"
 )
 
-// GetDataHome returns XDG_DATA_HOME.
-// GetDataHome returns $HOME/.local/share and nil error if XDG_DATA_HOME is not set.
+// DataHome (deprecated)
```
mtrmac, line 9: For all these name changes…

- I don’t really see enough benefit but also I don’t really care
- I do think the code should either:
  - Introduce a new name for symmetry, and intend to keep the old names forever. Then it’s not really deprecated.
  - Decide to deprecate the old name; in that case this should use the standard `Deprecated:` syntax so that tool make us migrate immediately.

mtrmac: … all these name changes apart from the real semantic change WRT `ConfigHome`.

Luap99: I rather not do the unrelated name changes, it just adds churn for no reason. If I look at a code and see the two different functions used I will always have to follow them to figure out if they are actual equal.

pkg/homedir/homedir_unix.go
```diff
@@ -115,33 +124,30 @@ func GetConfigHome() (string, error) {
...
-				return
-			} else if int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() {
-				cfgHomeDir = tmpDir
-			} else {
-				rootlessConfigHomeDirError = fmt.Errorf("path %q exists and it is not owned by the current user", tmpDir)
-				return
-			}
+			cfgHomeDir = filepath.Join(resolvedHome, ".config")
+		}
+		st, err := os.Stat(cfgHomeDir)
```
Luap99, line 129: I am still not convinced that the stat adds value here.

In fact if we keep the the stat then I see zero reason to change anything here because then most callers still have to ignore ENOENT errors which adds a lot of code and doesn't help the issue https://github.com/containers/podman/issues/23818.

Neither HomeDir() or GetCacheHome() validate permissions either.

pkg/homedir/homedir_unix.go
mtrmac: My mistake, this now propagates `ENOENT` to the caller. That’s discussed elsewhere…

pkg/homedir/homedir_unix.go
mtrmac: Also, right now, if the directory is missing, `ConfigHome` returns `("", ENOENT)`, so AFAICS this condition does nothing.

---

Shouldn’t all these functions, with their rare corner cases we don’t encounter on our workstations, have pretty good test coverage? I still mourn for tests lost in https://github.com/containers/storage/pull/1740 .

PR #2086, opened 2024-09-04
Update to Go 1.22

mtrmac:
… and take advantage of the new features where it’s easy to do.


mtrmac:
Cc: @Luap99

go.mod
```diff
@@ -1,7 +1,7 @@
-go 1.21
+go 1.22
```
Luap99, line 1: For consistency should we set this to 1.22.0 so that the toolchain line is removed like in the other projects?

I was thinking about updating the warning comment here to always say we never want the toolchain in go.mod

mtrmac: Both done.

Thanks!

Luap99 (approved):
/lgtm

PR #2087, opened 2024-09-04
Avoid manually-coded loops

mtrmac:
Use `slices` and `maps`, or other existing features, where appropriate.

This is basically a single pass looking for `append`.*`...`, and skimming all `for` loops.


store.go
```diff
@@ -1641,13 +1641,11 @@ func (s *store) CreateImage(id string, names []string, layer, metadata string, i
 				options.Metadata = iOptions.Metadata
 			}
 			options.BigData = append(options.BigData, copyImageBigDataOptionSlice(iOptions.BigData)...)
-			options.NamesHistory = append(options.NamesHistory, copyStringSlice(iOptions.NamesHistory)...)
+			options.NamesHistory = append(options.NamesHistory, iOptions.NamesHistory...)
```
cgwalters, line 1644: I only skimmed/spot checked but I noticed in this one we stopped copying...expected?

cgwalters (approved):
Looks great to me overall

PR #2090, opened 2024-09-11
Dedup fs-verity-digests metadata key definition

cgwalters:
I was just reading this code and found it odd that we had two definitions. This just makes it a public constant in the overlay code.


cgwalters:
Ah right, cross compilation. Messy. Nevermind.

PR #2096, opened 2024-09-17
Returns more than one additional image store

### LINKED ISSUE #2094: GraphOptions mixes image store with additional image stores (which are not returned)
Additional image stores are getting increasingly important for bootc. In https://github.com/containers/bootc/issues/644 I shared the desire to use `podman info` to query additional image stores to prevent bootc from parsing storage.conf files.

For testing, I configured /usr/share/containers/storage.conf as follows:
```
additionalimagestores = [
"/usr/lib/containers/storage",
"/var/lib/containers/storage",
"/var/lib/containers/storage123",
]
```

The returned data (`sudo podman info --format=json | jq .store`) looks as follows:
```json
{
  "configFile": "/usr/share/containers/storage.conf",
  "containerStore": {
    "number": 0,
    "paused": 0,
    "running": 0,
    "stopped": 0
  },
  "graphDriverName": "overlay",
  "graphOptions": {
    "overlay.imagestore": "/var/lib/containers/storage123",
    "overlay.mountopt": "nodev,metacopy=on"
  },
  "graphRoot": "/var/lib/containers/storage",
  "graphRootAllocated": 106219679744,
  "graphRootUsed": 3021795328,
  "graphStatus": {
    "Backing Filesystem": "btrfs",
    "Native Overlay Diff": "false",
    "Supports d_type": "true",
    "Supports shifting": "true",
    "Supports volatile": "true",
    "Using metacopy": "true"
  },
  "imageCopyTmpDir": "/var/tmp",
  "imageStore": {
    "number": 2
  },
  "runRoot": "/run/containers/storage",
  "volumePath": "/var/lib/containers/storage/volumes",
  "transientStore": false
}
```

What I observe is that the `overlay.imagestore` points to the last entry in the list.  That is because the "imagestore" key is returned three times and Podman takes the last one. However, an additional image store is not the "imagestore".

I suggest to fix that logic and add a new key for `options.additionalimagestores`.

Cc: @giuseppe

I may pick up the issue next week but feel free to assign it to someone else.

l0rd:
Closes #2094

### DIFF
--- a/types/options.go
+++ b/types/options.go
@@ -457,8 +457,11 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) erro
 	if config.Storage.RootlessStoragePath != "" {
 		storeOptions.RootlessStoragePath = config.Storage.RootlessStoragePath
 	}
-	for _, s := range config.Storage.Options.AdditionalImageStores {
-		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.imagestore=%s", config.Storage.Driver, s))
+	if len(config.Storage.Options.AdditionalImageStores) > 0 {
+		imageStore := config.Storage.Options.AdditionalImageStores[0]
+		additionalImageStores := strings.Join(config.Storage.Options.AdditionalImageStores, ",")
+		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.imagestore=%s", config.Storage.Driver, imageStore))
+		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.additionalimagestores=%s", config.Storage.Driver, additionalImageStores))
 	}
 	for _, s := range config.Storage.Options.AdditionalLayerStores {
 		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.additionallayerstore=%s", config.Storage.Driver, s))
--- a/types/options_test.go
+++ b/types/options_test.go
@@ -125,13 +125,52 @@ func TestGetRootlessStorageOpts2(t *testing.T) {
 }
 
 func TestReloadConfigurationFile(t *testing.T) {
-	content := bytes.NewBufferString("")
-	logrus.SetOutput(content)
-	var storageOpts StoreOptions
-	err := ReloadConfigurationFile("./storage_broken.conf", &storageOpts)
-	require.NoError(t, err)
-	assert.Equal(t, storageOpts.RunRoot, "/run/containers/test")
-	logrus.SetOutput(os.Stderr)
-
-	assert.Equal(t, strings.Contains(content.String(), "Failed to decode the keys [\\\"foo\\\" \\\"storage.options.graphroot\\\"] from \\\"./storage_broken.conf\\\"\""), true)
+	t.Run("broken", func(t *testing.T) {
+		content := bytes.NewBufferString("")
+		logrus.SetOutput(content)
+		var storageOpts StoreOptions
+		err := ReloadConfigurationFile("./storage_broken.conf", &storageOpts)
+		require.NoError(t, err)
+		assert.Equal(t, storageOpts.RunRoot, "/run/containers/test")
+		logrus.SetOutput(os.Stderr)
+		assert.Equal(t, strings.Contains(content.String(), "Failed to decode the keys [\\\"foo\\\" \\\"storage.options.graphroot\\\"] from \\\"./storage_broken.conf\\\"\""), true)
+	})
+	t.Run("imagestore-empty", func(t *testing.T) {
+		expectedStore := ""
+		expectedAdditionalStores := ""
+		var storageOpts StoreOptions
+		err := ReloadConfigurationFile("./storage_test.conf", &storageOpts)
+		require.NoError(t, err)
+		var actualStore, actualAdditionalStores string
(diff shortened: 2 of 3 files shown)


types/options.go
```diff
@@ -457,8 +457,11 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) erro
 	if config.Storage.RootlessStoragePath != "" {
 		storeOptions.RootlessStoragePath = config.Storage.RootlessStoragePath
 	}
-	for _, s := range config.Storage.Options.AdditionalImageStores {
-		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.imagestore=%s", config.Storage.Driver, s))
+	if len(config.Storage.Options.AdditionalImageStores) > 0 {
+		imageStore := config.Storage.Options.AdditionalImageStores[0]
+		additionalImageStores := strings.Join(config.Storage.Options.AdditionalImageStores, ",")
```
cgwalters, line 462: I'm not opposed to this way of doing things, but I think it'd be conceptually cleaner to add a new API field that was an array/list.

Squashing an array to a string like this comma-separated raises questions around e.g. "what happens if the store path has a comma in it" - to be clear not something that really we'd expect to happen in this case.

But as a *general* case these things can happen, and ad-hoc "stringification" can become a problem.

There's also the flip side in that in theory something could have been reading the prior field and relying on it only containing (the top?) store, and suddenly seeing multiple paths there would break a prior consumer.

cgwalters:
On the commit message: this repository like many others in this organization uses "Linux kernel style" commit messages, see the [git log](https://github.com/containers/storage/commits/main/) - [this blog](https://cbea.ms/git-commit/) is one I reference often.

In this case I'd say something like:

> List all image stores in serialized state
>
> The config format and internals support multiple stores, but the serialized state
> only returns one. Add a new field which lists all so that external callers
> can see all of them.

?

types/options.go
```diff
@@ -457,8 +457,11 @@ func ReloadConfigurationFile(configFile string, storeOptions *StoreOptions) erro
...
+		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.imagestore=%s", config.Storage.Driver, imageStore))
+		storeOptions.GraphDriverOptions = append(storeOptions.GraphDriverOptions, fmt.Sprintf("%s.additionalimagestores=%s", config.Storage.Driver, additionalImageStores))
```
mtrmac, line 464: I can’t see any parser of a driver `additionalimagestores` option; AFAICS that just doesn’t work.

---

All entries in `Options.AdditionalImageStores` have the same semantics (and that’s different from `Options.ImageStore`). I don’t know why we would treat the first item differently.

I _especially_ don’t know why we would now add the first item _twice_.

Also, if you look at the drivers’ option parsers, for the `overlay` driver both option names are treated the same, but for VFS, `additionalimagestore` is not recognized at all. So this breaks VFS.

types/options_test.go
```diff
@@ -125,13 +125,52 @@ func TestGetRootlessStorageOpts2(t *testing.T) {
...
+			option := strings.Split(o, "=")
+			switch option[0] {
+			case storageOpts.GraphDriverName + ".imagestore":
+				actualStore = option[1]
+			case storageOpts.GraphDriverName + ".additionalimagestores":
+				actualAdditionalStores = option[1]
+			}
+		}
+		assert.Equal(t, actualStore, expectedStore)
+		assert.Equal(t, actualAdditionalStores, expectedAdditionalStores)
```
mtrmac, line 155: (The parameter order is `(t, expected, actual)`.)

mtrmac (commented):
It seems to me that this is at best a workaround for this specific case.

The root cause, as I understand it, is that `GraphDriverOptions` is a list of (key, value) pairs, where nothing prohibits duplicate keys; meanwhile Podman treats it as a map of key: value, where duplicate keys are impossible.

This modifies _one_ code path where duplicate keys can happen; but `GraphDriverOptions` can contain user-specified items, via several different input paths. So this doesn’t actually fix things to accurately represent the configuration.

My best guess is that the `podman info` command needs to change its output format to be able to represent `graphOptions` as they truly exist; or, maybe, give up on that general extensible field entirely, and have a specific “image stores” output field, getting _only_ that data (not _parsing_ the config, but asking c/storage via `.AdditionalImageStores()` or maybe some to-be-added higher level interface over it) and presenting it in some user-appropriate way.

---

(Generally I’m seriously unhappy about the flow where we have well-typed `Storage.Options`, turn that into strings, and then parse strings in the graph driver again; but that’s not all that relevant for this PR, and because `GraphDriverOptions` can be specified by the user directly, a string parser must exist _somewhere_.)

types/options.go
giuseppe: this can be specified multiple times, e.g. `podman --storage-opt overlay.additionalimagestore=/tmp/1 --storage-opt overlay.additionalimagestore=/tmp/2 ...` works fine.

So instead of combining multiple additional image stores in a single option, we can have multiple `additionalimagestore`, one for each directory.

Said so, we don't currently support commas in the path, as we have an explicit `strings.Split(val, ",")` in `overlay.go`.

rhatdan:
@l0rd still working on this one?

l0rd:
@rhatdan yes. I mean I should resume working on this next week.

l0rd:
When I first looked at https://github.com/containers/storage/issues/2094, I thought we had to address something in `containers/storage`. After looking at it again, I see that the problem is in `container/podman`, and there is nothing to fix here. I am closing this PR in favor of https://github.com/containers/podman/pull/24731

PR #2097, opened 2024-09-18
pkg/directory: correctly ignore ENOENT while walking

Luap99:
As documented in (fs.DirEntry) Info() it can return ENOENT so we must ignore it there as well. Reported in
https://github.com/containers/podman/issues/23909

Given this is a race I don't think I can add a test here.

### DIFF
--- a/pkg/directory/directory_unix.go
+++ b/pkg/directory/directory_unix.go
@@ -4,8 +4,8 @@
 package directory
 
 import (
+	"errors"
 	"io/fs"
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@nalind @giuseppe @saschagrunert @flouthoc @mheon @vrothberg PTAL

saschagrunert (approved):

giuseppe (approved):
/lgtm

PR #2102, opened 2024-09-19
Assorted nits reported by revive linter

kolyshkin:
See individual commits for details.


rhatdan:
/approve
LGTM

nalind:
Should this be enabling the linter in `.golangci.yml`, too?

kolyshkin:
> Should this be enabling the linter in `.golangci.yml`, too?

This linter is sort of too opinionated, and thus it's best to run it manually and only fix those warnings that make sense.

We can, of course, annotate all the warnings that we don't want to fix, and then enable the linter, but I'm not sure it makes sense.

PR #2105, opened 2024-09-23
userns: fix off-by-one userns max size detection

giuseppe:
fix the detection for the maximum userns size from an image.

If the maximum ID used in an image is X, we need to use a user namespace with size X+1 to include UID=X.

Closes: https://github.com/containers/storage/issues/2104


eriksjolund:
I think this line needs a fix https://github.com/containers/podman/blob/186a2b8d49d9505ba48da373c566953aeae9c1dd/vendor/github.com/containers/storage/userns.go#L92
to check /etc/group

giuseppe:

mheon:
LGTM

Luap99 (approved):
/lgtm

Issue #2092, opened 2024-09-12
Fix `d.getStagingDir("")`


mtrmac:
Since #1784 , https://github.com/containers/storage/blob/10981abafe3e3ece4e0cc872761f7624cd1bef0e/drivers/overlay/overlay.go#L2188 is calling `getStagingDir("")`, i.e. `dir2` with a layer ID `""`.

Just from reading the code, that _happens_ to work as expected because it is not using the primary return value of `dir2`, but that’s extremely non-obvious and very likely to be broken during future maintenance. To the extent the cod is only looking for the “homeDir”, I think that should be modeled explicitly and not use the `id` value at all. Alternatively, if the `id` value mattered, then this needs to actually be implemented.

Cc: @giuseppe

giuseppe:
opened https://github.com/containers/storage/pull/2107

PR #2107, opened 2024-09-24
overlay: refactor getStagingDir helper

giuseppe:
simplify the implementation of getStagingDir() to just create a new staging directory and drop the check whether the correct staging base directory as anyway the rename operation would fail later.
    
Closes: https://github.com/containers/storage/issues/2092

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -869,11 +869,11 @@ func (d *Driver) pruneStagingDirectories() bool {
 
 	anyPresent := false
 
-	homeStagingDir := filepath.Join(d.home, stagingDir)
-	dirs, err := os.ReadDir(homeStagingDir)
+	stagingDirBase := filepath.Join(d.homeDirForImageStore(), stagingDir)
+	dirs, err := os.ReadDir(stagingDirBase)
 	if err == nil {
 		for _, dir := range dirs {
-			stagingDirToRemove := filepath.Join(homeStagingDir, dir.Name())
+			stagingDirToRemove := filepath.Join(stagingDirBase, dir.Name())
 			lock, err := lockfile.GetLockFile(filepath.Join(stagingDirToRemove, stagingLockFile))
 			if err != nil {
 				anyPresent = true
@@ -1205,17 +1205,22 @@ func (d *Driver) getAllImageStores() []string {
 	return additionalImageStores
 }
 
-func (d *Driver) dir2(id string, useImageStore bool) (string, string, bool) {
-	var homedir string
-
-	if useImageStore && d.imageStore != "" {
-		homedir = path.Join(d.imageStore, d.name)
-	} else {
-		homedir = d.home
+// homeDirForImageStore returns the home directory to use when an image store is configured
+func (d *Driver) homeDirForImageStore() string {
+	if d.imageStore != "" {
+		return path.Join(d.imageStore, d.name)
 	}
+	// If there is not an image store configured, use the same
+	// store
+	return d.home
+}
 
+func (d *Driver) dir2(id string, useImageStore bool) (string, string, bool) {
+	homedir := d.home
+	if useImageStore {
+		homedir = d.homeDirForImageStore()
+	}
 	newpath := path.Join(homedir, id)
-
 	if err := fileutils.Exists(newpath); err != nil {
 		for _, p := range d.getAllImageStores() {
 			l := path.Join(p, d.name, id)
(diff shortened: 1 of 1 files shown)


drivers/overlay/overlay.go
```diff
@@ -1205,17 +1205,17 @@ func (d *Driver) getAllImageStores() []string {
 	return additionalImageStores
 }
 
-func (d *Driver) dir2(id string, useImageStore bool) (string, string, bool) {
-	var homedir string
-
+func (d *Driver) getHomeDir(useImageStore bool) string {
```
rhatdan, line 1208: Can you drop the get?  homeDir()

drivers/overlay/overlay.go
```diff
@@ -2087,11 +2087,22 @@ func (g *overlayFileGetter) Close() error {
...
 
-func (d *Driver) getStagingDir(id string) string {
+// getStagingDirForLayer returns the path to the base staging directory used for the given layer.
+func (d *Driver) getStagingDirForLayer(id string) string {
 	_, homedir, _ := d.dir2(id, d.imageStore != "")
 	return filepath.Join(homedir, stagingDir)
 }
 
+// getNewStagingDir creates a new staging directory and returns the path to it.
+func (d *Driver) getNewStagingDir() (string, error) {
```
rhatdan, line 2097: Can you drop the get?  newStagingDir()

giuseppe:
@rhatdan thanks for the review.  Addressed the comments

drivers/overlay/overlay.go
```diff
@@ -2087,11 +2087,22 @@ func (g *overlayFileGetter) Close() error {
 	return errs.ErrorOrNil()
 }
 
-func (d *Driver) getStagingDir(id string) string {
+// stagingDirForLayer returns the path to the base staging directory used for the given layer.
+func (d *Driver) stagingDirForLayer(id string) string {
```
mtrmac, line 2091: Ultimately `id` (“for the given layer”) here should not be relevant as an input (to `ApplyDiffFromStagingDirectory`), because it was not used as an input when creating the staging directory in `ApplyDiffWithDiffer`.

The computation of `stagingDirBase` in `stagingDirForLayer` is different — shouldn’t that be the same code, or maybe a single shared implementation?

drivers/overlay/overlay.go
```diff
@@ -1205,17 +1205,17 @@ func (d *Driver) getAllImageStores() []string {
 	return additionalImageStores
 }
 
-func (d *Driver) dir2(id string, useImageStore bool) (string, string, bool) {
-	var homedir string
-
+func (d *Driver) homeDir(useImageStore bool) string {
```
mtrmac, line 1208: I’m confused by the terminology. We have `d.homeDir` and `d.home`, and those are different values with different semantics.

What do the values actually _mean_?

(I mean, I’m generally confused and scared by these code paths and options and indirections. This is just a very clear manifestation of my general confusion. But, still.)

---

`pruneStagingDirectories` hard-codes `filepath.Join(d.home, stagingDir)`, not using the `useImageStore` condition. Which of the two is correct?

mtrmac (commented):
Thanks!

This refactoring is correct as is, but it also exposes lingering questions…

drivers/overlay/overlay.go
giuseppe: The difference is that here we know what is the id for the layer, so we validate that we are using a correct staging directory.  Not sure it is worth to keep it around, I am going to drop the check, if the staging directory is not valid (for any reason), the atomic rename will fail anyway

drivers/overlay/overlay.go
giuseppe: thanks, I've renamed the function, hopefully it is clearer now.

`pruneStagingDirectories` was using the wrong location, I've fixed this too

rhatdan:
LGTM

drivers/overlay/overlay.go
mtrmac: My point really was that the `id` is not a factor on what should be a “correct” staging directory.

I don’t feel strongly about the the sanity check; removing it entirely does resolve the “different computation” concern.

drivers/overlay/overlay.go
mtrmac: I am still generally confused ;), but making the names are much more distinct does help.

mtrmac (commented):
/lgtm

Thanks again!

PR #2108, opened 2024-09-24
overlay: additional layer: check layer existence before using

ktock:
Currently *Driver.get doesn't check existence of additional layers. This result in an unclear error, like `mkdir /var/lib/containers/storage/overlay/a4fe7aa552629d15640e70d18b00359e6b58c413ba8aa6c57ec5a63a5742584d/diff: file exists`.
This commit fixes that function to check the existence of additional layers before using and return a clear error.


drivers/overlay/overlay.go
```diff
@@ -2698,12 +2701,17 @@ func notifyReleaseAdditionalLayer(al string) {
...
 // Store, it returns the original path without changes.
-func redirectDiffIfAdditionalLayer(diffPath string) (string, error) {
+func redirectDiffIfAdditionalLayer(diffPath string, checkExistence bool) (string, error) {
 	if ld, err := os.Readlink(diffPath); err == nil {
 		// diff is the link to Additional Layer Store
 		if !path.IsAbs(ld) {
 			return "", fmt.Errorf("linkpath must be absolute (got: %q)", ld)
 		}
+		if checkExistence {
+			if _, err := os.Stat(ld); err != nil {
```
giuseppe, line 2711: could you please use `fileutils.Exists()`?

ktock: Thanks for the review. Fixed to use fileutils.Exists.

giuseppe (approved):
LGTM

PR #2109, opened 2024-09-24
userns: skip "nogroup"

giuseppe:
the alpine image defines a "nogroup":

$ podman run --rm alpine grep nogroup /etc/group
nogroup:x:65533:

ignore it as we are already doing for the "nobody" user.


giuseppe:
testing the patch in https://github.com/containers/podman/pull/24054

Honny1 (approved):
/lgtm

PR #2116, opened 2024-09-27
Use tar-split/tar/asm.IterateHeaders now that it has been accepted

mtrmac:
... instead of our version which makes assumptions on the internal decisions of the tar-split project, and needs heuristics to guess where file padding ends.


kwilczynski:
@mtrmac, thank you!

PR #2117, opened 2024-09-27
Update copy.DirCopy to leave sockets in the file system

### LINKED ISSUE #2113: copy.DirCopy fails when encountering sockets
[We](https://github.com/warewulf/warewulf) use `copy.DirCopy` and have encountered a behavior that, if the source contains sockets, the copy operation fails with an error of the form:

```
lchown /var/lib/warewulf/chroots/gpuslurm/rootfs/run/user/0/gnupg/d.kg8ijih5tq41ixoeag4p1qup/S.gpg-agent: no such file or directory
```

Could `copy.DirCopy` be updated to ignore sockets in the source in stead?

Reported for Warewulf at https://github.com/warewulf/warewulf/issues/892

anderbubble:
`copy.DirCopy` contains code to create named sockets matching those on the source; however, this operation closes the socket, removing it implicitly from the destination and causing later chown and chmod operations to fail.
    
This commit creates matching file system nodes directly without opening a socket or invoking the auto-removal semantics of a listener.

- Closes #2113

### DIFF
--- a/drivers/copy/copy_linux.go
+++ b/drivers/copy/copy_linux.go
@@ -16,7 +16,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net"
 	"os"
 	"path/filepath"
 	"strings"
@@ -199,11 +198,9 @@ func DirCopy(srcDir, dstDir string, copyMode Mode, copyXattrs bool) error {
 			}
 
 		case mode&os.ModeSocket != 0:
-			s, err := net.Listen("unix", dstPath)
-			if err != nil {
+			if err := unix.Mknod(dstPath, stat.Mode, int(stat.Rdev)); err != nil {
 				return err
 			}
-			s.Close()
 
 		case mode&os.ModeDevice != 0:
 			if unshare.IsRootless() {
--- a/drivers/copy/copy_test.go
+++ b/drivers/copy/copy_test.go
@@ -5,6 +5,7 @@ package copy
 import (
 	"fmt"
 	"math/rand"
+	"net"
 	"os"
 	"path/filepath"
 	"syscall"
@@ -83,6 +84,11 @@ func randomMode(baseMode int) os.FileMode {
 
 func populateSrcDir(t *testing.T, srcDir string, remainingDepth int) {
 	if remainingDepth == 0 {
+		socketPath := filepath.Join(srcDir, "srcsocket")
+		s, err := net.ListenUnix("unix", &net.UnixAddr{Name: socketPath, Net: "unix"})
+		assert.NilError(t, err)
+		s.SetUnlinkOnClose(false)
+		s.Close()
 		return
 	}
 	aTime := time.Unix(rand.Int63(), 0)


anderbubble:
If possible, I'd like to see this backported to 1.55 (before the shift to go 1.22) so that we could use it in Warewulf on EL 8 and 9.

rhatdan:
@mtrmac @giuseppe @nalind @Luap99 PTAL

mtrmac (commented):
At a first glance this makes no sense to me as proposed: if we don’t close that socket, what _is_ going to close it?

This code needs to work in long-running processes like CRI-O, so leaking file descriptors is not acceptable.

(And, au contraire, if Go’s garbage collector somehow caused the socket to be closed eventually, then that means that the problematic chown/chmod operations could also fail, depending on GC timing.)

---

It might well be the case that we need to keep the socket open longer, and close it elsewhere; I’m not saying that the current code is correct, I didn’t look into that at the detail, at this point I just took the 1 minute to notice that the PR as currently proposed looks incorrect.

anderbubble:
I thought that this was just a difference in semantics between a TCP socket and a Unix domain socket; but turns out Go just mixes close with unlink. What I really want is UnlinkOnClose(false).

I'll fix it. Thanks for the review.

rhatdan:
/approve
LGTM

Luap99 (commented):
I am not familiar with this code but why are we copying socket files in the first place? I don't see a how any user could benefit from a not connected socket file in the target.
So would it not be simpler to skip socket files all together?

anderbubble:
> why are we copying socket files in the first place

I can’t speak to this: before I started investigating I expected the solution would be to skip sockets as well. But when I found code that was explicitly trying to copy them, but just failing to do so properly, I decided I would start by making the existing code work, rather than remove it, since I didn’t know the original intent.

in theory, I suppose the existence of the socket could be a sentinel to some other application. I also think the existing socket can be re-opened by later applications at the existing path.

Luap99:
> in theory, I suppose the existence of the socket could be a sentinel to some other application.

Maybe but this seems rather unlikely 

> I also think the existing socket can be re-opened by later applications at the existing path.

I do not see how this could work, binding an existing socket file will fail with EADDRINUSE even when there is no other program actively listening there. As such most programs do an explicit unlink() before calling bind().

Given the current version deletes the path again (and then fails to chown) it should be safe to assume that no user can depend on the behavior and I personally see skipping sockets to be much simpler than doing this here.

anderbubble:
The original socket code was added by @giuseppe in 029896beb38738b405c824acb891b994b3fb3968. No objection from me removing it / altering it to skip sockets in stead; but maybe he has insight into why it was added in the first place.

anderbubble:
Looks like the first commit at 6a9fa6b66e47fdec3d803e0d4e57b7e6bec2435e includes sockets and was authored by @nalind, so he might have insight into the original intent as well.

giuseppe:
I think this was done without thinking too much about its implications, so let's ignore them.  Should we do the same with FIFOs?

mtrmac:
In Docker, this exists all the way back in https://github.com/moby/moby/pull/7619 .

---

Note the existence of https://github.com/containers/storage/blob/4bf3f075cf3f2ddfa1f2909dbdbc03bf0b88d9d2/drivers/fsdiff.go#L45 ; just the presence of files might actually make a difference to the reported layer contents.

I don’t know that the files would be _useful_, but if a pull+push changed the filesystem contents, that might be unexpected.

drivers/copy/copy_linux.go
```diff
@@ -203,6 +203,9 @@ func DirCopy(srcDir, dstDir string, copyMode Mode, copyXattrs bool) error {
 			if err != nil {
 				return err
 			}
+			if unixListener, ok := s.(*net.UnixListener); ok {
```
mtrmac, line 206: I think this should fail if `!ok`; or, maybe better, use `net.ListenUnix` explicitly to make the cast unnecessary.

drivers/copy/copy_test.go
```diff
@@ -83,6 +84,13 @@ func randomMode(baseMode int) os.FileMode {
 
 func populateSrcDir(t *testing.T, srcDir string, remainingDepth int) {
 	if remainingDepth == 0 {
+		socketName := filepath.Join(srcDir, "srcsocket")
+		s, err := net.Listen("unix", socketName)
+		assert.NilError(t, err)
+		if unixListener, ok := s.(*net.UnixListener); ok {
+			unixListener.SetUnlinkOnClose(false)
+		}
+		s.Close()
```
mtrmac, line 93: Assert that the file still exists afterwards? Otherwise the test code and the test subject are exactly the same and might have the same bug.

mtrmac (commented):
(Let’s decide whether we need these files at all, first.)

Luap99:
I did take a quick look what gnu `cp` does. With `cp -r` it will copy the socket file, however it does not the socket(), bind(), close() thing we are doing here.
Instead it just calls mknodat() looking at the strace output
`mknodat(AT_FDCWD, "/tmp/test2/mysock", S_IFSOCK|0700) = 0` 

relevant source code: https://github.com/coreutils/coreutils/blob/e71b24cedc2ff640f94911298e53b24506f13062/src/copy.c#L3090-L3099

Thus I would say we should prefer a mknod() call here over the current fix.

---

Though this still doesn't answer if we have to keep socket files during copy. 

> I don’t know that the files would be useful, but if a pull+push changed the filesystem contents, that might be unexpected.

I guess if this is a practical concern then keeping them makes sense.

giuseppe:
> > I don’t know that the files would be useful, but if a pull+push changed the filesystem contents, that might be unexpected.
> 
> I guess if this is a practical concern then keeping them makes sense.

AFAICS a tarball cannot handle unix sockets (https://pkg.go.dev/archive/tar#pkg-constants), so that shouldn't be a concern.

> I did take a quick look what gnu `cp` does. With `cp -r` it will copy the socket file, however it does not the socket(), bind(), close() thing we are doing here. Instead it just calls mknodat() looking at the strace output `mknodat(AT_FDCWD, "/tmp/test2/mysock", S_IFSOCK|0700) = 0`

Using `mknod` seems like a good idea to fix the current code without changing its behavior

mtrmac:
> AFAICS a tarball cannot handle unix sockets (https://pkg.go.dev/archive/tar#pkg-constants)

Thanks, TIL.

rhatdan:
@anderbubble what is going on with this PR?

anderbubble:
@rhatdan I'll update it based on the conversation here; I just got pulled into some other things for a bit.

anderbubble:
@rhatdan I updated the pr to use mknod rather than opening a dangling socket.

I think it's good to go now; but let me know if you have any questions or concerns. Sorry for the delay!

rhatdan:
@giuseppe @Luap99 @mtrmac PTAL

drivers/copy/copy_test.go
Luap99: This comment still seems to be open, we should ensure the file exists after the copy.

anderbubble: I think the file is ensured to exist by way of https://github.com/containers/storage/pull/2117/files#diff-e80de23fd41197b1acaf1ebc6ec79d3df1d7f56704005df462246eef120b17e3L39-L74

giuseppe (approved):
LGTM

rhatdan:
/lgtm
Thanks @anderbubble

giuseppe:
it needs to be backported

anderbubble:
Backport PR at https://github.com/containers/storage/pull/2159. Let me know if that needs any adjustment.

Issue #2115, opened 2024-09-27
pkg/chunked/GetDiffer must differentiate between failures and errors that can be ignored


giuseppe:
we currently do not differentiate between critical errors (hard failures) and ignorable errors, so that the caller (containers/image in our case) can fallback to using the traditional pull code path.

Especially with composefs and fs-verity, if any error happens in this phase, the entire pull operation must fail and containers/image must not attempt any fallback.

Suggestion: define a new error typefor the errors that can be ignored and treat anything else as an hard failure, so that the caller can differentiate between them.

giuseppe:
@mtrmac what do you think?

mtrmac:
For context, what are the “ignorable” error situations? I can see:
- `c/image/internal/private.BadPartialRequestError` ~ `pkg/chunked/ErrBadRequest`: missing registry support
- `enable_partial_images` disabled
- layer is not chunked / missing TOC annotations.

And those are currently always ignorable, but might be fatal in an enforcing configuration, I guess.

Are there any _other_ significant classes of ignorable errors?

---

Yes, having a special error type (returned by `chunked.GetDiffer?`), and handling it specially in c/image so that other errors don’t trigger a fallback to the full layer copy, LGTM.

giuseppe:
yes, the only "ignorable" errors are the ones where the image doesn't support partial pulls and there is no conversion enabled.  Missing registry support should be a hard error when `convert_images` is configured

giuseppe:
thanks, opened a PR to implement it: https://github.com/containers/storage/pull/2118

PR #2118, opened 2024-09-30
chunked: define error for partial pulls not available

giuseppe:
define a new error type so that the caller can determine whether it is safe to ignore the error and retrieve the resource fully.

Closes: https://github.com/containers/storage/issues/2115

the relative patch for containers/image is something like:

```diff
diff --git a/copy/single.go b/copy/single.go
index 324785a8..b7921837 100644
--- a/copy/single.go
+++ b/copy/single.go
@@ -819,8 +819,13 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 				logrus.Debugf("Retrieved partial blob %v", srcInfo.Digest)
 				return true, updatedBlobInfoFromUpload(srcInfo, uploadedBlob), nil
 			}
-			logrus.Debugf("Failed to retrieve partial blob: %v", err)
-			return false, types.BlobInfo{}, nil
+			// On a "partial content not available" error, ignore it and retrieve the whole layer.
+			if chunkedToc.IsPartialContentNotAvailableError(err) {
+				logrus.Debugf("Failed to retrieve partial blob: %v", err)
+				err = nil
+			}
+			return false, types.BlobInfo{}, err
+
 		}()
 		if err != nil {
 			return types.BlobInfo{}, "", err
```

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -143,11 +143,13 @@ func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload *o
 }
 
 // GetDiffer returns a differ than can be used with ApplyDiffWithDiffer.
+// If it returns an error that implements IsErrFallbackToOrdinaryLayerDownload, the caller can
+// retry the operation with a different method.
 func GetDiffer(ctx context.Context, store storage.Store, blobDigest digest.Digest, blobSize int64, annotations map[string]string, iss ImageSourceSeekable) (graphdriver.Differ, error) {
 	pullOptions := store.PullOptions()
 
 	if !parseBooleanPullOption(pullOptions, "enable_partial_images", true) {
-		return nil, errors.New("enable_partial_images not configured")
+		return nil, newErrFallbackToOrdinaryLayerDownload(errors.New("partial images are disabled"))
 	}
 
 	zstdChunkedTOCDigestString, hasZstdChunkedTOC := annotations[internal.ManifestChecksumKey]
@@ -157,29 +159,54 @@ func GetDiffer(ctx context.Context, store storage.Store, blobDigest digest.Diges
 		return nil, errors.New("both zstd:chunked and eStargz TOC found")
 	}
 
-	if hasZstdChunkedTOC {
-		zstdChunkedTOCDigest, err := digest.Parse(zstdChunkedTOCDigestString)
-		if err != nil {
-			return nil, fmt.Errorf("parsing zstd:chunked TOC digest %q: %w", zstdChunkedTOCDigestString, err)
-		}
-		return makeZstdChunkedDiffer(store, blobSize, zstdChunkedTOCDigest, annotations, iss, pullOptions)
-	}
-	if hasEstargzTOC {
-		estargzTOCDigest, err := digest.Parse(estargzTOCDigestString)
-		if err != nil {
-			return nil, fmt.Errorf("parsing estargz TOC digest %q: %w", estargzTOCDigestString, err)
-		}
-		return makeEstargzChunkedDiffer(store, blobSize, estargzTOCDigest, iss, pullOptions)
+	convertImages := parseBooleanPullOption(pullOptions, "convert_images", false)
+
+	if !hasZstdChunkedTOC && !hasEstargzTOC && !convertImages {
+		return nil, newErrFallbackToOrdinaryLayerDownload(errors.New("no TOC found and convert_images is not configured"))
 	}
 
-	return makeConvertFromRawDiffer(store, blobDigest, blobSize, iss, pullOptions)
+	var err error
+	var differ graphdriver.Differ
+	// At this point one of hasZstdChunkedTOC, hasEstargzTOC or convertImages is true.
+	if hasZstdChunkedTOC {
+		zstdChunkedTOCDigest, err2 := digest.Parse(zstdChunkedTOCDigestString)
+		if err2 != nil {
+			return nil, err2
+		}
+		differ, err = makeZstdChunkedDiffer(store, blobSize, zstdChunkedTOCDigest, annotations, iss, pullOptions)
+		if err == nil {
+			logrus.Debugf("Created zstd:chunked differ for blob %q", blobDigest)
+			return differ, err
+		}
+	} else if hasEstargzTOC {
+		estargzTOCDigest, err2 := digest.Parse(estargzTOCDigestString)
+		if err2 != nil {
+			return nil, err
+		}
+		differ, err = makeEstargzChunkedDiffer(store, blobSize, estargzTOCDigest, iss, pullOptions)
+		if err == nil {
+			logrus.Debugf("Created eStargz differ for blob %q", blobDigest)
+			return differ, err
+		}
+	}
+	// If convert_images is enabled, always attempt to convert it instead of returning an error or falling back to a different method.
+	if convertImages {
+		logrus.Debugf("Created differ to convert blob %q", blobDigest)
+		return makeConvertFromRawDiffer(store, blobDigest, blobSize, iss, pullOptions)
+	}
+
+	logrus.Debugf("Could not create differ for blob %q: %v", blobDigest, err)
+
+	// If the error is a bad request to the server, then signal to the caller that it can try a different method.  This can be done
+	// only when convert_images is disabled.
+	var badRequestErr ErrBadRequest
+	if errors.As(err, &badRequestErr) {
+		err = newErrFallbackToOrdinaryLayerDownload(err)
+	}
+	return nil, err
 }
 
 func makeConvertFromRawDiffer(store storage.Store, blobDigest digest.Digest, blobSize int64, iss ImageSourceSeekable, pullOptions map[string]string) (*chunkedDiffer, error) {
-	if !parseBooleanPullOption(pullOptions, "convert_images", false) {
-		return nil, errors.New("convert_images not configured")
-	}
-
 	layersCache, err := getLayersCache(store)
 	if err != nil {
 		return nil, err
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -947,11 +947,9 @@ func (c *chunkedDiffer) retrieveMissingFiles(stream ImageSourceSeekable, dirfd i
 		}
 
 		if _, ok := err.(ErrBadRequest); ok {
-			// If the server cannot handle at least 64 chunks in a single request, just give up.
-			if len(chunksToRequest) < 64 {
+			if len(chunksToRequest) == 1 {
 				return err
 			}
-
 			// Merge more chunks to request
 			missingParts = mergeMissingChunks(missingParts, len(chunksToRequest)/2)
 			calculateChunksToRequest()
--- a/pkg/chunked/storage.go
+++ b/pkg/chunked/storage.go
@@ -23,3 +23,22 @@ type ErrBadRequest struct { //nolint: errname
 func (e ErrBadRequest) Error() string {
 	return "bad request"
 }
+
+// ErrFallbackToOrdinaryLayerDownload is a custom error type that
+// suggests to the caller that a fallback mechanism can be used
+// instead of a hard failure.
+type ErrFallbackToOrdinaryLayerDownload struct {
+	Err error
+}
+
+func (c ErrFallbackToOrdinaryLayerDownload) Error() string {
+	return c.Err.Error()
+}
+
+func (c ErrFallbackToOrdinaryLayerDownload) Unwrap() error {
+	return c.Err
+}
+
+func newErrFallbackToOrdinaryLayerDownload(err error) error {
+	return ErrFallbackToOrdinaryLayerDownload{Err: err}
+}


rhatdan:
LGTM
@nalind @Luap99 @mtrmac @mheon @vrothberg @saschagrunert PTAL

pkg/chunked/storage_linux.go
```diff
@@ -146,40 +146,48 @@ func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload *o
 func GetDiffer(ctx context.Context, store storage.Store, blobDigest digest.Digest, blobSize int64, annotations map[string]string, iss ImageSourceSeekable) (graphdriver.Differ, error) {
 	pullOptions := store.PullOptions()
 
+	convertImages := parseBooleanPullOption(pullOptions, "convert_images", false)
```
mtrmac, line 149: (I think this can be moved down after the first `enable_partial_images` check.)

pkg/chunked/storage_linux.go
```diff
@@ -146,40 +146,48 @@ func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload *o
...
 	zstdChunkedTOCDigestString, hasZstdChunkedTOC := annotations[internal.ManifestChecksumKey]
 	estargzTOCDigestString, hasEstargzTOC := annotations[estargz.TOCJSONDigestAnnotation]
 
 	if hasZstdChunkedTOC && hasEstargzTOC {
-		return nil, errors.New("both zstd:chunked and eStargz TOC found")
+		if !convertImages {
+			return nil, fmt.Errorf("both zstd:chunked and eStargz TOC found: %w", internal.ErrPartialContentNotAvailable)
+		}
+		// If a conversion is required, ignore both TOCs and retrieve the entire layer.
+		hasZstdChunkedTOC, hasEstargzTOC = false, false
```
mtrmac, line 163: I’d rather fail without fallback here; this is either a _severe_ misconfiguration or an attack.

(c/image is not ideal in this right now, https://github.com/containers/image/blob/93fa49b0f1fb78470512e0484012ca7ad3c5c804/copy/single.go#L892-L894 , but still, that shouldn’t result in an image with both TOC annotations.)

pkg/chunked/storage_linux.go
```diff
@@ -146,40 +146,48 @@ func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload *o
...
 
 	if hasZstdChunkedTOC {
 		zstdChunkedTOCDigest, err := digest.Parse(zstdChunkedTOCDigestString)
-		if err != nil {
-			return nil, fmt.Errorf("parsing zstd:chunked TOC digest %q: %w", zstdChunkedTOCDigestString, err)
+		if err == nil {
+			return makeZstdChunkedDiffer(store, blobSize, zstdChunkedTOCDigest, annotations, iss, pullOptions)
+		}
+		if !convertImages {
+			return nil, fmt.Errorf("parsing zstd:chunked TOC digest %q: %w", zstdChunkedTOCDigestString, internal.ErrPartialContentNotAvailable)
```
mtrmac, line 172: These error messages don’t look all that useful — and lose the actual text describing the reason for the failure.

WDYT about
```go
type Err$name struct {
   underlying error
}
```
implementing `Unwrap`? Then we can “mark” any error without affecting the type… and we can have a separate design conversation on whether the wrapping type should add any error text at all.

---

Alternatively, this works:
```go
return fmt.Errorf("%.0w parsing zstd:chunked TOC digest %q: %w", internal.ErrPartialContentNotAvailable, …)
```
where `%.0w` adds the error to `Unwrap` without affecting the text. (OTOH, as a matter of aesthetics / general downsides, I generally dislike the multi-error `Unwrap` — and we would need a comment for `%.0w`, explicit wrapping is easier to understand).

pkg/chunked/storage_linux.go
```diff
@@ -146,40 +146,48 @@ func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload *o
...
+		}
+		// If a conversion is required, ignore both TOCs and retrieve the entire layer.
+		hasZstdChunkedTOC, hasEstargzTOC = false, false
 	}
 
 	if hasZstdChunkedTOC {
 		zstdChunkedTOCDigest, err := digest.Parse(zstdChunkedTOCDigestString)
-		if err != nil {
-			return nil, fmt.Errorf("parsing zstd:chunked TOC digest %q: %w", zstdChunkedTOCDigestString, err)
+		if err == nil {
```
mtrmac, line 168: (I weakly think that if the TOC digest annotations exist but are invalid, we should fail without fallback. If nothing else, aesthetically, there are quite a few `if !convertImages` conditions here.)

pkg/chunked/storage_linux.go
```diff
@@ -146,40 +146,48 @@ func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload *o
...
+			return makeZstdChunkedDiffer(store, blobSize, zstdChunkedTOCDigest, annotations, iss, pullOptions)
```
mtrmac, line 169: https://github.com/containers/storage/issues/2115#issuecomment-2379738952 talks about the case where a registry does not support ranged requests; that should be detected somewhere in `makeZstdChunkedDiffer` and turned into the “fall back” error. Similarly for the other differ.

pkg/chunked/internal/errors.go
```diff
@@ -0,0 +1,8 @@
+package internal
+
+import (
+	"errors"
+)
+
+// ErrPartialContentNotAvailable is returned when the requested content does not support partial content retrieval
+var ErrPartialContentNotAvailable = errors.New("partial content not available")
```
mtrmac, line 8: This is correct as stated, but I think if the specific thing we want to control is _fallback_, the error name should say so. (Also the existence of `convert_images` complicates the discussion of “partial content”…)

`ErrFallbackToOrdinaryLayerDownload`?

pkg/chunked/storage_linux.go
```diff
@@ -146,40 +146,48 @@ func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload *o
 func GetDiffer(ctx context.Context, store storage.Store, blobDigest digest.Digest, blobSize int64, annotations map[string]string, iss ImageSourceSeekable) (graphdriver.Differ, error) {
```
mtrmac, line 146: The documentation of `GetDiffer` should refer to the special error type, and explain the semantics.

mtrmac (commented):
Starting from the c/image consumer, I’d prefer `copy/single.go` to not check for a specific c/storage error; c/image should define its own (private) error type for `PutBlobPartial`.

That, in turn, means that the c/storage code calling `GetDiffer` is going to already have a dependency on `pkg/chunked`; so we can put the error definition directly into `pkg/chunked`, and avoid the extra infrastructure to put it into `chunked/internal`+`chunked/toc`.

---

https://github.com/containers/storage/issues/2115 talks about adding a “require the `GetDiffer` code path to be used” option for fs-verity. We don’t need to do that right now, but how would that be added here?

It seems to me that we need a wrapping / unwrappable error mechanism (see `underlying error` in one of the comments), and then something like
```go
func GetDiffer {
   …
   differ, err := getDifferInternal()
   if fallbackMarker, ok := errIsTheFallbackMarker(err); ok && fallbackDisabled {
      err = fallbackMarker.Unwrap() // no fallback, hard failure per configuration
   }
   …
}
```

with `getDifferInternal` directly, or some of its callees, adding the marker on relevant code paths.

cgwalters:
> Starting from the c/image consumer, I’d prefer copy/single.go to not check for a specific c/storage error; c/image should define its own (private) error type for PutBlobPartial.

One thing I don't quite understand is the circular relationship between the two projects...are we getting much out of having them be in separate git repositories even?

mtrmac:
- There are c/image users who don’t use c/storage (registry clients, mirroring tools). In fact that’s the primary point of that concern — a fair bit of the generic c/image logic should not be hard-coding c/storage specifics.
- The relationship is not circular (apart from one callback from c/storage back, but I can remember only that one case where we want to invert control that way)
- The test mechanisms and CI costs are very different (sometimes in bad ways, e.g. how most of c/image tests live in the Skopeo repo); but anyway, paying the registry testing costs for c/storage, or the c/storage filesystem/driver test matrix costs for c/image, is not immediately attractive.

The relationship is not set in stone, but then again there’s a _substantial_ cost to integrate, and no obvious benefit to doing so.

pkg/chunked/storage_linux.go
giuseppe: IMO when partial pulls are not enforced, the entire feature is just an optimization and I think we should be as forgiving as possible.

In any case, we can relax it later if needed and return a hard failure for now

giuseppe:
> #2115 talks about adding a “require the `GetDiffer` code path to be used” option for fs-verity. We don’t need to do that right now, but how would that be added here?

The flag will probably look like `fsverity=disable|auto|require`.  The advantage with this PR is that `/pkg/chunked` can finally return an error that causes the pull to fail instead of falling back to the current pull mechanism.

pkg/chunked/storage_linux.go
```diff
@@ -949,6 +978,9 @@ func (c *chunkedDiffer) retrieveMissingFiles(stream ImageSourceSeekable, dirfd i
 		if _, ok := err.(ErrBadRequest); ok {
 			// If the server cannot handle at least 64 chunks in a single request, just give up.
 			if len(chunksToRequest) < 64 {
+				if c.canFallback {
```
mtrmac, line 981: Isn’t this too late? If the registry doesn’t support the feature, `readZstdChunkedManifest` / `readEstargzChunkedManifest` is going to fail, and _that_ needs to be annotated with the error type for fallback to be possible at all.

This `retrieveMissingFiles` is only called from `Differ.ApplyDiff`, that’s all after `GetDiffer` returns. (Are you proposing that `DifferApplyDiff` should _also_ have the special error type to trigger fallback? That would be possible but seems unnecessary to me.)

pkg/chunked/storage_linux.go
mtrmac: I don’t feel strongly about this. “First, do no harm” makes sense as a priority.

I’m thinking that if something is legitimately producing invalid digests, that’s _never_ going to trigger partial pulls and the authors of that image writer should be told, so that they fix things. But, to be fair, that is a notification of developers via users seeing failures, and users might have a good reason to not like that.

pkg/chunked/storage_linux.go
giuseppe: this entire code block is pure theoretical, as I don't remember ever hitting this case.  Should we just drop the `if len(chunksToRequest) < 64 {` and not have to worry about it?

giuseppe: > Isn’t this too late? If the registry doesn’t support the feature, `readZstdChunkedManifest` / `readEstargzChunkedManifest` is going to fail, and _that_ needs to be annotated with the error type for fallback to be possible at all.

in theory the registry could support less than 64 ranges in a request.  With `readZstdChunkedManifest / readEstargzChunkedManifest` we request only one

giuseppe: I added a patch to drop this condition

mtrmac: > Should we just drop the `if len(chunksToRequest) < 64 {` and not have to worry about it?

I have no opinion, I’m fine with dropping that logic … but …

pkg/chunked/storage_linux.go
```diff
@@ -947,11 +967,6 @@ func (c *chunkedDiffer) retrieveMissingFiles(stream ImageSourceSeekable, dirfd i
 		}
 
 		if _, ok := err.(ErrBadRequest); ok {
-			// If the server cannot handle at least 64 chunks in a single request, just give up.
-			if len(chunksToRequest) < 64 {
-				return err
-			}
-
```
mtrmac, line 954: AFAICS this can now loop infinitely, with a single-element `chunksToRequest`.

pkg/chunked/storage_linux.go
```diff
@@ -157,29 +159,47 @@ func GetDiffer(ctx context.Context, store storage.Store, blobDigest digest.Diges
...
+			return nil, err2
 		}
-		return makeZstdChunkedDiffer(store, blobSize, zstdChunkedTOCDigest, annotations, iss, pullOptions)
-	}
-	if hasEstargzTOC {
-		estargzTOCDigest, err := digest.Parse(estargzTOCDigestString)
-		if err != nil {
-			return nil, fmt.Errorf("parsing estargz TOC digest %q: %w", estargzTOCDigestString, err)
+		differ, err = makeZstdChunkedDiffer(store, blobSize, zstdChunkedTOCDigest, annotations, iss, pullOptions)
+		if err == nil {
```
mtrmac, line 177: If this fails because the registry does not support range requests, nothing marks the failure as eligible for fallback, AFAICS.

giuseppe: wouldn't it fail earlier when we request the TOC through a range request?  So we should treat it as a failure here?

pkg/chunked/storage_linux.go
giuseppe: It should not happen because if we got so far the server at least served the range request for the TOC, but let's be safe (could be a malicious server): I've added a check to fail if `chunksToRequest == 1`.

pkg/chunked/storage_linux.go
mtrmac: AFAICS we request the TOC inside this function, not earlier. Is that incorrect?

giuseppe: ah sorry, yes I got confused.

I've added the following snippet:

```diff
diff --git a/pkg/chunked/storage_linux.go b/pkg/chunked/storage_linux.go
index ce8f8a2ce..26e8f1aba 100644
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -196,6 +196,13 @@ func GetDiffer(ctx context.Context, store storage.Store, blobDigest digest.Diges
 	}
 
 	logrus.Debugf("Could not create differ for blob %q: %v", blobDigest, err)
+
+	// If the error is a bad request to the server, then signal to the caller that it can try a different method.  This can be done
+	// only when convert_images is disabled.
+	var badRequestErr ErrBadRequest
+	if errors.As(err, &badRequestErr) {
+		err = newErrFallbackToOrdinaryLayerDownload(err)
+	}
 	return nil, err
 }
```

mtrmac (commented):
LGTM. Thanks!

PR #2119, opened 2024-09-30
chunked: ignore ErrLayerUnknown when creating cache

giuseppe:
ignore the error if the layer is being deleted while we are processing it without a lock on the store.


Luap99 (commented):
LGTM

cgwalters:
/lgtm
(though always useful to know: did this come from a real world bug report? were you reading the code for other reasons?)

giuseppe:
> /lgtm (though always useful to know: did this come from a real world bug report? were you reading the code for other reasons?)

the flake was reported here: https://github.com/containers/podman/pull/24091#issuecomment-2379426629

Issue #2067, opened 2024-08-19
CI: `hack/git-validation.sh` fail 


Honny1:
During the lint CI job, `hack/git-validation.sh` failed with `fatal: Not a valid object name main`.

Log: https://api.cirrus-ci.com/v1/task/5169202091261952/logs/test.log

PR #2064, opened 2024-08-16
Prepare a golangci-lint installation for automatic update using renovate

Honny1:
This PR prepares a `golangci-lint` installation for automatic update using renovate.

PR #2121, opened 2024-10-01
CI: Fix `git-validation.sh`

Honny1:
This PR fixes `git-validation.sh`, failing in Cirrus CI. The problem was caused by the missing `main` branch. Cirrus CI performs a clone of a single branch with full Git history and no tags. 

Fixes: #2067

### DIFF
--- a/hack/git-validation.sh
+++ b/hack/git-validation.sh
@@ -5,5 +5,9 @@ if [ ! -x "$GIT_VALIDATION" ]; then
 	echo Try installing it with \"make install.tools\"
 	exit 1
 fi
-EPOCH_TEST_COMMIT=$(git merge-base ${DEST_BRANCH:-main} HEAD)
+
+EPOCH_TEST_COMMIT=$CIRRUS_BASE_SHA
+if [ -z "${EPOCH_TEST_COMMIT}" ]; then
+	EPOCH_TEST_COMMIT=$(git merge-base ${DEST_BRANCH:-main} HEAD)
+fi
 exec "$GIT_VALIDATION" -q -run DCO,short-subject -range "${EPOCH_TEST_COMMIT}..HEAD"


rhatdan:
LGTM
@cevich @edsantiago @lsm5 PTAL

hack/git-validation.sh
```diff
@@ -5,5 +5,9 @@ if [ ! -x "$GIT_VALIDATION" ]; then
 	echo Try installing it with \"make install.tools\"
 	exit 1
 fi
-EPOCH_TEST_COMMIT=$(git merge-base ${DEST_BRANCH:-main} HEAD)
+
+EPOCH_TEST_COMMIT=$CIRRUS_BASE_SHA
+if [ -z "${CIRRUS_CI}" ]; then
```
edsantiago, line 10: `CIRRUS_BASE_SHA` is, as we determined earlier this year, [worthless](https://github.com/containers/podman/pull/23081). 

I need to understand a lot more of this, like, what happened, what changed, why is `main` missing, what's the deal with `DEST_BRANCH`, and it's gonna take me a long time because I'm not familiar with this repo. Blocking until I get some time to understand.

lsm5:
lgtm. I'll defer to Ed and Chris (if he still bothers with PRs here) :D for final review and merge.

edsantiago (changes requested):
("Request changes" is unfair because I don't yet know what those changes are. Sorry)

hack/git-validation.sh
edsantiago: What a rabbit hole! TL;DR this appears to have been broken from the very beginning, and was only caught [by happenstance](https://github.com/containers/storage/pull/2064#pullrequestreview-2245280524) because the error was not (and still is not) causing a CI failure. 

Furthermore, the reason this works in podman is because of superconvoluted `clone_script` and other setup that has presumably been fine-tuned and tested over the years. We no longer have local expertise in such arcana. 

Proper solution is too complicated to get into here. It'll involve doing a git clone with proper branch tags for the given branch, and testing it under different branches, and under different conditions (CIRRUS_PR vs TAG vs MERGE vs PURPLEBUTTERFLIES or whatever) and different merge bases. I don't see a good ROI.

Good-enough solution: this approach, but please change as follows:
```diff
- if -z CIRRUS_CI
+ if -z EPOCH_TEST_COMMIT
```
Reason: yes, there's an implicit connection between `CIRRUS_BASE_SHA` and `CIRRUS_CI`, but it's much cleaner safer better to just actually test the actual actual, not some sort of distant cousin.

edsantiago (approved):
LGTM, thank you!

PR #2122, opened 2024-10-03
Fix copy/clone of empty value

Honny1:
This PR fixes the changed behavior in #2087. The `slices.Clone` function does not clone an empty slice as nil, breaking the tests in https://github.com/containers/common/pull/2182. Also the `maps.Clone` function clones a nil map as nil. Refactoring of map copying and functions that copy maps will be done by another PR.


Honny1:
PTAL @mtrmac

containers.go
```diff
@@ -164,17 +163,17 @@ type containerStore struct {
...
-		BigDataSizes:   maps.Clone(c.BigDataSizes),
-		BigDataDigests: maps.Clone(c.BigDataDigests),
+		BigDataNames:   copyStringSlice(c.BigDataNames),
+		BigDataSizes:   copyStringInt64Map(c.BigDataSizes),
+		BigDataDigests: copyStringDigestMap(c.BigDataDigests),
 		Created:        c.Created,
 		UIDMap:         copyIDMap(c.UIDMap),
 		GIDMap:         copyIDMap(c.GIDMap),
-		Flags:          maps.Clone(c.Flags),
+		Flags:          copyStringInterfaceMap(c.Flags),
```
mtrmac, line 176: Are the maps changes actually an improvement? I think a copy of a value with `nil`s should again have `nil`s, not allocated empty maps.

(Applies throughout)

store.go
```diff
@@ -1470,7 +1470,7 @@ func (s *store) putLayer(rlstore rwLayerStore, rlstores []roLayerStore, id, pare
 	if lOptions != nil {
 		options = *lOptions
 		options.BigData = copyLayerBigDataOptionSlice(lOptions.BigData)
-		options.Flags = maps.Clone(lOptions.Flags)
+		options.Flags = copyStringInterfaceMap(lOptions.Flags)
```
mtrmac, line 1473: I think this is unnecessary here; `nil`ness of `Flags` in itself carries no meaning, and the consumer in `layerStore.Create` is already using `copyStringInterfaceMap` to guarantee a non-`nil` value it needs.

store.go
```diff
@@ -1786,7 +1786,7 @@ func (s *store) CreateContainer(id string, names []string, image, layer, metadat
 		options.IDMappingOptions.UIDMap = copyIDMap(cOptions.IDMappingOptions.UIDMap)
 		options.IDMappingOptions.GIDMap = copyIDMap(cOptions.IDMappingOptions.GIDMap)
 		options.LabelOpts = copyStringSlice(cOptions.LabelOpts)
-		options.Flags = maps.Clone(cOptions.Flags)
+		options.Flags = copyStringInterfaceMap(cOptions.Flags)
```
mtrmac, line 1789: There is already an explicit check for `nil` below, so doing that here is unnecessary.

store.go
```diff
@@ -3688,17 +3688,28 @@ func copyStringSlice(slice []string) []string {
...
+	}
+	return maps.Clone(m)
+}
+
+func copyStringDigestMap(m map[string]digest.Digest) map[string]digest.Digest {
+	if m == nil {
+		return make(map[string]digest.Digest, len(m))
+	}
+	return maps.Clone(m)
 }
```
mtrmac, line 3706: If these functions need to be added at all:
- Please use a generic function instead of duplicating it
- Name and document the `nil` behavior clearly, so that callers see why `maps.Clone` is not being used. `createNonNilMapCloning` is a horrible name, there must be a better option – but something expressing that kind of thing at the call site.

store.go
```diff
@@ -3713,11 +3724,10 @@ func copyDigestSlice(slice []digest.Digest) []digest.Digest {
...
 // a non-pointer scalar value
 func copyStringInterfaceMap(m map[string]interface{}) map[string]interface{} {
-	ret := make(map[string]interface{}, len(m))
-	for k, v := range m {
-		ret[k] = v
+	if m == nil {
+		return make(map[string]interface{}, len(m))
 	}
-	return ret
+	return maps.Clone(m)
```
mtrmac, line 3730: (Long-term, I think it will pay off to split bug fixes from no-effect refactors into separate commits.)

mtrmac (commented):
Thanks, I didn’t realize this is turning `nil`s into empty slice allocations. Definitely worth fixing.

(I don’t think we really want to commit to a difference between `nil` and `[]…{}` as a matter of API — to the extent the c/common tests essentially test for `.Names == nil`, the semantics of `Names == nil` and `Names == []string{}` are exactly the same and the caller _should not care_ — but as an implementation concern, we should not be making those extra allocations all the time.)

containers.go
Honny1: Yes, you are right, but the previous behavior was that if the map was `nil`, it would return an empty map. This change hasn't led to a test failure yet, though, so we can use `maps.Clone` and hope that this change in behavior doesn't show up anywhere. WDYT?

store.go
Honny1: Does the `copyStringInterfaceMap` function have a specific purpose, or should it also be replaced by a generic function as you suggest in the previous comment?

containers.go
mtrmac: AFAICS none of these fields document any special semantics for `nil`, so callers should not be making any assumptions. Strictly speaking, that’s a bug in the callers, so we can do whatever we think best now that the question has come up.

I think it makes more sense to preserve `nil`ness:
- A “copy” changing that is just unexpected
- We do quite a lot of these copies (on every return from a c/storage API), so avoiding the allocations and GC cost seems desirable.

store.go
mtrmac: I don’t know that there is anything special here — this is all code written before Go had generics.

mtrmac (commented):
Thanks!

When creating separate commits for better understanding, that only works when the commits truly contain separate ideas, and precisely those.

Right now:
- The first commit says “fix” without details, but it combines an actual bug fix fix (for slice copies) with a no-change refactor to create `copyMap`; _and_ a behavior change where the `copyStringInterfaceMap` is newly called, it now allocates `Flags` where that was not allocated before. That’s three different things, and it’s not immediately obvious. (It might make sense to have a single commit like that, saying “move to new nil-ness rules”, and _documenting_ the 2/3 different rules. But users should not need to decipher that. I think splitting commits differently is easier for everyone.)
   - (BTW the documentation for `copyMap`, at that point, is not correct for non-`nil` empty maps.) 
- The second commit to remove `copyStringInterfaceMap` looks like a refactor on a cursory glance, but it actually changes behavior on `nil`. (That can be fine, but it would be better to document — except …)
- … The third commit then changes the behavior of `copyMap` _again_ on non-`nil` empty maps.

For the `Flags` field when cloning an object, that’s a _third behavior change in a row_, partially undoing the change from the second commit. That’s confusing to follow.

So I think some more restructuring (e.g. using `git rebase -i` / `git add --patch`) would be useful.

---

If we name the behaviors `copyMapPreferringNil` and `copyMapNeverNil`[1] (and `maps.Clone` in the middle), some of the `create` functions want `…NeverNil`; the object clone functions don’t want `…NeverNil`, and I don’t feel strongly between `…PreferringNil` and `maps.Clone` there (making fewer allocations vs. making a more faithful clone). 

([1] Those are not really good names, but I need names of the behaviors just for this discussion.)

---

What do we actually want to achieve here? I think that’s:
- Make c/common tests pass — could be also done in c/common
- Avoid the unwanted memory allocations
- We have learned something (about bugs I write — perhaps I’m unique, perhaps not). How do we avoid that in the future, by making bugs either impossible, or more visible?

To get only the first two, we can:
- One commit: Revert the `slices.Clone` part, no other code changes
- Optionally, one commit: Introduce `copyMap[PreferringNil]`, use it where appropriate, change nothing else
- (Not touch the `copyStringInterfaceMap` function or its remaining calls.)

To get all three, I think it would make sense to fix the codebase so that we don’t have `copy*` functions with _different_ `nil` preferences and nothing obvious in the name.
- Introduce `copySlicePreferringNil` (separate commit, or integrated into one of the following)
- One or more commits: refactor users of `copyStringSlice` / … several others to call `copySlicePreferringNil` instead, _no change to behavior_
- One commit: Use `copySlicePreferringNil` in the object copy functions — behavior change
- Introduce `copyMapNeverNil` / `newMapFrom`, refactor various users of `copy*Map` to call it instead, _no change in behavior_.
- One commit: Introduce `copyMapPreferringNil`, use it where appropriate — behavior change
- Optionally, a commit that adds comments to `maps.Clone`/`slices.Clone` about their `nil` behavior — probably overkill

I think it’s perfectly fine to have a smaller PR that does only the first two; in that case I’ll probably want to work on the other renames / refactors just to protect myself from doing the same mistake, but that’s my personal preference, not something that should block fixing the c/common tests.

Honny1:
@mtrmac I did revert the `slices.Clone` part. I will do further refactoring in a separate PR.

PR #2123, opened 2024-10-04
drivers/zfs: Handle missing layers during image deletion in ZFS file system

kimminss0:
This commit fixes the following issue:

**Issue Summary:** An error occurs when attempting to delete images with non-existing ZFS layers, resulting in unintended behavior during image removal in Podman.

**Steps to reproduce:**

**Preconditions:**

- A system running ZFS file system (tested on a FreeBSD machine).
- ZFS dataset `zroot/containers` is mounted on `/var/db/containers`
- [optional] No existing images and containers (recommended for easier identification of the pulled image in the following steps)

**Steps:**

```.bash
# Replace <image> with any other images, such as docker.io/dougrabson/hello
$ podman pull <image>

# Check the dataset name created under `zroot/containers`
$ zfs list -d 1 zroot/containers

# Manually destroy dataset
$ zfs destroy zroot/containers/<dataset_name>

# Error may occur; this might be expected
$ podman images <image>

# Manually remove the corrupted image
$ podman rmi <image> --force

# Error reproduces here - not intended
$ podman images <image>
```

**Note:** For recovery, you can recreate the dataset with the same name, and then remove the image with Podman.


rhatdan:
/approve
LGTM
@nalind @saschagrunert @dfr @giuseppe @mheon PTAL

dfr:
LGTM

kimminss0:
rebased

drivers/zfs/zfs.go
```diff
@@ -392,12 +392,19 @@ func (d *Driver) Remove(id string) error {
...
 	dataset := zfs.Dataset{Name: name}
 	err := dataset.Destroy(zfs.DestroyRecursive)
-	if err == nil {
-		d.Lock()
-		delete(d.filesystemsCache, name)
-		d.Unlock()
+	if err != nil {
+		// We must be tolerant in case the image has already been removed,
+		// for example, accidentally by hand.
+		mountpoint := d.mountPath(id)
```
kimminss0, line 398: I think I should have used `zfsPath` instead of `mountPath`, since `mountPath` is intended for the mount point of a ZFS dataset.

drivers/zfs/zfs.go
```diff
@@ -392,12 +392,18 @@ func (d *Driver) Remove(id string) error {
...
 	dataset := zfs.Dataset{Name: name}
 	err := dataset.Destroy(zfs.DestroyRecursive)
-	if err == nil {
-		d.Lock()
-		delete(d.filesystemsCache, name)
-		d.Unlock()
+	if err != nil {
+		// We must be tolerant in case the image has already been removed,
+		// for example, accidentally by hand.
+		if _, err1 := zfs.GetDataset(name); err1 == nil {
```
kimminss0, line 398: Also, `zfs.GetDataset` seems more appropriate here than `zfs.Datasets` in my previous commits, since we don't need to search its children recursively.

PR #2128, opened 2024-10-09
Refactor copy of slices and maps

Honny1:
This PR refactors:

- Operations with slices:
   - Removes duplicate copy*Slice functions using a generic copy function or replaces them with the `slices.Clone` function. 
   - Also simplifies the `stringSliceWithoutValue` function. 
   - **These changes should not change the behavior.**
- Copying maps:
   - Removes the duplicate copy*Map function using the general function `newMapFrom`. 
   - Reduces the allocation of empty maps using the `copyMapPrefferingNil` function. 
   - **This change may affect the behavior so that instead of an empty allocated map, a `nil` will be returned.**

Follows up on: #2122


store.go
```diff
@@ -1783,13 +1759,13 @@ func (s *store) CreateContainer(id string, names []string, image, layer, metadat
...
-		options.LabelOpts = copyStringSlice(cOptions.LabelOpts)
+		options.IDMappingOptions.UIDMap = copySlicePreferringNil(cOptions.IDMappingOptions.UIDMap)
+		options.IDMappingOptions.GIDMap = copySlicePreferringNil(cOptions.IDMappingOptions.GIDMap)
+		options.LabelOpts = copySlicePreferringNil(cOptions.LabelOpts)
 		options.Flags = maps.Clone(cOptions.Flags)
-		options.MountOpts = copyStringSlice(cOptions.MountOpts)
+		options.MountOpts = copySlicePreferringNil(cOptions.MountOpts)
 		options.StorageOpt = copyStringStringMap(cOptions.StorageOpt)
-		options.BigData = copyContainerBigDataOptionSlice(cOptions.BigData)
+		options.BigData = slices.Clone(cOptions.BigData)
```
kolyshkin: `slices.Clone` makes a shallow clone, i.e. `Data` slice is not really copied with this change.

kolyshkin (changes requested):

rhatdan:
/approved
LGTM

store.go
Honny1: The previous `copyContainerBigDataOptionSlice` function used the `slices.Clone` function to copy data. So I would expect similar behavior.

nalind: It used `slices.Clone()` to copy the `Data` field in each element in the slice.  The godoc for `slices.Clone()` notes that it only copies using assignment, so a new slice would not be created for `Data` fields.

rhatdan:
/approve
@kolyshkin @nalind @mtrmac @Luap99 @flouthoc @giuseppe PTAL

nalind:
LGTM

PR #2129, opened 2024-10-10
Fix assignment of quota project IDs

mheon:
This PR reverts #1921 which completely removed the PROJINHERIT flag from our quota support. That did not actually fix the problem, it just papered over a few of the cracks. The actual problem is that configuring quota support on the system using our recommended commands sets the PROJINHERIT flag on the top-level directory in some (but not all, for unclear reasons?) cases. We set a project ID on the base directory to give us a base value for subsequent ID assignments, but if the base directory has PROJINHERIT that base ID prevents us from setting new IDs on any subdirectories. Strip the flag from the top-level directory if present to resolve.

There is a second problem, in that Podman is choosing the wrong directory to assign quotas to, but that must be fixed in Podman after this merges.

Fixes https://issues.redhat.com/browse/RHEL-18038

### DIFF
--- a/drivers/quota/projectquota_supported.go
+++ b/drivers/quota/projectquota_supported.go
@@ -18,6 +18,16 @@ package quota
 #include <linux/quota.h>
 #include <linux/dqblk_xfs.h>
 
+#ifndef FS_XFLAG_PROJINHERIT
+struct fsxattr {
+	__u32		fsx_xflags;
+	__u32		fsx_extsize;
+	__u32		fsx_nextents;
+	__u32		fsx_projid;
+	unsigned char	fsx_pad[12];
+};
+#define FS_XFLAG_PROJINHERIT	0x00000200
+#endif
 #ifndef FS_IOC_FSGETXATTR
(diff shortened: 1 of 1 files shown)


rhatdan:
/approved
LGTM
@giuseppe @nalind @mtrmac @saschagrunert @flouthoc PTAL

mtrmac (commented):
LGTM, at least this root cause explanation is consistent with my earlier research in https://github.com/containers/storage/pull/1921#issuecomment-2115947327 .

I’d prefer failing on unexpected setups instead of `stripProjectInherit`: we either have a bug somewhere else that should be detected and fixed, or we have Podman users with some setups/expectations we are not planning for, and we should find out what they are and discuss how to deal with them.

But this PR seems to be an improvement over the current state either way.

mtrmac:
Yes, I was thinking to report an error, and force the user to investigate. But that has the obvious downsides — on balance I’m fine with what this PR is doing.

giuseppe (approved):
LGTM

mtrmac:
/lgtm

for the tools

PR #2130, opened 2024-10-10
Correctly compute UncompressedSize on zstd:chunked pull, don’t set it on estargz

mtrmac:
The current value obtained by summing the sizes of regular file contents does not match the size of the uncompressed layer tarball.

We don't have a convenient source to compute the correct size for estargz without pulling the full layer and defeating the point.

For recent zstd:chunked images, we have the full tar-split, so we can compute the correct size; <s>for now, this doesn't do that. That might slow down image size computation.</s>

---

<s>Absolutely untested, and we probably do want the tar-split-based computation to happen.</s>

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -2513,7 +2513,7 @@ func (r *layerStore) applyDiffFromStagingDirectory(id string, diffOutput *driver
 		return err
 	}
 
-	if len(diffOutput.TarSplit) != 0 {
+	if diffOutput.TarSplit != nil {
 		tsdata := bytes.Buffer{}
 		compressor, err := pgzip.NewWriterLevel(&tsdata, pgzip.BestSpeed)
 		if err != nil {
--- a/layers.go
+++ b/layers.go
@@ -136,9 +136,12 @@ type Layer struct {
 	TOCDigest digest.Digest `json:"toc-digest,omitempty"`
 
 	// UncompressedSize is the length of the blob that was last passed to
-	// ApplyDiff() or create(), after we decompressed it.  If
-	// UncompressedDigest is not set, this should be treated as if it were
-	// an uninitialized value.
+	// ApplyDiff() or create(), after we decompressed it.
+	//
+	// - If UncompressedDigest is set, this must be set to a valid value.
+	// - Otherwise, if TOCDigest is set, this is either valid or -1.
+	// - If neither of this digests is set, this should be treated as if it were
+	//   an uninitialized value.
 	UncompressedSize int64 `json:"diff-size,omitempty"`
 
 	// CompressionType is the type of compression which we detected on the blob
@@ -1214,8 +1217,8 @@ func (r *layerStore) Size(name string) (int64, error) {
 	// We use the presence of a non-empty digest as an indicator that the size value was intentionally set, and that
 	// a zero value is not just present because it was never set to anything else (which can happen if the layer was
 	// created by a version of this library that didn't keep track of digest and size information).
-	if layer.TOCDigest != "" || layer.UncompressedDigest != "" {
-		return layer.UncompressedSize, nil
+	if layer.UncompressedDigest != "" || layer.TOCDigest != "" {
+		return layer.UncompressedSize, nil // This may return -1 if only TOCDigest is set
 	}
 	return -1, nil
 }
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -288,6 +288,36 @@ func ensureTOCMatchesTarSplit(toc *internal.TOC, tarSplit []byte) error {
 	return nil
 }
 
+// tarSizeFromTarSplit computes the total tarball size, using only the tarSplit metadata
+func tarSizeFromTarSplit(tarSplit []byte) (int64, error) {
+	var res int64 = 0
+
+	unpacker := storage.NewJSONUnpacker(bytes.NewReader(tarSplit))
+	for {
+		entry, err := unpacker.Next()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return -1, fmt.Errorf("reading tar-split entries: %w", err)
+		}
+		switch entry.Type {
+		case storage.SegmentType:
+			res += int64(len(entry.Payload))
+		case storage.FileType:
+			// entry.Size is the “logical size”, which might not be the physical size for sparse entries;
+			// but the way tar-split/tar/asm.WriteOutputTarStream combines FileType entries and returned files contents,
+			// sparse files are not supported.
+			// Also https://github.com/opencontainers/image-spec/blob/main/layer.md says
+			// > Sparse files SHOULD NOT be used because they lack consistent support across tar implementations.
+			res += entry.Size
+		default:
+			return -1, fmt.Errorf("unexpected tar-split entry type %q", entry.Type)
+		}
+	}
+	return res, nil
+}
+
 // ensureTimePointersMatch ensures that a and b are equal
 func ensureTimePointersMatch(a, b *time.Time) error {
 	// We didn’t always use “timeIfNotZero” when creating the TOC, so treat time.IsZero the same as nil.
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -139,7 +139,7 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, 
 }
 
 // readZstdChunkedManifest reads the zstd:chunked manifest from the seekable stream blobStream.
-// Returns (manifest blob, parsed manifest, tar-split blob, manifest offset).
+// Returns (manifest blob, parsed manifest, tar-split blob or nil, manifest offset).
 func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Digest, annotations map[string]string) ([]byte, *internal.TOC, []byte, int64, error) {
 	offsetMetadata := annotations[internal.ManifestInfoKey]
 	if offsetMetadata == "" {
@@ -214,7 +214,7 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
 		return nil, nil, nil, 0, fmt.Errorf("unmarshaling TOC: %w", err)
 	}
 
-	decodedTarSplit := []byte{}
+	var decodedTarSplit []byte = nil
 	if toc.TarSplitDigest != "" {
 		if tarSplitChunk.Offset <= 0 {
 			return nil, nil, nil, 0, fmt.Errorf("TOC requires a tar-split, but the %s annotation does not describe a position", internal.TarSplitInfoKey)
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -1153,7 +1153,7 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 
 	var compressedDigest digest.Digest
 	var uncompressedDigest digest.Digest
-	var convertedBlobSize int64
+	var uncompressedSize int64 = -1
 
 	if c.convertToZstdChunked {
 		fd, err := unix.Open(dest, unix.O_TMPFILE|unix.O_RDWR|unix.O_CLOEXEC, 0o600)
@@ -1185,7 +1185,7 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 		if err != nil {
 			return graphdriver.DriverWithDifferOutput{}, err
 		}
-		convertedBlobSize = tarSize
+		uncompressedSize = tarSize
 		// fileSource is a O_TMPFILE file descriptor, so we
 		// need to keep it open until the entire file is processed.
 		defer fileSource.Close()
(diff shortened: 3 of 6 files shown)


mtrmac:
Cc: @giuseppe . This should fix https://github.com/containers/skopeo/issues/2437#issuecomment-2404991419 .

edsantiago:
I picked this into my pet pr, and it works slightly better, but now it breaks in the podman [additional-store test](https://api.cirrus-ci.com/v1/artifact/task/5006990613676032/html/sys-podman-fedora-40-root-host-sqlite.log.html#t--00041):
```
...
<+011ms> # # podman pull -q quay.io/libpod/testimage:20241010
<+070ms> # 9c6e6209f54a048342fd899e1e0885be64dfc836ed3664b33d6d07bcb4fc1c51
         # 263622a183c94c2433a43be5464a954b9c4e5b0a77cb177f6fe60dacfff66f80
         # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
         # #|     FAIL: pull -q quay.io/libpod/testimage:20241010, using storage.conf
         # #| expected: '263622a183c94c2433a43be5464a954b9c4e5b0a77cb177f6fe60dacfff66f80'
         # #|   actual: '9c6e6209f54a048342fd899e1e0885be64dfc836ed3664b33d6d07bcb4fc1c51'
         # #|         > '263622a183c94c2433a43be5464a954b9c4e5b0a77cb177f6fe60dacfff66f80'
         # #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Basically, `podman pull` now spits out two hashes instead of one. Fails very reliably on my laptop too:
```console
$ hack/bats -T --root 010
...
same failure, down to the same hashes
```

edsantiago:
Skopeo is involved, though (for the copy to the store dir), and I haven't rebuilt skopeo, so it's possible that the bad thing is happening in that step.

mtrmac:
Thanks! I can‘t see how the two-ID output can happen.

I’ll set up a VM and test (a better version of) this PR, as well as those tests, tomorrow.

mtrmac:
PR updated to compute the size correctly for zstd:chunked with tar-split, manually tested per https://github.com/containers/skopeo/issues/2437#issuecomment-2404991419 (and inspecting the created layer metadata).

Ready for review.

mtrmac:
> ```
> $ hack/bats -T --root 010
> ...
> same failure, down to the same hashes
> ```

Testing that on a slightly unclean VM, Podman commit 2aacd4e212525db4ee06be8e44e4405400d4df9d + this c/storage fix passes 010 successfully in `make localsystem`; the command above fails with a lot of unexpected, different, errors, e.g. about `localhost/podman-pause`. It’s very possible there’s something wrong about that environment, I know very little about Podman tests.

And, on the command line, `podman pull -q quay.io/libpod/testimage:20241010` outputs just `9c6e6209f54a048342fd899e1e0885be64dfc836ed3664b33d6d07bcb4fc1c51`

mtrmac:
Scratch that last part, that’s not invoking the tested version.

edsantiago:
Ugh. The podman-pause thing is my fault, I couldn't find a good solution to a nasty problem. Please try this: 
https://github.com/containers/podman/pull/23275/commits/9d8e3b058bc22758d394b4297090f0c09c7b7e9e

pkg/chunked/compression_linux.go
```diff
@@ -288,6 +288,36 @@ func ensureTOCMatchesTarSplit(toc *internal.TOC, tarSplit []byte) error {
 	return nil
 }
 
+// tarSizeFromTarSplit computes the total tarball size, computing only tarSplit
```
cgwalters, line 291:
```suggestion
// tarSizeFromTarSplit computes the total tarball size using only the tarSplit metadata
```

pkg/chunked/compression_linux.go
```diff
@@ -288,6 +288,36 @@ func ensureTOCMatchesTarSplit(toc *internal.TOC, tarSplit []byte) error {
...
+			}
+			return -1, fmt.Errorf("reading tar-split entries: %w", err)
+		}
+		switch entry.Type {
+		case storage.SegmentType:
+			res += int64(len(entry.Payload))
+		case storage.FileType:
+			// entry.Size is the “logical size”, which might not be the physical size for sparse entries;
+			// but the way tar-split/tar/asm.WriteOutputTarStream combines FileType entries and returned files contents,
+			// sparse files are not supported.
```
cgwalters, line 310: But we don't error out today?

cgwalters (commented):
What about adding the uncompressed size to the ToC? Seems like it'd possibly simplify things?

This looks sane to me but I only gave it a superficial review.

pkg/chunked/compression_linux.go
mtrmac: If we should be erroring out, that’s something that should happen at tar-split construction.

The physical size outright isn’t available in the tar-split format, so here this consumer is only documenting the impact of that assumption.

(As is, sparse files are barely supported by archive/tar or the tar-split fork: they are indicated _either_ by a special file type, or as a regular file a special PAX record; and there is no API to expose the data / hole segments. Also, `c/storage/pkg/archive` does not special-case them at all.)

mtrmac:
> What about adding the uncompressed size to the ToC? Seems like it'd possibly simplify things?

- It might be a bit late for that. The format is documented “as normal” with no warning that we might want to change the format and break images. IIRC we sort of did that when introducing the tar-split element, but already at that time we had to relent and re-introduce at least the previous level of support of the no-tar-split format. Still, if this were the only concern, we should at least document the idea around `pkg/chunked/internal.TOC` for future consideration.
- If we did include the size in the TOC, we would have to worry about producers recording incorrect sizes. I mean, we have to worry about that anyway… but the tar-split metadata must be correct exactly for the operation which primarily relies on the uncompressed size value, so that is not introducing a _new_ assumption. Compare also how we have 3 different sources of metadata in zstd:chunked files — it seems to me that reducing redundancy is a better trade-off here anyway.

Anyway, I’m open to adding a comment to the `TOC` type.

mtrmac:
@edsantiago 

On a VM manually building things:

All of:
- Podman `main` commit 11ab0b72465432d93c5421d566ad0501f072a954 without this PR
- https://github.com/containers/podman/commit/9d8e3b058bc22758d394b4297090f0c09c7b7e9e without this PR
- Podman `main` commit 11ab0b72465432d93c5421d566ad0501f072a954 with this PR
- https://github.com/containers/podman/commit/9d8e3b058bc22758d394b4297090f0c09c7b7e9e with this PR

pass `make localsystem`, apart from `[500] podman networking: port with --userns=keep-id for rootless or --uidmap=* for rootful` failing to find `rootlessport`, very likely unrelated.

Similarly, https://github.com/containers/podman/commit/9d8e3b058bc22758d394b4297090f0c09c7b7e9e passes `hack/bats -T --root 010`, with or without this PR; and the podman `main` commit fails, with or without this PR.

Either way, I haven’t been able to reproduce this:
> Basically, `podman pull` now spits out two hashes instead of one.

I have now filed https://github.com/containers/podman/pull/24265 to exercise that in the usual environment.

Is there anything else I should be looking at?

edsantiago:
@mtrmac please forgive me, I've switched gears for today and have commitments I must attend to. The crucial element I don't see in your comment is running the full CI suite using `testimage:20241010` (or `1009`). Those two images were pushed with `zstd` and are the ones that cause all the problems. Thing is, the failing tests use `skopeo` for one step, and I'm pretty sure you might also need to patch that. I never was able to figure that out, and can't look into it now.

mtrmac:
Thanks, I was looking for references to `testimage:20241010` and I couldn’t find any. I’ll revisit tomorrow.

edsantiago:
Try (untested!) in podman dir
```
$ grep -R 20241011 . | xargs sed -i -e 's/20241011/20241010/g'
```
Try running the `010` tests with that change, they should fail.

Then try running with that change _and_ your new code. If they pass, push a podman PR and see how CI does.

Again, I'm speaking completely off the cuff here, no testing or validating, YMMV, please double-check everything for sanity.

edsantiago:
Oops! I screwed up; I thought #24238 had merged. It hasn't, so `grep 20241011` is a NOP. 

So, for your purposes, change `20241011` above to `20240123`, and import the 11-12-13 changes to `320-system-df`, and skip (or ignore) the `apiv2` tests.

giuseppe (approved):
LGTM

edsantiago:
And, once again, testing in podman, [weird errors](https://api.cirrus-ci.com/v1/artifact/task/6567703043899392/html/sys-podman-fedora-40-root-host-sqlite.log.html) including the two-ID one.

mtrmac:
Thanks!

The zstd:chunked implementation changes how image IDs are computed: So far, (for schema2 and OCI), the image ID == config digest. With zstd:chunked, partially-pulled images and fully-pulled images (depending on the code path and other options) have different IDs, there can be 2^(layers) different IDs for “the same” image.

That’s going to break assumptions, but it was ~unavoidable for security. 

The image ID equality checks will probably need to be replaced, maybe by comparing the config digest. (Alternatively, the tests could ensure that both tested writes use the “full” path, but testing something closer to the default seems preferable to me.)

---

> 41 [010] podman pull image with additional store

Ah, I see what is going on here. That’s indeed zstd:chunked-related.

The `skopeo copy` from storage to storage means one store contains a partially-pulled image, one a fully-(stored) image

Then, after a pull, `Runtime.imagesIDsForManifest` assumes that the image stored in all stores will have the same ID, and that’s no longer the case.

There’s really no benefit to returning multiple values from that method anyway, I’ll fix that part. Still, the image ID equality check will probably need changing.

> 42 [010] podman images with concurrent removal

fails because `localhost/podman-pause` exists, that seems unrelated.

>  124 [120] podman load - from URL

Image ID changed, partially-pulled image vs. fully-(loaded) image. The image ID equality test will need to be replace.

edsantiago:
Thank you. That's really pretty huge.

layers.go
```diff
@@ -136,9 +136,12 @@ type Layer struct {
...
 	// UncompressedSize is the length of the blob that was last passed to
-	// ApplyDiff() or create(), after we decompressed it.  If
-	// UncompressedDigest is not set, this should be treated as if it were
-	// an uninitialized value.
+	// ApplyDiff() or create(), after we decompressed it.
+	//
+	// - If UncompressedDigest is set, this must be set to a valid value.
+	// - Otherwise, if TOCDigest is set, this is either valid or -1.
+	// - If neither of this digests is set, this should be treated as if it were
+	//   an uninitialized value.
```
kolyshkin, line 144: You need indentation for list items (same as for code blocks, i.e. an extra space).

```suggestion
	//  - If UncompressedDigest is set, this must be set to a valid value.
	//  - Otherwise, if TOCDigest is set, this is either valid or -1.
	//  - If neither of this digests is set, this should be treated as if it were
	//    an uninitialized value.
```

pkg/chunked/storage_linux.go
```diff
@@ -229,24 +231,33 @@ func makeZstdChunkedDiffer(store storage.Store, blobSize int64, tocDigest digest
 	if err != nil {
 		return nil, fmt.Errorf("read zstd:chunked manifest: %w", err)
 	}
+	var uncompressedTarSize int64 = -1
+	if tarSplit != nil {
+		uncompressedTarSize, err = tarSizeFromTarSplit(tarSplit)
+		if err != nil {
+			return nil, fmt.Errorf("computing size from tar-split")
```
kolyshkin, line 238: nit: `errors.New`, or `fmt.Errorf("computing site from tar-split: %w", err)`, or just `err`.

mtrmac:
> So, for your purposes, change `20241011` above to `20240123`, and import the 11-12-13 changes to `320-system-df`, and skip (or ignore) the `apiv2` tests.

Note to self (and in case it is useful to others): Commit “DO NOT MERGE: Test with a zstd:chunked testimage” in https://github.com/containers/podman/pull/24287 .

layers.go
mtrmac: Yeah, `gofmt` didn’t want to help with this (and struct field comments are formatted in HTML just as an uninterpreted code block), so I punted.

You’re right, this is the right thing to do. Fixed in https://github.com/containers/storage/pull/2136 .

pkg/chunked/storage_linux.go
mtrmac: Thanks! Fixed in https://github.com/containers/storage/pull/2136 .

cgwalters:
> The zstd:chunked implementation changes how image IDs are computed: So far, (for schema2 and OCI), the image ID == config digest. With zstd:chunked, partially-pulled images and fully-pulled images (depending on the code path and other options) have different IDs, there can be 2^(layers) different IDs for “the same” image.

Is there more info on that? Sounds worth adding to `docs/containers-storage-composefs.md` perhaps.

mtrmac:
@cgwalters That’s not directly a c/storage property; it’s a c/image choice of a a deterministic image ID, in https://github.com/containers/image/blob/cba49408c0ea237a6aa6dba5e81b74f4a8f23480/storage/storage_dest.go#L671-L684 .

Yes, we might eventually need a user-facing explanation; I’m not sure where is a good place for it, the internal c/* projects are not too likely to be known to users. Maybe a Podman blog post.

PR #2134, opened 2024-10-14
Use securejoin.SecureJoin when forming userns paths

mheon:
We need to read /etc/passwd and /etc/group in the container to get an idea of how many UIDs and GIDs we need to allocate for a user namespace when `--userns=auto` is specified. We were forming paths for these using filepath.Join, which is not safe for paths within a container, resulting in this CVE allowing crafted symlinks in the container to access paths on the host instead.

Addresses CVE-2024-9676

### DIFF
--- a/userns.go
+++ b/userns.go
@@ -1,18 +1,21 @@
+//go:build linux
+
 package storage
 
 import (
 	"fmt"
 	"os"
 	"os/user"
-	"path/filepath"
 	"strconv"
 
 	drivers "github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/unshare"
 	"github.com/containers/storage/types"
+	securejoin "github.com/cyphar/filepath-securejoin"
 	libcontainerUser "github.com/moby/sys/user"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 // getAdditionalSubIDs looks up the additional IDs configured for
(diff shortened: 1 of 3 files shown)


userns.go
```diff
@@ -1,18 +1,21 @@
+//go:build linux
+
```
Luap99, line 2: Just to make sure does this still build the freebsd build in podman?

Another thing if we now that this will only ever be used on linux we can rename userns.go to userns_linux.go which I think is a bit better, also for the test userns_linux_test.go

mtrmac (commented):
LGTM

Luap99 (commented):
just some minor nit, the fix itself looks good to me though I haven't tried to reproduce it

userns.go
mheon: Should be fine, FreeBSD doesn't support user namespaces last I checked.

mheon:
Cross builds:
```
env CGO_ENABLED=0 GOOS=freebsd GOARCH=amd64 go build -compiler gc -tags " " ./...
env CGO_ENABLED=0 GOOS=freebsd GOARCH=amd64 go build -compiler gc -tags " " -o containers-storage.freebsd.amd64 ./cmd/containers-storage
env CGO_ENABLED=0 GOOS=freebsd GOARCH=arm64 go build -compiler gc -tags " " ./...
env CGO_ENABLED=0 GOOS=freebsd GOARCH=arm64 go build -compiler gc -tags " " -o containers-storage.freebsd.arm64 ./cmd/containers-storage
```

So I think we're fine

Luap99 (commented):
yes this looks good then, thanks

LGTM

nalind:
LGTM

mheon:
@rhatdan PTAL and merge

mtrmac:
/lgtm

Three reviewers should be enough :)

nalind:
This cherry-picks cleanly onto the release-1.55 branch if #2105 and #2109 are cherry-picked first.

userns.go
```diff
@@ -309,3 +331,14 @@ func getAutoUserNSIDMappings(
 	gidMap := append(availableGIDs.zip(requestedContainerGIDs), additionalGIDMappings...)
 	return uidMap, gidMap, nil
 }
+
+// Securely open (read-only) a file in a container mount.
+func secureOpen(containerMount, file string) (*os.File, error) {
+	tmpFile, err := securejoin.OpenInRoot(containerMount, file)
```
kwilczynski, line 337: @mheon, would it be possible to use the older API here?

Why? Using `filepath-securejoin` **3.x** would bring the internal use of the slices package, which requires Go 1.21 or newer. Which is something we don't have for when building images for older releases of OpenShift.

mheon:
I like having the new version on main, but replacing the new API with a
securejoin and os.Open for older versions should be fine

kwilczynski:
> I like having the new version on main, but replacing the new API with a securejoin and os.Open for older versions should be fine

@mheon, sounds good!

I will made the necessary changes and then ask you for a quick review, to make sure everything checks out.

userns.go
kwilczynski: Related:

- https://github.com/cri-o/cri-o/pull/8677#pullrequestreview-2372994785

PR #2135, opened 2024-10-14
[release-1.55] backport fix for CVE-2026-9676

mheon:
Also drags in a backport of #2105 to make the diff apply cleanly. I don't think there's harm in doing so.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -120,7 +120,7 @@ lint_task:
     env:
         CIRRUS_WORKING_DIR: "/go/src/github.com/containers/storage"
     container:
-        image: golang
+        image: golang:1.21
     modules_cache:
         fingerprint_script: cat go.sum
         folder: $GOPATH/pkg/mod
--- a/userns.go
+++ b/userns.go
@@ -1,18 +1,21 @@
+//go:build linux
+
 package storage
 
 import (
 	"fmt"
 	"os"
 	"os/user"
-	"path/filepath"
 	"strconv"
 
 	drivers "github.com/containers/storage/drivers"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/unshare"
 	"github.com/containers/storage/types"
+	securejoin "github.com/cyphar/filepath-securejoin"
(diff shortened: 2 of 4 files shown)


mheon:
This takes care of release-1.55 which takes care of the current Podman and Buildah release branches

mheon:
@nalind Any idea what's going on with lint over here? I'm seeing the same locally, but I'm pretty sure they aren't caused by this PR either.

nalind:
I'm not sure why, but `.cirrus.yml` is specifying `golang` as the image for some tasks (including the linting task) and `golang:1.21` for others.  Changing it produces a better result for me, but not quite passing.

mheon:
Do we have a minimum supported Go version of 1.21 for the branch? My preliminary investigation suggests 1.22 might pass.

nalind:
It tends to be dictated by its consumers.  Organization-wide, I think we tend to either use "the oldest currently supported golang", or "the version available on the distributions where we still need to be able to build".  The buildah 1.37 branch and podman 5.2 branches appear to be using 1.21, fwiw.

mheon:
A lot of these seem completely erroneous. It's complaining that `ContainerOptions` doesn't have certain fields, for example - those are brought in by the included `types.IDMappingOptions` in `ContainerOptions`. It'd be a compile error if those were missing. I think most of them boil down to things that are allowed by the compiler, but the linter is refusing to handle.

mheon:
Well, whatever I was doing locally was clearly incorrect, because 1.21 failed locally but seems to pass here. Fascinating.

nalind:
LGTM

mheon:
@containers/storage-maintainers PTAL

kolyshkin:
I guess there's a typo and it should be CVE-2024-9676.

mheon:
That is indeed a typo. At least I got the identifier within the year correct.

lsm5:
Need a new release cut with this and vendored into the usual places.

mheon:
Ack, I'll start vendoring this afternoon.

danishprakash:
It would be helpful to retitle this to CVE-2024-9676 for those trying to search for this CVE in the repo.

PR #2145, opened 2024-10-23
composefs: use mount from file where supported

giuseppe:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=fb176750266a3d7f42ebdcf28e8ba40350b27847 adds the possibility to mount a EROFS file system directly from its backing file instead of using a loopback device.


giuseppe:
@cgwalters PTAL

cgwalters (approved):
/lgtm

PR #2146, opened 2024-10-23
[release-1.51] Backport CVE-2024-9676 fix

mheon:
This backports #2134 to release-1.51 (with some bugfix patches from Giuseppe to make it apply cleanly). release-1.51 is used by Podman 4.9.x, which we are still supported on F39 for a couple more months.

### DIFF
--- a/userns.go
+++ b/userns.go
@@ -99,7 +99,7 @@ func parseMountedFiles(containerMount, passwdFile, groupFile string) uint32 {
 		for _, u := range users {
 			// Skip the "nobody" user otherwise we end up with 65536
 			// ids with most images
-			if u.Name == "nobody" {
+			if u.Name == "nobody" || u.Name == "nogroup" {
 				continue
 			}
 			if u.Uid > size && u.Uid != nobodyUser {
@@ -114,7 +114,7 @@ func parseMountedFiles(containerMount, passwdFile, groupFile string) uint32 {
 	groups, err := libcontainerUser.ParseGroupFile(groupFile)
 	if err == nil {
 		for _, g := range groups {
-			if g.Name == "nobody" {
+			if g.Name == "nobody" || g.Name == "nogroup" {
 				continue
 			}
 			if g.Gid > size && g.Gid != nobodyUser {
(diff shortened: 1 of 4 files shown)


mheon:
Too old securejoin to have the new API, reverting to the old one

userns.go
```diff
@@ -309,3 +331,13 @@ func getAutoUserNSIDMappings(
...
+
+// Securely open (read-only) a file in a container mount.
+func secureOpen(containerMount, file string) (*os.File, error) {
+	finalPath, err := securejoin.SecureJoin(containerMount, file)
+	if err != nil {
+		return nil, err
+	}
+
+	return os.Open(finalPath)
+}
```
kwilczynski, line 343: @mheon, would it be possible to use a similar code to the one from:

- https://github.com/cri-o/cri-o/pull/8677/files#r1803416996

Just to make things consistent.

mheon: Done

TomSweeneyRedHat:
LGTM

userns.go
cyphar: FYI, this is not sufficient if there is a risk of a racing process changing the filesystem from underneath you. You should instead do something like (untested):

```go
func somewhatSafeOpen(root, unsafePath string, mode int) (_ *os.File, Err error) {
	if mode&(os.O_CREAT|os.O_TRUNC) != 0 {
		return nil, fmt.Errorf("O_CREAT|O_TRUNC cannot be used safely")
	}
	path, err := securejoin.SecureJoin(root, unsafePath)
	if err != nil {
		return nil, err
	}
	handle, err := os.OpenFile(path, mode)
	if err != nil {
		return nil, err
	}
	defer func() {
		if Err != nil {
			_ = handle.Close()
		}
	}()
	// At this point, we aren't sure that the handle we grabbed is the same
	// path we expected. So double-check it using procfs.
	actualPath, err := os.Readlink(fmt.Sprintf("/proc/self/fd/%d", handle.Fd()))
	if err != nil {
		return nil, fmt.Errorf("failed to check handle path for %q: %w", path, err)
	}
	if path != actualPath {
		return nil, fmt.Errorf("container breakout detected: tried to open %q but instead opened %q", path, actualPath)
	}
	return handle, nil
}
```

PR #2147, opened 2024-10-23
fsdiff: mount both layers are read-only

giuseppe:
there is no reason to mount the first layer as writeable.  If fixes mounting layers that are in a read-only location like an additional image store, or are backed by composefs.


vrothberg (commented):
LGTM, nice catch

PR #2148, opened 2024-10-23
overlay: use private directory for composefs mounts

giuseppe:
similarly to what we are doing for the merged directory we need to use private directories for the composefs temporary mounts.  Otherwise when the id is in an additional store, we attempt to perform the mount without holding an exclusive lock there.

Could help with https://github.com/containers/storage/issues/2042.  Can't confirm as I wasn't able to find a reproducer yet.

@cgwalters PTAL, another fix for composefs

### DIFF
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -1579,7 +1579,6 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 	}()
 
 	composeFsLayers := []string{}
-	composeFsLayersDir := filepath.Join(dir, "composefs-layers")
 	maybeAddComposefsMount := func(lowerID string, i int, readWrite bool) (string, error) {
 		composefsBlob := d.getComposefsData(lowerID)
 		if err := fileutils.Exists(composefsBlob); err != nil {
@@ -1594,7 +1593,7 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 			return "", fmt.Errorf("cannot mount a composefs layer as writeable")
 		}
 
-		dest := filepath.Join(composeFsLayersDir, strconv.Itoa(i))
+		dest := d.getStorePrivateDirectory(id, dir, fmt.Sprintf("composefs-layers/%d", i), inAdditionalStore)
 		if err := os.MkdirAll(dest, 0o700); err != nil {
 			return "", err
 		}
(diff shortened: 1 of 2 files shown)


drivers/overlay/overlay.go
```diff
@@ -1580,6 +1580,12 @@ func (d *Driver) get(id string, disableShifting bool, options graphdriver.MountO
 
 	composeFsLayers := []string{}
 	composeFsLayersDir := filepath.Join(dir, "composefs-layers")
+	if inAdditionalStore {
+		// Similarly to getMergedDir, we need to make sure we use a private directory for
+		// the composefs layers if in an additional store.  Otherwise, we could end up attempting
+		// a mount without having an exclusive lock on the additional store.
+		composeFsLayersDir = filepath.Join(d.runhome, id, "composefs-layers")
```
cgwalters, line 1587: There's a lot of hard to follow logic for me happening here. When I'm reviewing this PR I went to ask "ok what is `dir` here" - and it's the thing that gives us `inAdditionalStore`.

Then we have `mergedDir` which is computed using `inAdditionalStore` - and is seemingly written to solve *exactly* this issue, it says:

```
	// To prevent races with other processes mounting or unmounting the layer,
	// use a private directory under the main store rundir, not the "merged" 
```

So...would things be cleaner if we did something like:

```
var idDir string
if inAdditionalStore {
   idDir = path.Join(d.runhome, id)
} else {
   idDir = dir
}
```

There's probably a better name than `idDir`.

rhatdan:
LGTM

cgwalters (commented):
I can believe the patch fixes or helps with the reported bug, but I don't know this code well enough to just drop a lgtm myself yet, and I have some suggested cleanups.

drivers/overlay/overlay.go
cgwalters: Or maybe `getMergedDir()` changes to `getMountDir(..., "merged")` and `getMountDir(..., "composefs-layers")

giuseppe: thanks for the feedback.

Pushed a new version to refactor the `if inAdditionalStore { idDir = path.Join(d.runhome, id)} else {   idDir = dir }` in a separate function

vrothberg:
@giuseppe lint is failing

giuseppe:
rebased, it passes now

vrothberg (approved):
/lgtm

PR #2155, opened 2024-10-30
Compute an uncompressed digest for chunked layers

mtrmac:
… to ensure we can always use traditional layer IDs and image IDs, and use the uncompressed digests to avoid zstd:chunked view ambiguity, in c/image.

<s>Absolutely untested.</s>

### DIFF
--- a/docs/containers-storage.conf.5.md
+++ b/docs/containers-storage.conf.5.md
@@ -124,6 +124,27 @@ The `storage.options.pull_options` table supports the following keys:
   It is an expensive operation so it is not enabled by default.
   This is a "string bool": "false"|"true" (cannot be native TOML boolean)
 
+**insecure_allow_unpredictable_image_contents="false"|"true"**
+  This should _almost never_ be set.
+  It allows partial pulls of images without guaranteeing that "partial
+  pulls" and non-partial pulls both result in consistent image contents.
+  This allows pulling estargz images and early versions of zstd:chunked images;
+  otherwise, these layers always use the traditional non-partial pull path.
+
+  This option should be enabled _extremely_ rarely, only if _all_ images that could
+  EVER be conceivably pulled on this system are _guaranteed_ (e.g. using a signature policy)
+  to come from a build system trusted to never attack image integrity.
+
+  If this consistency enforcement were disabled, malicious images could be built
+  in a way designed to evade other audit mechanisms, so presence of most other audit
+  mechanisms is not a replacement for the above-mentioned need for all images to come
+  from a trusted build system.
+
+  As a side effect, enabling this option will also make image IDs unpredictable
+  (usually not equal to the traditional value matching the config digest).
+
+  This is a "string bool": "false"|"true" (cannot be native TOML boolean)
+
 ### STORAGE OPTIONS FOR AUFS TABLE
 
 The `storage.options.aufs` table supports the following options:
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -132,7 +132,11 @@ func parsePullOptions(store storage.Store) pullOptions {
 		{&res.convertImages, "convert_images", false},
 		{&res.useHardLinks, "use_hard_links", false},
 	} {
-		*e.dest = parseBooleanPullOption(options, e.name, e.defaultValue)
+		if value, ok := options[e.name]; ok {
+			*e.dest = strings.ToLower(value) == "true"
+		} else {
+			*e.dest = e.defaultValue
+		}
 	}
 	res.ostreeRepos = strings.Split(options["ostree_repos"], ":")
 
@@ -1070,13 +1074,6 @@ type hardLinkToCreate struct {
 	dirfd    int
 	mode     os.FileMode
 	metadata *fileMetadata
-}
-
-func parseBooleanPullOption(pullOptions map[string]string, name string, def bool) bool {
-	if value, ok := pullOptions[name]; ok {
-		return strings.ToLower(value) == "true"
-	}
-	return def
 }
 
 type findAndCopyFileOptions struct {
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -1120,7 +1120,7 @@ func (c *chunkedDiffer) findAndCopyFile(dirfd int, r *fileMetadata, copyOptions 
 func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
 	var new []fileMetadata
 
-	hashes := make(map[string]string)
+	knownFlatPaths := make(map[string]struct{})
 	for i := range mergedEntries {
 		if mergedEntries[i].Type != TypeReg {
 			continue
@@ -1133,13 +1133,14 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
 			return nil, err
 		}
 		d := digest.Encoded()
+		path := fmt.Sprintf("%s/%s", d[0:2], d[2:])
 
-		if hashes[d] != "" {
+		if _, known := knownFlatPaths[path]; known {
 			continue
 		}
-		hashes[d] = d
+		knownFlatPaths[path] = struct{}{}
 
-		mergedEntries[i].Name = fmt.Sprintf("%s/%s", d[0:2], d[2:])
+		mergedEntries[i].Name = path
 		mergedEntries[i].skipSetAttrs = true
 
 		new = append(new, mergedEntries[i])
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -1130,10 +1130,12 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
 		}
 		digest, err := digest.Parse(mergedEntries[i].Digest)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("invalid digest %q for %q: %w", mergedEntries[i].Digest, mergedEntries[i].Name, err)
 		}
-		d := digest.Encoded()
-		path := fmt.Sprintf("%s/%s", d[0:2], d[2:])
+		path, err := path.RegularFilePathForValidatedDigest(digest)
+		if err != nil {
+			return nil, fmt.Errorf("determining physical file path for %q: %w", mergedEntries[i].Name, err)
+		}
 
 		if _, known := knownFlatPaths[path]; known {
 			continue
@@ -1449,7 +1451,8 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 		}
 		createdDirs := make(map[string]struct{})
 		for _, e := range mergedEntries {
-			d := e.Name[0:2]
+			// This hard-codes an assumption that RegularFilePathForValidatedDigest creates paths with exactly one directory component.
+			d := filepath.Dir(e.Name)
 			if _, found := createdDirs[d]; !found {
 				if err := unix.Mkdirat(dirfd, d, 0o755); err != nil {
 					return output, &fs.PathError{Op: "mkdirat", Path: d, Err: err}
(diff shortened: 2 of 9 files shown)


pkg/chunked/storage_linux.go
```diff
@@ -1734,3 +1754,63 @@ func validateChunkChecksum(chunk *internal.FileMetadata, root, path string, offs
...
+			}
+			if e.Digest == "" {
+				return nil, fmt.Errorf("missing digest for %q", e.Name)
+			}
+			digest, err := digest.Parse(e.Digest)
+			if err != nil {
+				return nil, err
+			}
+			d := digest.Encoded()
+			pathMap[filepath.Clean(e.Name)] = fmt.Sprintf("%s/%s", d[0:2], d[2:])
```
mtrmac, line 1775: The file naming logic should not be duplicated.

pkg/chunked/storage_linux.go
```diff
@@ -1734,3 +1754,63 @@ func validateChunkChecksum(chunk *internal.FileMetadata, root, path string, offs
...
+		}
+	}
+
+	dirFD2, err := unix.Dup(dirFD)
+	if err != nil {
+		return nil, err
+	}
+	rootDir := os.NewFile(uintptr(dirFD2), "staging root directory")
```
mtrmac, line 1783: This is inelegant, maybe the `*os.File` creation can happen in the caller, making _that_ the primary holder, and the FD integer a derived value? At least the `Dup` should be avoidable.

pkg/chunked/storage_linux.go
```diff
@@ -1609,6 +1615,20 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
...
 
+	if output.UncompressedDigest == "" && output.TarSplit != nil {
+		metadata := tsStorage.NewJSONUnpacker(bytes.NewReader(output.TarSplit))
+		fg, err := newStagedFileGetter(dirfd, differOpts, entriesWithOriginalNames)
+		if err != nil {
+			return output, fmt.Errorf("preparing stagedFileGetter: %w", err)
+		}
+		defer fg.Close()
+		digester := digest.Canonical.Digester()
+		if err := asm.WriteOutputTarStream(fg, metadata, digester.Hash()); err != nil {
```
cgwalters, line 1626: This is the expensive part where we're basically re-synthesizing the tarball from the unpacked files and the tar split-data just to get the checksum, right?

I know this is just a PoC but it seems very much worth a comment at the top of this conditional, something like:

```
// If we don't yet have the uncompressed digest, compute it now. This is needed
// for zstd:chunked for example to ensure callers identifying images by this
// have the data.
```

pkg/chunked/storage_linux.go
```diff
@@ -1734,3 +1754,63 @@ func validateChunkChecksum(chunk *internal.FileMetadata, root, path string, offs
...
+			digest, err := digest.Parse(e.Digest)
+			if err != nil {
+				return nil, err
+			}
+			d := digest.Encoded()
+			pathMap[filepath.Clean(e.Name)] = fmt.Sprintf("%s/%s", d[0:2], d[2:])
+		}
+	}
+
+	dirFD2, err := unix.Dup(dirFD)
```
cgwalters, line 1779: This doesn't have `O_CLOEXEC` set

pkg/chunked/storage_linux.go
```diff
@@ -1734,3 +1754,63 @@ func validateChunkChecksum(chunk *internal.FileMetadata, root, path string, offs
...
+
+func (fg *stagedFileGetter) Get(filename string) (io.ReadCloser, error) {
+	if fg.pathMap != nil {
+		path, ok := fg.pathMap[filepath.Clean(filename)]
+		if !ok {
+			return nil, fmt.Errorf("no path mapping exists for tar entry %q", filename)
+		}
+		filename = path
+	}
+	pathFD, err := securejoin.OpenatInRoot(fg.rootDir, filename)
```
cgwalters, line 1806: (A little surprising securejoin doesn't have a high level `securejoin.OpenReader` or something)

pkg/chunked/storage_linux.go
```diff
@@ -1734,3 +1754,63 @@ func validateChunkChecksum(chunk *internal.FileMetadata, root, path string, offs
 
 	return digester.Digest() == digest
 }
+
+func newStagedFileGetter(dirFD int, differOpts *graphdriver.DifferOptions, entriesWithOriginalNames []fileMetadata) (*stagedFileGetter, error) {
```
cgwalters, line 1758: I don't know this codebase well but it seems surprising to me that we wouldn't have an equivalent of this function somewhere else, there has to already be code that's serializing zstd:chunked to tar in another place which has to do the same thing?

pkg/chunked/storage_linux.go
mtrmac: > This is the expensive part where we're basically re-synthesizing the tarball from the unpacked files and the tar split-data just to get the checksum, right?

Yes.

pkg/chunked/storage_linux.go
mtrmac: There is a getter in `layers.go` which doesn’t support the composefs digest mapping (in that case we fall back to a different expensive way to generate the diff, IIRC).

mtrmac: I do agree that having those `FileGetter` implementations scattered around the codebase suggest that a better design might be possible. OTOH the path/digest mapping is expensive to maintain in a ready-to-use form.

cgwalters: > OTOH the path/digest mapping is expensive to maintain in a ready-to-use form.

On the composefs side that's basically what the composefs blob is (plus all inline metadata) and it's pretty cheap to read back a deserialized abstract tree from a composefs blob, and a big bonus is that because fsverity is enabled on that file, validation is automatic.

mtrmac: … yes, that’s a great point. There already is a special `overlayFileGetter` with composefs support, but it uses a temporary mount instead of reading the data directly. 

It seems likely that reading the data in user-space would be better, and that could be reused here.

But that’s, also, not the focus of this PR.

mtrmac: Cc: @giuseppe

cgwalters: It's also possible to deserialize a composefs in userspace via `composefs-info dump`; for the use case of turning "composefs into tar" that can be done without mounting it at a kernel level if desired. We have Rust code wrapping that in our crates. I suspect what would be helpful is to try to create a more extensive Go library for composefs separate from this project that fleshes things out like this.

Basically a cool thing about composefs is that it can be natively kernel mounted and provide full verity and things like that, but also because it's just metadata you can think of it much like a tar-split that can be handled in userspace too and efficiently merged etc.

pkg/chunked/storage_linux.go
```diff
@@ -1687,6 +1705,30 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
...
+	// to one UncompressedDigest value for each layer, and that will avoid the ambiguity (in consumers who validate layers against DiffID).
+	//
+	// c/image also uses the UncompressedDigest as a layer ID, allowing it to use the traditional layer and image IDs.
+	//
+	// This is, sadly, quite costly: Up to now we might have only have had to write, and digest, only the new/modified files.
+	// Here we need to read, and digest, the whole layer, even if almost all of it was already present locally previously.
+	//
+	// Also, layers without a tar-split (estargz layers and old zstd:chunked layers) can't produce an UncompressedDigest that
+	// matches the expected RootFS.DiffID; they will need to use a special TOC-based layer ID, and we need to somehow (??)
+	// ensure unambiguity - either by always falling back to full pulls, or by never falling back.
```
mtrmac, line 1721: @giuseppe @cgwalters note this.

zstd:chunked layers have tar-split in the current format since https://github.com/containers/storage/commit/b5413c2bd64742a08384300f09616a66231b21d1 , i.e. c/storage 1.54 == Podman 5.1.10.

I’m not 100% sure what to do here. I’m leaning towards outright refusing to partially pull estargz and old chunked layers (falling back to full pull), that is “clearly correct” and always possible. (Users who insist on always using partial pulls, e.g. for composefs, can rebuild their images.) Alternatively, we could always do a partial pull and refuse a fallback, but then we would need to break pulling those layers on other graph drivers (or, eventually, implement partial pulls for those graph drivers) — and we still would have the signing ambiguity.

mtrmac (commented):
Updated:
- Introduced `composefs.RegularFilePathForValidatedDigest(d)` to centralize the path mapping logic
- Avoided the `Dup()` of the top-level directory file handle
- Added more comments.

@giuseppe RFC. Still mostly untested, but closer to the final form, I think.

pkg/chunked/storage_linux.go
mtrmac: Introduced `pkg/chunked/internal/composefs.RegularFilePathForValidatedDigest` for that.

pkg/chunked/storage_linux.go
mtrmac: Introduced `dirFile`.

(Eventually, it seems to me that it might be clean to have a `stagingDestination struct { dirFile , …}` holder, with various functions like `openDestinationFile` becoming methods on this new type, instead of passing a raw file descriptor as a parameter all over the place. That would clearly be a separate PR.)

pkg/chunked/storage_linux.go
mtrmac: Comment added, discussing more of the goals, costs and tradeoffs.

pkg/chunked/storage_linux.go
mtrmac: This `Dup` is no longer necessary.

pkg/chunked/storage_linux.go
mtrmac: (We have an `userns.secureOpen` internally with two other callers — centralizing that, and proposing that upstream, does sound attractive.)

giuseppe:
changes LGTM

Anything more you'd like to do before it is ready for review/merge?

mtrmac:
This could be merged as is, but the plan of record is to add an opt-out option.

But, most importantly, I want to test that this, along with containers/image#2613 , actually works.

pkg/chunked/storage_linux.go
mtrmac: estargz and old zstd:chunked layers now fall back to full pulls by default, unless the user opts out of layer integrity enforcement.

mtrmac:
Updated:
- estargz and old zstd:chunked layers (with no tar-split) fall back to full pulls
- Added an `"insecure_allow_unpredictable_image_contents"` `storage.conf` option to opt out of the layer consistency enforcement. Suggestions on better names (espcially more scary names) welcome. This also allows estargz  and old zstd:chunked layers again. The documentation intentionally does not advertise the performance benefit.
- Moved `RegularFilePathForValidatedDigest` into the just-added `pkg/chunked/internal/path` package.

Still untested ATM.

@giuseppe PTAL for an early review.

mtrmac:
Temporarily includes #2197.

mtrmac:
This PR is now ready for review / merging.

Testing in https://github.com/containers/podman/pull/24419#issuecomment-2573977213 (not yet automated) confirms that this PR works as expected.

@giuseppe PTAL

docs/containers-storage.conf.5.md
```diff
@@ -124,6 +124,27 @@ The `storage.options.pull_options` table supports the following keys:
   It is an expensive operation so it is not enabled by default.
   This is a "string bool": "false"|"true" (cannot be native TOML boolean)
 
+**insecure_allow_unpredictable_image_contents="false"|"true"**
+  This should _almost never_ be set.
+  It allows partial pulls of images without guaranteeing that "partial
+  pulls" and non-partial pulls both result in consistent image contents.
```
mtrmac, line 130: BTW note the phrasing: without _guaranteeing_ consistency. I.e. this disables the _computation_ of uncompressed digest on partial pulls.

But images are still required to set correct DiffIDs: If we know the uncompressed digest for some reason (e.g. from a non-partial pull of the same blob, when the annotations were missing or the registry did not support partial pulls), and we see that it does not match DiffIDs, we are going to fail the pull.

giuseppe:
@rhatdan @cgwalters PTAL

PR #2156, opened 2024-10-30
Disable partial pulls (zstd:chunked) by default

nalind:
Disable partial pulls by default.
Add a helper command to print the library configuration after parsing everything which may be useful for diagnostic purposes.
Update the `apply-diff-from-staging-directory` integration test to call the test helper binary directly, so that the configuration file the test writes won't have its settings overridden by command line options that the `storage()` test helper function adds.

Resolves https://issues.redhat.com/browse/RUN-2335

### DIFF
--- a/cmd/containers-storage/config.go
+++ b/cmd/containers-storage/config.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/containers/storage"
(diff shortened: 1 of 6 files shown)


cgwalters (approved):

mtrmac (commented):
LGTM. (This will need to be undone in the various test PRs.)

@giuseppe FYI

PR #2160, opened 2024-11-05
chunked: add check for more parts

giuseppe:
validate that there are no other parts returned by the client, and do not ignore other errors that could have happened later.

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -1141,6 +1141,32 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
 	return new, nil
 }
 
+// checkNoStreams checks that there are no other streams on the channel and that there are no errors.
+func checkNoStreams(streams chan io.ReadCloser, errs chan error) error {
+	var err error
+	closed := 0
+	for {
+		select {
+		case _, ok := <-streams:
+			if !ok {
+				closed++
+			} else if err == nil {
+				// shouldn't happen as this is already checked in containers/image
+				err = fmt.Errorf("invalid stream returned by the server")
+			}
+		case err1, ok := <-errs:
+			if !ok {
+				closed++
+			} else if err == nil {
+				err = err1
+			}
+		}
+		if closed == 2 {
+			return err
+		}
+	}
+}
+
 func (c *chunkedDiffer) copyAllBlobToFile(destination *os.File) (digest.Digest, error) {
 	var payload io.ReadCloser
 	var streams chan io.ReadCloser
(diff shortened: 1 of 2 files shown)


pkg/chunked/storage_linux.go
```diff
@@ -1176,6 +1202,10 @@ func (c *chunkedDiffer) copyAllBlobToFile(destination *os.File) (digest.Digest,
 	// copy the entire tarball and compute its digest
 	_, err = io.CopyBuffer(destination, r, c.copyBuffer)
```
cgwalters, line 1203: Aren't we ignoring this error now?

pkg/chunked/storage_linux.go
```diff
@@ -1141,6 +1141,32 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
...
+				err = fmt.Errorf("invalid stream returned by the server")
+			}
+		case err1, ok := <-errs:
+			if !ok {
+				closed++
+			} else if err == nil {
+				err = err1
+			}
+		}
+		if closed == 2 {
```
cgwalters, line 1164: I am confused why we have this counter instead of checking `err` and `err1`

cgwalters (commented):
It'd be useful to have a bit more context about this - did you find this through code auditing or did it come out of a user hitting it in reality?

pkg/chunked/storage_linux.go
```diff
@@ -1141,6 +1141,32 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
...
 
+// checkNoStreams checks that there are no other streams on the channel and that there are no errors.
+func checkNoStreams(streams chan io.ReadCloser, errs chan error) error {
+	var err error
+	closed := 0
+	for {
+		select {
+		case _, ok := <-streams:
+			if !ok {
+				closed++
```
mtrmac, line 1152: AFAICS once this happens, this `case` will _always_ trigger and this function will consume 100% of a CPU until the other channel is closed.

pkg/chunked/storage_linux.go
```diff
@@ -1141,6 +1141,32 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
...
+			} else if err == nil {
```
mtrmac, line 1153: Nothing guarantees that we read the error from `errs` before we read from this channel. `select` is explicitly designed to choose randomly.

mtrmac (changes requested):
I think this doesn’t quite work as is, and https://issues.redhat.com/browse/OCPBUGS-43968 suggests that we need either a redesign or a much more careful error handling.

---

Also, unless the code becomes somehow trivial, I think the consumption of `(streams, errs)`, starting with the first `select` already, should be strongly centralized in a helper, not copy&pasted. (We do have some helpers, but those are already copy&pasted instead of used universally.)

pkg/chunked/storage_linux.go
mtrmac: … and `close` will keep incrementing.

giuseppe:
I close this PR for now.  I'll redesign to address https://issues.redhat.com/browse/OCPBUGS-43968

giuseppe:
alternative fix: https://github.com/containers/storage/pull/2162

PR #2162, opened 2024-11-06
chunked: rework GetBlobAt usage

giuseppe:
rewrite how the result from GetBlobAt is used, to make sure 1) that the streams are always closed, and 2) that any error is processed.

Closes: https://issues.redhat.com/browse/OCPBUGS-43968

### DIFF
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -20,6 +20,12 @@ import (
 	expMaps "golang.org/x/exp/maps"
 )
 
+const (
+	// maxTocSize is the maximum size of a blob that we will attempt to process.
+	// It is used to prevent DoS attacks from layers that embed a very large TOC file.
+	maxTocSize = (1 << 20) * 50
+)
+
 var typesToTar = map[string]byte{
 	TypeReg:     tar.TypeReg,
 	TypeLink:    tar.TypeLink,
@@ -77,7 +83,7 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, 
 
 	size := int64(blobSize - footerSize - tocOffset)
 	// set a reasonable limit
-	if size > (1<<20)*50 {
+	if size > maxTocSize {
 		return nil, 0, errors.New("manifest too big")
 	}
 
@@ -106,7 +112,7 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, 
 					return err
 				}
 				// set a reasonable limit
-				if header.Size > (1<<20)*50 {
+				if header.Size > maxTocSize {
 					return errors.New("manifest too big")
 				}
 
@@ -166,10 +172,10 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
 	}
 
 	// set a reasonable limit
-	if manifestChunk.Length > (1<<20)*50 {
+	if manifestChunk.Length > maxTocSize {
 		return nil, nil, nil, 0, errors.New("manifest too big")
 	}
-	if manifestLengthUncompressed > (1<<20)*50 {
+	if manifestLengthUncompressed > maxTocSize {
 		return nil, nil, nil, 0, errors.New("manifest too big")
 	}
 
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -44,25 +44,21 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, 
 	if blobSize <= footerSize {
 		return nil, 0, errors.New("blob too small")
 	}
-	chunk := ImageSourceChunk{
-		Offset: uint64(blobSize - footerSize),
-		Length: uint64(footerSize),
-	}
-	parts, errs, err := blobStream.GetBlobAt([]ImageSourceChunk{chunk})
+
+	footer := make([]byte, footerSize)
+	streamsOrErrors, err := getBlobAt(blobStream, ImageSourceChunk{Offset: uint64(blobSize - footerSize), Length: uint64(footerSize)})
 	if err != nil {
 		return nil, 0, err
 	}
-	var reader io.ReadCloser
-	select {
-	case r := <-parts:
-		reader = r
-	case err := <-errs:
-		return nil, 0, err
-	}
-	defer reader.Close()
-	footer := make([]byte, footerSize)
-	if _, err := io.ReadFull(reader, footer); err != nil {
-		return nil, 0, err
+
+	for soe := range streamsOrErrors {
+		if soe.stream != nil {
+			_, err = io.ReadFull(soe.stream, footer)
+			_ = soe.stream.Close()
+		}
+		if soe.err != nil && err == nil {
+			err = soe.err
+		}
 	}
 
 	/* Read the ToC offset:
@@ -85,44 +81,50 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, 
 		return nil, 0, errors.New("manifest too big")
 	}
 
-	chunk = ImageSourceChunk{
-		Offset: uint64(tocOffset),
-		Length: uint64(size),
-	}
-	parts, errs, err = blobStream.GetBlobAt([]ImageSourceChunk{chunk})
+	streamsOrErrors, err = getBlobAt(blobStream, ImageSourceChunk{Offset: uint64(tocOffset), Length: uint64(size)})
 	if err != nil {
 		return nil, 0, err
 	}
 
-	var tocReader io.ReadCloser
-	select {
-	case r := <-parts:
-		tocReader = r
-	case err := <-errs:
-		return nil, 0, err
-	}
-	defer tocReader.Close()
+	var manifestUncompressed []byte
 
-	r, err := pgzip.NewReader(tocReader)
-	if err != nil {
-		return nil, 0, err
-	}
-	defer r.Close()
+	for soe := range streamsOrErrors {
+		if soe.stream != nil {
+			err1 := func() error {
+				defer soe.stream.Close()
 
-	aTar := archivetar.NewReader(r)
+				r, err := pgzip.NewReader(soe.stream)
+				if err != nil {
+					return err
+				}
+				defer r.Close()
 
-	header, err := aTar.Next()
-	if err != nil {
-		return nil, 0, err
-	}
-	// set a reasonable limit
-	if header.Size > (1<<20)*50 {
-		return nil, 0, errors.New("manifest too big")
-	}
+				aTar := archivetar.NewReader(r)
 
-	manifestUncompressed := make([]byte, header.Size)
-	if _, err := io.ReadFull(aTar, manifestUncompressed); err != nil {
-		return nil, 0, err
+				header, err := aTar.Next()
+				if err != nil {
+					return err
+				}
+				// set a reasonable limit
+				if header.Size > (1<<20)*50 {
+					return errors.New("manifest too big")
+				}
+
+				manifestUncompressed = make([]byte, header.Size)
+				if _, err := io.ReadFull(aTar, manifestUncompressed); err != nil {
+					return err
+				}
+				return nil
+			}()
+			if err == nil {
+				err = err1
+			}
+		} else if err == nil {
+			err = soe.err
+		}
+	}
+	if manifestUncompressed == nil {
+		return nil, 0, errors.New("manifest not found")
 	}
 
 	manifestDigester := digest.Canonical.Digester()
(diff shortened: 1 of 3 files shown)


giuseppe:
testing in https://github.com/containers/podman/pull/24478

pkg/chunked/compression_linux.go
```diff
@@ -48,21 +48,21 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
 		Offset: uint64(blobSize - footerSize),
 		Length: uint64(footerSize),
 	}
-	parts, errs, err := blobStream.GetBlobAt([]ImageSourceChunk{chunk})
+
+	footer := make([]byte, footerSize)
```
mtrmac, line 52: IIRC, a lot of the code could benefit from a single “ask for one chunk, read it into memory, validate against a digest” utility.

Arguably that’s a cleanup that should not block this PR…

pkg/chunked/compression_linux.go
```diff
@@ -89,44 +89,55 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
...
+			}
+		}
+		if soe.err != nil && err == nil {
+			err = soe.err
+		}
 	}
 
 	manifestDigester := digest.Canonical.Digester()
 	manifestChecksum := manifestDigester.Hash()
+	if manifestUncompressed == nil {
```
mtrmac, line 138: Nit: Checking this immediately after the loop would be a bit cleaner.

pkg/chunked/compression_linux.go
```diff
@@ -175,26 +186,29 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
...
 		chunks = append(chunks, tarSplitChunk)
 	}
-	parts, errs, err := blobStream.GetBlobAt(chunks)
+
+	streamsOrErrors, err := getBlobAt(blobStream, chunks, len(chunks))
 	if err != nil {
 		return nil, nil, nil, 0, err
 	}
 
+	defer func() {
```
mtrmac, line 195: This works.

Maybe another way to structure the consumers would be to:
```go
defer consumeRemainingStreamsOrErrors(streamsOrErrors) // this loop, ignoring all errors. Also see elsewhere — maybe this should be a context cancellation instead.
// consume data: if there is only one chunk:
reader, err := getNextStream(streamsOrErrors) // if this contains an error, or is closed, returns an error
// Optionally, after all data is consumed: Alternatively, this can be enforced in `getBlobAt` — if we are fine with ignoring errors after all expected data is consumed, like excess chunks.
err := ensureAllDone(streamsOrErrors) // Ensure straemsOrErrors is at EOF, there are no unexpected items, there are no errors reported .

File: pkg/chunked/compression_linux.go [hunk 4]
```diff
@@ -175,26 +186,29 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
...
+	}()
+
 	readBlob := func(len uint64) ([]byte, error) {
-		var reader io.ReadCloser
-		select {
-		case r := <-parts:
-			reader = r
-		case err := <-errs:
-			return nil, err
+		soe := <-streamsOrErrors
```
mtrmac, line 204: This returns all-0 on EOF

pkg/chunked/storage_linux.go
```diff
@@ -1141,41 +1141,93 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
...
+
+func getBlobAt(is ImageSourceSeekable, chunksToRequest []ImageSourceChunk, maxStreams int) (chan streamOrErr, error) {
+	streams, errs, err := is.GetBlobAt(chunksToRequest)
+	if err != nil {
+		return nil, err
+	}
+	stream := make(chan streamOrErr)
+	go func() {
+		tooManyStreams := false
+		defer close(stream)
```
mtrmac, line 1157: I’m paranoid about goroutines failing in hard-to-diagnose ways.

```go
go getBlobAtConverterGoroutine()…
…

func getBlobAtConverterGoroutine(stream) {
   err = errors.New("Unexpected error in getBlobAtGoroutine")
   defer func() {
      if err != nil { stream <- err }
      close(err)
   }()
   err = getBlobAtConverter(…)
}
```

would ensure that any panic in the child is caught and reported.

pkg/chunked/storage_linux.go
```diff
@@ -1141,41 +1141,93 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
...
+				stream <- streamOrErr{stream: p}
+			}
+		}
+		if errs != nil {
+			for err := range errs {
+				stream <- streamOrErr{err: err}
+			}
+		}
+		if tooManyStreams {
+			stream <- streamOrErr{err: fmt.Errorf("too many streams returned, requested %d, got %d", len(chunksToRequest), streamsSoFar)}
```
mtrmac, line 1199: `streamsSoFar` is not incremented beyond `maxStreams` due to the `continue`s.

pkg/chunked/storage_linux.go
```diff
@@ -1141,41 +1141,93 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
...
-	var payload io.ReadCloser
-	var streams chan io.ReadCloser
-	var errs chan error
-	var err error
+type streamOrErr struct {
+	stream io.ReadCloser
+	err    error
+}
+
+func getBlobAt(is ImageSourceSeekable, chunksToRequest []ImageSourceChunk, maxStreams int) (chan streamOrErr, error) {
```
mtrmac, line 1149: AFAICS `maxStreams` is always `len(…)`, no need for the caller to provide it.

pkg/chunked/compression_linux.go
mtrmac, line 195: Having to consume everything there is, generally, an undesired cost.

The typical handling for that is to use a cancellable context, and to have the consumer here `defer cancelTheContext`; and for the goroutine to observe that context and terminate on cancellation.

Sadly that is not possible in `ImageSourceSeekable` (no `Context` parameter), but maybe we could build the `streamsOrErrors` the idiomatic way, and have `getBlobAt` deal with draining the produced output, instead of doing that in every consumer.

(I didn’t try, this might be too hard to do. Notably channel sends in `getBlobAt` would also have to `select` vs. the context terminating.)

pkg/chunked/compression_linux.go
mtrmac: … on a closed channel, rather.

pkg/chunked/compression_linux.go
```diff
@@ -89,40 +89,50 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
...
+				defer r.Close()
+
+				aTar := archivetar.NewReader(r)
+
+				header, err := aTar.Next()
+				if err != nil {
+					return err
+				}
+				// set a reasonable limit
+				if header.Size > (1<<20)*50 {
```
cgwalters, line 117: Not new, but I think these types of things should be a `const` somewhere with a bit of documented rationale.

pkg/chunked/storage_linux.go
```diff
@@ -1141,41 +1141,116 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
...
+		if soe.stream != nil {
+			_ = soe.stream.Close()
+		} else if retErr == nil {
+			retErr = soe.err
+		}
+	}
+	return
+}
 
+func getBlobAtConverterGoroutine(stream chan streamOrErr, streams chan io.ReadCloser, errs chan error, maxStreams int) {
```
cgwalters, line 1161: This function is quite complex and could probably use a brief explanatory comment.

giuseppe:
@cgwalters thanks for the review.  Addressed the comments and pushed a new version

pkg/chunked/storage_linux.go
```diff
@@ -1141,41 +1141,121 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
...
+			stream <- streamOrErr{err: err}
+		}
+	}
+	if tooManyStreams {
+		stream <- streamOrErr{err: fmt.Errorf("too many streams returned, got more than %d", maxStreams)}
+	}
+	err = nil
+}
+
+func getBlobAt(is ImageSourceSeekable, chunksToRequest []ImageSourceChunk) (chan streamOrErr, error) {
```
mtrmac, line 1223:
```suggestion
// getBlobAt provides a much more convenient way to consume data returned by ImageSourceSeekable.GetBlobAt.
// GetBlobAt returns two channels, forcing a caller to `select` on both of them — and in Go, reading a closed channel
// always succeeds in select.
// Instead, getBlobAt provides a single channel with all events, which can be consumed conveniently using `range`.
func getBlobAt(is ImageSourceSeekable, chunksToRequest []ImageSourceChunk) (chan streamOrErr, error) {
```

I think that’s what we have learned, and it seems worth preserving that for future maintainers.

pkg/chunked/storage_linux.go
```diff
@@ -1141,41 +1141,121 @@ func makeEntriesFlat(mergedEntries []fileMetadata) ([]fileMetadata, error) {
...
+	}()
+loop:
+	for {
+		select {
+		case p, ok := <-streams:
+			if !ok {
+				streams = nil
+				break loop
+			}
+			if maxStreams > 0 && streamsSoFar >= maxStreams {
```
mtrmac, line 1186: `maxStreams > 0` should always be set — there is only one caller, and the value is a `len()`. Or can we ever get here with a zero-chunks request? If so, that should probably be handled somewhere earlier than this — and with `maxStreams == 0`, arguably the rest of the condition is correct anyway.

pkg/chunked/storage_linux.go
mtrmac, line 1223: Absolutely non-blocking:

With
```suggestion
func getBlobAt(is ImageSourceSeekable, chunksToRequest ImageSourceChunk...) (chan streamOrErr, error) {
```
the simple 1-chunk or 2-chunk callers could embed the chunk data directly into the function call, without having to construct a slice explicitly, making the call sites a bit more concise.

pkg/chunked/storage_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+	streams chan io.ReadCloser
+	errors  chan error
+}
+
+func (m *mockImageSource) GetBlobAt(chunks []ImageSourceChunk) (chan io.ReadCloser, chan error, error) {
+	return m.streams, m.errors, nil
+}
+
+type mockReadCloser struct {
+	reader io.Reader
```
mtrmac, line 25: (Absolutely non-blocking:

```go
type mockReadCloser struct {
   io.Reader
   closed bool
}
```

would allow removing the manually-written `mockReadCloser.Read` forwarder. This works just fine as is.

pkg/chunked/storage_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+	i := 0
+	for result := range resultChan {
+		assert.NoError(t, result.err)
+		buf := new(bytes.Buffer)
+		_, _ = buf.ReadFrom(result.stream)
+		result.stream.Close()
+		assert.Equal(t, expectedStreams[i], buf.String())
+		i++
+	}
+	assert.Equal(t, len(expectedStreams), i)
```
mtrmac, line 71: Absolutely non-blocking, this works perfectly fine:
- `assert.Len` exists
- Alternatively, the loop could collect a `returnedContents [][]byte`, and afterwards there could be a single `assert.Equal(t, expectedStreams, returnedContents)` comparing both the length and data.

pkg/chunked/storage_linux_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	for _, s := range streamsObjs {
+		streams <- s
+	}
+	close(streams)
+	close(errors)
+
+	is := &mockImageSource{streams: streams, errors: errors}
+
+	resultChan, err := getBlobAt(is, nil)
```
mtrmac, line 59: The `nil` here is … non-representative. Is this the only reason for the `maxStreams > 0` condition in the code?

mtrmac (commented):
ACK overall, and the tests are great.

giuseppe:
@mtrmac addressed your comments, pushed a new version

pkg/chunked/storage_linux_test.go
```diff
@@ -0,0 +1,172 @@
...
+	errorsC := make(chan error, 1)
+
+	streams <- mockReadCloserFromContent("stream1")
+	errorsC <- errors.New("error1")
+	close(streams)
+	close(errorsC)
+
+	is := &mockImageSource{streams: streams, errors: errorsC}
+
+	resultChan, err := getBlobAt(is)
```
mtrmac, line 158: I liked the previous version, where “the first” stream is successfully read, and “the second one” fails, better. Asking for 2 chunks here would return to that behavior, I think.

mtrmac (commented):
Code LGTM. Can I ask for one last test change?

giuseppe:
> Code LGTM. Can I ask for one last test change?

what do you think of the last version?

mtrmac:
Sure, that works.

---

The `lint` CI step is failing, and the error message is not very helpful.

mtrmac:
@giuseppe Thanks!

Issue #2174, opened 2024-11-18
Setting shared xattr fails on images with named pipes


m3nu:
### Issue Description

This is a bit of an edge case. When extracting an image that has a named pip inside, setting the xattr `user.containers.override_stat` for `force_mask=shared` fails. Using `setfattr` directly also fails. 

I'm assuming this is a limitation of extended attributes, but wonder if Podman should ignore such errors instead of failing to pull the image.

### Steps to reproduce the issue

Steps to reproduce the issue. (All run as root user)

Pulling the image with `podman pull`
```
 # podman --storage-opt=overlay.force_mask=shared pull docker.io/privatebin/fs:1.7.5

Trying to pull docker.io/privatebin/fs:1.7.5...
Getting image source signatures
Copying blob 4f4fb700ef54 skipped: already exists
Copying blob 06ad7d3e6d44 done   |
Copying blob da9db072f522 skipped: already exists
Copying blob a1e61df148cb done   |
Copying blob f6ba44d80dfe done   |
Error: copying system image from manifest list: writing blob: adding layer with blob "sha256:a1e61df148cbf56a174660fbc701191105f58fa4dcd60ca7ba56ce5e42b6e485"/""/"sha256:4bef5d03909cda3717dbf5caa1fc1f64a63b5e3b4072b87cf6c281cb20997f88": unpacking failed (error: exit status 1; output: lsetxattr /etc/s6/services/nginx/supervise/control: operation not permitted)
```

Using `setfattr` directly fails on the same file:
```
# setfattr -n user.myattribute -v "value" .../diff/etc/s6/services/nginx/supervise/control
setfattr: .../diff/etc/s6/services/nginx/supervise/control: Operation not permitted
```

File seems to be a named pipe:
```
# ls -lh .../diff/etc/s6/services/nginx/supervise/control
prw-r--r--. 1 165533 100081 0 Nov 16 07:41 .../diff/etc/s6/services/nginx/supervise/control
```

### Describe the results you received

Pulling the image fails with `force_mask=shared` setting.

### Describe the results you expected

Pulling the image should succeed, even if the xattr of this one file isn't set.

### podman info output

```yaml
# podman info
host:
  arch: amd64
  buildahVersion: 1.37.3
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  - rdma
  - misc
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
[...]

PR #2175, opened 2024-11-19
archive: honor ignoreChownErrors with ForceMode

giuseppe:
when ignore_chown_errors is set, ignore also failures from setting the override mode xattr.

Closes: https://github.com/containers/storage/issues/2174

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -734,7 +734,10 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 			Mode: hdrInfo.Mode() & 0o7777,
 		}
 		if err := idtools.SetContainersOverrideXattr(path, value); err != nil {
-			return err
+			if !ignoreChownErrors {
+				return err
+			}
+			logrus.Warnf("Setting override extended attribute failed for %q. Ignoring due to ignoreChownErrors flag: %v", path, err)
 		}
 	}
 
--- a/pkg/archive/archive_test.go
+++ b/pkg/archive/archive_test.go
@@ -1247,3 +1247,20 @@ func readFileFromArchive(t *testing.T, archive io.ReadCloser, name string, expec
 	assert.NoError(t, err)
 	return string(content)
 }
+
+func TestCreateFifo(t *testing.T) {
+	hdr := tar.Header{Typeflag: tar.TypeFifo}
+	tmpDir := t.TempDir()
+	buffer := make([]byte, 1<<20)
+	err := createTarFile(filepath.Join(tmpDir, "fifo1"), tmpDir, &hdr, nil, true, nil, false, false, nil, buffer)
+	assert.NoError(t, err)
+
+	forceMask := os.FileMode(0o755)
+	err = createTarFile(filepath.Join(tmpDir, "fifo2"), tmpDir, &hdr, nil, true, nil, false, false, &forceMask, buffer)
+	// expect an error to chown a fifo
+	assert.Error(t, err)
+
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM

giuseppe:
@mtrmac PTAL

mtrmac (changes requested):
I don’t see why this makes sense.
- Just formally, this PR is underdocumented:
  - `ignore_chown_errors` is documented to _only_ affect file ownership. Nothing says it interacts with `ContainersOverrideXattr`, and it hasn’t behaved that way so far.
  - `ContainersOverrideXattr` is documented for the `force_mask` option, and nothing says that `ignore_chown_errors` should affect that behavior.
- What is the use case? With `force_mask`, AFAICS `ContainersOverrideXattr` _must_ be set for the container to behave as expected. The documentation explicitly talks about sharing stores with other users, or using private stores by unprivileged users — in both cases I think the users _must_ be told if setting `ContainersOverrideXattr` fails, at least by default.
  - I can imagine a hypothetical option that means “I don’t care about any attributes at all, don’t set UIDs, don’t set modes, don’t bother with the extended attribute either” … but, what for? It isn’t useful for running containers in general (ownership and permissions matter), it isn’t necessary for specially-restricted containers that don’t care (they can be specially built to only be root:root 700:700). My best guess is that this would be useful for some kind of “inspect the filesystem” operation, but then that can be done by extracting the tarball without any privileges at all, and possibly without writing to the filesystem, so I don’t think an extra option is warranted.
- Investigating further, per https://github.com/containers/storage/issues/2174#issuecomment-2506849902 , this has nothing to do with `ignore_chown_errors`: setting `user.…` extended attributes on devices/pipes/symlinks _never_ works, even if the user has `CAP_SYS_ADMIN`. This is a fundamental limitation of the current `force_mask` design. I don’t know what we want to do here.
  - Hypothetically, we could have an option that means “I want to use `force_mask` with images that contain devices/pipes and I’m fine with those objects having incorrect ownership and permissions”. That option has different meaning from `ignore_chown_errors`, so it should be a new option — but, anyway, I don’t know why anyone _should_ set this option. Users who have these objects in their container filesystems, presumably, _need them_; if not, they can always build (squashed?) images with the objects removed.

giuseppe:
from an user PoV, with `ignore_chown_errors` the expectation is that the files inside a container/image might not have the correct ownership.  Since the container might behave differently in this case, we opted for an explicit setting instead of doing it transparently.  For example, `ignore_chown_errors` is useful when the user has not enough IDs allocated, and the end result is that all the files are owned only by a subset of IDs in the image, depending how many IDs are available.

That situation is similar to using `force_mask` and failing to set the `ContainersOverrideXattr`.  The end result is that the files inside the container have the wrong ownership.

Ideally, images should not have special files, but still there are images using them and we need to handle them in some way, I don't think this should be done transparently but must be an explicit setting.  Sure we could add yet another option to make it clearer, but the semantic would be identical and it will just cover what seems to be a corner case.

mtrmac:
> from an user PoV, with `ignore_chown_errors` the expectation is that the files inside a container/image might not have the correct ownership. Since the container might behave differently in this case, we opted for an explicit setting instead of doing it transparently.

Hum, having `ignore_chown_errors` effectively mean “I am opting in into not representing the image correctly” (and include not just ownership, but permissions) does make some sense, as long as it is documented that way.

But https://github.com/containers/storage/issues/2174#issuecomment-2507487073 seems much nicer to me — that makes things _work_ correctly and then we won’t have to worry users about choosing between non-ideal options.

giuseppe:
closed in favor of: https://github.com/containers/storage/pull/2183

I am still working on it, but opened to get some early feedback

PR #2176, opened 2024-11-20
store: new Dedup() API

giuseppe:
offer a way to deduplicate files in the local storage using reflinks when possible.

For now, it deduplicates only identical files, even if reflinks support deduplication only of a block aligned portion of a file


giuseppe:
cleaned up some code I was using just for local testing.

@rhatdan @nalind PTAL

pkg/dedup/dedup_linux.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	defer srcFile.Close()
+
+	dstFile, err := os.OpenFile(dst, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o666)
+	if err != nil {
+		return fmt.Errorf("failed to create destination file: %w", err)
+	}
+	defer dstFile.Close()
+
+	_, _, errno := unix.Syscall(unix.SYS_IOCTL, dstFile.Fd(), C.FICLONE, srcFile.Fd())
```
nalind, line 36: Use `unix.IoctlFileClone()` here?

giuseppe: yes thanks, that is better

mtrmac (commented):
Only a brief drive-by skim, I didn’t read the PR carefully I’m afraid.

pkg/dedup/dedup.go
```diff
@@ -0,0 +1,93 @@
...
+
+var notSupported = errors.New("reflinks are not supported on this platform")
+
+type DedupResult struct {
+	// Deduped is the number of deduped bytes
+	Deduped uint64
+}
+
+func getFileChecksum(path string) (uint64, error) {
+	c := crc64.New(crc64.MakeTable(crc64.ECMA))
```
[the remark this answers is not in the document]
mtrmac: OK, but then consider the ABCD scenario above: The code as is would only try to deduplicate AB AC AD, but not BC.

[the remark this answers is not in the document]
giuseppe: I've added an option to use a cryptographic function if desired.

cgwalters:
If we're only trying to dedup whole files, I think we could generally just skip the userspace hashing (whether crc or cryptographic) and just find files with identical size and try calling the dedup ioctl on them, letting the kernel do the actual comparison, right?

(Skipping zero sized, and also potentially skipping size < 4096 as probably low value and much more likely to collide)

giuseppe:
> If we're only trying to dedup whole files, I think we could generally just skip the userspace hashing (whether crc or cryptographic) and just find files with identical size and try calling the dedup ioctl on them, letting the kernel do the actual comparison, right?
> 
> (Skipping zero sized, and also potentially skipping size < 4096 as probably low value and much more likely to collide)

added a way to choose among "size", "crc" and "sha256sum".

I see worse performance using the file size as the hash function.  There are more conflicts even when skipping small files (tried up to 4096 bytes).    Could be something wrong in the code, I'll look at it on Monday.

```shell
$ podman images
REPOSITORY                         TAG         IMAGE ID      CREATED       SIZE
registry.fedoraproject.org/fedora  latest      b889b3ce322a  39 hours ago  164 MB
docker.io/library/debian           latest      63ba939a68fd  11 days ago   121 MB
docker.io/library/ubuntu           latest      fec8bfd95b54  5 weeks ago   80.6 MB
docker.io/library/nginx            latest      60c8a892f36f  7 weeks ago   196 MB
docker.io/library/alpine           latest      63b790fccc90  2 months ago  8.09 MB

$ hyperfine './containers-storage dedup --hash-method=size'
Benchmark 1: ./containers-storage dedup --hash-method=size
  Time (mean ± σ):     546.8 ms ±  46.2 ms    [User: 1363.3 ms, System: 2340.5 ms]
  Range (min … max):   503.9 ms … 647.4 ms    10 runs

$ hyperfine './containers-storage dedup --hash-method=crc'
Benchmark 1: ./containers-storage dedup --hash-method=crc
  Time (mean ± σ):     322.5 ms ±  38.9 ms    [User: 1285.4 ms, System: 724.2 ms]
  Range (min … max):   224.9 ms … 353.9 ms    13 runs

$ hyperfine './containers-storage dedup --hash-method=sha256'
Benchmark 1: ./containers-storage dedup --hash-method=sha256
  Time (mean ± σ):     822.4 ms ±  77.3 ms    [User: 4562.1 ms, System: 802.3 ms]
  Range (min … max):   635.5 ms … 912.4 ms    10 runs
```

cgwalters:
One thing that'd help a little bit is a bit more of the "why" - is there a production user of VFS that really wants this? Is this a side project/experiment?

giuseppe:
ready for review

pkg/dedup/dedup_linux.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	sSrc, ok := stSrc.Sys().(*syscall.Stat_t)
+	if !ok {
+		return 0, fmt.Errorf("unable to get raw syscall.Stat_t data")
+	}
+	sDest, ok := fiDst.Sys().(*syscall.Stat_t)
+	if !ok {
+		return 0, fmt.Errorf("unable to get raw syscall.Stat_t data")
+	}
+	if sSrc.Ino == sDest.Ino {
```
mtrmac, line 91: A comment pointing at the code that handles `Dev` matching, explaining that’s why we don’t need to compare it here, would be useful.

store.go
```diff
@@ -3843,3 +3853,41 @@ func (s *store) MultiList(options MultiListOptions) (MultiListResult, error) {
...
+func (s *store) Dedup(req DedupArgs) (drivers.DedupResult, error) {
+	imgs, err := s.Images()
+	if err != nil {
+		return drivers.DedupResult{}, err
+	}
+	var topLayers []string
+	for _, i := range imgs {
+		topLayers = append(topLayers, i.TopLayer)
+		topLayers = append(topLayers, i.MappedTopLayers...)
+	}
```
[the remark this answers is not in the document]
giuseppe: we should not deduplicate writeable layers, as that can slow down containers at runtime.  If a container has its own copy of a file, and we deduplicate it, then the next time the container tries to write to it, the file system needs to create a copy.  The current logic instead has no such disadvantages because read-only layers referenced by an image can't change

tests/dedup.bats
```diff
@@ -0,0 +1,78 @@
...
+	deduped=$(jq -r .Deduped <<< $output)
+	[[ $deduped -gt 0 ]]
+
+	for METHOD in size crc sha256; do
+		# Test that it always returns the same value with any hash-method.
+		for i in $(seq 10); do
+			run storage --debug=false dedup -j --hash-method=$METHOD
+			[ "$status" -eq 0 ]
+			actual=$(jq -r .Deduped <<< $output)
+			[[ $deduped = $actual ]]
```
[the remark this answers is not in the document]
giuseppe: that maps how ioctl(FIDEDUPERANGE) works, it reports how many bytes are shared no matter if they were already shared before the call.  I don't see any easy way to see only what changed.

From the man page:

```
       Upon  successful  completion  of this ioctl, the number of bytes successfully deduplicated is returned in bytes_deduped
       and a status code for the deduplication operation is returned in status.  If even a single byte in the range  does  not
       match,  the  deduplication  operation  request will be ignored and status set to FILE_DEDUPE_RANGE_DIFFERS.  The status
       code is set to FILE_DEDUPE_RANGE_SAME for success, a negative error code in case of error, or FILE_DEDUPE_RANGE_DIFFERS
       if the data did not match.
```

pkg/dedup/dedup.go
```diff
@@ -0,0 +1,153 @@
...
+					dedupBytes += deduped
+					break
+				}
+				logrus.Debugf("Failed to deduplicate: %v", err)
+				if errors.Is(err, notSupported) {
+					item.lock.Unlock()
+					return err
+				}
+			}
+			if dedupBytes == 0 {
```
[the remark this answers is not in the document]
giuseppe: from the documentation it seems that it cannot happen, it is either satisfies the entire request or it fails:

```
 If even a single byte in the range  does  not match,  the  deduplication  operation  request will be ignored and status set to FILE_DEDUPE_RANGE_DIFFERS
```

but I've not verified if it is really the case

giuseppe:
@mtrmac thanks for the review.  Addressed your comments, and pushed a new version

store.go
```diff
@@ -3843,3 +3853,41 @@ func (s *store) MultiList(options MultiListOptions) (MultiListResult, error) {
...
+	return writeToLayerStore(s, func(rlstore rwLayerStore) (drivers.DedupResult, error) {
+		layers := make(map[string]interface{})
+		for _, i := range topLayers {
+			cur := i
+			for cur != "" {
+				layers[cur] = nil
```
mtrmac, line 3873: (Considering that this code is already handling `ErrLayerUnknown`, it would be a tiny bit cleaner to only add a layer to `layers` after the `.Get()` below succeeds. Ultimately, the two implemented graph drivers filter out unknown layers anyway, but doing it centrally here at the top level would cost us nothing.)

pkg/dedup/dedup.go
```diff
@@ -0,0 +1,153 @@
...
+				lock.Unlock()
+				return nil
+			}
+			item.lock.Lock()
+			lock.Unlock()
+
+			var dedupBytes uint64
+			for _, src := range item.paths {
+				deduped, err := dedup.Dedup(src, path, info)
+				if err == nil && deduped == size {
```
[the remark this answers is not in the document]
mtrmac: That feels like a platform-specific detail to me. Some other platform could hypothetically deduplicate partially; I think this code could accept that smaller match and report that as a success.

At this level, it seems incorrect for the code to report “failed” when deduplication worked on 50% of the file. There already is a Linux-specific backend; if there should be an assertion that the ioctl returns either 0 or all of `size`, it seems to me that assertion would be better placed in that Linux-specific backend specific to this ioctl (but, even in that case, I’m doubtful that the assertion would be all that valuable).

pkg/dedup/dedup.go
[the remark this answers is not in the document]
mtrmac: See elsewhere, I think that might make sense for a Linux-specific implementation, but this is at least trying to build a platform-agnostic frontend, and I think it costs ~nothing at run time (just these review comments) for it to be ready for partial matches.

pkg/dedup/dedup_linux.go
```diff
@@ -0,0 +1,139 @@
...
+func (d *dedupFiles) recordInode(dev, ino uint64) (bool, error) {
+	d.lock.Lock()
+	defer d.lock.Unlock()
+
+	// record the device used by the first inode, do not allow
+	// recording inodes from different devices
+	if d.dev == 0 {
+		d.dev = dev
+	} else if d.dev != dev {
+		logrus.Warningf("Ignored inode from different device: %d (expected %v)", dev, d.dev)
```
mtrmac, line 37: I’m tempted to read the `%d` vs. `%v` difference as an indication that something non-obvious is going on; if it isn’t the case, please use the same directive for both.

mtrmac (commented):
I’m still a bit worried about making this a public API — also some of the comments related to that (cross-device matching, not forcing the caller to deal with `HashMethod`) are still outstanding.

pkg/dedup/dedup.go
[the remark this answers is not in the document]
mtrmac: BTW the documentation of the Linux ioctl says that there is a size limit to matches — so it’s easy to imagine that (even if we only wanted an exact match and if we were not considering a rsync-like rolling hash) we might want to run this ioctl repeatedly for parts of the same file, 8 MB at a time or so. That could easily result in partial deduplication success.

giuseppe:
> I’m still a bit worried about making this a public API — also some of the comments related to that (cross-device matching, not forcing the caller to deal with `HashMethod`) are still outstanding.

can a file on a different device really happen though?  We only deduplicate from the main storage, so it is expected to be on the same device, and even with partial pulls, from a different device we perform a fully copy since there is no other method we can use (both hard links and reflinks require src and dest on the same file system).

mtrmac:
To clarify, I’m primarily talking about making `pkg/dedup` public. I 100% agree that `Store.Dedup` must be public.

Calling deduplication across filesystems can’t happen when deduplicating _image storage_. But if `c/storage/pkg/dedup` becomes public, it may well be called on other inputs — in fact it _only_ becomes valuable to make it public if we anticipate it would ever get an external caller that uses it on other inputs.

So, I think the options are:
- Make `pkg/dedup` work with inputs across filesystems
- Make `pkg/dedup` cleanly fail, somehow, when inputs cross filesystems
- Move `pkg/dedup` to `internal/dedup` (probably requires relocating/restructuring `HashMethod`).

I think the last one is simplest to do. We can always make the package public, and more general, later, after a possible user comes up.

---

Or do you plan to call `pkg/dedup` from Podman directly?

\*shrug\* documenting it as experimental / not to be used by outside projects is also an option.

giuseppe:
> * 
> * Move `pkg/dedup` to `internal/dedup` (probably requires relocating/restructuring `HashMethod`).

what do you think of the new version?

It adds some duplication in the `store.go` file, but perhaps that is acceptable?

Alternatively, we have `types/`  and we could move these common types there.  From what I can see though, the current types deal mostly with the storage configuration file

store.go
```diff
@@ -166,6 +167,26 @@ type flaggableStore interface {
 
 type StoreOptions = types.StoreOptions
 
+type HashMethod = dedup.HashMethod
+
+const (
+	HashInvalid  = dedup.HashInvalid
```
mtrmac, line 173: `DedupHash…` for all of this, please, to reduce ambiguity within the large `storage` namespace.

pkg/dedup/dedup.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+var notSupported = errors.New("reflinks are not supported on this platform")
+
+const (
+	HASH_CRC = iota
+	HASH_FILE_SIZE
+	HASH_SHA256
+)
```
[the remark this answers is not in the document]
mtrmac: I’d still like there to be a default when the caller doesn’t have a preference, but, meh, we can always add that later.

pkg/dedup/dedup.go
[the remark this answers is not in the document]
mtrmac: Still worth cleaning up I think.

tests/dedup.bats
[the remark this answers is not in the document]
mtrmac: Outstanding.

pkg/dedup/dedup_linux.go
mtrmac: I’d prefer this to be cleaned up, maybe reverting `recordInode` to fail again if it sees a cross-device candidate, and this code to point at that `recordInode` condition.

mtrmac:
> > * Move `pkg/dedup` to `internal/dedup` (probably requires relocating/restructuring `HashMethod`).
> 
> what do you think of the new version?

Great, works for me.

> It adds some duplication in the `store.go` file, but perhaps that is acceptable?
> 
> Alternatively, we have `types/` and we could move these common types there. From what I can see though, the current types deal mostly with the storage configuration file

(I’m not that worried about these simple types with no dependencies and no runtime cost; we can always add aliases from other packages to make the API more convenient — as long as the types/fields make sense to commit to.) Here the caller is certainly going to import the top-level `storage` package, so I think having them present directly in `storage` is the most convenient option for users.

mtrmac (commented):
I’d prefer addressing also at least https://github.com/containers/storage/pull/2176#discussion_r1866416654 and https://github.com/containers/storage/pull/2176#discussion_r1868141714 , while we still remember this context.

giuseppe:
documented that the dedup result accounts for previous runs, and changed the code to deal with multiple file systems

giuseppe:

internal/dedup/dedup.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return res, err
+	}
+
+	for _, dir := range dirs {
+		logrus.Debugf("Deduping directory %s", dir)
+		if err := pwalkdir.Walk(dir, func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				logrus.Errorf("Failed to read file %q: %v", path, err)
+				return nil
```
Luap99, line 88: Why does this not return the proper error? That just makes it so the API caller never gets an error if an invalid path was passed?

Luap99: Also I guess you may want to ignore ENOENT here unless that is considered a fatal error? Not sure if anything can race for your use case bu in a generic API that likely should not cause an error?

internal/dedup/dedup.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+			if !d.Type().IsRegular() {
+				return nil
+			}
+			info, err := d.Info()
+			if err != nil {
+				logrus.Errorf("Failed to get file info for %q: %v", path, err)
+				return nil
+			}
```
Luap99, line 97: This can also fail with ENOENT if the file was removed after we read the dir.

internal/dedup/dedup.go
```diff
@@ -0,0 +1,167 @@
...
+				return nil
+			} else if visited {
+				return nil
+			}
+
+			h, err := getFileChecksum(options.HashMethod, path, info)
+			if err != nil {
+				logrus.Errorf("Failed to get checksum for %q: %v", path, err)
+				return nil
+			}
```
Luap99, line 116: Looks like this also opens the file path so it should handle ENOENT as well.

And similar to the other comment it is really no clear to me why logrus.Error is used here. If this is ever used from the podman API it will be invisible to users and scripts will not get exit > 0 potentially.
So if this is expected I would expect a comment saying so why.

Luap99 (commented):
Not really a full review as I am not familiar with the code base and intentions here but error handling could need some work I believe

internal/dedup/dedup.go
giuseppe: we are locking the store, and running the deduplication only on read-only layers.  How would ENOENT happen?  In that case, unless we decide to support running without any lock, wouldn't be better to know that something weird happened?

Ignoring messages was done deliberately so that the process can keep going and try to deduplicate as much as possible.

But I've no strong opinion about it, if you prefer to just return the error on the first failure, I am fine to change the current behavior

internal/dedup/dedup.go
giuseppe: that should not happen, the deduplication is done only with read-only layers

internal/dedup/dedup.go
giuseppe: I've changed the implementation to return immediately an error

mtrmac:
> documented that the dedup result accounts for previous runs, and changed the code to deal with multiple file systems

ACK

internal/dedup/dedup.go
giuseppe: I think we can't ignore ENOENT, since it works only with read-only layers and the package is private anyway, so can't be used outside of c/storage

internal/dedup/dedup.go
Luap99: Ack, if we know everything is locked and there can be no ENOENT race there then it is fine, as I said I don't know enough of the context of how/when it is run.

As for errors I think immediately retuning seems better as it returns a clear error to the caller.

Luap99 (commented):
LGTM for the error parts I looked at least. For the actual storage details I defer to others

PR #2170, opened 2024-11-11
add support for unpacking layers with xattrs set on FreeBSD

kimminss0:
Closes #2168

This PR implements system call wrappers for `extattr` on FreeBSD, and adds support for unpacking layers with external attributes(`xattr`). The implementation mimics how FreeBSD's linux compatibility layer and BSD tar handles xattr using extattr system calls, while both has some limitations that only two of the namespaces `user` and `system` are supported among the four – the rest, `trusted` and `security` remain unsupported.

PR #2177, opened 2024-11-20
pkg/system: add mimicked xattr syscalls on FreeBSD

kimminss0:
This PR adds mimicked `xattr` system calls on FreeBSD using `extattr`. Previous PR #2170 enabled that compatibility on FreeBSD, but only for a very specific area.

EDIT: Sync with the revised changes


rhatdan:
/approve
LGTM
@nalind @mtrmac @giuseppe @saschagrunert @flouthoc PTAL

rhatdan:
@dfr PTAL

dfr:
LGTM

nalind:
Why is this new package wrapping the functions with the same names as ones in `pkg/system` on darwin and linux, instead of adding the freebsd implementation directly to that package?  It's not clear to me what benefit we're gaining from the added layer of indirection here.

kimminss0:
The original intention was to let `pkg/system` be only a set of trivial wrappers of system calls; xattr implementation using extattr system call just seemed non-trivial to me.  But I agree that introducing this additional layer might be unnecessary. I'll merge them into `pkg/system`.

drivers/chown_unix.go
```diff
@@ -101,7 +101,8 @@ func (c *platformChowner) LChown(path string, info os.FileInfo, toHost, toContai
 	}
 	if uid != int(st.Uid) || gid != int(st.Gid) {
 		cap, err := system.Lgetxattr(path, "security.capability")
-		if err != nil && !errors.Is(err, system.EOPNOTSUPP) && !errors.Is(err, system.EOVERFLOW) && err != system.ErrNotSupportedPlatform {
+		if err != nil && !errors.Is(err, system.EOPNOTSUPP) && !errors.Is(err, syscall.ENOTSUP) &&
```
mtrmac, line 104: The direct reference to `syscall.E…` here breaks the abstraction to an extent; could that error be published by the `system` subpackage as well?

(Alternatively, do we really need all three of `system.ENOTSUP`, `system.EOPNOTSUPP`, `system.ErrNotSupportedPlatform`? I know very little about extended attribute implementations/compatibility.)

mtrmac (commented):
I like the general direction of this work.

drivers/chown_unix.go
kimminss0: I inspected `lgetxattr(2)`, `lsetxattr(2)`, and `llistxattr(2)` on Linux, as well as `getxattr(2)`, `setxattr(2)`, and `listxattr(2)` on MacOS. I found that these system calls actually return `ENOTSUP` instead of `EOPNOTSUPP`. Hence, `system.ENOTSUP` and `system.EOPNOTSUPP` are sufficient.

Note that `EOPNOTSUPP` and `ENOTSUP` have the same value in Linux according to `errno(3)`.

mtrmac: For the record, note to self:
- The UNIX spec says `EOPNOTSUPP` is for sockets, `ENOTSUP` for other situations (and allows the two to be the same value).
- macOS apparently differentiates the two in the kernel, but maps them back to the same value in libc in most modes
- Linux kernel uses `EOPNOTSUPP` exclusively (and in the extended attribute code), glibc defines `ENOTSUP` to be equal.

PR #2181, opened 2024-11-27
pkg/system: add fromStatT for NetBSD

tydavis:
Identical to openbsd's implementation, but is not included when compiling on netbsd due to using named files rather than build arguments.

This adds a file which is part of the patches (to the `vendor` directory) for the pkgsrc build.

### DIFF
--- a/pkg/system/stat_netbsd.go
+++ b/pkg/system/stat_netbsd.go
@@ -1,0 +1,13 @@
+package system
+
+import "syscall"
+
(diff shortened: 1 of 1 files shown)


tydavis:
Origin of the patch file: http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/sysutils/podman/patches/#dirlist

mtrmac (commented):
Thanks!

/lgtm
/approve

PR #2183, opened 2024-11-29
archive: store the override xattr with the inode type

giuseppe:
Closes: https://github.com/containers/storage/issues/2174

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -655,12 +655,20 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 	// so use hdrInfo.Mode() (they differ for e.g. setuid bits)
 	hdrInfo := hdr.FileInfo()
 
+	typeFlag := hdr.Typeflag
 	mask := hdrInfo.Mode()
+
+	// update also the implementation of ForceMask in pkg/chunked
 	if forceMask != nil {
 		mask = *forceMask
+		// If we have a forceMask, force the real type to either be a directory,
+		// a link, or a regular file.
+		if typeFlag != tar.TypeDir && typeFlag != tar.TypeSymlink && typeFlag != tar.TypeLink {
+			typeFlag = tar.TypeReg
+		}
 	}
 
-	switch hdr.Typeflag {
+	switch typeFlag {
 	case tar.TypeDir:
 		// Create directory unless it exists as a directory already.
 		// In that case we just want to merge the two
@@ -728,16 +736,6 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 		return fmt.Errorf("unhandled tar header type %d", hdr.Typeflag)
 	}
 
-	if forceMask != nil && (hdr.Typeflag != tar.TypeSymlink || runtime.GOOS == "darwin") {
-		value := idtools.Stat{
-			IDs:  idtools.IDPair{UID: hdr.Uid, GID: hdr.Gid},
-			Mode: hdrInfo.Mode() & 0o7777,
-		}
-		if err := idtools.SetContainersOverrideXattr(path, value); err != nil {
-			return err
-		}
-	}
-
 	// Lchown is not supported on Windows.
 	if Lchown && runtime.GOOS != windows {
 		if chownOpts == nil {
@@ -803,6 +801,18 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 				errs = append(errs, err.Error())
 				continue
 			}
+			return err
+		}
+	}
+
+	if forceMask != nil && (typeFlag == tar.TypeReg || typeFlag == tar.TypeDir || runtime.GOOS == "darwin") {
+		value := idtools.Stat{
+			IDs:   idtools.IDPair{UID: hdr.Uid, GID: hdr.Gid},
+			Mode:  hdrInfo.Mode(),
+			Major: int(hdr.Devmajor),
+			Minor: int(hdr.Devminor),
+		}
+		if err := idtools.SetContainersOverrideXattr(path, value); err != nil {
 			return err
 		}
 	}
@@ -1149,11 +1159,11 @@ loop:
 	}
 
 	if options.ForceMask != nil {
-		value := idtools.Stat{Mode: 0o755}
+		value := idtools.Stat{Mode: os.ModeDir | os.FileMode(0o755)}
 		if rootHdr != nil {
 			value.IDs.UID = rootHdr.Uid
 			value.IDs.GID = rootHdr.Gid
-			value.Mode = os.FileMode(rootHdr.Mode)
+			value.Mode = os.ModeDir | os.FileMode(rootHdr.Mode)
 		}
 		if err := idtools.SetContainersOverrideXattr(dest, value); err != nil {
 			return err
@@ -1379,7 +1389,7 @@ func remapIDs(readIDMappings, writeIDMappings *idtools.IDMappings, chownOpts *id
 			uid, gid = hdr.Uid, hdr.Gid
 			if xstat, ok := hdr.PAXRecords[PaxSchilyXattr+idtools.ContainersOverrideXattr]; ok {
 				attrs := strings.Split(string(xstat), ":")
-				if len(attrs) == 3 {
+				if len(attrs) >= 3 {
 					val, err := strconv.ParseUint(attrs[0], 10, 32)
 					if err != nil {
 						uid = int(val)
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -1262,6 +1262,24 @@ func (c *chunkedDiffer) copyAllBlobToFile(destination *os.File) (digest.Digest, 
 	return originalRawDigester.Digest(), err
 }
 
+func typeToOsMode(typ string) (os.FileMode, error) {
+	switch typ {
+	case TypeReg, TypeLink:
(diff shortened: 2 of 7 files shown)


giuseppe:
related fuse-overlayfs change: https://github.com/containers/fuse-overlayfs/pull/434

pkg/archive/archive.go
```diff
@@ -660,7 +660,17 @@ func createTarFile(path, extractDir string, hdr *tar.Header, reader io.Reader, L
 		mask = *forceMask
 	}
 
-	switch hdr.Typeflag {
+	typeFlag := hdr.Typeflag
+
+	if forceMask != nil {
+		// If we have a forceMask, force the real type to either be a directory,
+		// or a regular file.
+		if typeFlag != tar.TypeDir {
```
mtrmac, line 668: What about symlinks? This turns them into regular files, and the destination seems to be lost.

pkg/chunked/storage_linux.go
```diff
@@ -1487,7 +1487,7 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 	for i := range mergedEntries {
 		r := &mergedEntries[i]
 		if options.ForceMask != nil {
-			value := idtools.FormatContainersOverrideXattr(r.UID, r.GID, int(r.Mode))
+			value := idtools.FormatContainersOverrideXattrDevice(r.UID, r.GID, int(r.Mode), int(r.Devmajor), int(r.Devminor))
```
mtrmac, line 1490: Doesn’t this need to handle `r.Type`?

And there is a `SetContainersOverrideXattr` call in this file that seems to need updating.

pkg/idtools/idtools.go
```diff
@@ -369,54 +370,124 @@ func checkChownErr(err error, name string, uid, gid int) error {
...
+			if err != nil {
+				return stat, err
+			}
+		} else if strings.HasPrefix(typ, "char") {
+			stat.Mode |= os.ModeCharDevice
+			stat.Major, stat.Minor, err = parseDevice(typ)
+			if err != nil {
+				return stat, err
+			}
+		}
```
mtrmac, line 466: Else fail?

pkg/idtools/idtools.go
```diff
@@ -436,7 +507,7 @@ func SafeChown(name string, uid, gid int) error {
 				}
 			}
 		}
-		value := Stat{IDPair{uid, gid}, mode}
+		value := Stat{IDPair{uid, gid}, mode, 0, 0}
```
mtrmac, line 510: Does this need to handle file types?

pkg/idtools/idtools.go
```diff
@@ -464,7 +535,7 @@ func SafeLchown(name string, uid, gid int) error {
...
```
mtrmac, line 538: Does this need to handle file types?

pkg/idtools/idtools.go
```diff
@@ -436,7 +507,7 @@ func SafeChown(name string, uid, gid int) error {
 				}
```
mtrmac, line 507: The parser above expects exactly 3 fields. Same in `SafeLchown`.

mtrmac (commented):
ACK conceptually, but there seems even more to be done.

- symbolic links are now not handled consistently.
- `check.go` interprets `force_mask`; it needs to be taught about this
- Doesn’t the code in `overlay.Driver.create` around creating `diff` need to also set `os.ModeDir` now?
- The parser in `pkg/archive.remapIDs` seems to need updating (preferably consolidate all parsers to one function)

(It gets confusing… if only the OS had separate types for “permission bits” and “the full mode including a `ModeType`.)

giuseppe:
pushed a new version, and verified manually both partial pulls and with regular pulls

pkg/idtools/idtools_test.go
```diff
@@ -273,3 +276,155 @@ func TestIsContiguous(t *testing.T) {
...
+		minor    int
+		expected string
+	}{
+		{"regular file", 1000, 1001, 0o644, 0, 0, "1000:1001:0644:file"},
+		{"directory", 1000, 1001, 0o755 | int(os.ModeDir), 0, 0, "1000:1001:0755:dir"},
+		{"symlink", 1000, 1001, 0o777 | int(os.ModeSymlink), 0, 0, "1000:1001:0777:symlink"},
+		{"pipe", 1000, 1001, 0o666 | int(os.ModeNamedPipe), 0, 0, "1000:1001:0666:pipe"},
+		{"socket", 1000, 1001, 0o700 | int(os.ModeSocket), 0, 0, "1000:1001:0700:socket"},
+		{"block device", 1000, 1001, 0o600 | int(os.ModeDevice), 8, 0, "1000:1001:0600:block-8-0"},
+		{"char device", 1000, 1001, 0o600 | int(os.ModeDevice|os.ModeCharDevice), 1, 3, "1000:1001:0600:char-1-3"},
```
mtrmac, line 389: Also test permission handling - at least `ISUID`/`ISGID`/`ISVTX`, or possibly each of the 12 bits.

mtrmac (commented):
ACK to the symlink semantics change.

Note also the outstanding items in the top-level review comment in https://github.com/containers/storage/pull/2183#pullrequestreview-2473801817 .

pkg/idtools/idtools_test.go
```diff
@@ -273,3 +278,180 @@ func TestIsContiguous(t *testing.T) {
...
+		{"directory", 1001, 1012, fs.FileMode(0o755) | os.ModeDir, 0, 0, "1001:1012:0755:dir"},
+		{"symlink", 1002, 1013, fs.FileMode(0o777) | os.ModeSymlink, 0, 0, "1002:1013:0777:symlink"},
+		{"pipe", 1003, 1014, fs.FileMode(0o666) | os.ModeNamedPipe, 0, 0, "1003:1014:0666:pipe"},
+		{"socket", 1004, 1015, fs.FileMode(0o700) | os.ModeSocket, 0, 0, "1004:1015:0700:socket"},
+		{"block device", 1005, 1016, fs.FileMode(0o600) | os.ModeDevice, 8, 0, "1005:1016:0600:block-8-0"},
+		{"char device", 1006, 1017, fs.FileMode(0o600) | os.ModeDevice | os.ModeCharDevice, 1, 3, "1006:1017:0600:char-1-3"},
+
+		{"setuid bit", 1012, 1022, fs.FileMode(0o4755), 0, 0, "1012:1022:4755:file"},
+		{"setgid bit", 1013, 1023, fs.FileMode(0o2755), 0, 0, "1013:1023:2755:file"},
+		{"sticky bit", 1014, 1024, fs.FileMode(0o1777), 0, 0, "1014:1024:1777:file"},
```
mtrmac, line 395: `07000` are not the right `fs.Mode*` values; AFAICS they have no assigned meaning.

mtrmac (commented):
Don’t `SafeChown`/`SafeLchown` still need a bit of an update?

giuseppe:
> Don’t `SafeChown`/`SafeLchown` still need a bit of an update?

I've changed them further. I don't really know how they are used on darwin.  I see the same data is also set on the file itself.

mtrmac:
IIRC the general idea has been to allow unprivileged macOS users to extract volumes, but I’d need to look up the git history to confirm.

pkg/idtools/idtools_test.go
```diff
@@ -273,3 +278,180 @@ func TestIsContiguous(t *testing.T) {
...
+		{"directory", 1001, 1012, fs.FileMode(0o755) | os.ModeDir, 0, 0, "1001:1012:0755:dir"},
+		{"symlink", 1002, 1013, fs.FileMode(0o777) | os.ModeSymlink, 0, 0, "1002:1013:0777:symlink"},
+		{"pipe", 1003, 1014, fs.FileMode(0o666) | os.ModeNamedPipe, 0, 0, "1003:1014:0666:pipe"},
+		{"socket", 1004, 1015, fs.FileMode(0o700) | os.ModeSocket, 0, 0, "1004:1015:0700:socket"},
+		{"block device", 1005, 1016, fs.FileMode(0o600) | os.ModeDevice, 8, 0, "1005:1016:0600:block-8-0"},
+		{"char device", 1006, 1017, fs.FileMode(0o600) | os.ModeDevice | os.ModeCharDevice, 1, 3, "1006:1017:0600:char-1-3"},
+
+		{"setuid bit", 1012, 1022, fs.FileMode(0o4755) | fs.ModeSetuid, 0, 0, "1012:1022:4755:file"},
+		{"setgid bit", 1013, 1023, fs.FileMode(0o2755) | fs.ModeSetgid, 0, 0, "1013:1023:2755:file"},
+		{"sticky bit", 1014, 1024, fs.FileMode(0o1777) | fs.ModeSticky, 0, 0, "1014:1024:1777:file"},
```
mtrmac, line 395:
```suggestion
		{"setuid bit", 1012, 1022, fs.FileMode(0o755) | fs.ModeSetuid, 0, 0, "1012:1022:4755:file"},
		{"setgid bit", 1013, 1023, fs.FileMode(0o755) | fs.ModeSetgid, 0, 0, "1013:1023:2755:file"},
		{"sticky bit", 1014, 1024, fs.FileMode(0o777) | fs.ModeSticky, 0, 0, "1014:1024:1777:file"},
```
?

mtrmac (commented):
LGTM otherwise.

pkg/idtools/idtools_test.go
giuseppe: amended

Issue #2191, opened 2024-12-07
[zstd:chunked, rootless] `staging a partially-pulled layer: mkdirat /: no such file or directory`


cgwalters:
I have both zstd:chunked and composefs enabled in my rootless storage (so really right now, just zstd:chunked), i.e. in my `/etc/containers/storage.conf`:

```
[storage.options.pull_options]
enable_partial_images = "true"
convert_images = "true"
[storage.options.overlay]
use_composefs = "true"
```

Recently this seems to have basically started reliably failing to some images. Now here's the thing...because of bootc requirements 90% of my podman usage is actually podman-machine where none of this is enabled by default, so I don't have a rough sense of how recently this actually broke. (The non-podman-machine rootless use cases are mostly toolbox containers for me)

```
$ Client:       Podman Engine
Version:      5.3.1
API Version:  5.3.1
Go Version:   go1.23.3
Built:        Wed Nov 20 19:00:00 2024
OS/Arch:      linux/amd64
$ podman --log-level debug pull -q quay.io/rhel-devel-tools/rhel-developer-toolbox
...
DEBU[0002] Downloading /v2/rhel-devel-tools/rhel-developer-toolbox/blobs/sha256:65370aed77e699049343d487c0cb354188d87709caddece2324eedca262c8d1f 
DEBU[0002] GET https://quay.io/v2/rhel-devel-tools/rhel-developer-toolbox/blobs/sha256:65370aed77e699049343d487c0cb354188d87709caddece2324eedca262c8d1f 
INFO[0002] Error loading cache file for layer "345778062ceff8d733b1b965f15736af7afb36d5ae484b313ce88bf1021a3404": EOF 
DEBU[0002] Written lookaside cache for layer "345778062ceff8d733b1b965f15736af7afb36d5ae484b313ce88bf1021a3404" with length 68 
INFO[0002] Error loading cache file for layer "8991c80d55c1b1cb5c97b1f9ebab59a40595b30b529f2ed15ddf5e99809c67ab": EOF 
DEBU[0002] Written lookaside cache for layer "8991c80d55c1b1cb5c97b1f9ebab59a40595b30b529f2ed15ddf5e99809c67ab" with length 68 
DEBU[0002] Applying differ in /var/home/walters/.local/share/containers/storage/overlay/staging/1902067082/dir 
[...]

PR #2194, opened 2024-12-09
chunked: handle creating root directory

giuseppe:
if the file name is the root directory, do not attempt to use the base component since it is the empty string.

Closes: https://github.com/containers/storage/issues/2191

### DIFF
--- a/pkg/chunked/dump/dump.go
+++ b/pkg/chunked/dump/dump.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/containers/storage/pkg/chunked/internal"
+	storagePath "github.com/containers/storage/pkg/chunked/internal/path"
 	"golang.org/x/sys/unix"
 )
 
@@ -103,18 +104,8 @@ func getStMode(mode uint32, typ string) (uint32, error) {
 	return mode, nil
 }
 
-func sanitizeName(name string) string {
-	path := filepath.Clean(name)
-	if path == "." {
-		path = "/"
-	} else if path[0] != '/' {
-		path = "/" + path
-	}
-	return path
-}
-
 func dumpNode(out io.Writer, added map[string]*internal.FileMetadata, links map[string]int, verityDigests map[string]string, entry *internal.FileMetadata) error {
-	path := sanitizeName(entry.Name)
+	path := storagePath.CleanAbsPath(entry.Name)
 
 	parent := filepath.Dir(path)
 	if _, found := added[parent]; !found && path != "/" {
@@ -172,7 +163,7 @@ func dumpNode(out io.Writer, added map[string]*internal.FileMetadata, links map[
 		if entry.Type == internal.TypeSymlink {
 			payload = entry.Linkname
 		} else {
-			payload = sanitizeName(entry.Linkname)
+			payload = storagePath.CleanAbsPath(entry.Linkname)
 		}
 	} else {
 		if len(entry.Digest) > 10 {
--- a/pkg/chunked/filesystem_linux.go
+++ b/pkg/chunked/filesystem_linux.go
@@ -16,6 +16,7 @@ import (
 	driversCopy "github.com/containers/storage/drivers/copy"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/chunked/internal"
+	storagePath "github.com/containers/storage/pkg/chunked/internal/path"
 	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/vbatts/tar-split/archive/tar"
 	"golang.org/x/sys/unix"
@@ -49,10 +50,37 @@ type fileMetadata struct {
 	skipSetAttrs bool
 }
 
+// splitPath takes a file path as input and returns two components: dir and base.
+// Differently than filepath.Split(), this function handles some edge cases.
+// If the path refers to a file in the root directory, the returned dir is "/".
+// The returned base value is never empty, it never contains any slash and the
+// value "..".
+func splitPath(path string) (string, string, error) {
+	path = storagePath.CleanAbsPath(path)
+	dir, base := filepath.Split(path)
+	if base == "" {
+		base = "."
+	}
+	// Remove trailing slashes from dir, but make sure that "/" is preserved.
+	dir = strings.TrimSuffix(dir, "/")
+	if dir == "" {
+		dir = "/"
+	}
+
+	if strings.Contains(base, "/") {
+		// This should never happen, but be safe as the base is passed to *at syscalls.
+		return "", "", fmt.Errorf("internal error: splitPath(%q) contains a slash", path)
+	}
+	return dir, base, nil
+}
+
 func doHardLink(dirfd, srcFd int, destFile string) error {
-	destDir, destBase := filepath.Split(destFile)
+	destDir, destBase, err := splitPath(destFile)
+	if err != nil {
+		return err
+	}
 	destDirFd := dirfd
-	if destDir != "" && destDir != "." {
+	if destDir != "/" {
 		f, err := openOrCreateDirUnderRoot(dirfd, destDir, 0)
 		if err != nil {
 			return err
@@ -72,7 +100,7 @@ func doHardLink(dirfd, srcFd int, destFile string) error {
 		return nil
 	}
 
-	err := doLink()
+	err = doLink()
 
 	// if the destination exists, unlink it first and try again
 	if err != nil && os.IsExist(err) {
@@ -281,8 +309,11 @@ func openFileUnderRootFallback(dirfd int, name string, flags uint64, mode os.Fil
 	// If O_NOFOLLOW is specified in the flags, then resolve only the parent directory and use the
 	// last component as the path to openat().
 	if hasNoFollow {
-		dirName, baseName := filepath.Split(name)
-		if dirName != "" && dirName != "." {
+		dirName, baseName, err := splitPath(name)
+		if err != nil {
+			return -1, err
+		}
+		if dirName != "/" {
 			newRoot, err := securejoin.SecureJoin(root, dirName)
 			if err != nil {
 				return -1, err
@@ -449,9 +480,12 @@ func appendHole(fd int, name string, size int64) error {
 }
 
 func safeMkdir(dirfd int, mode os.FileMode, name string, metadata *fileMetadata, options *archive.TarOptions) error {
-	parent, base := filepath.Split(name)
+	parent, base, err := splitPath(name)
+	if err != nil {
+		return err
+	}
 	parentFd := dirfd
-	if parent != "" && parent != "." {
+	if parent != "/" {
 		parentFile, err := openOrCreateDirUnderRoot(dirfd, parent, 0)
 		if err != nil {
 			return err
@@ -507,9 +541,12 @@ func safeLink(dirfd int, mode os.FileMode, metadata *fileMetadata, options *arch
 }
 
 func safeSymlink(dirfd int, metadata *fileMetadata) error {
-	destDir, destBase := filepath.Split(metadata.Name)
+	destDir, destBase, err := splitPath(metadata.Name)
+	if err != nil {
+		return err
+	}
 	destDirFd := dirfd
-	if destDir != "" && destDir != "." {
+	if destDir != "/" {
 		f, err := openOrCreateDirUnderRoot(dirfd, destDir, 0)
 		if err != nil {
 			return err
@@ -543,9 +580,12 @@ func (d whiteoutHandler) Setxattr(path, name string, value []byte) error {
 }
 
 func (d whiteoutHandler) Mknod(path string, mode uint32, dev int) error {
-	dir, base := filepath.Split(path)
+	dir, base, err := splitPath(path)
+	if err != nil {
+		return err
+	}
 	dirfd := d.Dirfd
-	if dir != "" && dir != "." {
+	if dir != "/" {
 		dir, err := openOrCreateDirUnderRoot(d.Dirfd, dir, 0)
 		if err != nil {
 			return err
--- a/pkg/chunked/filesystem_linux.go
+++ b/pkg/chunked/filesystem_linux.go
@@ -409,7 +409,8 @@ func openOrCreateDirUnderRoot(dirfd int, name string, mode os.FileMode) (*os.Fil
 
 	if errors.Is(err, unix.ENOENT) {
 		parent := filepath.Dir(name)
-		if parent != "" {
+		// do not create the root directory, it should always exist
+		if parent != name {
 			pDir, err2 := openOrCreateDirUnderRoot(dirfd, parent, mode)
 			if err2 != nil {
 				return nil, err
--- a/pkg/chunked/filesystem_linux_test.go
+++ b/pkg/chunked/filesystem_linux_test.go
@@ -127,7 +127,7 @@ func TestSafeMkdir(t *testing.T) {
 	require.NoError(t, err)
 
 	dir, err := openFileUnderRoot(rootFd, dirName, syscall.O_DIRECTORY|syscall.O_CLOEXEC, 0)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	err = dir.Close()
 	assert.NoError(t, err)
 }
@@ -167,7 +167,7 @@ func TestSafeLink(t *testing.T) {
 
 	// validate it was created
 	newFile, err := openFileUnderRoot(rootFd, linkName, syscall.O_RDONLY, 0)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	st := syscall.Stat_t{}
 	err = syscall.Fstat(int(newFile.Fd()), &st)
@@ -216,7 +216,7 @@ func TestSafeSymlink(t *testing.T) {
 
 	// validate it was created
 	newFile, err := openFileUnderRoot(rootFd, linkName, syscall.O_RDONLY, 0)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	st2 := syscall.Stat_t{}
 	err = syscall.Fstat(int(newFile.Fd()), &st2)
--- a/pkg/chunked/filesystem_linux_test.go
+++ b/pkg/chunked/filesystem_linux_test.go
@@ -2,6 +2,7 @@ package chunked
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"os"
 	"path"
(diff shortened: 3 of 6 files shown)


pkg/chunked/storage_linux.go
```diff
@@ -1280,6 +1280,14 @@ func typeToOsMode(typ string) (os.FileMode, error) {
 	return 0, fmt.Errorf("unknown file type %q", typ)
 }
 
+func isRootDirectory(path string) bool {
+	path = filepath.Clean(path)
+	for strings.HasPrefix(path, "../") {
```
rhatdan, line 1285: Why not 

path=strings.Prefix(s, ".//")

Do we care about "./"?

giuseppe: '/' is automatically removed by `filepath.Clean()` AFAICS, I've added a new test case to validate a sequence of `./`.

yes `strings.CutPrefix` is better.  Changed and pushed a new version

pkg/chunked/filesystem_linux.go
```diff
@@ -458,6 +458,9 @@ func safeMkdir(dirfd int, mode os.FileMode, name string, metadata *fileMetadata,
 		defer parentFile.Close()
 		parentFd = int(parentFile.Fd())
 	}
+	if base == "" {
```
mtrmac, line 461: Please add a comment on when this can happen.

pkg/chunked/storage_linux.go
```diff
@@ -1280,6 +1280,18 @@ func typeToOsMode(typ string) (os.FileMode, error) {
 	return 0, fmt.Errorf("unknown file type %q", typ)
 }
 
+func isRootDirectory(path string) bool {
```
mtrmac, line 1283: I’m rather unhappy about this extra and somewhat unusual code path. Could we have `safeMkdir` return a “resolved to `dirfd`” value which precisely matches its decision?

giuseppe: we don't need this for safeMkdir but to detect what tar entry refers to the root, so we can store its file mode and report it to the caller

pkg/chunked/filesystem_linux.go
```diff
@@ -458,10 +458,15 @@ func safeMkdir(dirfd int, mode os.FileMode, name string, metadata *fileMetadata,
...
-
-	if err := unix.Mkdirat(parentFd, base, uint32(mode)); err != nil {
-		if !os.IsExist(err) {
-			return &fs.PathError{Op: "mkdirat", Path: name, Err: err}
+	// It can happen if name is a path in a form like "./".  When it happens, set base to point to
+	// the current directory.
+	if base == "" {
+		base = "."
+	} else {
+		if err := unix.Mkdirat(parentFd, base, uint32(mode)); err != nil {
```
mtrmac, line 466: ⚠️  AFAICS with `r.Name == "../..`, this ends up doing `Mkdirat(dirfd, "..")`.

That’s, luckily, guaranteed to fail with `EEXIST`, so I don’t think there is any actual vulnerability, but it shows that the logic is imprecise. This, similarly, applies to some other code here, I think, and probably needs a universal solution.

And below, we are trying to `setFileAttrs` on `dirfd/..` — in that case I open `openFileUnderRoot` prevents that from breaking out.

pkg/chunked/storage_linux.go
mtrmac: Yes — but see https://github.com/containers/storage/pull/2194/files#r1876699526 . I think all of the root, and root`/..` , handling, needs _some_ kind of cleanup / restructuring, and I suspect that would be fairly likely to make this function unnecessary.

My first instinct would be to factor out the repeated 
```go
	parent, base := filepath.Split(name)
	parentFd := dirfd
	if parent != "" && parent != "." {
		parentFile, err := openOrCreateDirUnderRoot(dirfd, parent, 0)
		if err != nil {
			return err
		}
		defer parentFile.Close()
		parentFd = int(parentFile.Fd())
	}
```
into a single helper (preparePathForCreation returning an object?). And then, I don’t know, _maybe_ call that in `ApplyDiff`, and centralize that way?

I haven’t given this much thought now.

pkg/chunked/filesystem_linux.go
```diff
@@ -458,10 +458,15 @@ func safeMkdir(dirfd int, mode os.FileMode, name string, metadata *fileMetadata,
 		defer parentFile.Close()
 		parentFd = int(parentFile.Fd())
 	}
-
-	if err := unix.Mkdirat(parentFd, base, uint32(mode)); err != nil {
-		if !os.IsExist(err) {
-			return &fs.PathError{Op: "mkdirat", Path: name, Err: err}
+	// It can happen if name is a path in a form like "./".  When it happens, set base to point to
```
mtrmac, line 461: Per
```go
package main

import (
	"fmt"
	"path/filepath"
)

func main() {
	for _, in := range []string{"a", "a/b", "a//b", "a/b/", "/a", "/a/b", "/a//b", "a/b/", ".", "./", "/", "", "../", "..", "../..", "../../"} {
		p := filepath.Clean(in)
		a, b := filepath.Split(p)
		fmt.Printf("%q\t%q\t%q\t%q\n", in, p, a, b)
	}
}
```
`./` does not trigger this — but a pure `/` does.

pkg/chunked/filesystem_linux.go
giuseppe: yes `openFileUnderRoot` prevents opening files above dirfd.

I've modified the code to prevent using "..", even if it is not a security problem

pkg/chunked/filesystem_linux.go
```diff
@@ -49,8 +49,31 @@ type fileMetadata struct {
...
+func splitPath(path string) (string, string) {
+	path = filepath.Clean(path)
+	dir, base := filepath.Split(path)
+	if isRootDirectory(dir) {
+		dir = ""
+		if base == ".." {
+			base = "."
+		}
+	} else {
+		dir = removeDotDotPrefix(filepath.Clean(dir))
```
mtrmac, line 67: (I think the `filepath.Clean` here does nothing, but it doesn’t hurt.)

pkg/chunked/filesystem_linux.go
```diff
@@ -49,8 +49,31 @@ type fileMetadata struct {
...
+		base = "."
+	}
+	return dir, base
+}
+
 func doHardLink(dirfd, srcFd int, destFile string) error {
-	destDir, destBase := filepath.Split(destFile)
+	destDir, destBase := splitPath(destFile)
 	destDirFd := dirfd
 	if destDir != "" && destDir != "." {
```
mtrmac, line 78: AFAICS, and per the added documentation, `destDir` can never be `"."` now (adding a few tests to `TestSplitPath` to confirm this would be nice), so in all callers of `splitPath`, this condition can be simplified.

mtrmac (commented):
Thanks! This is nice.

pkg/chunked/filesystem_linux.go
giuseppe: was used as a lazy replacement for ` strings.TrimSuffix(path, "/")` and have a consistent result for the tests.  I'll replace it.

pkg/chunked/filesystem_linux.go
giuseppe: thanks, I'll simplify the checks.

pkg/chunked/storage_linux.go
```diff
@@ -1280,6 +1280,13 @@ func typeToOsMode(typ string) (os.FileMode, error) {
 	return 0, fmt.Errorf("unknown file type %q", typ)
 }
 
+// cleanAbsDirectory removes any ".." and "." from the path
+// and ensures it starts with a "/".  If the path refers to the root
+// directory, it returns "/".
+func cleanAbsDirectory(path string) string {
```
mtrmac, line 1286: Nice, this is elegant.

---

- `s/cleanAbsDirectory/cleanAbsPath`, it is used for non-directories as well
- This change, effectively, changes the semantics of `fsVerityDigestsKey`. That’s fine, and AFAICS it works correctly — but it seems that changing `dump.sanitizeName` to this implementation, for consistency, would be nice. Actually, I think moving this function to `pkg/chunked/internal/composefs` (or a new subpackage of `…/internal`), and calling the function in both places, would be best.

pkg/chunked/storage_linux_test.go
```diff
@@ -211,3 +211,39 @@ func TestTypeToOsMode(t *testing.T) {
 		})
 	}
 }
+
+func TestCleanAbsDirectory(t *testing.T) {
+	tests := []struct {
+		path     string
+		expected string
+	}{
+		{"", "/"},
```
mtrmac, line 220: (Absolutely non-blocking: Consider adding entries for `"."`, `".."`, and maybe `"foo/../.."`,`"/foo/../.."`)

pkg/chunked/filesystem_linux_test.go
```diff
@@ -334,3 +338,48 @@ func createTempFile(t *testing.T, dir, name string) *os.File {
...
+		{"/foo/", "/", "foo"},
+		{"/foo/bar", "/foo", "bar"},
+		{"/foo/bar/", "/foo", "bar"},
+		{"/foo/////bar/", "/foo", "bar"},
+		{"/home/foo/file.txt", "/home/foo", "file.txt"},
+		{"/home/foo////file.txt", "/home/foo", "file.txt"},
+		{"file", "/", "file"},
+		{"foo/", "/", "foo"},
+		{"foo/.", "/", "foo"},
+		{"foo/..", "/", "."},
```
mtrmac, line 371: (Absolutely non-blocking: Consider testing `"foo/bar/.."`.)

pkg/chunked/filesystem_linux.go
```diff
@@ -49,10 +49,33 @@ type fileMetadata struct {
...
+// empty string.
+// The returned base value is never empty, it never contains any slash and the
+// value "..".
+func splitPath(path string) (string, string, error) {
+	path = filepath.Clean(path)
+	dir, base := filepath.Split(path)
+	dir = cleanAbsDirectory(dir)
+	if base == "" || base == ".." {
+		base = "."
+	}
```
mtrmac, line 64:
```suggestion
	path = cleanAbsDirectory(path)
	dir, base := filepath.Split(path)
	if base == "" {
		base = "."
	}
```
would do the same thing a bit simpler way.

pkg/chunked/filesystem_linux.go
```diff
@@ -49,10 +49,33 @@ type fileMetadata struct {
 	skipSetAttrs bool
 }
 
+// splitPath takes a file path as input and returns two components: dir and base.
+// Differently than filepath.Split(), this function handles some edge cases.
+// If the path refers to a file in the root directory, the returned dir is the
+// empty string.
```
mtrmac, line 55: This is no longer correct.

pkg/chunked/storage_linux.go
mtrmac: `cleanAbsDirectory` + a single comparison to `"/"` is nice, short and elegant.

pkg/chunked/filesystem_linux.go
giuseppe: the difference is that it won't remove trailing slashes from `dir`, which probably doesn't matter.  I'll adjust to your version

mtrmac: Actually, wouldn’t a trailing slash break the `openOrCreateDirUnderRoot` calls to create a parent directory? `filepath.Dir("/a/")) == "/a"`. So either keep the code as is, or call `cleanAbsDirectory` twice, or…

[Somewhere in here is a discussion about using `filepath.Dir`/`filepath.Base` instead of the fairly naive `filepath.Split`, but I suspect neither of us has much taste for another complete rewrite.]

pkg/chunked/filesystem_linux.go
```diff
@@ -49,10 +49,31 @@ type fileMetadata struct {
 	skipSetAttrs bool
 }
 
+// splitPath takes a file path as input and returns two components: dir and base.
+// Differently than filepath.Split(), this function handles some edge cases.
+// If the path refers to a file in the root directory, the returned dir is "/".
+// The returned base value is never empty, it never contains any slash and the
+// value "..".
+func splitPath(path string) (string, string, error) {
```
cgwalters, line 57: In Rust the first would be `Option<str>` or so (e.g. ref [Path::parent](https://doc.rust-lang.org/std/path/struct.Path.html#method.parent)) which would have then forced all call sites to clearly handle its absence or presence in a single canonical form.

pkg/chunked/storage_linux.go
```diff
@@ -1544,10 +1544,10 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 			}
 		}
 
-		r.Name = filepath.Clean(r.Name)
+		r.Name = internal.CleanAbsPath(r.Name)
```
cgwalters, line 1547: One thing I didn't quite understand here is why these ones need to be absolute paths?

pkg/chunked/dump/dump.go
```diff
@@ -103,18 +103,8 @@ func getStMode(mode uint32, typ string) (uint32, error) {
...
-		path = "/"
-	} else if path[0] != '/' {
-		path = "/" + path
-	}
-	return path
-}
-
 func dumpNode(out io.Writer, added map[string]*internal.FileMetadata, links map[string]int, verityDigests map[string]string, entry *internal.FileMetadata) error {
-	path := sanitizeName(entry.Name)
+	path := internal.CleanAbsPath(entry.Name)
```
cgwalters, line 107: Right, composefs wants these to be absolute paths, so this makes sense.

(That said, it's kind of tempting to lift that and support relative paths too just on the principle that relative paths are IMO more "canonical" to use in use cases like tar)

cgwalters (approved):
I didn't do any deep review but looks sane to me.

pkg/chunked/filesystem_linux_test.go
```diff
@@ -334,3 +338,49 @@ func createTempFile(t *testing.T, dir, name string) *os.File {
...
+	}{
+		{"", "/", "."},
+		{".", "/", "."},
+		{"..", "/", "."},
+		{"../..", "/", "."},
+		{"../../..", "/", "."},
+		{"../../../foo", "/", "foo"},
+		{"../../../foo/..", "/", "."},
+		{"../../../foo/./../foo/bar/baz", "/foo/bar/", "baz"},
+		{"../../../foo/bar", "/foo/", "bar"},
```
mtrmac, line 356: Is the trailing `/` in `dir` acceptable? With https://github.com/containers/storage/pull/2194#discussion_r1878843841 , it seems to me that it would interfere with the automatic creation of parent directories.

pkg/chunked/internal/path.go
```diff
@@ -0,0 +1,12 @@
+package internal
+
+import (
+	"path/filepath"
+)
+
+// CleanAbsPath removes any ".." and "." from the path
+// and ensures it starts with a "/".  If the path refers to the root
+// directory, it returns "/".
+func CleanAbsPath(path string) string {
```
mtrmac, line 10: `pkg/chunked/internal` is included in all c/image callers (incl. ones that don’t use the c/storage transport, e.g. when they only want to use the registry client), compare the comment in `pkg/chunked/compression.go`. Please use some other subpackage.

pkg/chunked/filesystem_linux.go
mtrmac: What do you propose as a Go alternative?

AFAICS the primary value of optionals is that the _compiler_ understands and enforces value presence guarantees on various code paths. We just don’t have that. (The closes option is a pointer type, but Go doesn’t have a “non-`nil` pointer” type in the type system, so, again, no control flow enforcement.)

pkg/chunked/storage_linux.go
mtrmac: They don’t need to be, this is a very concise way to eliminate leading `../` entries (at the syntax level, which matters for `*at()` system calls; something still needs to care about symlinks trying to break out).

pkg/chunked/filesystem_linux.go
cgwalters: It was one of my usual “aside” comments, not actionable in this PR. I will try to prefix with “Aside:” or so more consistently.

mtrmac: No worries — I’m a similarly vocal Go hater, it’s just, as you say, frequently not actionable.

pkg/chunked/storage_linux.go
cgwalters: > this is a very concise way to eliminate leading ../ entries (at the syntax level, which matters for *at() system calls; something still needs to care about symlinks trying to break out).

Hmm, only if we're not using `RESOLVE_IN_ROOT` though...which, digging through we are, at least if it's supported.

Anyways, marking this conversation as resolved.

mtrmac: `RESOLVE_IN_ROOT` helps for creating a single file; we may need to create all parent directories.

cgwalters: AFAIK with RESOLVE_IN_ROOT, `../` is just going to be the same as `/` in that root, so any leading/unnecessary `../` is harmless.

But anyways either way I don't think it matters, this sub-thread is just for building up our own shared understanding.

(On that topic, currently. https://github.com/containers/composefs-rs builds up a userspace in-memory VFS that represents the target composefs metadata file, so the semantics of the kernel openat() as it applies to potentially hostile or just weird user space paths don't come into play)

pkg/chunked/filesystem_linux_test.go
giuseppe: that is a direct consequence of applying https://github.com/containers/storage/pull/2194#discussion_r1878704500

Added a  `strings.TrimSuffix` to drop any trailing slash, except for the root directory

pkg/chunked/internal/path.go
giuseppe: I assumed it was an acceptable cost.  Moved to `storage/internal`

mtrmac: I agree that the cost of this one function is trivial.

Breaking the precedent, and opening the door to adding more compression-unrelated functionality or dependencies into that subpackage, would be a long-term problem.

Arguably, with the addition of `pkg/chunked/internal/*` subpackages, this special property of `pkg/chunked/internal` itself is becoming less clear. Shall I file a PR to move that code into `pkg/chunked/internal/minimal` or something like that?

mtrmac (commented):
/lgtm

Thanks!

pkg/chunked/internal/path.go
giuseppe: > Arguably, with the addition of `pkg/chunked/internal/*` subpackages, this special property of `pkg/chunked/internal` itself is becoming less clear. Shall I file a PR to move that code into `pkg/chunked/internal/minimal` or something like that?

yes thanks, that would be better than what we have now

mtrmac: > Shall I file a PR to move that code into `pkg/chunked/internal/minimal` or something like that?

https://github.com/containers/storage/pull/2196 .

PR #2196, opened 2024-12-12
Move pkg/chunked/internal to pkg/chunked/internal/minimal

mtrmac:
We now have several `internal` subpackages of `pkg/chunked`, so delineate more explicitly the parts that should be kept as small as possible because the c/image compression package depends on them.

Should not change behavior.

---

Previously discussed in https://github.com/containers/storage/pull/2194#discussion_r1881081145 .

I don’t like the `internal/minimal` subpackage name much; I don’t know how to choose a name that both expresses “be careful not to add dependencies here” and that looks good at the call sites (which don’t all care about the minimizing dependencies). Suggestions for an alternative are very welcome.

### DIFF
--- a/pkg/chunked/cache_linux.go
+++ b/pkg/chunked/cache_linux.go
@@ -16,7 +16,7 @@ import (
 
 	storage "github.com/containers/storage"
 	graphdriver "github.com/containers/storage/drivers"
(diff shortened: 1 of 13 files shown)


giuseppe (approved):
LGTM

PR #2197, opened 2024-12-12
Fix `SafeChown` when `ContainersOverrideXattr` is unset

mtrmac:
This happens e.g. in `idtools.MkdirAllAndChownNew` .

@giuseppe PTAL, this is necessary for c/image tests on macOS.


giuseppe (approved):
/lgtm

PR #2208, opened 2025-01-10
[release-1.51] Use newer release of securejoin package to secure userns paths

kwilczynski:
The backport to release 1.51 brought about implementation of a fix to the [CVE-2024-9676](https://nvd.nist.gov/vuln/detail/CVE-2024-9676) / https://github.com/advisories/GHSA-wq2p-5pc6-wpgf that while deemed sufficient wasn't completely secure as there are potential side effects to it that would need to be fixed as identified by @cyphar. The alternative fix would also involve a hand-rolled implementation, which might or might not be sufficient. 

However, while discussing a possible way forward, it has become apparent that the latest available implementation within the securejoin package would be the preferred one anyway. The only reason why we were unable to use it was Go version compatibility issue—a too-new Go version was required for some ancillary changes, not the core implementation, and thus, the package could not be used.

That said, as part of some valiant effort, @cyphar was able to remove the blocker that prevented us from using a newer release of securejoin package with an older version of Go.

Thus, as there are no more reasons not to, update to the supported release of the securejoin package, which allows us to use the more secure implementation of the `secureOpen()` helper as seen in more modern releases.

Related:

- https://github.com/containers/storage/pull/2134
- https://github.com/containers/storage/pull/2135
- https://github.com/cri-o/cri-o/pull/8677
- https://github.com/cyphar/filepath-securejoin/pull/37
- https://github.com/cyphar/filepath-securejoin/releases/tag/v0.3.6


mheon:
LGTM on my end
@nalind @giuseppe PTAL

giuseppe (approved):
LGTM

PR #2216, opened 2025-01-21
Revert "fix(deps): update module github.com/cyphar/filepath-securejoin to v0.4.0"

Luap99:
This reverts commit 05dcfd3caa2b1f89b8ff75c8aafcc39ee9238257.

In particular this commit doesn't build when vendored into podman because runc fails to build with it.
https://github.com/opencontainers/runc/pull/4590

This is not visible in storage because runc is not imported here.


Luap99:
@mtrmac @nalind @rhatdan @mheon @TomSweeneyRedHat PTAL
This will block the podman v5.4 vendor dance if we do not merge this.

PR #2228, opened 2025-01-23
Graphdriver test cleanups

hanwen-flow:
These are a couple of things I saw when perusing the test code.

### DIFF
--- a/drivers/graphtest/graphtest_unix.go
+++ b/drivers/graphtest/graphtest_unix.go
@@ -65,6 +65,14 @@ func newDriver(t testing.TB, name string, options []string) *Driver {
 	runroot, err := os.MkdirTemp("", "storage-graphtest-")
 	require.NoError(t, err)
 
+	defer func() {
+		// Cannot use t.Cleanup(), some test files persist the
+		// driver across test functions.
+		if t.Failed() || t.Skipped() {
+			os.RemoveAll(runroot)
+			os.RemoveAll(root)
+		}
+	}()
 	return &Driver{newGraphDriver(t, name, options, root, runroot), root, runroot, 1}
 }
 
--- a/drivers/graphtest/graphtest_unix.go
+++ b/drivers/graphtest/graphtest_unix.go
@@ -65,7 +65,6 @@ func newDriver(t testing.TB, name string, options []string) *Driver {
 	runroot, err := os.MkdirTemp("", "storage-graphtest-")
 	require.NoError(t, err)
 
-	require.NoError(t, os.MkdirAll(root, 0o755))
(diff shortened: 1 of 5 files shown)


drivers/graphtest/graphtest_unix.go
```diff
@@ -47,6 +47,8 @@ func newGraphDriver(t testing.TB, name string, options []string, root string, ru
 	d, err := graphdriver.GetDriver(name, graphdriver.Options{DriverOptions: options, Root: root, RunRoot: runroot})
 	if err != nil {
 		t.Logf("graphdriver: %v\n", err)
+		os.RemoveAll(runroot)
+		os.RemoveAll(root)
```
kolyshkin: I think the correct way is to call `os.RemoveAll` in the same function which calls `MkdirTemp`, i.e. `newDriver`.

In there, you can do something like
```go
if t.Failed() || t.Skipped() {
  os.RemoveAll...
}
```

kolyshkin: Nah, that won't work.

In any case, it would be better to place the cleanup into the same function as the setup, and it might require some refactoring.

hanwen-flow: PTAL

kolyshkin: Yes, this looks good and is actually working (I was wrong it won't), thanks!

kolyshkin (approved):
lgtm

hanwen-flow:
> lgtm

thanks! the 'tide' check says that it still needs the 'lgtm' label.

drivers/graphtest/graphtest_unix.go
```diff
@@ -62,23 +62,32 @@ func newGraphDriver(t testing.TB, name string, options []string, root string, ru
...
-	require.NoError(t, os.MkdirAll(root, 0o755))
-	return &Driver{newGraphDriver(t, name, options, root, runroot), root, runroot, 1}
+	defer func() {
+		// Cannot use t.Cleanup(), some test files persist the
+		// driver across test functions.
+		if t.Failed() || t.Skipped() {
+			os.RemoveAll(runRoot)
+			os.RemoveAll(root)
+		}
+	}()
```
mtrmac, line 75: I can’t see how this could possibly work: the `defer` is going to run immediately after this function returns, before anything could get the opportunity to use the created driver. Am I missing something?

hanwen-flow: newGraphDriver can fail. In that case, there is no driver, but there are the temp dirs created in l.63 and l.65

mtrmac: Thanks, I did miss that.

PR #2229, opened 2025-01-23
Archive tests

hanwen-flow:
Some random cleanups I spotted when reading over the code.

### DIFF
--- a/pkg/archive/archive_test.go
+++ b/pkg/archive/archive_test.go
@@ -18,15 +18,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var tmp string
-
-func init() {
-	tmp = "/tmp/"
-	if runtime.GOOS == windows {
-		tmp = os.Getenv("TEMP") + `\`
-	}
-}
-
 var defaultArchiver = NewDefaultArchiver()
 
 func defaultTarUntar(src, dst string) error {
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

hanwen-flow:
> LGTM
thanks! the 'tide' check says that it still needs the 'lgtm' label.

nalind:
LGTM, thanks.  The package-scope "tmp" appears to be unused after these changes, might as well remove it.

hanwen-flow:
> LGTM, thanks. The package-scope "tmp" appears to be unused after these changes, might as well remove it.

there are a couple of uses under TestUntarPathWithInvalidDest and below. It's for Windows, and I don't have a windows machine to do tests.

nalind:
> > LGTM, thanks. The package-scope "tmp" appears to be unused after these changes, might as well remove it.
> 
> there are a couple of uses under TestUntarPathWithInvalidDest and below. It's for Windows, and I don't have a windows machine to do tests.

A spot-check with `GOOS=windows go test -c` reveals some breakage introduced by #2199 (#2246 hopes to fix it), but after working those out, I couldn't find the uses.  What am I missing?

hanwen-flow:
> What am I missing?

nothing. I was confused. Sorry!

nalind:
/lgtm
Thanks!

hanwen-flow:
> reveals some breakage introduced by

@nalind  - shouldn't this repo have some CI setup as well? It's vendored into important other packages, but the vendoring drops the tests.

as an aside, IIUC, the tests assume they run as root. Would it be useful to add some t.Skip() or t.Fail() statements to the effect?

nalind:
Ideally, CI would cover more things, yes.  Tests that we know will fail without certain privileges could stand to skip themselves if they know they don't have them, and at that point, having CI ensure that they run both with and without those privileges would be helpful.

PR #2231, opened 2025-01-27
chunked: use fallback mechanism on non-linux platforms

giuseppe:
inform the caller to fallback to an ordinary copy when not running on Linux.


dfr:
Looks good. I tested with podman built from the main branch with this change patched in and it successfully pulls the alpine image which was causing me problems.

PR #2233, opened 2025-01-29
homedir: check runtime dir validity using system boot_id

giuseppe:
Add a boot_id check to ensure the runtime directory is not reused after a system reboot.  This prevents the use of stale runtime directories when the runtime directory is created by pkh/homedir because the user is running without a proper user session.

If the file /proc/sys/kernel/random/boot_id is not present, the check is ignored.

### DIFF
--- a/pkg/homedir/homedir_unix.go
+++ b/pkg/homedir/homedir_unix.go
@@ -6,6 +6,7 @@ package homedir
 // NOTE: this package has originally been copied from github.com/docker/docker.
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -14,6 +15,7 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/unshare"
 	"github.com/sirupsen/logrus"
 )
@@ -133,6 +135,33 @@ func GetConfigHome() (string, error) {
 	return rootlessConfigHomeDir, rootlessConfigHomeDirError
 }
 
+// checkBootId checks if the boot_id of the system has changed since the last time the runtime directory was created.
+func checkBootId(path string) error {
+	systemBootId, err := os.ReadFile("/proc/sys/kernel/random/boot_id")
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			// The system does not have a boot_id, so ignore the check
+			return nil
+		}
+		return err
+	}
+
+	bootIdFile := filepath.Join(path, "bootid")
+	bootId, err := os.ReadFile(bootIdFile)
+	if err != nil {
+		// If the file does not exist, create it
+		if errors.Is(err, os.ErrNotExist) {
+			return ioutils.AtomicWriteFile(bootIdFile, systemBootId, 0o600)
+		}
+		return err
+	}
+
+	if string(bootId) != string(systemBootId) {
+		return fmt.Errorf("The boot_id of the system has changed, the runtime directory %q is not valid anymore, please remove it first", path)
+	}
+	return nil
+}
+
 // GetRuntimeDir returns a directory suitable to store runtime files.
 // The function will try to use the XDG_RUNTIME_DIR env variable if it is set.
 // XDG_RUNTIME_DIR is typically configured via pam_systemd.
@@ -170,6 +199,7 @@ func GetRuntimeDir() (string, error) {
 			st, err := os.Lstat(tmpDir)
 			if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() && isWriteableOnlyByOwner(st.Mode().Perm()) {
 				runtimeDir = tmpDir
+				rootlessRuntimeDirError = checkBootId(runtimeDir)
 			} else {
 				rootlessRuntimeDirError = fmt.Errorf("path %q exists and it is not writeable only by the current user", tmpDir)
 				return


pkg/homedir/homedir_unix.go
```diff
@@ -133,6 +134,33 @@ func GetConfigHome() (string, error) {
...
+			return nil
+		}
+		return err
+	}
+
+	bootIdFile := filepath.Join(path, "bootid")
+	bootId, err := os.ReadFile(bootIdFile)
+	if err != nil {
+		// If the file does not exist, create it
+		if os.IsNotExist(err) {
```
Luap99, line 152: should use errors.Is()

pkg/homedir/homedir_unix.go
```diff
@@ -133,6 +134,33 @@ func GetConfigHome() (string, error) {
...
+			return os.WriteFile(bootIdFile, systemBootId, 0o600)
```
Luap99, line 153: there are no locks for this so in theory there is a super unlikely race between two callers calling this check at the same time.

WriteFile() can cause a empty/partial written file and then the other caller would not read the full content thus failing the check below incorrectly.
So in short I think this must use the atomic write function we have here that uses rename to ensure we have no partial file.

Luap99 (commented):
Don't we already have the same check in podman
https://github.com/containers/podman/commit/3560ccd5dfa4816aae02ec323d7967beae2755ee?

Doing it twice seems a bit pointless though I guess in c/storage other users (i.e. buildah would benefit too)

giuseppe:
> Don't we already have the same check in podman [containers/podman@3560ccd](https://github.com/containers/podman/commit/3560ccd5dfa4816aae02ec323d7967beae2755ee)?
> 
> Doing it twice seems a bit pointless though I guess in c/storage other users (i.e. buildah would benefit too)

yeah you are right, we already check that in Podman.  I forgot about that.

The advantage of the check in Podman is that it is done only once, while in c/storage it covers more users.  So that isn't the reason for https://github.com/containers/podman/issues/25144

giuseppe:
> That said I have no experience how other users of these paths behave when it is not removed? Is there something critical in c/storage or buildah that does not work properly when the paths are not removed on boot?

the XDG_RUNTIME_DIR is passed also to the OCI runtime, so in general it would be better to clean it up but I am not aware of anything critical that would fail if the directory contains some garbage.

Wouldn't it be a problem for Buildah when it uses `common/libnetwork`?

Luap99:
> Wouldn't it be a problem for Buildah when it uses common/libnetwork?

Not yet at least, I never ported buildah over to use the rootless netns code. It currently only uses the plain pasta/slirp4netns as rootless. But yeah you are right once I allow custom networks in rootless buildah (not much work) then it would use the same code.

giuseppe:
@mheon what do you think about this PR?

pkg/homedir/homedir_unix.go
```diff
@@ -133,6 +135,33 @@ func GetConfigHome() (string, error) {
 	return rootlessConfigHomeDir, rootlessConfigHomeDirError
 }
 
+// checkBootId checks if the boot_id of the system has changed since the last time the runtime directory was created.
+func checkBootId(path string) error {
+	systemBootId, err := os.ReadFile("/proc/sys/kernel/random/boot_id")
```
mtrmac, line 140: This is not supported on all Unixes, at least macOS does not have `/proc`. I didn’t check FreeBSD.

pkg/homedir/homedir_unix.go
```diff
@@ -133,6 +135,33 @@ func GetConfigHome() (string, error) {
...
+	bootId, err := os.ReadFile(bootIdFile)
+	if err != nil {
+		// If the file does not exist, create it
+		if errors.Is(err, os.ErrNotExist) {
+			return ioutils.AtomicWriteFile(bootIdFile, systemBootId, 0o600)
+		}
+		return err
+	}
+
+	if string(bootId) != string(systemBootId) {
```
mtrmac, line 159: (Absolutely non-blocking: `bytes.Compare` expresses the same thing without implying a copy, and is clearer about the intent (we don’t care whether the value is valid text). This works fine as is, and it is not performance-sensitive.)

pkg/homedir/homedir_unix.go
```diff
@@ -170,6 +199,7 @@ func GetRuntimeDir() (string, error) {
 			st, err := os.Lstat(tmpDir)
 			if err == nil && int(st.Sys().(*syscall.Stat_t).Uid) == os.Geteuid() && isWriteableOnlyByOwner(st.Mode().Perm()) {
 				runtimeDir = tmpDir
+				rootlessRuntimeDirError = checkBootId(runtimeDir)
```
mtrmac, line 202: Non-blocking:

This works… but it introduces a new semantics of this code path ("set an error, but continue to the success case of `rootlessRuntimeDir = runtimeDir`”) and it’s getting a bit unnecessarily unwieldy.

I think using a very ordinary function with a very ordinary “if we see an error, return it immediately” control flow would be easier to follow and maintain. Then the `rootlessRuntimeDirOnce` closure would just do `rootlessRuntimeDir, rootlessRuntimeDirError = theExpensiveLogicWeDontWantToRepeatedlyRun()` (and `theExpensive…Run` could be a top-level function, easier to unit test).

(Also, the existing `rootlessRuntimeDir*` could be replaced with a single `sync.OnceValues` variable — I don’t feel strongly about that at all.)

pkg/homedir/homedir_unix.go
```diff
@@ -133,6 +135,33 @@ func GetConfigHome() (string, error) {
...
+	if err != nil {
+		// If the file does not exist, create it
+		if errors.Is(err, os.ErrNotExist) {
+			return ioutils.AtomicWriteFile(bootIdFile, systemBootId, 0o600)
+		}
+		return err
+	}
+
+	if string(bootId) != string(systemBootId) {
+		return fmt.Errorf("The boot_id of the system has changed, the runtime directory %q is not valid anymore, please remove it first", path)
```
mtrmac, line 160: “The boot_id of the system has changed” is an implementation explanation, but it is not directly in terms the user works with (my first instinct might be to search for ways how `boot_id` can be preserved).

Maybe something like “the runtime directory %q has persisted across a reboot”? I can’t think of a short enough explanation.

Absolutely non-blocking: Would we want two versions of this message, one for can’t-happen situations (`/run/user` persisted; just say so) and one for “sad but inevitable situations” (`/var/tmp/storage-run-500`, which we fell back to, was not cleared; give more guidance about what to set up? I have no idea how frequently users run into this.

pkg/homedir/homedir_unix.go
giuseppe: when the file doesn't exist, the check is not performed (line 142)

mtrmac: Good point, but, still, I think it would be better to avoid that; it leaves noise in syscall traces, and inattentive readers might not notice that the protection is not working on those systems.

giuseppe:
since it is not clear whether the patch is needed, I am closing it

Issue #2237, opened 2025-01-30
Possible handle leak


Bbulatov:
Good day!
During the static analysis was found possible handle leak:
1) The handle 'diff' is created at pkg/chunked/storage_linux.go:153 by calling function 'archive.DecompressStream' and lost at pkg/chunked/storage_linux.go:160 or at pkg/chunked/storage_linux.go:170.

![Image](https://github.com/user-attachments/assets/139398f3-a9ca-48de-9ba1-cbdbdb90b018)

Bbulatov:
I apologize for the large number of pull requests.

PR #2238, opened 2025-01-30
Fix handle leak

Bbulatov:
Possible solve

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -150,6 +150,17 @@ func parsePullOptions(store storage.Store) pullOptions {
 }
 
 func (c *chunkedDiffer) convertTarToZstdChunked(destDirectory string, payload *os.File) (int64, *seekableFile, digest.Digest, map[string]string, error) {
+	diff, err := archive.DecompressStream(payload)
+        if err != nil {
+                return 0, nil, "", nil, err
+        }
+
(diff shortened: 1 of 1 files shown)


mtrmac (changes requested):
This
- doesn’t fix the issue at all, AFAICS
- Introduces a different leak

This is what `defer`, and maybe the existing `if !succeeded` pattern, is for.

mtrmac:
Reference #2237 .

pkg/chunked/storage_linux.go
```diff
@@ -151,9 +151,15 @@ func parsePullOptions(store storage.Store) pullOptions {
...
-		return 0, nil, "", nil, err
-	}
+        if err != nil {
+                return 0, nil, "", nil, err
+        }
+
+        defer func() {
+		if err := diff.Close(); err != nil {
+			return 0, nil, "", nil, err
+		}
```
mtrmac, line 161: I’m not sure this can fail at all, but if it did, we should prefer the original error that caused this function to return, not a possibly-derived error that happens during a cleanup.

I think just a `defer diff.Close()` here would be good enough.

mtrmac (commented):
https://github.com/containers/storage/blob/main/CONTRIBUTING.md#sign-your-prs please, we can’t really even look at PRs with unclear copyright status.

PR #2240, opened 2025-01-30
Bump maxTocSize to 150 MB

mtrmac:
We have seen an image with:
- total size 1.43 GB
- uncompressed zstd:chunked manifest size of 91.7 MB
- uncompressed tar-split size (not constrained by maxTocSize) 310 MB

Without more infrastructure, we are just guessing about what the system we are running on can support, so, for now, *shrug*, bump the number.

Eventually we should stream the data from/to disk, making this much less relevant; that makes building the infrastructure to estimate available memory unattractive.

---

Per previous conversation in #2230 . Cc: @giuseppe @TomSweeneyRedHat .


TomSweeneyRedHat (commented):
LGTM

PR #2246, opened 2025-02-04
Fix compilation of pkg/archive tests on non-Linux

nalind:
Unixy systems have Lutimes(), and we should have stubbed it out for non-Unixy systems.

### DIFF
--- a/pkg/archive/changes_test.go
+++ b/pkg/archive/changes_test.go
@@ -12,7 +12,6 @@ import (
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/system"
 	"github.com/stretchr/testify/require"
-	"golang.org/x/sys/unix"
 )
 
 func copyDir(src, dst string) error {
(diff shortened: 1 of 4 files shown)


pkg/archive/changes_unix_test.go
```diff
@@ -0,0 +1,14 @@
+//go:build !windows
+
```
kolyshkin, line 2: Strictly speaking,
1. it should be `unix` rather than `!windows` (otherwise it won't work on e.g. plan9).
2. the build tag is redundant as we have `_unix_test.go` in the file name (supported since Go 1.19)

pkg/archive/changes_unix_test.go
```diff
@@ -0,0 +1,14 @@
...
+package archive
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+// resetSymlinkTimesLinux sets the atime and mtime of a symlink to a known value, to test
```
kolyshkin, line 9:
```suggestion
// resetSymlinkTimes sets the atime and mtime of a symlink to a known value, to test
```

nalind: Oof, shoulda caught that.  Thanks!

pkg/archive/changes_unix_test.go
nalind: I'll have to remember that.  Thanks!

pkg/archive/changes_test.go
```diff
@@ -42,16 +41,6 @@ type FileData struct {
 	permissions os.FileMode
 }
 
-// resetSymlinkTimesLinux sets the atime and mtime of a symlink to a known value, to test
-// whether changes to the symlink target are detected correctly.
-func resetSymlinkTimesLinux(path string) error {
-	if runtime.GOOS != linux {
```
kolyshkin, line 48: nit: with that, `const linux` in archive.go (introduced by commit f5c1ad912) is now unused.

PR #2250, opened 2025-02-06
overlay: Add error prefixing for mount.MakePrivate

cgwalters:
I'm hitting this error path in
https://github.com/containers/buildah/issues/5976
and it was not obvious to me which bit of code was failing. Add an error prefix so it's easy to find.


mtrmac (commented):
LGTM

Luap99 (commented):
LGTM

PR #2251, opened 2025-02-07
pkg/fileutils: add LinkOrCopy()

Luap99:
Add a function to copy via reflink if possible or fall back to the normal io.Copy().
c/image uses this function internally[1] however I like to export this and use it in podman and I think this package here in c/storage is the most logical location for it. Note compared to the c/image version I am calling unix.IoctlFileClone() directly instead of the raw syscall function.

[1] https://github.com/containers/image/tree/main/internal/reflink

### DIFF
--- a/pkg/fileutils/reflink_linux.go
+++ b/pkg/fileutils/reflink_linux.go
@@ -1,0 +1,20 @@
+package fileutils
+
+import (
+	"io"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// ReflinkOrCopy attempts to reflink the source to the destination fd.
+// If reflinking fails or is unsupported, it falls back to io.Copy().
+func ReflinkOrCopy(src, dst *os.File) error {
+	err := unix.IoctlFileClone(int(dst.Fd()), int(src.Fd()))
+	if err == nil {
+		return nil
+	}
+
+	_, err = io.Copy(dst, src)
+	return err
+}
--- a/pkg/fileutils/reflink_unsupported.go
+++ b/pkg/fileutils/reflink_unsupported.go
@@ -1,0 +1,15 @@
+//go:build !linux
+
+package fileutils
+
+import (
+	"io"
+	"os"
+)
+
+// ReflinkOrCopy attempts to reflink the source to the destination fd.
+// If reflinking fails or is unsupported, it falls back to io.Copy().
+func ReflinkOrCopy(src, dst *os.File) error {
+	_, err := io.Copy(dst, src)
+	return err
+}


Luap99:
@mtrmac @giuseppe PTAL

pkg/fileutils/reflink_linux.go
```diff
@@ -0,0 +1,20 @@
...
+
+// LinkOrCopy attempts to reflink the source to the destination fd.
+// If reflinking fails or is unsupported, it falls back to io.Copy().
+func LinkOrCopy(src, dst *os.File) error {
+	err := unix.IoctlFileClone(int(dst.Fd()), int(src.Fd()))
+	if err == nil {
+		return nil
+	}
+
+	_, err = io.Copy(dst, src)
```
giuseppe, line 18: could we use Copy from pkg/pools?  Would the new dependency be a problem?

Luap99: Mhh does this help us here? Would this one allocation for the buffer from io.Copy() even be noticeable. Also io.Copy() uses interfaces such as WriteTo() which if you pass a file should already try to use sendfile() and not need a buffer

Thus I am not sure if using a pre existing buffer would help us here. Also would there not be contention if many goroutines try to use pkg/pool at the same time? 

The more practical concern `pkg/ioutils` is imported from pkg/pools so we cannot import it here and would need to move the function into pkg/pool.

giuseppe (approved):
LGTM

pkg/fileutils/reflink_linux.go
giuseppe: Not worth the effort. Thanks

pkg/fileutils/reflink_linux.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"io"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// LinkOrCopy attempts to reflink the source to the destination fd.
+// If reflinking fails or is unsupported, it falls back to io.Copy().
+func LinkOrCopy(src, dst *os.File) error {
```
mtrmac, line 12: `ReflinkOrCopy`? I would read this as “hard link or copy”, which is frequently undesirable.

mtrmac (commented):
Thanks! I agree c/storage is the right place for this.

I wonder whether we want the “…OrCopy” part; it seems somewhat attractive to me to have `drivers/copy.CopyRegularToFile` call this new function. (And there’s… friction: that function needs to explicitly implement `doCopyWithFileRange` _only_ because it relies on `pools.Copy`, which defeats the equivalent of `doCopyWithFileRange` inside the standard library.)

I think we can start with this `…OrCopy` function for now, and we can always add a reflink-only function later — I’m pointing this out in case others think differently.

pkg/fileutils/reflink_linux.go
Luap99: agree, `ReflinkOrCopy()` sounds better

cgwalters:
On modern Linux `copy_file_range` automatically uses reflinks if available, but is also generally useful since if reflinks aren't available, it avoids the copy to and from userspace for the file data. So I think it should be generally preferred over the `FICLONE` ioctl nowadays.

BTW Rust automatically uses that by default via specialization for its [std::io::copy](https://doc.rust-lang.org/std/io/fn.copy.html#platform-specific-behavior)

mtrmac (commented):
/lgtm

Thanks!

Issue #2257, opened 2025-02-13
Podman invocations using different `--root`s cannot share an `--imagestore`


seanlafferty-ibm:
### Issue Description

I suspect this was never an intended feature, but our use case is CI runners. We have hundreds of replicas, each of which has podman installed. Jobs always start with container pulls (which is the slowest part of the job). If a job pulls an image, we'd like for it to be written to a shared imagestore for subsequent jobs to reuse.

I created a `imgs` volume on the host, which I mounted into all of the runners. Then, all the runners were given a storage.conf with the contents:
```
[storage]
  driver = "overlay"
  imagestore = "/home/runner/imgs"
```

My expectation was that any podman pull inside of a runner would read/write this shared imagestore. However, the pulls step on each other, corrupting the imagestore/root.

### Steps to reproduce the issue

```
# pull with root1, shared imgstore (works)
podman --root /tmp/root1 --imagestore /tmp/imgs pull mirror.gcr.io/library/bash

# pull with root2, shared imgstore (doesn't work)
podman --root /tmp/root2 --imagestore /tmp/imgs pull mirror.gcr.io/library/bash

ERRO[0000] Image mirror.gcr.io/library/bash exists in local storage but may be corrupted (remove the image to resolve the issue): layer not known
```

### Describe the results you received

Corrupted imagestore

### Describe the results you expected

Non-corrupted imagestore

### podman info output

```yaml
host:
  arch: amd64
  buildahVersion: 1.37.6
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  - rdma
  - misc
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.12-1.el9.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.12, commit: c0564282e9befb7804c3642230f8e94f1b2ba9f8'
  cpuUtilization:
    idlePercent: 95.6
    systemPercent: 1.94
    userPercent: 2.47
  cpus: 2
  databaseBackend: sqlite
  distribution:
    distribution: rhel
    version: "9.4"
  eventLogger: journald
  freeLocks: 2033
  hostname: slaffy-github1.fyre.ibm.com
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.14.0-427.42.1.el9_4.x86_64
  linkmode: dynamic
  logDriver: journald
  memFree: 357359616
  memTotal: 3837394944
[...]

PR #2258, opened 2025-02-14
layers: write read only layers to imagestore

giuseppe:
when an imagestore is used, a R/O layer must be written to the layers.json under the imagestore, not graphroot.

The lock on the imagestore is already taken through the multipleLockFile{} mechanism in place.

Closes: https://github.com/containers/storage/issues/2257

### DIFF
--- a/layers.go
+++ b/layers.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"maps"
+	"math/bits"
 	"os"
 	"path"
 	"path/filepath"
@@ -46,11 +47,13 @@ const (
 
 type layerLocations uint8
 
-// The backing store is split in two json files, one (the volatile)
-// that is written without fsync() meaning it isn't as robust to
-// unclean shutdown
+// The backing store is split in three json files.
+// The volatile store is written without fsync() meaning it isn't as robust to unclean shutdown.
+// Optionally, an image store can be configured to store RO layers.
+// The stable store is used for the remaining layers that don't go into the other stores.
 const (
 	stableLayerLocation layerLocations = 1 << iota
+	imageStoreLayerLocation
 	volatileLayerLocation
 
 	numLayerLocationIndex = iota
@@ -58,6 +61,10 @@ const (
 
 func layerLocationFromIndex(index int) layerLocations {
 	return 1 << index
+}
+
+func indexFromLayerLocation(location layerLocations) int {
+	return bits.TrailingZeros(uint(location))
 }
 
 // A Layer is a record of a copy-on-write layer that's stored by the lower
@@ -164,8 +171,8 @@ type Layer struct {
 	// ReadOnly is true if this layer resides in a read-only layer store.
 	ReadOnly bool `json:"-"`
 
-	// volatileStore is true if the container is from the volatile json file
-	volatileStore bool `json:"-"`
+	// location is the location of the store where the layer is present.
+	location layerLocations `json:"-"`
 
 	// BigDataNames is a list of names of data items that we keep for the
 	// convenience of the caller.  They can be large, and are only in
@@ -430,14 +437,6 @@ type layerStore struct {
 	driver drivers.Driver
 }
 
-// The caller must hold r.inProcessLock for reading.
-func layerLocation(l *Layer) layerLocations {
-	if l.volatileStore {
-		return volatileLayerLocation
-	}
-	return stableLayerLocation
-}
-
 func copyLayer(l *Layer) *Layer {
 	return &Layer{
 		ID:                 l.ID,
@@ -455,7 +454,7 @@ func copyLayer(l *Layer) *Layer {
 		TOCDigest:          l.TOCDigest,
 		CompressionType:    l.CompressionType,
 		ReadOnly:           l.ReadOnly,
-		volatileStore:      l.volatileStore,
+		location:           l.location,
 		BigDataNames:       copySlicePreferringNil(l.BigDataNames),
 		Flags:              copyMapPreferringNil(l.Flags),
 		UIDMap:             copySlicePreferringNil(l.UIDMap),
@@ -658,7 +657,11 @@ func (r *layerStore) layersModified() (lockfile.LastWrite, bool, error) {
 	// modified manually, then we have to reload the storage in
 	// any case.
 	for locationIndex := 0; locationIndex < numLayerLocationIndex; locationIndex++ {
-		info, err := os.Stat(r.jsonPath[locationIndex])
+		rpath := r.jsonPath[locationIndex]
+		if rpath == "" {
+			continue
+		}
+		info, err := os.Stat(rpath)
 		if err != nil && !os.IsNotExist(err) {
 			return lockfile.LastWrite{}, false, fmt.Errorf("stat layers file: %w", err)
 		}
@@ -794,6 +797,9 @@ func (r *layerStore) load(lockedForWriting bool) (bool, error) {
 	for locationIndex := 0; locationIndex < numLayerLocationIndex; locationIndex++ {
 		location := layerLocationFromIndex(locationIndex)
 		rpath := r.jsonPath[locationIndex]
+		if rpath == "" {
+			continue
+		}
 		info, err := os.Stat(rpath)
 		if err != nil {
 			if !os.IsNotExist(err) {
@@ -820,9 +826,7 @@ func (r *layerStore) load(lockedForWriting bool) (bool, error) {
 				continue // skip invalid duplicated layer
 			}
 			// Remember where the layer came from
-			if location == volatileLayerLocation {
-				layer.volatileStore = true
-			}
+			layer.location = location
 			layers = append(layers, layer)
 			ids[layer.ID] = layer
 		}
@@ -843,7 +847,7 @@ func (r *layerStore) load(lockedForWriting bool) (bool, error) {
 			if conflict, ok := names[name]; ok {
 				r.removeName(conflict, name)
 				errorToResolveBySaving = ErrDuplicateLayerNames
-				modifiedLocations |= layerLocation(conflict)
+				modifiedLocations |= conflict.location
 			}
 			names[name] = layers[n]
 		}
@@ -939,7 +943,7 @@ func (r *layerStore) load(lockedForWriting bool) (bool, error) {
 				incompleteDeletionErrors = errors.Join(incompleteDeletionErrors,
 					fmt.Errorf("deleting layer %#v: %w", layer.ID, err))
 			}
-			modifiedLocations |= layerLocation(layer)
+			modifiedLocations |= layer.location
 		}
 		if err := r.saveLayers(modifiedLocations); err != nil {
 			return false, err
@@ -1008,7 +1012,7 @@ func (r *layerStore) save(saveLocations layerLocations) error {
 // The caller must hold r.lockfile locked for writing.
 // The caller must hold r.inProcessLock for WRITING.
 func (r *layerStore) saveFor(modifiedLayer *Layer) error {
-	return r.save(layerLocation(modifiedLayer))
+	return r.save(modifiedLayer.location)
 }
 
 // The caller must hold r.lockfile locked for writing.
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@nalind @mtrmac @Luap99 PTAL

store.go
```diff
@@ -1545,6 +1547,9 @@ func (s *store) putLayer(rlstore rwLayerStore, rlstores []roLayerStore, id, pare
 			GIDMap:         copySlicePreferringNil(gidMap),
 		}
 	}
+	if !writeable && s.imageStoreDir != "" {
+		options.ImageStore = true
+	}
```
mtrmac, line 1552: I don’t generally see why this logic should exist at this layer of the call stack. `layerStore` was already initialized with the imageStore location, it seems to me that it can make this decision on its own.

Alternatively… for `imageStore`, the search, and the decision where to write is managed at the store level. There is no `multipleLockFile` for images. This looks _so_ different, and I don’t understand the design nearly enough to tell whether the difference is necessary.

layers.go
```diff
@@ -1180,6 +1192,7 @@ func newROLayerStore(rundir string, layerdir string, driver drivers.Driver) (roL
 		mountsLockfile: nil,
 		rundir:         rundir,
 		jsonPath: [numLayerLocationIndex]string{
+			filepath.Join(layerdir, "layers.json"),
 			filepath.Join(layerdir, "layers.json"),
```
mtrmac, line 1196: The code does not anticipate this. Saving would be completely broken (although that shouldn’t happen for the RO store, sure), and loads call this invalid (“There should be no duplicated ids between json files, but lets check to be sure”).

layers.go
```diff
@@ -51,6 +51,7 @@ type layerLocations uint8
 // unclean shutdown
 const (
```
mtrmac, line 52: The comment above needs updating.

layers.go
```diff
@@ -456,6 +463,7 @@ func copyLayer(l *Layer) *Layer {
 		CompressionType:    l.CompressionType,
 		ReadOnly:           l.ReadOnly,
 		volatileStore:      l.volatileStore,
+		imageStore:         l.imageStore,
```
mtrmac, line 466: I think using the location index and/or enum directly would ultimately be simpler than 2 separate bools. Now we have to worry about both being set.

mtrmac (commented):
This does not touch drivers/overlay where I am refusing to review PRs adding new conditionals, but it’s pretty close… please find a reviewer who understands this.

I’m basically only reviewing the `store.go` API part.

---

As for the core of the thing, changing where layer metadata is stored seems like a Big Deal. If we were doing the wrong thing all this time, how come no-one has noticed until now?!

I suppose, this is preexisting (and fixing this will mean touching drivers/overlay…): I’m also pretty seriously uncomfortable about this code making a decision where to record metadata, vs. the overlay driver making a decision where to create the physical layer (or, if there are multiple layers with the same ID, which one to use), _completely independently_. That seems to be one of those things that is guaranteed to drift out of sync.

store.go
giuseppe: the difference is necessary because images are always directed to the image store.  Instead for layers, we write only the RO layers to the imagestore; container layers are written to the main store.

layers.go
giuseppe: we end up here only when setting up the additional stores, so there is no split of `image store` and `container store` there

layers.go
giuseppe: I'll change to use directly the location in the layer, but we will still need to handle the two bools that we get from the options (since Volatile is part of the API at this point)

giuseppe:
> As for the core of the thing, changing where layer metadata is stored seems like a Big Deal. If we were doing the wrong thing all this time, how come no-one has noticed until now?!

nobody tried to use the image store as a self store, instead of an additional store.

> I suppose, this is preexisting (and fixing this will mean touching drivers/overlay…): I’m also pretty seriously uncomfortable about this code making a decision where to record metadata, vs. the overlay driver making a decision where to create the physical layer (or, if there are multiple layers with the same ID, which one to use), _completely independently_. That seems to be one of those things that is guaranteed to drift out of sync.

I was against this feature when it was first added, because it is a mess and makes everything more complicated.   IMO this could have been an additional store and the decision where to pull an image handled at a higher level, but it was needed for Kubernetes and now we have to deal with it.

layers.go
mtrmac: That doesn’t make the consumers of `jsonPath` any more ready to accept such inputs.

To be explicit, I think the situation should be modeled precisely the way we mean it: perhaps make the `jsonPath` entry `""` (or turn `jsonPath` into a map with missing entries?), and then handle that situation in all consumers (skipping loads, refusing saves, and so on.)

layers.go
mtrmac: Well… we _could_ push that decision all the way to `store.go`, but really https://github.com/containers/storage/pull/2258#discussion_r1958485813 needs to be decided first.

As is, I think one location mapping from bools to the enum, and the simplification in other places, is worth it already.

mtrmac:
> now we have to deal with it.

… :| I am deeply dissatisfied with how the current implementation requires `store.go`, `layers.go` and `drivers/overlay` _precisely_ orchestrate the various layer stores, lookups, and locking, with no explicit visibility into the interdependencies. But, also, I don’t know what a good design would look like and I don’t know what it would take to figure that out.

[I guess something along the lines of: would want to just have one “ordinary” `layerStore` per directory, and have lookups etc. all managed by the “search list” in `store go`? The part that does not trivially work would, I think, be mounting a layer, which would need to be passed some kind of a “handle” to all parent layers across several stores. I don’t know what that would look like (locking? locking order?), and I can’t immediately see how to split that redesign into smaller steps.]

mtrmac:
> > now we have to deal with it.
> 
> … :| I am deeply dissatisfied with how the current implementation requires `store.go`, `layers.go` and `drivers/overlay`  to _precisely_ orchestrate the various layer stores, lookups, and locking, with no explicit visibility into the interdependencies. But, also, I don’t know what a good design would look like and I don’t know what it would take to figure that out.
> 
> [I guess something along the lines of: would want to just have one “ordinary” `layerStore` per directory, and have lookups etc. all managed by the “search list” in `store go`? The part that does not trivially work would, I think, be mounting a layer, which would need to be passed some kind of a “handle” to all parent layers across several stores. I don’t know what that would look like (locking? locking order?), and I can’t immediately see how to split that redesign into smaller steps.]

store.go
giuseppe: should remapped images be always created on the image store?

I think it is better if they are created on the same store where their parent layer is.  I've added a patch to the PR to implement this behavior

layers.go
giuseppe: used "" in this case and changed the users of `jsonPath` to ignore that entry

layers.go
```diff
@@ -1033,12 +1031,15 @@ func (r *layerStore) saveLayers(saveLocations layerLocations) error {
 			continue
 		}
 		rpath := r.jsonPath[locationIndex]
+		if rpath == "" {
+			continue
```
mtrmac, line 1035: This should be impossible to reach — so it should fail instead of silently not saving the layer.

layers.go
```diff
@@ -1144,6 +1145,7 @@ func (s *store) newLayerStore(rundir, layerdir, imagedir string, driver drivers.
 		rundir:         rundir,
 		jsonPath: [numLayerLocationIndex]string{
 			filepath.Join(layerdir, "layers.json"),
+			filepath.Join(imagedir, "layers.json"),
```
mtrmac, line 1148: I don’t think doing this unconditionally works — for `imagedir == ""`, this sets the path to "layers.json"`.

store.go
mtrmac: Fine, I can see how that is a good location to choose

… but now, again, we have one part of the code that decides the location in `layers.go` and another here in `store.go`.
WDYT about `layerStore` explicitly tracking whether imageDir is set (compare https://github.com/containers/storage/pull/2258#discussion_r1960040690 ), and then this condition can be moved inside the new `pickLocation`.

layers.go
```diff
@@ -1399,6 +1402,27 @@ func (r *layerStore) create(id string, parentLayer *Layer, names []string, mount
 		selinux.ReserveLabel(mountLabel)
 	}
 
+	pickLocation := func(volatile, imageStore bool, parentLayer *Layer) layerLocations {
```
mtrmac, line 1405: Can this be made a top-level method? That would, at least in principle, allow writing unit tests :)

store.go
mtrmac: > and then this condition can be moved

no, that doesn’t work _that_ easily because, AFAICS, we can’t currently tell between “ordinary” RO layers and the ID-mapping layers from `imageTopLayerForMapping`; we would still need to add a new option to `create`. Still, an `thisIsAnIDMappingLayer` option set in `store.Go`, and all location decisions centralized in `pickLocations`, seems worth considering.

giuseppe:
took care of the last comments and pushed a new version

mtrmac:
AFAICS https://github.com/containers/storage/pull/2258#discussion_r1958493677 and some variant of https://github.com/containers/storage/pull/2258#discussion_r1958485813 are still outstanding.

giuseppe:
> AFAICS [#2258 (comment)](https://github.com/containers/storage/pull/2258#discussion_r1958493677) and some variant of [#2258 (comment)](https://github.com/containers/storage/pull/2258#discussion_r1958485813) are still outstanding.

if the goal is to not have duplication of the logic in `store.go`, then I think we can just pass down `writeble`, so that now all the logic is in the same function in `layers.go`

giuseppe:
working on a fix for the failing tests...

layers.go
```diff
@@ -60,6 +63,10 @@ func layerLocationFromIndex(index int) layerLocations {
 	return 1 << index
 }
 
+func indexFromLayerLocation(location layerLocations) int {
```
mtrmac, line 66: ACK… Aesthetically, I think that using a bitmap for `layerLocations` has turned out to be too clever, and that using an ordinary enum, and an ordinary set type [… which… doesn’t exist in Go standard library…] for the parameter of `saveLayers`, would be simpler with negligible performance downside.

But now that this code has been written, that local aesthetics is not a reason to ask for a change, please keep it as it is — just recording the thought for possible benefit of future investigation.

layers.go
```diff
@@ -1329,6 +1342,17 @@ func (r *layerStore) PutAdditionalLayer(id string, parentLayer *Layer, names []s
 	return copyLayer(layer), nil
 }
 
+func pickStoreLocation(jsonPath *[numLayerLocationIndex]string, volatile, writeable bool) layerLocations {
```
mtrmac, line 1345: (Absolutely non-blocking: This could be a method on `*layerStore`)

layers.go
```diff
@@ -1329,6 +1342,17 @@ func (r *layerStore) PutAdditionalLayer(id string, parentLayer *Layer, names []s
...
+	switch {
+	case volatile:
+		return volatileLayerLocation
+	case !writeable && jsonPath[indexFromLayerLocation(imageStoreLayerLocation)] != "":
```
mtrmac, line 1349: I think a previous version tried to place the ID-mapping layer to the same location as a parent … and, second thought, without more direct coordination with the overlay driver (see other parts of this review), that would be out of sync with the overlay driver? Fine.

mtrmac (commented):
Thanks!

LGTM, just please check that I understand the removal of the “inherit from parent” code. If so, feel free to merge.

layers.go
giuseppe: fixed and pushed a new version

layers.go
giuseppe: yes, we kind of use `writeable` as `container layer`.  Not sure if it makes sense to specify differently, but would be nice to move this logic out of the graph driver driver somehow so that it gets an explicit location where the layer must be created

mtrmac (commented):
/lgtm

Thanks!

I guess I tricked myself into doing a full review after all :)

PR #2261, opened 2025-02-18
Write tar fixes

hanwen-flow:


hanwen-flow:
@giuseppe - let me propose the error handling commit in a separate PR, because it evidently needs more discussion.

pkg/archive/archive.go
```diff
@@ -842,178 +842,188 @@ func Tar(path string, compression Compression) (io.ReadCloser, error) {
...
+// `options.IncludeFiles` (if non-nil) or not in
+// `options.ExcludePatterns`.
+func TarWithOptionsTo(dest io.WriteCloser, srcPath string, options *TarOptions) error {
 	// Fix the source path to work with long path names. This is a no-op
 	// on platforms other than Windows.
 	srcPath = fixVolumePathPrefix(srcPath)
 
+	// make changes thread-safe.
+	copy := *options
+	options = &copy
```
giuseppe, line 856: are you protecting from something like `go TarWithOptionsTo(...)`?  In this case, `options` could still change between the time the function is called and here?  Shouldn't be the caller responsibility to take care of this?

hanwen-flow: it's the caller's responsibility not to mutate data it passes into a goroutine. We're allowing the following usage without race conditions:

```
opts := TarOptions { ... }
go TarWithOptions(dir1, &opts)
go TarWithOptions(dir2, &opts)
```

hanwen-flow: ping. @giuseppe

giuseppe: thanks, I had missed the part where we change the changes inside the goroutine itself

giuseppe (approved):
LGTM

hanwen-flow:
> LGTM

thanks; this needs another LGTM, I think. Do you know someone who could provide this?

rhatdan:
@mtrmac @nalind PTAL

mtrmac (commented):
- Do we _need_ to make `TarWithOptionsTo` public? c/storage attempts to have a stable API, every new function is a commitment. See also `NewPatternMatcher`.
- This does not actually substantially improve error handling, because errors in the goroutine are only logged and don’t cause a failure in the consumer.

pkg/archive/archive.go
```diff
@@ -856,179 +856,189 @@ func Tar(path string, compression Compression) (io.ReadCloser, error) {
...
+	go func() {
+		if err := TarWithOptionsTo(compressWriter, srcPath, options); err != nil {
+			logrus.Errorf("Can't create tar archive: %s", err)
+		}
+		if err := compressWriter.Close(); err != nil {
+			logrus.Errorf("Can't close compress writer: %s", err)
+		}
+		if err := pipeWriter.Close(); err != nil {
+			logrus.Errorf("Can't close pipe writer: %s", err)
 		}
```
mtrmac, line 1042: Moving these things out of `defer` is incorrect, panics will now cause the consuming goroutine to hang.

I think the right pattern for a pipe-producing goroutine is to:
- Split as much as possible into an “ordinary” function.
- Add a single-purpose goroutine function whose _only purpose_ is to _guarantee_ error handling, i.e. that the goroutine will exit, that the consuming goroutine will not block, and that errors, if any, are correctly reported.

Compare https://github.com/containers/image/blob/e4a0c90bdc94a22db79aad8d278c203819ffab42/copy/compression.go#L414 .

pkg/archive/archive.go
```diff
@@ -856,179 +856,189 @@ func Tar(path string, compression Compression) (io.ReadCloser, error) {
...
-					if err == io.ErrClosedPipe {
-						return err
-					}
-				}
-				return nil
-			}); err != nil {
-				logrus.Errorf("%s", err)
-				return
-			}
+	compressWriter, err := CompressStream(pipeWriter, options.Compression)
```
mtrmac, line 1028: AFAICS `compressWriter` is ~entirely local to the pipe-writing goroutine; the reading goroutine doesn’t need to know it exists.

pkg/archive/archive.go
```diff
@@ -856,179 +856,189 @@ func Tar(path string, compression Compression) (io.ReadCloser, error) {
...
 }
 
-// TarWithOptions creates an archive from the directory at `path`, only including files whose relative
-// paths are included in `options.IncludeFiles` (if non-nil) or not in `options.ExcludePatterns`.
-func TarWithOptions(srcPath string, options *TarOptions) (io.ReadCloser, error) {
+// TarWithOptionsTo writes `srcPath` as an uncompressed tar archive to
+// `dest`, only including files whose relative paths are included in
+// `options.IncludeFiles` (if non-nil) or not in
+// `options.ExcludePatterns`.
+func TarWithOptionsTo(dest io.WriteCloser, srcPath string, options *TarOptions) error {
```
mtrmac, line 863: Given how this function silently ignores many errors (and, it seems, the plan is to _change that_), I’m very uncomfortable with making it a public API at this point.

pkg/archive/archive.go
```diff
@@ -856,179 +856,189 @@ func Tar(path string, compression Compression) (io.ReadCloser, error) {
...
 	// Fix the source path to work with long path names. This is a no-op
 	// on platforms other than Windows.
 	srcPath = fixVolumePathPrefix(srcPath)
 
+	// make changes thread-safe.
```
mtrmac, line 868: “thread-safe” is not really the point; this is equally problematic for an unthreaded caller which doesn’t expect `options` to be modified.

… and in this case, I think just having a local variable for `IncludeFiles` would also work, without making a large copy of the whole `TarOptions`; that copy ~negates all benefits from passing it by reference.

pkg/archive/archive.go
```diff
@@ -1215,6 +1225,10 @@ func untarHandler(tarArchive io.Reader, dest string, options *TarOptions, decomp
 	dest = filepath.Clean(dest)
 	if options == nil {
 		options = &TarOptions{}
+	} else {
+		// make writes thread-safe.
```
mtrmac, line 1229: “thread-safe” is not really the point.

Also, looking at how `ExcludePatterns` is used, AFAICS the change from `nil` to `{}` is completely unnecessary, and could be removed everywhere instead.

mtrmac (commented):
.

pkg/archive/archive.go
mtrmac: >  “ordinary” function.

By this I mean a function that returns `error` the usual way, without having to use some other error reporting/propagating mechanism. Then that unusual error reporting mechanism only exists in _one_ easily-verifiable place and users who add new error paths to the large function don’t risk forgetting to do something unusual.

pkg/archive/archive.go
```diff
@@ -856,179 +856,189 @@ func Tar(path string, compression Compression) (io.ReadCloser, error) {
...
-					if rebaseName != string(filepath.Separator) {
-						// Special case the root directory to replace with an
-						// empty string instead so that we don't end up with
-						// double slashes in the paths.
-						replacement = rebaseName
-					}
+// TarWithOptions creates an archive from the directory at `srcPath`, only including files whose relative
+// paths are included in `options.IncludeFiles` (if non-nil) or not in `options.ExcludePatterns`.
+func TarWithOptions(srcPath string, options *TarOptions) (io.ReadCloser, error) {
+	if _, err := fileutils.NewPatternMatcher(options.ExcludePatterns); err != nil {
```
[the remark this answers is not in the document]
mtrmac: Alternatively, if the pipe writer were modified to use `CloseWithError` (which, I, think, should definitely happen), we don’t need to specifically check here: we can just start the pipe writer goroutine, and receive the pattern matcher error through the pipe.

pkg/archive/archive.go
mtrmac: It is already _doing_ the compression: `compressWriter` is a layer of indirection that does the compression within its `Write` implementation, and `compressWriter.Write` is ultimately called within the producer goroutine.

(After all, `compressWriter` is writing to `pipeWriter`; the caller of `TarWithOptions` is, most of the time, blocked on reading from `pipeReader`, and it has no opportunity to write to `pipeWriter` in the meantime.)

pkg/archive/archive.go
mtrmac: I think the “uncontested parts” can be merged ~immediately. 

I didn’t look at the failures triggered by adding error handling ~at all yet. I understand that they matter to you (don’t  they?), so they block the overall effort _anyway_. It think that’s the next step.

And making the API public would follow, or be a part of the “add error handling” PR.

---

> I think it would be better to export something like `archive.tarWriter`.

I have ~no opinion on that in principle, I’d defer to actual c/storage maintainers.

At a high level, I’m a bit unclear on what is the point of a granular API for external consumers. `archive.Tar`, sure, a very simple one-shot call. `archive.TarWithOptions`, fine, for various tunables (although many would be pretty specific to needs of c/storage callers).

For the individual `tarWriter.addFile`… I’m tempted to question why a caller specialized _enough

mtrmac: I think the “uncontested parts” can be merged ~immediately. 

I didn’t look at the failures triggered by adding error handling ~at all yet. I understand that they matter to you (don’t  they?), so they block the overall effort _anyway_. It think that’s the next step.

And making the API public would follow, or be a part of the “add error handling” PR.

---

> I think it would be better to export something like `archive.tarWriter`.

I have ~no opinion on that in principle, I’d defer to actual c/storage maintainers.

At a high level, I’m a bit unclear on what is the point of a granular API for external consumers. `archive.Tar`, sure, a very simple one-shot call. `archive.TarWithOptions`, fine, for various tunables (although many would be pretty specific to needs of c/storage callers).

For the individual `tarWriter.addFile`… I’m tempted to question whether there are callers specialized _enough_ that using just `archive.TarWithOptions` is insufficient, but not so much that they would want to use the standard-library `archive/tar.Writer`, and whether the intersection of these conditions is enough to worry about.

---

> Then, some of the functionality of TarOptions could be expressed in code by callers

I don’t know what this means. Do you mean some other parameter-passing convention (e.g. “functional options”?) Using ordinary function parameters instead of a struct? I’d be rather opposed to the latter.

---

I am primarily interested in creating tar files (in the Podman Checkpoint command) without the intermediate buffering introduced by `io.Pipe`. I’m curious, does that have a measurable performance impact?

pkg/archive/archive.go
mtrmac: If you are referring to the use around `reexec.Command("storage-applyLayer", …)`, that is a communication mechanism entirely internal and specific to a single version of (indeed a single executable binary o) c/storage, and we can change it at any time.

hanwen-flow: :+1:

pkg/archive/archive.go
hanwen-flow: > I understand that they matter to you (don’t they?), 

The failures in the test suite (caused by ignoring errors) don't matter to me. What matters to me is that in case of _any_ error, the tar creation fails.

> they would want to use the standard-library archive/tar.Writer, and whether the intersection of these conditions is enough to worry about.

The `archive/tar` package falls short because it doesn't  connect the file system to the Writer/Reader, handling among others 

* extended attributes
* file types (mkdir vs mknod vs creat)
* hard-links

Note that the std library `archive/tar.Writer` is structured as a type with methods, offering greater flexibility to callers, rather than a function with a zillion options. 

> I’m curious, does that have a measurable performance impact?

See https://github.com/containers/podman/pull/25050/commits/40e49cce65464643aeedc83a8ba0e12c1f4065bd and https://github.com/containers/podman/issues/24826#issuecomment-2601996867

for writing large files to BTRFS, there is an effective speedup of up to 1000x.

pkg/archive/archive.go
hanwen-flow: I'm not sure what your original point was. Can you rephrase it?

pkg/archive/archive.go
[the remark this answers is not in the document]
hanwen-flow: it's somewhat incompatible, though: the error will only be propagated once reading on the stream starts. If the caller does something like

```
tarStream, err := Tar(...)
if err != nil { return err }
dest := os.Create("file.tar")
io.Copy(dest, tarStream)
``` 

then your proposed behavior would now leave around "file.tar" in case of a PatternMAtcher failure.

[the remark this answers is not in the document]
mtrmac: > the error will only be propagated once reading on the stream starts

That’s a fair point. There are several ways to structure that in a private function; for the eventual public API, I don’t know.

pkg/archive/archive.go
mtrmac: c/storage is a library, and used from other projects; Podman does seem to have a panic handler, at least at the top level.

But you are right that that does not prevent this function from crashing the whole process.

pkg/archive/archive.go
mtrmac: > You mean: the writing routine (which becomes TarWithOptionsTo in this PR) should also do compression?

Yes. `TarWithOptionsTo( options: { Compression: … } )` can be set, and it seems simpler to implement it than to specifically refuse the option (or, worse, to silently ignore it while `TarWithOptions` implements it).

OTOH that is similar to the `PatternMatcher` situation, where moving the initialization into the writing goroutine would delay error reporting.

mtrmac: … and would the compression interfere with your plans to use reflinks?

hanwen-flow: >  and would the compression interfere with your plans to use reflinks?

Compression, either in the file system, or on the tar stream obviously breaks reflink copying.

Unfortunately, the `CompressStream` function does more than compression; it also introduces buffering for the `Uncompressed` case by means of the `NewWriteCloserWrapper`. AFAICT, the wrapper will hide the `ReadFrom` method of the underlying writer.

I'm not sure why the buffering is there. The `tarWriter` has its own buffer.

pkg/archive/archive.go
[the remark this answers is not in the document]
hanwen-flow: as an aside, the pattern matcher could be O(1) in the number of patterns by rewriting

NewPatternMatcher("a", "b")

to generate the regexp '(a|b)'

pkg/archive/archive.go
```diff
@@ -1026,10 +1017,26 @@ func TarWithOptions(srcPath string, options *TarOptions) (io.ReadCloser, error)
...
+	pipeReader, pipeWriter := io.Pipe()
+	compressWriter, err := CompressStream(pipeWriter, options.Compression)
+	if err != nil {
+		return nil, err
+	}
+
+	go func() {
+		err := tarWithOptionsTo(compressWriter, srcPath, options)
+		if err != nil {
+			logrus.Errorf("Can't create tar archive: %s", err)
```
mtrmac, line 1035: Is this necessary? The consumer can receive the error from the pipe, and then choose for itself whether to log the error.

pkg/archive/archive.go
```diff
@@ -859,43 +859,34 @@ func Tar(path string, compression Compression) (io.ReadCloser, error) {
...
 		)
 		ta.WhiteoutConverter = GetWhiteoutConverter(options.WhiteoutFormat, options.WhiteoutData)
 		ta.CopyPass = options.CopyPass
 
+		includeFiles := options.IncludeFiles
 		defer func() {
 			// Make sure to check the error on Close.
 			if err := ta.TarWriter.Close(); err != nil {
 				logrus.Errorf("Can't close tar writer: %s", err)
 			}
```
mtrmac, line 886: Hum, `TarWriter` is now closed in two places. The current implementation of `tar.Writer` makes redundant `Close` calls always succeed, so that’s fair enough, but it’s going to be confusing to maintain.

WDYT about dropping this call? If the whole function fails, the archive is not usable anyway — and `tarWithOptionsTo` must have returned an error, so, via `pipeWriter.CloseWithError`, the consumer will be told.

pkg/archive/archive.go
mtrmac: https://github.com/moby/moby/pull/7542 claims a fairly significant performance improvement, although that’s not directly attributed to exactly these lines.

I don’t see `archive.Tar` doing all that much buffering.

mtrmac (commented):
So, I think we are now at
- some refactoring towards a better design — good
- some, but not all, error paths now returning errors

That’s a good bit of progress.

@giuseppe PTanotherL.

pkg/archive/archive.go
hanwen-flow: the moby PR has a bunch of anecdotal data, along with a TarUntar benchmark. The latter looks rather unrepresentative, creating 100 4 byte files in /tmp, and then repeatedly tarring them.

I made a quick test that tars a dir on a real file system. The buffering in CompressStream makes a ~10% impact when going from ext4 to ext4,

```
$ TARDEST=$HOME/tardest.tar TARDIR=$HOME/vc/linux go test -run TarLinux -bench TarLinux 
2025/03/11 13:41:38 tarring "/home/hanwen/vc/linux"
2025/03/11 13:41:45 7.221876687s
2025/03/11 13:41:55 9.783580435s
2025/03/11 13:42:03 8.324693106s
2025/03/11 13:42:13 9.535414505s
2025/03/11 13:42:21 8.34720174s
2025/03/11 13:42:31 9.622281549s
2025/03/11 13:42:39 8.726733474s
2025/03/11 13:42:48 9.115060653s
2025/03/11 13:42:57 8.690927914s
2025/03/11 13:43:06 9.37279054s
2025/03/11 13:43:06 With 8.34720174s, without 9.535414505s
PASS
ok  	github.com/containers/storage/pkg/archive	88.748s
``` 

suprprisingly, with my local hack for zero-transport copy, the buffering (which disables zero-transport) actually makes things go faster: 

```
$ TARDEST=$HOME/.cache/tardest.tar TARDIR=$HOME/.cache/linux go test -run TarLinux -bench TarLinux 
2025/03/11 13:47:06 tarring "/home/hanwen/.cache/linux"
2025/03/11 13:47:15 9.059872899s
2025/03/11 13:47:30 14.882930582s
2025/03/11 13:47:37 6.968169459s
2025/03/11 13:47:53 16.237515011s
2025/03/11 13:48:00 7.07192259s
2025/03/11 13:48:15 15.151443769s
2025/03/11 13:48:23 7.171133439s
2025/03/11 13:48:38 15.060790794s
2025/03/11 13:48:46 8.301557066s
2025/03/11 13:49:01 14.920596331s
2025/03/11 13:49:01 With 7.171133439s, without 15.060790794s
PASS
ok  	github.com/containers/storage/pkg/archive	114.837s
```

for tarring larger files, zero copy does bring benefits, 

``` 
 $ TARDEST=$HOME/.cache/tardest.tar TARDIR=$HOME/.cache/lib64 go test -run TarLinux -bench TarLinux 
2025/03/11 13:54:09 tarring "/home/hanwen/.cache/lib64"
2025/03/11 13:54:12 2.802090087s
2025/03/11 13:54:14 1.640192869s
2025/03/11 13:54:16 2.12680278s
2025/03/11 13:54:18 1.61872927s
2025/03/11 13:54:20 2.151480276s
2025/03/11 13:54:21 1.619020427s
2025/03/11 13:54:24 2.168449473s
2025/03/11 13:54:25 1.584594405s
2025/03/11 13:54:27 2.132707088s
2025/03/11 13:54:29 1.617399592s
2025/03/11 13:54:29 With 2.151480276s, without 1.61872927s
```

it seems I'll need a more sophisticated strategy here (using zero-copy only for large files), and right now, there is no good argument for dropping the buffering from CompressStream.

pkg/archive/archive.go
hanwen-flow: sgtm.

pkg/archive/archive.go
hanwen-flow: to me, it's not necessary, but I didn't want to change behavior.

mtrmac (commented):
LGTM.

giuseppe (approved):
/lgtm

hanwen-flow:
@mtrmac , @giuseppe - PTAL, had to rebase. (I used the GH UI to rebase)

giuseppe (approved):
/lgtm

PR #2264, opened 2025-02-25
Recursively set project ID for quota driver

mheon:
We have previously relied on the PROJINHERIT flag for XFS quotas, which causes the ID of the parent directory to be recursively applied to subdirectories under the volume's parent directory. However, PROJINHERIT only applies to directories created after the project ID was first set. Pre-existing directories do not get the project ID if we only set it on the parent. This means that quota enforcement is not complete unless we recurse to subdirectories and apply the project ID to those as well.

Fixes https://github.com/containers/podman/issues/25368

### DIFF
--- a/drivers/quota/projectquota_supported.go
+++ b/drivers/quota/projectquota_supported.go
@@ -190,7 +190,8 @@ func NewControl(basePath string) (*Control, error) {
 }
 
 // SetQuota - assign a unique project id to directory and set the quota limits
-// for that project id
+// for that project id.
+// targetPath must exist, must be a directory, and must be empty.
 func (q *Control) SetQuota(targetPath string, quota Quota) error {
 	var projectID uint32
 	value, ok := q.quotas.Load(targetPath)
@@ -200,10 +201,20 @@ func (q *Control) SetQuota(targetPath string, quota Quota) error {
 	if !ok {
 		projectID = q.nextProjectID
 
+		// The directory we are setting an ID on must be empty, as
+		// the ID will not be propagated to pre-existing subdirectories.
+		dents, err := os.ReadDir(targetPath)
+		if err != nil {
+			return fmt.Errorf("reading directory %s: %w", targetPath, err)
+		}
+		if len(dents) > 0 {
+			return fmt.Errorf("can only set project ID on empty directories, %s is not empty", targetPath)
+		}
+
 		//
 		// assign project id to new container directory
 		//
-		err := setProjectID(targetPath, projectID)
+		err = setProjectID(targetPath, projectID)
 		if err != nil {
 			return err
 		}


drivers/quota/projectquota_supported.go
```diff
@@ -370,6 +370,20 @@ func setProjectID(targetPath string, projectID uint32) error {
...
+	// While we are setting the PROJINHERIT flag, it only applies to files
+	// and directories created after the fact.
+	// Any pre-existing subdirectories will still need to have the project
+	// ID set manually, so recurse onto them.
+	if dents, err := os.ReadDir(targetPath); err == nil {
+		for _, dent := range dents {
+			if dent.IsDir() {
+				if err := setProjectID(filepath.Join(targetPath, dent.Name()), projectID); err != nil {
+					return fmt.Errorf("recursively setting project ID on subdirectories of %s: %w", targetPath, err)
+				}
```
Luap99, line 382: I am not sure if it matters for our use case but if we walk a dir and then try to open the path it is possible that it has been removed between the listing and the open call.
As such I think this API needs to ignore ErrNotExist (though as this uses cgo I have not checked if matching ErrNotExist will actually match the right error in that case, I think it does but better double check)

mheon:
Once this merges, I'm going to vendor c/s main into Podman and add testing for quotas to CI there.

drivers/quota/projectquota_supported.go
```diff
@@ -370,6 +371,20 @@ func setProjectID(targetPath string, projectID uint32) error {
 		return fmt.Errorf("failed to set projid for %s: %w", targetPath, errno)
 	}
 
+	// While we are setting the PROJINHERIT flag, it only applies to files
+	// and directories created after the fact.
```
mtrmac, line 375: Nothing here changes properties of non-directories. Doesn’t the project ID have to be set for every file as well?

drivers/quota/projectquota_supported.go
```diff
@@ -370,6 +371,20 @@ func setProjectID(targetPath string, projectID uint32) error {
...
+	// Any pre-existing subdirectories will still need to have the project
+	// ID set manually, so recurse onto them.
+	if dents, err := os.ReadDir(targetPath); err == nil {
```
mtrmac, line 378: If we should do this, I’d expect to take advantage of existing `WalkDir`.

mtrmac (commented):
I don’t really understand the issue or how this is supposed to help.

As a structural concern, what if the existing files in that directory are _already larger than the quota_?

---

To be more specific about the situation, this code is called:
- in `drivers/overlay.Driver.create`; in that case the target does not exist (or is explicitly removed before proceeding)
- in `Runtime.newVolume`, which AFAICS does not proceed if the volume already exists.

So, how do we get into a situation where there are pre-existing files/directories? At a high level, that seems to be something that shouldn’t happen (or I misunderstand what “create a volume” means, quite possible — I didn’t read the whole of the Podman issue).

---

As a _guess_, would reordering `Runtime.newVolume` to set the quota _first_ before creating `_data` help?

drivers/quota/projectquota_supported.go
mheon: No, it is a per-directory construct

mheon:
@mtrmac RE: your structural question - if the directory is already populated with files whose size exceeds the quota, all further writes fail with ENOSPC until sufficient files are deleted to bring you below quota (per a brief experiment I just conducted). Interestingly, `xfs_quota` does not include these files in its size calculation, so quota shows as entirely unused while simultaneously denying writes due to being full. Fascinating.

I would be amenable to reordering things in Podman to avoid the need for this, but if so we should update the docs here to be clear that quotas should only be set on empty directories. While it seems like things don't break if we have pre-existing files, they also don't seem to work as expected, and pre-existing subdirectories definitely break things.

drivers/quota/projectquota_supported.go
mtrmac: I agree the _flag_ is per-directory. The project ID itself?

mtrmac:
> Interestingly, `xfs_quota` does not include these files in its size calculation, so quota shows as entirely unused while simultaneously denying writes due to being full.

That seems hard to manage.

(For comparison, I think https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/quota/project.c#n174 means that they only affect that one directory, with no other checking.)


> I would be amenable to reordering things in Podman to avoid the need for this, but if so we should update the docs here to be clear that quotas should only be set on empty directories.

I’d even prefer if the `SetQuota` function checked, and failed with a non-empty target.

drivers/quota/projectquota_supported.go
Luap99: Why? WalkDir seems to do more work we don't need here, this version seems to be much simpler than calling WalkDir() here.

Luap99:
Personally I would think it would be good to see a podman PR with it vendored first with the test you promise. Then it is much easier for me to check before/after without having to figure out how to setup quotas on a test system

mheon:
Reworked to block setting quotas entirely if the directory is not empty. I'll reorder quota setting on the Podman side to account for this.

giuseppe:
LGTM

/retest

mheon:
Testing in https://github.com/containers/podman/pull/25417

mheon:
Podman-side tests are green, I think this is good to merge

giuseppe (approved):
/lgtm

PR #2273, opened 2025-03-06
loop: remove hardcoded constants

giuseppe:
follow-up for commit eebb044f5607269a434b07bdf4660118432374ac

remove the hardcoded constants and re-use the values defined in the unix package.

@mtrmac PTAL


mtrmac (commented):
/lgtm

Thank you!

PR #2277, opened 2025-03-10
subid: drop support for old version

giuseppe:
the new ABI has been around for more than 3 years now (shadow-utils v4.11).

Closes: https://github.com/containers/storage/issues/2260

### DIFF
--- a/hack/libsubid_tag.sh
+++ b/hack/libsubid_tag.sh
@@ -10,14 +10,12 @@ cc -o "$tmpdir"/libsubid_tag -l subid -x c - > /dev/null 2> /dev/null << EOF
 #include <stdio.h>
 #include <stdlib.h>
 
-const char *Prog = "test";
-FILE *shadow_logfd = NULL;
-
 int main() {
 	struct subid_range *ranges = NULL;
 #if SUBID_ABI_MAJOR >= 4
 	subid_get_uid_ranges("root", &ranges);
 #else
+	libsubid_init("root", stderr);
 	get_subuid_ranges("root", &ranges);
 #endif
 	free(ranges);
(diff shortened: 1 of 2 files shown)


Luap99 (changes requested):
This breaks build support for RHEL 9
```
[root@vm-10-0-187-181 ~]# ls /usr/lib64/libsubid.so.3
/usr/lib64/libsubid.so.3
[root@vm-10-0-187-181 ~]# ls /usr/lib64/libsubid.so.4
ls: cannot access '/usr/lib64/libsubid.so.4': No such file or directory
[root@vm-10-0-187-181 ~]# cat /etc/os-release 
NAME="Red Hat Enterprise Linux"
VERSION="9.6 (Plow)"
...
```

I don't think we can drop support for RHEL 9 at this point?

giuseppe:
> This breaks build support for RHEL 9
> 
> ```
> [root@vm-10-0-187-181 ~]# ls /usr/lib64/libsubid.so.3
> /usr/lib64/libsubid.so.3
> [root@vm-10-0-187-181 ~]# ls /usr/lib64/libsubid.so.4
> ls: cannot access '/usr/lib64/libsubid.so.4': No such file or directory
> [root@vm-10-0-187-181 ~]# cat /etc/os-release 
> NAME="Red Hat Enterprise Linux"
> VERSION="9.6 (Plow)"
> ...
> ```
> 
> I don't think we can drop support for RHEL 9 at this point?

shurg, I assumed we moved on from that old version.  Thanks for checking it.  I've fixed the patch

Luap99 (commented):
Now that I look at it I remembered https://issues.redhat.com/browse/RHEL-21366.
I think this would fix that as we correctly use the init function now?

Luap99 (approved):
LGTM I suppose but I have not tried to compile test it

giuseppe:
@kolyshkin PTAL

giuseppe:
@mtrmac PTAL

giuseppe:
@flouthoc PTAL

flouthoc (approved):
LGTM

PR #2283, opened 2025-03-14
Remove some do-nothing files

mtrmac:
This should not make (much?) difference at run-time, but let’s save everyone the effort looking at these in the future.


rhatdan:
LGTM
@nalind PTAL

giuseppe (approved):
/lgtm

Issue #2009, opened 2024-07-10
Delete the Windows storage driver?


mtrmac:
Ever since 1eba391bd3cde8005087b661a2f07e803ecba390 in Sep 2017 the driver’s test can’t compile because a `addAceToSddlDacl` function is missing.

Clearly no-one is even trying to make it functional.

Can we remove that code? If we ever did want to add a Windows support, the old code can always be fished out of git.

rhatdan:
SGTM

nalind:
That call is in the unit test which only run on Windows, so CI doesn't run them.  The driver itself compiles, for whatever that's worth.

Luap99:
To add another reason, just noticed ` go.opencensus.io` has an expired SSL cert and thus fails new vendor in CI (or well locally if this version was not downloaded before). To work around that we must use the golang proxy to vendor and cannot use the direct mode (fedora golang default), see https://github.com/containers/common/pull/2300. 

Looking into why we import this package which has been archived a while ago points back here in c/storage to the windows driver. Dropping it would solve that as well AFAICS.

```
 go mod why go.opencensus.io
# go.opencensus.io
github.com/containers/storage/drivers/windows
github.com/Microsoft/hcsshim
github.com/Microsoft/hcsshim/internal/hcs
go.opencensus.io/trace
go.opencensus.io/internal
go.opencensus.io
```

mtrmac:
https://github.com/microsoft/hcsshim/pull/2136 tracks removing the dependency.

rhatdan:
Why not just delete the support from C/S?

mtrmac:
See what removing the graph driver would do https://github.com/containers/storage/pull/2284 :

But, also, https://github.com/containers/buildah/issues/4010 is live again.

PR #2284, opened 2025-03-14
RFC: Remove drivers/windows

mtrmac:
Compare https://github.com/containers/storage/issues/2009 : The code could never build its unit tests since September 2017, and nobody seems to have complained.

Also, see how this impacts `go.mod`: many of the largest (`google.golang.org/grpc`), obsolete (`github.com/gogo/protobuf`) or outright frozen and unmaintained (`go.opencensus.io`), or just aesthetically unappealing (`github.com/containerd/…`) dependencies go away.

[And then we could discuss the unmaintained `github.com/json-iterator/go`, a codebase full of `reflect` and `unsafe`. That’s also _fun_.]

To be fair, `containers/ocicrypt` drags GRPC into Podman anyway, and several dependencies add various implementations of `protobuf`.

I’m not going to strongly advocate merging this, but I did want to show that, look, this is clearly not in a good shape, and the cost is not all that small, just as one more data point for the conversation (so, marking as draft, ~indefinitely).

Cc: @giuseppe @nalind @mheon 

But then again, would this impact plans for https://github.com/containers/buildah/issues/4010 ? Do we instead need to start seriously caring? Cc: @sebsoto


giuseppe:
> Cc: @giuseppe @nalind @mheon
> 
> But then again, would this impact plans for [containers/buildah#4010](https://github.com/containers/buildah/issues/4010) ? Do we instead need to start seriously caring? Cc: @sebsoto

I am in favor of simplifying code that is out of maintenance (as it appears to be in this case).  (+1. -89,831) is a significant improvement, even if it doesn't affect the binary size

mheon:
LGTM on my end

giuseppe (approved):
/lgtm

PR #2285, opened 2025-03-17
store: use canonical image mapping when mounting

giuseppe:
Ensure that a mapped layer with the canonical mapping is created if one doesn't already exist.  Previously, any layer was used regardless of its mapping, which resulted in the mounted image having incorrect file ownership, depending on what mapping was found.

### DIFF
--- a/cmd/containers-storage/mount.go
+++ b/cmd/containers-storage/mount.go
@@ -22,7 +22,7 @@ func mount(flags *mflag.FlagSet, action string, m storage.Store, args []string) 
 	moes := []mountPointOrError{}
 	for _, arg := range args {
 		if paramReadOnly {
-			result, err := m.MountImage(arg, []string{"noexec", "nosuid"}, paramMountLabel)
+			result, err := m.MountImage(arg, []string{}, paramMountLabel)
 			errText := ""
 			if err != nil {
 				errText = err.Error()
--- a/store.go
+++ b/store.go
@@ -2830,7 +2830,11 @@ func (s *store) Version() ([][2]string, error) {
 	return [][2]string{}, nil
 }
 
-func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
+func (s *store) MountImage(id string, mountOpts []string, mountLabel string) (string, error) {
+	if err := validateMountOptions(mountOpts); err != nil {
+		return "", err
+	}
+
 	// We need to make sure the home mount is present when the Mount is done, which happens by possibly reinitializing the graph driver
 	// in startUsingGraphDriver().
 	if err := s.startUsingGraphDriver(); err != nil {
@@ -2842,57 +2846,61 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if options.UidMaps != nil || options.GidMaps != nil {
-		options.DisableShifting = !s.canUseShifting(options.UidMaps, options.GidMaps)
-	}
+	var imageHomeStore roImageStore
 
-	// function used to have a scope for rlstore.StopWriting()
-	tryMount := func() (string, error) {
-		if err := rlstore.startWriting(); err != nil {
+	if err := rlstore.startWriting(); err != nil {
+		return "", err
+	}
+	defer rlstore.stopWriting()
+	for _, s := range lstores {
+		if err := s.startReading(); err != nil {
 			return "", err
 		}
-		defer rlstore.stopWriting()
-		if rlstore.Exists(id) {
-			return rlstore.Mount(id, options)
-		}
-		return "", nil
+		defer s.stopReading()
 	}
-	mountPoint, err := tryMount()
-	if mountPoint != "" || err != nil {
-		return mountPoint, err
+	if err := s.imageStore.startWriting(); err != nil {
+		return "", err
+	}
+	defer s.imageStore.stopWriting()
+
+	cimage, err := s.imageStore.Get(id)
+	if err == nil {
+		imageHomeStore = s.imageStore
+	} else {
+		for _, s := range s.roImageStores {
+			if err := s.startReading(); err != nil {
+				return "", err
+			}
+			defer s.stopReading()
+			cimage, err = s.Get(id)
+			if err == nil {
+				imageHomeStore = s
+				break
+			}
+		}
+	}
+	if cimage == nil {
+		return "", fmt.Errorf("locating image with ID %q: %w", id, ErrImageUnknown)
 	}
 
-	// check if the layer is in a read-only store, and return a better error message
-	for _, store := range lstores {
-		if err := store.startReading(); err != nil {
-			return "", err
-		}
-		exists := store.Exists(id)
-		store.stopReading()
-		if exists {
-			return "", fmt.Errorf("mounting read/only store images is not allowed: %w", ErrStoreIsReadOnly)
-		}
+	idmappingsOpts := types.IDMappingOptions{
+		HostUIDMapping: true,
+		HostGIDMapping: true,
 	}
-
-	return "", ErrLayerUnknown
-}
-
-func (s *store) MountImage(id string, mountOpts []string, mountLabel string) (string, error) {
-	// Append ReadOnly option to mountOptions
-	img, err := s.Image(id)
+	ilayer, err := s.imageTopLayerForMapping(cimage, imageHomeStore, rlstore, lstores, idmappingsOpts)
 	if err != nil {
 		return "", err
 	}
 
-	if err := validateMountOptions(mountOpts); err != nil {
-		return "", err
+	if len(ilayer.UIDMap) > 0 || len(ilayer.GIDMap) > 0 {
+		return "", fmt.Errorf("cannot create an image with canonical UID/GID mappings in a read-only store")
 	}
+
 	options := drivers.MountOpts{
 		MountLabel: mountLabel,
 		Options:    append(mountOpts, "ro"),
 	}
-
-	return s.mount(img.TopLayer, options)
+	return rlstore.Mount(ilayer.ID, options)
 }
 
 func (s *store) Mount(id, mountLabel string) (string, error) {
@@ -2914,7 +2922,43 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
 			}
 		}
 	}
-	return s.mount(id, options)
+
+	// We need to make sure the home mount is present when the Mount is done, which happens by possibly reinitializing the graph driver
+	// in startUsingGraphDriver().
+	if err := s.startUsingGraphDriver(); err != nil {
+		return "", err
+	}
+	defer s.stopUsingGraphDriver()
+
+	rlstore, lstores, err := s.bothLayerStoreKindsLocked()
+	if err != nil {
+		return "", err
+	}
+	if options.UidMaps != nil || options.GidMaps != nil {
+		options.DisableShifting = !s.canUseShifting(options.UidMaps, options.GidMaps)
+	}
+
+	if err := rlstore.startWriting(); err != nil {
+		return "", err
+	}
+	defer rlstore.stopWriting()
+	if rlstore.Exists(id) {
+		return rlstore.Mount(id, options)
+	}
+
+	// check if the layer is in a read-only store, and return a better error message
+	for _, store := range lstores {
+		if err := store.startReading(); err != nil {
+			return "", err
+		}
+		exists := store.Exists(id)
+		store.stopReading()
+		if exists {
+			return "", fmt.Errorf("mounting read/only store images is not allowed: %w", ErrStoreIsReadOnly)
+		}
+	}
+
+	return "", ErrLayerUnknown
 }
 
 func (s *store) Mounted(id string) (int, error) {
@@ -2938,7 +2982,23 @@ func (s *store) UnmountImage(id string, force bool) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	return s.Unmount(img.TopLayer, force)
+
+	return writeToLayerStore(s, func(lstore rwLayerStore) (bool, error) {
+		for _, layerID := range img.MappedTopLayers {
+			l, err := lstore.Get(layerID)
+			if err != nil {
+				if err == ErrLayerUnknown {
+					continue
+				}
+				return false, err
+			}
+			// check if the layer with the canonical mapping is in the mapped top layers
+			if len(l.UIDMap) == 0 && len(l.GIDMap) == 0 {
+				return lstore.unmount(l.ID, force, false)
+			}
+		}
+		return lstore.unmount(img.TopLayer, force, false)
+	})
 }
 
 func (s *store) Unmount(id string, force bool) (bool, error) {
--- a/tests/idmaps.bats
+++ b/tests/idmaps.bats
@@ -290,7 +290,17 @@ load helpers
 	done
 	# Create new containers based on the layer.
 	imagename=idmappedimage
-	storage create-image --name=$imagename $lowerlayer
+	run storage create-image --name=$imagename $lowerlayer
+	[ "$status" -eq 0 ]
+
+	run storage --debug=false mount -r $imagename
+	[ "$status" -eq 0 ]
+	mountpoint="$output"
+	run stat -c %u:%g $mountpoint/file1
+	# make sure the file shows the correct ownership after the image is mounted
+	[ "$output" == "0:0" ]
+	run storage umount $imagename
+	[ "$status" -eq 0 ]
 
 	run storage --debug=false create-container --hostuidmap --hostgidmap $imagename
 	[ "$status" -eq 0 ]


giuseppe:
so this seems to fix the problem: https://github.com/containers/podman/pull/25606

@Luap99

mtrmac (commented):
I understand ~nothing about the problem space.

> With idmapped mounts, there is no extra cost to change the IDs at runtime when using overlay.

This code has no condition for “using overlay”, so I don’t know how that relates. Doesn’t it break the other graph drivers?

> This allows the image to be mounted directly with VFS, avoiding any mismatch that could occur when it was previously created with an explicit ID mapping.

AFAICS for non-writeable == image layers, `uidMap`, `gidMap` come from the options passed by c/image; and c/image passes _nothing_, so the existing condition is ~equivalent to `Driver.SupportsShifting`. That sounds, intuitively, about right? (See above about “using overlay”).

Also, assuming that `uidMap`/`gidMap` are _always_ `nil` during pulls, the two `if` cases only differ in `Host?IDMapping` fields … and those fields are not used _anywhere_ in the callees, AFAICS. So how does this make a difference? [Reusing the same types for different purposes is a mess…]

So… I’m confused. I appreciate that there are test PRs that suggest this does _something_, but just looking at this PR, I can’t see what, or how it possibly could.

(If other reviewers _understand_ what is going on, that’s fine, this does not need to block on me.)

giuseppe:
> I understand ~nothing about the problem space.
> 
> > With idmapped mounts, there is no extra cost to change the IDs at runtime when using overlay.
> 
> This code has no condition for “using overlay”, so I don’t know how that relates. Doesn’t it break the other graph drivers?
> 
> > This allows the image to be mounted directly with VFS, avoiding any mismatch that could occur when it was previously created with an explicit ID mapping.
> 
> AFAICS for non-writeable == image layers, `uidMap`, `gidMap` come from the options passed by c/image; and c/image passes _nothing_, so the existing condition is ~equivalent to `Driver.SupportsShifting`. That sounds, intuitively, about right? (See above about “using overlay”).
> 
> Also, assuming that `uidMap`/`gidMap` are _always_ `nil` during pulls, the two `if` cases only differ in `Host?IDMapping` fields … and those fields are not used _anywhere_ in the callees, AFAICS. So how does this make a difference? [Reusing the same types for different purposes is a mess…]
> 
> So… I’m confused. I appreciate that there are test PRs that suggest this does _something_, but just looking at this PR, I can’t see what, or how it possibly could.
> 
> (If other reviewers _understand_ what is going on, that’s fine, this does not need to block on me.)

the reason why we have this code is that if you run something like:

`podman run --uidmap A:B:C ... IMAGE`

and the IMAGE is not present in the local storage, it is pulled directly with the specified mapping.  In this way we don't need to create another copy of the image before running the container.

This is not an issue anymore with overlay, since with idmapped mounts we already store the "canonical" image, since `s.canUseShifting(uidMap, gidMap)` returns true for overlay.

With VFS it does not matter, because for the container we always need to create a full copy anyway.

The issue we had in Podman is that if you run something like:

```
podman --storage-driver vfs run --uidmap A:B:C ... IMAGE
podman --storage-driver vfs image mount IMAGE
```

we would end up mounting the image with the mapping A:B:C, which is of course wrong.  An alternative solution would be to create a copy of the image at `MountImage()` time if there is not already a copy with the "canonical" mapping.

I've picked this alternative because it is easier to implement as we already have the logic to handle it, and because I don't see the value in doing this anymore, since the driver we care the most about (overlay) can do the mapping at runtime without any additional setup cost.

mtrmac:
Oh, I see what I missed:
```go
if !options.HostUIDMapping && len(options.UIDMap) == 0 {
			uidMap = s.uidMap
		}
```
in `putLayer` comes from the Podman CLI, without c/image involvement.

Luap99:
> This is not an issue anymore with overlay, since with idmapped mounts we already store the "canonical" image, since s.canUseShifting(uidMap, gidMap) returns true for overlay.

Is that true for rootless? I thought rootless can never use idmapped mounts (unless we use fuse-overlayfs). As such this optimization would be lost for rootless pulls then?
Not sure if that matters because many ways how the images are pulled may not include the idmapped options via podman anyway (e.g. podman pull or the system service)

giuseppe:
> Is that true for rootless? I thought rootless can never use idmapped mounts (unless we use fuse-overlayfs). As such this optimization would be lost for rootless pulls then?
> Not sure if that matters because many ways how the images are pulled may not include the idmapped options via podman anyway (e.g. podman pull or the system service)

that is a good point!  No that is not true for rootless and it might impact toolbx

mtrmac:
> The issue we had in Podman is that if you run something like:
> ```
> podman --storage-driver vfs run --uidmap A:B:C ... IMAGE
> podman --storage-driver vfs image mount IMAGE
> ```
> we would end up mounting the image with the mapping A:B:C, which is of course wrong. 

So, approximately, `MountImage` is missing something like `imageTopLayerForMapping`?

---

The current approach ~relies on `podman image mount` not having an `--uidmap` option, so that we store layers in the format which `podman image mount` prefers.

But what about `Store.Mount(containerID)`? If different containers can have different ID maps, that effectively allows doing `… mount --uidmap $arbitrary`; so, from just locally reading the code, shouldn’t we be supporting the general case, with arbitrary mappings, anyway?

And I see a Podman caller of `Mount(containerID)`.

---

> I've picked this alternative because it is easier to implement as we already have the logic to handle it, and because I don't see the value in doing this anymore, since the driver we care the most about (overlay) can do the mapping at runtime without any additional setup cost.

The issue I have with this (_if_ I understand things, which is _unlikely_) is that it is extremely indirect and invisible to readers of the code.

- This assumes we only have graph drivers like overlay (with trivially cheap ID mapping) or VFS (which do a full copy anyway). What about all the other graph drivers, then? That’s Btrfs, ZFS, AUFS, Windows. I mean, if we absolutely don’t care whether they work, we should just delete them right now.
- Even for overlay, `Driver.SupportsShifting` has ~4 conditions that need to be true. What about all of the other cases? Should we delete all of that code, and refuse to start?
- We have `PutLayer` nominally accepting ID mapping options… and it’s going to just ignore them because it assumes that they don’t matter. “How does c/storage.Store know?”; at this API layer, it seems to be possible for a caller to do PutLayer(nonDefaultIDMapping); Mount(), and the caller would be legitimate in expecting the ID mapping to be processed.

---

I’m all for simplifying things and not processing irrelevant options on unreachable code paths. But then the code paths should be _recognizably_ unreachable. E.g. deprecate the public `LayerOptions.*IDMap*`, replace internal data structures with smaller better-focused structs that only carry the relevant fields, etc.

I suppose that’s a lot of work, quite possibly more work than making the top-level mapping work as expected for this use case.

But then again, I understand very little about all this code.

mtrmac:
> What about all the other graph drivers, then? 

… and if either of the graph drivers decided to change these features (maybe VFS grew some kind of “fast copy” that required the UID maps to match), how would anyone notice that this `putLayer` needs updating? That’s what I mean by “invisible”.

giuseppe:
> * This assumes we only have graph drivers like overlay (with trivially cheap ID mapping) or VFS (which do a full copy anyway). What about all the other graph drivers, then? That’s Btrfs, ZFS, AUFS, Windows. I mean, if we absolutely don’t care whether they work, we should just delete them right now.

I don't know much about these drivers.  If changes in c/storage would depend on these drivers we would just stop doing any.

Your analysis is correct, with containers we use `imageTopLayerForMapping` but then there is a fallback when the image is in an additional image store, and we can't do a copy.  So the remapping is done in the container layer (https://github.com/containers/storage/blob/main/store.go#L1742-L1747).  With images we can't do that, since we don't create a new layer, so I am not sure how we can deal with images in an additional image store

giuseppe:
> I think the VFS situation can be modeled ~precisely by:
> 
> * Adding a `Driver` property saying what is special about VFS, `ContainerCreationAlwaysExpensivelyCopiesLayer` — and with _that_ property, never shifting image layers
> * Checking the ID mapping options on all other APIs (to do: enumerate), and _failing_ if they specify any other ID mapping when using image layers.
> 
> And we would let the other graph drivers do what they have always done, blindly hoping that it is correct.
> 
> _But_ I don’t see how this is desirable for the overlay cases where shifting is not available.

I don't think we want to keep the current behavior in any case, an optimization makes sense only when the result is correct.  If I run `podman image mount fedora`, I expect the files to be owned by root, not some random ID that was used for a user namespace.  That is completely unexpected.

mtrmac:
I agree “all image layers are always using the default mapping” is a reasonable and consistent design, in general. Now that we have the other one, switching could potentially have performance implications, by requiring the ID-mapped container layer to be now expensively created. I don’t know how much we care and I don’t know how to tell.

(Does not doing any mapping for image layers work for rootless? I guess it does because everything happens inside a namespace, but just to confirm.)

giuseppe:
pushed a new version, now a canonical mapping is created by MountImage if one doesn't already exist.  It fails if the image exists only in an additional image store without the canonical mapping, but I think that is a weird configuration anyway.

Luap99:
> pushed a new version, now a canonical mapping is created by MountImage if one doesn't already exist. It fails if the image exists only in an additional image store without the canonical mapping, but I think that is a weird configuration anyway.

Sounds reasonable to me, can you please test this version again in podman to make sure there is no new regression.

giuseppe:
pushed now: https://github.com/containers/podman/pull/25606

giuseppe:
seems to break `podman run --mount image`

cmd/containers-storage/mount.go
```diff
@@ -22,7 +22,7 @@ func mount(flags *mflag.FlagSet, action string, m storage.Store, args []string)
 	moes := []mountPointOrError{}
 	for _, arg := range args {
 		if paramReadOnly {
-			result, err := m.MountImage(arg, []string{"noexec", "nosuid"}, paramMountLabel)
+			result, err := m.MountImage(arg, []string{}, paramMountLabel)
```
mtrmac, line 25: … I don’t know. This behavior seems to be undocumented, looking at the commit history, I can’t find any discussion or rationale.

But then again, suddenly allowing things is risky. Do we know if anyone uses this command in production?

Adding an extra option to disable/override these defaults would be safer.

store.go
```diff
@@ -2914,7 +2930,44 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
...
+		options.DisableShifting = !s.canUseShifting(options.UidMaps, options.GidMaps)
+	}
+
+	if err := rlstore.startWriting(); err != nil {
+		return "", err
+	}
+	defer rlstore.stopWriting()
+	if rlstore.Exists(id) {
+		mountPoint, err := rlstore.Mount(id, options)
+		if mountPoint != "" || err != nil {
```
mtrmac, line 2953: AFAICS there is no code path where this can be false, so this can be simplified.

Alternatively, preserve the `tryMount` closure, to decrease the scope of the lock; then there really is a “N/A” code path. But the lock scope only matters on an error path, and simpler code is also valuable — I don’t care either way.

store.go
```diff
@@ -2842,57 +2846,69 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
...
-	tryMount := func() (string, error) {
-		if err := rlstore.startWriting(); err != nil {
-			return "", err
-		}
-		defer rlstore.stopWriting()
-		if rlstore.Exists(id) {
-			return rlstore.Mount(id, options)
-		}
-		return "", nil
+	if err := validateMountOptions(mountOpts); err != nil {
```
mtrmac, line 2850: This can happen before doing any operations that take locks.

store.go
```diff
@@ -2842,57 +2846,69 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
...
+	id = ilayer.ID
 
-	if err := validateMountOptions(mountOpts); err != nil {
-		return "", err
-	}
-	options := drivers.MountOpts{
-		MountLabel: mountLabel,
-		Options:    append(mountOpts, "ro"),
+	if len(ilayer.UIDMap) > 0 || len(ilayer.GIDMap) > 0 {
+		return "", fmt.Errorf("mounting images with user mappings is not allowed")
```
mtrmac, line 2908: This reads as a caller error, but it’s really due to `!createMappedLayer`, isn’t it? If so, I think the error message should explain the combination of factors that makes this fail.

---

… but, also, I think `!createMappedLayer` can only happen for layers in additional image stores, and those can’t be mounted _in any case_ (because `layerStore.Mount` requires a write lock).

If that’s correct, I think there should be some opportunity to simplify — either just this error message, or possibly the whole thing — but I didn’t now think this through. (OTOH, I don’t think e.g. creating a duplicate only a bit smaller version of `imageTopLayerForMapping` would be worth it.)

store.go
```diff
@@ -2914,7 +2930,44 @@ func (s *store) Mount(id, mountLabel string) (string, error) {
 			}
 		}
 	}
-	return s.mount(id, options)
+
+	if err := s.startUsingGraphDriver(); err != nil {
```
mtrmac, line 2934: Please preserve/duplicate the comment about the home mount.

store.go
```diff
@@ -2842,57 +2846,69 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
...
+	var cimage *Image               // Set if image != ""
 
-	// check if the layer is in a read-only store, and return a better error message
-	for _, store := range lstores {
+	if err := rlstore.startWriting(); err != nil {
+		return "", err
+	}
+	defer rlstore.stopWriting()
+	for _, s := range lstores {
+		store := s
```
mtrmac, line 2865: (Loop iteration variables are now fresh in each copy, so this should be no longer required anywhere.)

mtrmac (commented):
This makes sense to me overall. I didn’t investigate the test failure, and I didn’t review _too carefully_ the new `MountImage`.

store.go
mtrmac: (Also `for _, s := range s.roImageStores `)

giuseppe:
> seems to break `podman run --mount image`

tests fixed: https://github.com/containers/podman/pull/25606

giuseppe:
another flake happening in the "sys podman fedora-41 root host boltdb" CI worth investigating

Luap99:
@giuseppe One thing I just realized I have used the image mount API as part of my shell completion code for `podman run/create $IMAGE <PATH>` where the paths after the image are auto completed based on the real image content.
Now I assume when we only have a idmapped image as argument given without canonical image in the store we might get an expensive (slow) copy, right? That is clearly undesirable for shell completion which should generally be fast and not cause side effects (i.e. double the store space in worse case).

I do think this generally make sense for the Mount API for storage to force the canonical image but maybe it would be good to have a new option that a caller can set to just error out if the right image version does not exist for the mount. With that the shell completion logic can be updated to set this and then we don't have to worry about causing problems for the shell completion.
Or do you see another way? I personally like the shell completion can suggest the actual image paths, i.e
```
$ podman create quay.io/libpod/testimage:20241011 cat /etc/a[TAB]
/etc/alpine-release  /etc/apk/
```
This only shows the real paths in the image not from the /etc on your host.

giuseppe:
> @giuseppe One thing I just realized I have used the image mount API as part of my shell completion code for `podman run/create $IMAGE <PATH>` where the paths after the image are auto completed based on the real image content. Now I assume when we only have a idmapped image as argument given without canonical image in the store we might get an expensive (slow) copy, right? That is clearly undesirable for shell completion which should generally be fast and not cause side effects (i.e. double the store space in worse case).

How does it work with rootless?  Does it use `podman unshare` each time?

store.go
```diff
@@ -2938,6 +2984,19 @@ func (s *store) UnmountImage(id string, force bool) (bool, error) {
...
+			if err == ErrLayerUnknown {
+				continue
+			}
+			return false, err
+		}
+		if len(l.UIDMap) == 0 && len(l.GIDMap) == 0 {
+			return s.Unmount(l.ID, force)
+		}
+	}
 	return s.Unmount(img.TopLayer, force)
```
mtrmac, line 3000: Should this outright fail if we don’t find the layer that we think we have mounted?

Or do we need to preserve this so that (mount; upgrade Podman; unmount) works? … maybe that can’t work anyway, `podman --uidmap pull && podman run` creates a layer with the canonical mapping, then `podman mount` mounts the original one, we upgrade, and after an upgrade, `podman unmount` will find the canonical mapping.

giuseppe: `podman run --uidmap && podman image mount` will cause the TopLayer that has the specified mappings to be mounted.  This wouldn't work if we upgrade podman after the image is mounted.

giuseppe: I'd not usually worry about such a corner case, but it is so trivial to fix that we could perhaps consider doing it?

mtrmac: Ah, right, `old/podman` will mount exactly `TopLayer`.

If keeping this last line handles that corner case, I think it’s valuable — just please add a comment explaining why this is inconsistent with the mount part.

giuseppe: If there are no `MappedTopLayers` because the image is pulled in the canonical form, don't we need the last line anyway?

giuseppe:
tests are finally passing for Podman.  We also need: https://github.com/containers/common/pull/2374

Luap99:
> > @giuseppe One thing I just realized I have used the image mount API as part of my shell completion code for `podman run/create $IMAGE <PATH>` where the paths after the image are auto completed based on the real image content. Now I assume when we only have a idmapped image as argument given without canonical image in the store we might get an expensive (slow) copy, right? That is clearly undesirable for shell completion which should generally be fast and not cause side effects (i.e. double the store space in worse case).
> 
> How does it work with rootless? Does it use `podman unshare` each time?

No, the shell completion is part of the podman binary so we don't have to unshare. The cobra cli abstract most completion code into the go code and then just has small wrapper scripts for each shell it supports to call the binary to get the completions, search for `AutocompleteCreateRun`. In short I mount mount the image then just list the content there and then return this to the shell which then does the actual suggestions.
There are some tests for that even in test/system/600-completion.bats, i.e. `podman shell completion for paths in container/image`

giuseppe:
thanks, we could add a functionality that just returns `SupportsShifting() || hasCanonicalMappingReady(...)` to use for the CLI

Luap99:
> thanks, we could add a functionality that just returns `SupportsShifting() || hasCanonicalMappingReady(...)` to use for the CLI.
> 
> EDIT: we can do that directly in Podman without adding another API.
> 
> ```
> if (store.GraphDriver().SupportsShifting())
> return true;
> 
> iterate image layers `image.MappedTopLayers + image.TopLayer` and find one without mappings
> ```

Sounds good to me, I am happy to do that in the shell completion code once this is merged.

store.go
mtrmac: Doh! Yes, thanks.

cmd/containers-storage/mount.go
mtrmac: This is still outstanding.

store.go
```diff
@@ -2842,57 +2846,63 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
...
+	ilayer, err := s.imageTopLayerForMapping(cimage, imageHomeStore, rlstore, lstores, idmappingsOpts)
 	if err != nil {
 		return "", err
 	}
+	id = ilayer.ID
 
-	if err := validateMountOptions(mountOpts); err != nil {
-		return "", err
+	if len(ilayer.UIDMap) > 0 || len(ilayer.GIDMap) > 0 {
+		return "", fmt.Errorf("cannot create an image with canonical UID/GID mappings in a read-only store: %w", ErrImageUnknown)
```
mtrmac, line 2898: Should this be `ErrImageUnknown`?

I’d start with making this an untyped error, we can add a specific type for this situation if we find a user that requires that.

mtrmac:
> EDIT: we can do that directly in Podman without adding another API.

I’d much rather add a specialized `MountImageIsAlwaysCheap` API, or something like that.

One reason is the general design aspect: it’s better to make _precisely_ the promise the caller cares about, and not expose implementation details that affect that promise but don’t exactly match, where we wouldn’t notice when changing c/storage that the Podman CLI also needs updating.

An equally important reason is that we have all these conversations about redesigning the layer store backing (bootc), and redesigning layer identities (sha512); both almost certainly require changes to the driver API, so I’m thinking we should make it internal-only and deprecate the public `store.GraphDriver()` entirely. Last time I’ve looked, there’s ~1 user in each of Podman, c/common, CRI-O. So, I’d prefer not adding any more users.

cmd/containers-storage/mount.go
giuseppe: We do not ship it in any package, I've always considered it only a debugging/testing tool. You can't stop people from using it.

We can argue this is relaxing the old behavior, but the old one seems wrong.  It was setting some flags that were not supposed to be there in any case.

store.go
giuseppe: dropped the type

cmd/containers-storage/mount.go
mtrmac: I can live with that. @nalind WDYT?

nalind: I'd be fine with that - "noexec" is certainly a surprise.  I'm more surprised that whether we call `MountImage()` or `Mount()` depends on whether or not the read-only flag was passed in at the command line.  Anyway, a quick grep didn't turn up any use of this option in the tests, so there doesn't appear to be anything that depends on this.

giuseppe:
@mtrmac fine to merge?

giuseppe:
@nalind ok to merge (together with https://github.com/containers/common/pull/2374)?

store.go
```diff
@@ -2842,57 +2846,63 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if options.UidMaps != nil || options.GidMaps != nil {
-		options.DisableShifting = !s.canUseShifting(options.UidMaps, options.GidMaps)
-	}
+	var imageHomeStore roImageStore // Set if image != ""
+	var cimage *Image               // Set if image != ""
```
mtrmac, line 2850: The `Set if image != ""` comments are not applicable here.

(Non-blocking nit: The variables could be moved immediately before their first use (or, in case of `cimage`, use :=`).)

store.go
```diff
@@ -2842,57 +2846,63 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
...
-		store.stopReading()
-		if exists {
-			return "", fmt.Errorf("mounting read/only store images is not allowed: %w", ErrStoreIsReadOnly)
+	defer s.imageStore.stopWriting()
+	cimage, err = s.imageStore.Get(id)
+	if err == nil {
+		imageHomeStore = s.imageStore
+	} else {
+		for _, s := range s.roImageStores {
+			store := s
```
mtrmac, line 2871: (This is now unnecessary.)

store.go
```diff
@@ -2842,57 +2846,63 @@ func (s *store) mount(id string, options drivers.MountOpts) (string, error) {
...
-	img, err := s.Image(id)
+	idmappingsOpts := types.IDMappingOptions{
+		HostUIDMapping: true,
+		HostGIDMapping: true,
+	}
+	ilayer, err := s.imageTopLayerForMapping(cimage, imageHomeStore, rlstore, lstores, idmappingsOpts)
 	if err != nil {
 		return "", err
 	}
+	id = ilayer.ID
```
mtrmac, line 2895: I’m unhappy about using the same `id` variable for both image and layer IDs. Either use two variables with explicit names, or this can be inlined directly into the only user.

(And non-blocking nit: If a variable remains, deferring this after the “did we get the mapping we want” check would be easier to follow.)

store.go
```diff
@@ -2938,6 +2984,19 @@ func (s *store) UnmountImage(id string, force bool) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	// check if the layer with the canonical mapping is in the mapped top layers
+	for _, layerID := range img.MappedTopLayers {
+		l, err := s.Layer(layerID)
```
mtrmac, line 2989: This works, but it takes O(`len(MappedTopLayers)*len(roImageStores)`) locks, _and_ it can later fail due to a race with a “layer unknown”.

Consider (adding a comment to note that `MountImage` only creates mounts in the primary store and) manually locking the primary store, and looking up layers only in that store. That would ~require inlining `s.Unmount`.

store.go
```diff
@@ -2934,11 +2978,35 @@ func (s *store) Mounted(id string) (int, error) {
...
+		return false, err
+	}
+	defer s.stopUsingGraphDriver()
+
+	if err := s.imageStore.startReading(); err != nil {
+		return false, err
+	}
+	defer s.imageStore.stopReading()
+
 	img, err := s.Image(id)
```
mtrmac, line 2991: `s.Image` does its own locking, this is unexpected.

Blocker: And the locking order now violates the locking hierarchy.

I’m fine with calling ordinary `s.Image()` here and only doing manual locking for layers afterwards; that’s much simpler, and no worse than what we did previously.

store.go
```diff
@@ -2934,11 +2978,35 @@ func (s *store) Mounted(id string) (int, error) {
...
 		return false, err
 	}
-	return s.Unmount(img.TopLayer, force)
+
+	return writeToLayerStore(s, func(rlstore rwLayerStore) (bool, error) {
+		// check if the layer with the canonical mapping is in the mapped top layers
+		for _, layerID := range img.MappedTopLayers {
+			l, err := rlstore.Get(layerID)
+			if err != nil {
+				return false, err
```
mtrmac, line 3001: The previous version silently ignored `ErrLayerUnknown`. Is the change intentional?

I don’t have much of an opinion, failing immediately could be argued to be better, I just want to avoid accidents.

store.go
```diff
@@ -2934,11 +2978,35 @@ func (s *store) Mounted(id string) (int, error) {
...
+			}
+			// check if the layer with the canonical mapping is in the mapped top layers
```
mtrmac, line 3003: (This comment exists twice in the function, that’s probably unnecessary.)

mtrmac (changes requested):

store.go
mtrmac: The primary thing I’d like to see fixed here is removing the incorrect/inapplicable comments.

store.go
giuseppe: before we were running without keeping the lock, so that error was expected.  Now, I am not sure we should ignore it.

store.go
```diff
@@ -2934,11 +2978,39 @@ func (s *store) Mounted(id string) (int, error) {
...
 func (s *store) UnmountImage(id string, force bool) (bool, error) {
-	img, err := s.Image(id)
+	lstore, err := s.getLayerStore()
+	if err != nil {
+		return false, err
+	}
+	if err := lstore.startWriting(); err != nil {
+		return false, err
+	}
+	defer lstore.stopWriting()
```
mtrmac, line 2988: `writeToLayerStore`, please.

store.go
```diff
@@ -2934,11 +2978,39 @@ func (s *store) Mounted(id string) (int, error) {
...
+
+	// images can be mounted only in the rw-store, so don't
+	// check in the additional stores.
```
mtrmac, line 2991: I thought image mounts are restricted to the primary _layer_ store.

AFAICS `MountImage` would now be perfectly content with finding an image in a RO _image_ store. (How would that happen? Start with an ~empty RO image store, pull an image to primary store, untag it without deleting layers, then update the RO store to include the same image. Should unmounting such an RO-store-tagged-only image be expected to work? Meh, I don’t know, but when mounting does work…)

---

(IF _both_ layer and image stores were restricted to the RW store, that would be best commented at the start of the function.)

giuseppe:
podman tests pass with the last version

store.go
mtrmac: Thanks, good point.

We are back to loading the image, unlocking, and then looking at layers, so we might again hit a race here. Should this go back to the previous behavior?

mtrmac (commented):
LGTM.

Please consider https://github.com/containers/storage/pull/2285#discussion_r2012441521 , and https://github.com/containers/storage/pull/2285#discussion_r2010791595 .

And after looking at that, feel free to merge without another review.

giuseppe:
> Please consider [#2285 (comment)](https://github.com/containers/storage/pull/2285#discussion_r2012441521) , and [#2285 (comment)](https://github.com/containers/storage/pull/2285#discussion_r2010791595) .
> 
> And after looking at that, feel free to merge without another review.

thanks, pushed a new version.   I'll merge once the CI is green.

Fine to merge https://github.com/containers/common/pull/2374 too?

mtrmac (commented):
/lgtm

mtrmac (approved):

PR #2288, opened 2025-03-18
RFC: Use Go 1.23 iterators for locking+traversing stores.

mtrmac:
RFC: this is a ~demo, only migrating one function out of several.

This replaces one kind of boilerplate with another, but overall seems a bit less repetitive. The primary benefit is that the code working within the locked stores can `return` directly, without boilerplate, and there is no need to manually declare a `func(…)` for the loop body.

The `for store, err := range` syntax is a bit weird, admittedly, but maybe worth it.

The primary downside is that this really doesn’t work for the exactly-one-closure-call helpers (`writeTo…Store`):
- Using `for store, err := range writeToSingleStore` is very unintuitive, but maybe acceptable with good helper names
- But Go doesn’t recognize that the loop body executes exactly once, and that if the loop body always `return`s, there is no need for the code afterwards to `return` as well. So every caller would need to add an unreachable invented `return` there… and that seems a bit much for me.

So, we would end up with two rather different helper patterns instead of ~one. I’m really unsure whether it is worth it. It might well make sense to leave the code as is, and to wait another 1/5/10 years for Go to gain type inference and a convenient closure syntax.

### DIFF
--- a/store.go
+++ b/store.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"iter"
 	"maps"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


Luap99 (commented):
I am not an active c/storage contributor but I read the code regularly (i.e. when debugging a podman issue) and this seems to me clearly an improvement over the callback. It feels much more natural to read the callers as loops so I like this a lot.

store.go
```diff
@@ -1237,41 +1238,36 @@ func (s *store) allLayerStores() ([]roLayerStore, error) {
...
-// If all layer stores are processed without setting done == true, it returns ({}, false, nil).
+// If reading any layer store fails, it yields (nil, err) and terminates.
 //
 // Typical usage:
 //
-//	if res, done, err := s.readAllLayerStores(store, func(…) {
-//		…
-//	}; done {
-//		return res, err
+//	for rlstore, err := range s.readAllLayerStores() {
```
mtrmac, line 1247: Note to self: This looks almost _too easy_, without warning callers about the semantics. Maybe `range iterateLockingAllLayerStores` ?

mtrmac: `iterateReadLockingAllLayerStores`? As usual, I tend towards unwieldy names.

PR #2289, opened 2025-03-19
Improve error handling in archive creation

hanwen-flow:


hanwen-flow:
@mtrmac @giuseppe 

FYI - I have found a different way to deal with tar file overhead in checkpoint/restore , so in the end I don't need tarWithOptionsTo as a public function. It might still be a good addition though. Most calls to TarWithOptions copy the the stream to a destination tar file.

pkg/archive/archive.go
```diff
@@ -616,25 +619,36 @@ func (ta *tarWriter) addFile(path, name string) error {
...
 				return err
 			}
-			if hdr.Typeflag == tar.TypeReg && hdr.Size > 0 {
-				return fmt.Errorf("tar: cannot use whiteout for non-empty file")
-			}
-			hdr = wo
+			hdr = headers.whiteout
+		} else {
+			logrus.Infof("tar: cannot use whiteout for non-empty file %s", hdr.Name)
 		}
```
mtrmac, line 652: I don‘t understand the “pkg/archive: remove "whiteout for non-empty file" error handling” commit, I think the original code was correct.

If the converter returns a new whiteout header, the quoted code calls `WriteHeader` on the original `hdr`, and then immediately follows `WriteHeader(theNewWhiteout)`. In that case, the `Size > 0` situation (as you say) would cause an invalid archive to be created —  so why does that commit turns that problematic situation from an error into a warning?

pkg/archive/archive.go
```diff
@@ -661,6 +675,11 @@ func (ta *tarWriter) addFile(path, name string) error {
...
 	}
 
+	if !headers.fi.IsDir() && hasHardlinks(headers.fi) {
+		// NOSUBMIT: hdr.Name or headers.hdr.Name?
```
mtrmac, line 679: “NOSUBMIT”?

pkg/archive/archive.go
```diff
@@ -528,38 +528,46 @@ func canonicalTarName(name string, isDir bool) (string, error) {
...
+type addFileData struct {
+	path     string
+	hdr      *tar.Header
+	whiteout *tar.Header
+	fi       os.FileInfo
+}
+
+// prepareAddFile generates the tar file header(s) for adding a file
+// from `path` as `name` to the tar archive.
+func (ta *tarWriter) prepareAddFile(path, name string) (*addFileData, error) {
```
mtrmac, line 540: If this can return `nil` to mean “skip”, it would be useful to document that.

pkg/archive/archive_test.go
```diff
@@ -1266,3 +1267,45 @@ func TestTimestamp(t *testing.T) {
 	// we expect the ones with a fixed timestamp to be the same
 	assert.Equal(t, origTarEpochOptions, laterTarEpochOptions)
 }
+
+type errorBuf struct {
+	bytes.Buffer
+
+	try  int
+	fail int
+	err  error
```
mtrmac, line 1276: These fields would benefit from a bit more descriptive names.

mtrmac (commented):
Thanks!

Reading this line by line, this is _fair enough_, but I’m afraid that with this as is, future readers/maintainers would be _more_ confused than with the current code.

Can you add documentation to `addFileData`? Primarily answering why is the “add file” functionality split into two functions (and, if adding new features in the future, in which of the two functions those features should go), and maybe anything else that maintainers should know.

Looking at https://github.com/containers/storage/pull/2261#issuecomment-2667985410 , I _guess_ the answer is “we have existing tests that require some failures to be ignored, so this is a way to structure the code to allow that, without ignoring others”. That could be a fair answer as far as it goes, but now it’s not obvious from the code without such a documentation…

… but, _also_, I’m inclined to push for details on the “require write failures to be ignored” part. I’m _not at all saying_ this is incorrect, but someone needs to understand that this is the case, and that we should preserve that behavior — that it is not a bug in the tests where we are now doing a lot of work to reinforce that bug. And I’ll fully admit that I didn’t look into this: I assume you did, so please write down what you found and where we are. (On this part, I’m completely ignorant and I’d happily defer to maintainers more familiar with the code.)

pkg/archive/archive.go
hanwen-flow: See https://github.com/containers/storage/blob/main/pkg/archive/archive.go#L634,

```
			if err := ta.TarWriter.WriteHeader(hdr); err != nil {
				return err
			}
			if hdr.Typeflag == tar.TypeReg && hdr.Size > 0 {
				return fmt.Errorf("tar: cannot use whiteout for non-empty file")
			}
			hdr = wo
```

the check flags a problematic situation, but it should have been made before writing the header with Size > 0, because the current approach breaks the entire archive without flagging an error to the caller. I don't know why it was written this way; given the state of this code, I suspect carelessness. Removing the check doesn't make a test fail, and moby/moby has this erroneous bit until today.

It was introduced in https://github.com/moby/moby/commit/daa70195171d798cdfb6c4033b33a06a8f7ebe6c . It already ignore the error return from addFile.

for us, we have to decide between:

* write the header, write file, then write whiteout 
* ignore the whiteout if file.size >0
* ignore the file + whiteout if file.size > 0
* fail the archive creation

I can't tell which one is the right solution , but clearly the current code is broken.

hanwen-flow: looking a bit further, the only WhiteoutConverter is the one in archive_linux.go. Its `ConvertWrite` only takes action on directories and char devices, so this condition can never trigger.

hanwen-flow:
>… but, also, I’m inclined to push for details on the “require write failures to be ignored” part. I’m not at all saying this is incorrect, but someone needs to understand that this is the case, and that we should preserve that behavior

these functions run on live containers, so another process may be mutating  the file we're trying to copy out, so there can always be errors, eg. the path we're trying to archive was just removed by another process, so os.Lstat returns ENOENT. 

This code could be much more resistant to races (eg. using openat(2) and friends to fix a file entry to export in just one place.), but it can't be fundamentally fixed: the container may truncate a file while we're writing it to the tar, causing a short read, and there is no way to recover from this. 

With this change, we can be sure that we notice before handing back a corrupted file to the caller.

hanwen-flow:
> these functions run on live containers, so another process may be mutating the file we're trying to copy out, so there can always be erro

I added this as a comment to TarWithOptions.

hanwen-flow:
> Oh and the lint task is failing, with some inscrutable failure… is that an infrastructure problem, or are we somehow throwing away the logs in CI?

with some local tweaks:

```
$ make local-validate
make -C tests/tools GOLANGCI_LINT_VERSION=1.64.7
make[1]: Entering directory '/home/hanwen/vc/containers/storage/tests/tools'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/home/hanwen/vc/containers/storage/tests/tools'
./hack/git-validation.sh
+ export GIT_VALIDATION=tests/tools/build/git-validation
+ GIT_VALIDATION=tests/tools/build/git-validation
+ '[' '!' -x tests/tools/build/git-validation ']'
+ EPOCH_TEST_COMMIT=
+ '[' -z '' ']'
++ git merge-base main HEAD
+ EPOCH_TEST_COMMIT=b87bac8fb133a3570c5b11542bf791ef550f308c
+ exec time tests/tools/build/git-validation -q -run DCO,short-subject -range b87bac8fb133a3570c5b11542bf791ef550f308c..HEAD
1.53user 4.91system 0:06.21elapsed 103%CPU (0avgtext+0avgdata 12544maxresident)k
0inputs+5616outputs (0major+643812minor)pagefaults 0swaps
```


Looks like it's taking 6 seconds on my 16-core machine to check if I have a signed-off line in my commits. :man_shrugging: 

note that my 'main' branch is not upstream's main branch.

pkg/archive/archive.go
mtrmac: I’m not familiar with the problem space, just a minimal pointer for “how do whiteouts work”: https://github.com/opencontainers/image-spec/blob/main/layer.md#whiteouts is the current spec for what they are and how they work — but it’s quite possible that this code is trying to support more formats.

mtrmac: Not to self and/or other reviewers: This needs a careful look.

mtrmac:
> I added this as a comment to TarWithOptions.

I like this, as well as the comments on the two functions around `addFileData`. That’s clear thinking and a clear rationale. Thanks!

pkg/archive/archive.go
hanwen-flow: > Yes. But, sadly, it’s a public API, so we can’t change WhiteoutConverter itself.

the converter is a public API, but it is only selected through TarOptions.WhiteoutFormat, whose interpretation in GetWhiteoutConverter we control.

I did some more reading. The AUFS and overlayfs file systems have different ways of indicating deletions. This functionality is converts between them (where overlayfs -> aufs may introduce 2 entries for a single directory). The assumption therefore is that deletions are never marked by a non-empty files, and that a single whiteout in one format may only expand to 2 entries (without content) in the other format.  

I'll add a comment.

pkg/archive/archive.go
hanwen-flow: the fileinfo `fi` is for headers.hdr.Name, so that's the one we should store.

pkg/archive/archive.go
hanwen-flow: > Where the disconnect seems to be is that previously the code reported a failure, 

it returned an error, but then that error only resulted in a log message.

hanwen-flow:
PTAL

giuseppe (approved):
LGTM

hanwen-flow:
@mtrmac ping?

hanwen-flow:
> I like this, as well as the comments on the two functions around addFileData. That’s clear thinking and a clear rationale. Thanks!

@mtrmac  - would you be able to approve this? Or is there someone else who can give the 2nd LGTM?

hanwen-flow:
Rebased. Needs LGTM again. @giuseppe , @mtrmac

giuseppe (approved):
LGTM

pkg/archive/archive.go
```diff
@@ -850,13 +850,19 @@ func extractTarFileEntry(path, extractDir string, hdr *tar.Header, reader io.Rea
 }
 
 // Tar creates an archive from the directory at `path`, and returns it as a
-// stream of bytes.
+// stream of bytes. This is a convenience wrapper for TarWithOptions.
```
kolyshkin, line 895: nit: may use markup while at it, i.e.
```suggestion
// stream of bytes. This is a convenience wrapper for [TarWithOptions].
```

so that go doc will link to this function.

pkg/archive/archive.go
```diff
@@ -850,13 +850,19 @@ func extractTarFileEntry(path, extractDir string, hdr *tar.Header, reader io.Rea
...
 func Tar(path string, compression Compression) (io.ReadCloser, error) {
 	return TarWithOptions(path, &TarOptions{Compression: compression})
 }
 
-// TarWithOptions creates an archive from the directory at `path`, only including files whose relative
-// paths are included in `options.IncludeFiles` (if non-nil) or not in `options.ExcludePatterns`.
+// TarWithOptions creates an archive from the directory at `path`,
+// only including files whose relative paths are included in
+// `options.IncludeFiles` (if non-nil) or not in
+// `options.ExcludePatterns`.
```
kolyshkin: Generally speaking, it makes no sense to use `backticks` in docstrings.

pkg/archive/archive.go
```diff
@@ -528,38 +528,56 @@ func canonicalTarName(name string, isDir bool) (string, error) {
 	return name, nil
 }
 
-// addFile adds a file from `path` as `name` to the tar archive.
-func (ta *tarWriter) addFile(path, name string) error {
+type addFileData struct {
+	// path from which to read contents
```
kolyshkin, line 532: nit: missing period

pkg/archive/archive.go
```diff
@@ -528,38 +528,56 @@ func canonicalTarName(name string, isDir bool) (string, error) {
...
+	path string
+
+	// os.Stat for the above
```
kolyshkin, line 535: nit: missing period

pkg/archive/archive_test.go
```diff
@@ -1266,3 +1267,45 @@ func TestTimestamp(t *testing.T) {
...
+}
+
+func (b *errorBuf) Close() error {
+	return nil
+}
+
+func TestTarErrorHandling(t *testing.T) {
+	dir := t.TempDir()
+
+	for i := 0; i < 10; i++ {
```
kolyshkin, line 1294: nit: can use `for i := range 10`

kolyshkin (approved):
left some minor nits; SGTM overall

pkg/archive/archive_test.go
hanwen-flow: done.

pkg/archive/archive.go
hanwen-flow: done.

pkg/archive/archive.go
hanwen-flow: done.

pkg/archive/archive.go
hanwen-flow: removed. (I was following conventions from rest of file.)

pkg/archive/archive.go
hanwen-flow: done.

hanwen-flow:
@kolyshkin ping?

hanwen-flow:
@giuseppe - seems to need an LGTM (again?)

giuseppe (approved):
/lgtm

PR #2301, opened 2025-03-27
idmap: fix first argument to open_tree

giuseppe:
0 is a valid file descriptor, to ignore the argument we need to use AT_FDCWD.


cgwalters (approved):
Looks correct to me but how was this not caught before? Presumably you saw a failure somewhere else (podman CI? just guessing?) - I personally like to have at least a basic cookie like that in commit messages.

mtrmac (commented):
/lgtm

PR #2302, opened 2025-03-28
fix minor review comment about the driver mutex and a minor lint issue

Luap99:
driver/overlay: inline the mutex in driver

There is no need for the extra pointer indirection. With this Driver can
no longer be copied but that was never supported and now lint should
even catch a copy and warn as copying a Mutex is not allowed.

---

pkg/idtools: fix lint gofumpt error

CI doesn't seem to cover the libsubid build tag.

---

CI: add libsubid to lint task

This should ensure we run lint with the libsubid build tag.
Now this is not perfect either, as we do not cover the !libsubid but
that is only one single simple file.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -126,7 +126,7 @@ lint_task:
         folder: $GOPATH/pkg/mod
     build_script: |
       apt-get update
-      apt-get install -y libbtrfs-dev
+      apt-get install -y libbtrfs-dev libsubid-dev
     test_script: |
       make TAGS=regex_precompile local-validate
       make lint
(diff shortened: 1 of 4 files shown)


Luap99:
> tests/tools/build/golangci-lint run --build-tags="  "

Looks like it still doesn't run in CI with the tag...

giuseppe (approved):
LGTM

Luap99:
Ok, now it worked. The

> tests/tools/build/golangci-lint run --build-tags=" libsubid "

`hack/libsubid_tag.sh` was not correctly working either.

mtrmac (commented):
/lgtm

Thank you!

PR #2304, opened 2025-03-29
Use any instead of interface{}

kolyshkin:
It's available since Go 1.18 (see https://pkg.go.dev/builtin#any).


layers.go
```diff
@@ -1031,7 +1031,7 @@ func (r *layerStore) saveLayers(saveLocations layerLocations) error {
 	}
 	r.lastWrite = lw
 
-	for locationIndex := 0; locationIndex < numLayerLocationIndex; locationIndex++ {
+	for locationIndex := range numLayerLocationIndex {
```
mtrmac, line 1034: You have filed separate #2305 for this

kolyshkin: My bad; moved this hunk to #2305.

mtrmac (commented):
/lgtm

Thanks!

PR #2305, opened 2025-03-29
Use for range over integers

kolyshkin:
This form is available since Go 1.22 (see https://tip.golang.org/ref/spec#For_range) and will probably be seen more and more in the new code.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -458,7 +458,7 @@ func (r *containerStore) load(lockedForWriting bool) (bool, error) {
 
 	ids := make(map[string]*Container)
 
-	for locationIndex := 0; locationIndex < numContainerLocationIndex; locationIndex++ {
+	for locationIndex := range numContainerLocationIndex {
 		location := containerLocationFromIndex(locationIndex)
 		rpath := r.jsonPath[locationIndex]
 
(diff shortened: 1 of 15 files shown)


mtrmac (commented):
LGTM WRT correctness of the changes.

As for the style … Aesthetically, I’m not personally much of a fan.

`for _ := range 10` to express “10 times”, sure. The full `for i := range 10`, has, to my taste, a bit too much implicit and hidden about the interval this operates over.

Ultimately, I think it might be easier to given in than to continue fight default linters’ settings over this. That is, alone, a good enough reason to merge this.

I’ll, at least, let someone else chime in, on the small chance that my dislike of this syntax were universally and strongly shared.

kolyshkin:
> LGTM WRT correctness of the changes.
> 
> As for the style … Aesthetically, I’m not personally much of a fan.
> 
> `for _ := range 10` to express “10 times”, sure.

By the way, this can (and should) be shortened to `for range 10`.

> The full `for i := range 10`, has, to my taste, a bit too much implicit and hidden about the interval this operates over.
> 

I'm with you on that. To me, `for i := 0; i < 10; i++` is way easier to read than `for i := range 10`.

I guess though this is mostly because I'm so much used to reading (and writing) in the (old/classical) way, thus I'm just biased. Over time, as more and more new style code will appear, the perception should change.

> Ultimately, I think it might be easier to given in than to continue fight default linters’ settings over this. That is, alone, a good enough reason to merge this.

One other reason is, we all have to adopt to this new style (as this is now standard), and this PR is an attempt to speed up this process.

> I’ll, at least, let someone else chime in, on the small chance that my dislike of this syntax were universally and strongly shared.

Would also like to hear more opinions on this one.

rhatdan:
I am fine with the new style.
/lgtm

PR #2306, opened 2025-03-29
Use `slices` and `maps` more where appropriate

kolyshkin:
Please see individual commits for details.

### DIFF
--- a/cmd/containers-storage/main.go
+++ b/cmd/containers-storage/main.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"slices"
 
 	"github.com/containers/storage"
 	"github.com/containers/storage/internal/opts"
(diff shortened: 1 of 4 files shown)


rhatdan (approved):
/approve
LGTM

cmd/containers-storage/main.go
```diff
@@ -84,63 +85,61 @@ func main() {
 	cmd := args[0]
 
 	for _, command := range commands {
```
mtrmac, line 87: Non-blocking: This works, and it is an improvement already, but processing `commands` in a loop is not quite what we _mean_.

Consider
```go
i := slices.IndexFunc(commands, func(c) { slices.Contains(c.names, cmd) })
if i == -1 { /* fail */ }
command := commands[i]
// straight-line top-level code processing the command we found
```

mtrmac (commented):
LGTM

cmd/containers-storage/main.go
kolyshkin: You are right, this looks weird.

Rewrote (I left `for range` as it seems more readable than `slices.IndexFunc` in this particular case).

cmd/containers-storage/main.go
```diff
@@ -81,69 +82,70 @@ func main() {
...
+	name := args[0]
+	command := func() *command {
+		for i := range commands {
+			if slices.Contains(commands[i].names, name) {
+				return &commands[i]
 			}
 		}
+		fmt.Printf("%s: unrecognized command.\n", name)
+		os.Exit(1)
+		return nil // To satisfy linters.
```
mtrmac, line 94: (Absolutely non-blocking: Aesthetically, I’d prefer `return nil` here, and having `os.Exit` at the top level, but this works perfectly fine.)

mtrmac (commented):
LGTM.

PR #2308, opened 2025-03-31
dirvers/btrfs: remove btrfs_noversion tag

Luap99:
The header should be included since btrfs-progs 3.16.1, it is over 10 years old now so I think we can just drop the tag and assume it is always there. This allows us to simplify the buildtag script as well.


mtrmac:
FWIW this can be also removed in users after the current vendor dance.
- https://github.com/containers/image/pull/2831
- https://github.com/containers/skopeo/pull/2581

PR #2312, opened 2025-04-09
chunked: use temporary file for tar-split data

giuseppe:
Replace the in-memory buffer with a O_TMPFILE file.  This reduces the memory requirements for a partial pull since the tar-split data can be written to disk.

c/image needs the following patch:

```
index d1c5acef..e66e0ccc 100644
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -425,6 +425,7 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 	if err != nil {
 		return private.UploadedBlob{}, err
 	}
+	defer differ.Close()
 
 	out, err := s.imageRef.transport.store.PrepareStagedLayer(nil, differ)
 	if err != nil {
```

### DIFF
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -267,6 +267,7 @@ type DifferOptions struct {
 // This API is experimental and can be changed without bumping the major version number.
 type Differ interface {
 	ApplyDiff(dest string, options *archive.TarOptions, differOpts *DifferOptions) (DriverWithDifferOutput, error)
+	Close() error
 }
 
 // DriverWithDiffer is the interface for direct diff access.
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -216,7 +216,7 @@ type DriverWithDifferOutput struct {
 	CompressedDigest   digest.Digest
 	Metadata           string
 	BigData            map[string][]byte
-	TarSplit           []byte // nil if not available
+	TarSplit           *os.File // nil if not available
 	TOCDigest          digest.Digest
 	// RootDirMode is the mode of the root directory of the layer, if specified.
 	RootDirMode *os.FileMode
--- a/layers.go
+++ b/layers.go
@@ -2550,10 +2550,14 @@ func (r *layerStore) applyDiffFromStagingDirectory(id string, diffOutput *driver
 		if err != nil {
 			compressor = pgzip.NewWriter(&tsdata)
 		}
+		if _, err := diffOutput.TarSplit.Seek(0, 0); err != nil {
+			return err
+		}
+
 		if err := compressor.SetConcurrency(1024*1024, 1); err != nil { // 1024*1024 is the hard-coded default; we're not changing that
 			logrus.Infof("setting compression concurrency threads to 1: %v; ignoring", err)
 		}
-		if _, err := compressor.Write(diffOutput.TarSplit); err != nil {
+		if _, err := diffOutput.TarSplit.WriteTo(compressor); err != nil {
 			compressor.Close()
 			return err
 		}
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -433,19 +433,26 @@ func ensureFileMetadataAttributesMatch(a, b *minimal.FileMetadata) error {
 	return nil
 }
 
-func decodeAndValidateBlob(blob []byte, lengthUncompressed uint64, expectedCompressedChecksum string) ([]byte, error) {
+func validateBlob(blob []byte, expectedCompressedChecksum string) error {
 	d, err := digest.Parse(expectedCompressedChecksum)
 	if err != nil {
-		return nil, fmt.Errorf("invalid digest %q: %w", expectedCompressedChecksum, err)
+		return fmt.Errorf("invalid digest %q: %w", expectedCompressedChecksum, err)
 	}
 
 	blobDigester := d.Algorithm().Digester()
 	blobChecksum := blobDigester.Hash()
 	if _, err := blobChecksum.Write(blob); err != nil {
-		return nil, err
+		return err
 	}
 	if blobDigester.Digest() != d {
-		return nil, fmt.Errorf("invalid blob checksum, expected checksum %s, got %s", d, blobDigester.Digest())
+		return fmt.Errorf("invalid blob checksum, expected checksum %s, got %s", d, blobDigester.Digest())
+	}
+	return nil
+}
+
+func decodeAndValidateBlob(blob []byte, lengthUncompressed uint64, expectedCompressedChecksum string) ([]byte, error) {
+	if err := validateBlob(blob, expectedCompressedChecksum); err != nil {
+		return nil, err
 	}
 
 	decoder, err := zstd.NewReader(nil) //nolint:contextcheck
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"maps"
+	"os"
 	"slices"
 	"strconv"
 	"time"
@@ -18,6 +19,7 @@ import (
 	"github.com/vbatts/tar-split/archive/tar"
 	"github.com/vbatts/tar-split/tar/asm"
 	"github.com/vbatts/tar-split/tar/storage"
+	"golang.org/x/sys/unix"
 )
 
 const (
@@ -157,10 +159,33 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64, 
 	return manifestUncompressed, tocOffset, nil
 }
 
+func openTmpFile(tmpDir string) (*os.File, error) {
+	file, err := os.OpenFile(tmpDir, unix.O_TMPFILE|unix.O_RDWR|unix.O_CLOEXEC|unix.O_EXCL, 0o600)
+	if err == nil {
+		return file, nil
+	}
+	return openTmpFileNoTmpFile(tmpDir)
+}
+
+// openTmpFileNoTmpFile is a fallback used by openTmpFile when the underlying file system does not
+// support O_TMPFILE.
+func openTmpFileNoTmpFile(tmpDir string) (*os.File, error) {
+	file, err := os.CreateTemp(tmpDir, ".tmpfile")
+	if err != nil {
+		return nil, err
+	}
+	// Unlink the file immediately so that only the open fd refers to it.
+	_ = os.Remove(file.Name())
+	return file, nil
+}
+
 // readZstdChunkedManifest reads the zstd:chunked manifest from the seekable stream blobStream.
-// Returns (manifest blob, parsed manifest, tar-split blob or nil, manifest offset).
+// tmpDir is a directory where the tar-split temporary file is written to.  The file is opened with
+// O_TMPFILE so that it is automatically removed when it is closed.
+// Returns (manifest blob, parsed manifest, tar-split file or nil, manifest offset).  The opened tar-split file
+// points to the end of the file (equivalent to Seek(0, 2)).
 // It may return an error matching ErrFallbackToOrdinaryLayerDownload / errFallbackCanConvert.
-func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Digest, annotations map[string]string) (_ []byte, _ *minimal.TOC, _ []byte, _ int64, retErr error) {
+func readZstdChunkedManifest(tmpDir string, blobStream ImageSourceSeekable, tocDigest digest.Digest, annotations map[string]string) (_ []byte, _ *minimal.TOC, _ *os.File, _ int64, retErr error) {
 	offsetMetadata := annotations[minimal.ManifestInfoKey]
 	if offsetMetadata == "" {
 		return nil, nil, nil, 0, fmt.Errorf("%q annotation missing", minimal.ManifestInfoKey)
@@ -245,7 +270,7 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
 		return nil, nil, nil, 0, fmt.Errorf("unmarshaling TOC: %w", err)
 	}
 
-	var decodedTarSplit []byte = nil
+	var decodedTarSplit *os.File
 	if toc.TarSplitDigest != "" {
 		if tarSplitChunk.Offset <= 0 {
 			return nil, nil, nil, 0, fmt.Errorf("TOC requires a tar-split, but the %s annotation does not describe a position", minimal.TarSplitInfoKey)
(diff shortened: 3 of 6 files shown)


pkg/chunked/compression_linux.go
```diff
@@ -254,14 +258,20 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
 		if err != nil {
 			return nil, nil, nil, 0, err
 		}
-		decodedTarSplit, err = decodeAndValidateBlob(tarSplit, tarSplitLengthUncompressed, toc.TarSplitDigest.String())
+		fd, err := unix.Open(tmpDir, unix.O_TMPFILE|unix.O_RDWR|unix.O_CLOEXEC, 0o600)
```
Luap99, line 261: just a drive by comment
The open man page says O_TMPFILE needs support from the file system. Is that a real problem we would need to care about? Could this affect any users on an uncommon fs that does not have support for it?

pkg/chunked/compression_linux.go
```diff
@@ -254,14 +258,20 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
...
 		if err != nil {
+			return nil, nil, nil, 0, err
```
Luap99, line 263: The error produced by open is just the errno, I suggest you wrap this with something like
`&fs.PathError{Op: "open O_TMPFILE", Path: tmpDir, Err: err}`

pkg/chunked/compression_linux.go
giuseppe: good point.  I've added a fallback when the underlying file system doesn't support it and adapted the other user of `O_TMPFILE`

pkg/chunked/compression_linux.go
giuseppe: Thanks, fixed

pkg/chunked/compression_linux.go
```diff
@@ -157,10 +162,36 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
...
+		name := fmt.Sprintf(".tmpfile-%d", rand.Int63())
+		path := filepath.Join(tmpDir, name)
+
+		fd, err := unix.Open(path, unix.O_RDWR|unix.O_CREAT|unix.O_EXCL|unix.O_CLOEXEC, 0o600)
+		if err == nil {
+			// Unlink the file immediately so that only the open fd refers to it.
+			_ = os.Remove(path)
+			return fd, nil
+		}
+		if !errors.Is(err, os.ErrNotExist) {
```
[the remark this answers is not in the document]
giuseppe: yes thanks for catching this silly bug.  As punishment, I've added a unit test

pkg/chunked/compression_linux.go
cgwalters: Is there really no well-known Go library out there wrapping this? In my Rust projects we use https://docs.rs/cap-tempfile/latest/cap_tempfile/struct.TempFile.html and I've contributed to the crate.

pkg/chunked/compression_linux.go
```diff
@@ -157,10 +162,41 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
 	return manifestUncompressed, tocOffset, nil
 }
 
+func openTmpFile(tmpDir string) (fd int, err error) {
+	fd, err = unix.Open(tmpDir, unix.O_TMPFILE|unix.O_RDWR|unix.O_CLOEXEC, 0o600)
```
cgwalters, line 166: Super minor but for this use case you can also add `O_EXCL` which signals that the file can never be linked (i.e. given a name later), in theory this allows optimizing things a bit.

giuseppe: thanks.  Added now

pkg/chunked/compression_linux.go
```diff
@@ -157,10 +162,41 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
 	return manifestUncompressed, tocOffset, nil
 }
 
+func openTmpFile(tmpDir string) (fd int, err error) {
+	fd, err = unix.Open(tmpDir, unix.O_TMPFILE|unix.O_RDWR|unix.O_CLOEXEC|unix.O_EXCL, 0o600)
```
liamstask, line 166: minor, but should be able to use https://pkg.go.dev/os#OpenFile here and below to create an `*os.File` directly - can use the flags from https://pkg.go.dev/syscall similar to the re-exported constants in pkg/os (https://pkg.go.dev/os#pkg-constants)

drivers/driver.go
```diff
@@ -216,7 +216,7 @@ type DriverWithDifferOutput struct {
 	CompressedDigest   digest.Digest
 	Metadata           string
 	BigData            map[string][]byte
-	TarSplit           []byte // nil if not available
+	TarSplit           *os.File // nil if not available
```
liamstask, line 219: maybe not a performance bottleneck, but it's probably worth wrapping in https://pkg.go.dev/bufio Reader/Writer to reduce syscall traffic on lots of tiny reads/writes

liamstask (commented):
found this PR while I happened to be following a few threads related to this, and noticed that using buffered i/o would likely be worthwhile - feel free to disregard, of course!

pkg/chunked/compression_linux.go
giuseppe: thanks, amended now

drivers/driver.go
giuseppe: I think it should not really matter because we don't perform lots of read/writes (so unless it is really worth it, I'd prefer to avoid another buffer since it goes against the goal of this PR)

giuseppe:
@Luap99 @mtrmac the PR is ready, PTAL

pkg/chunked/compression_linux.go
```diff
@@ -254,14 +292,19 @@ func readZstdChunkedManifest(blobStream ImageSourceSeekable, tocDigest digest.Di
 		if err != nil {
 			return nil, nil, nil, 0, err
 		}
-		decodedTarSplit, err = decodeAndValidateBlob(tarSplit, tarSplitLengthUncompressed, toc.TarSplitDigest.String())
+		decodedTarSplit, err = openTmpFile(tmpDir)
 		if err != nil {
+			return nil, nil, nil, 0, err
+		}
+		if err := decodeAndValidateBlobToStream(tarSplit, decodedTarSplit, toc.TarSplitDigest.String()); err != nil {
+			decodedTarSplit.Close()
```
Luap99, line 300: I have not fully understood all the code but it seem many callers no call seek(0,0) on the tarsplit, would it makes sense to seek once after this here instead? Seems kinda unintuitive and easy to forget for a caller otherwise that they must seek the returned file?

pkg/chunked/compression_linux.go
```diff
@@ -157,10 +161,44 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
...
+		return file, nil
+	}
+	return openTmpFileNoTmpFile(tmpDir)
+}
+
+// openTmpFileNoTmpFile is a fallback used by openTmpFile when the underlying file system does not
+// support O_TMPFILE.
+func openTmpFileNoTmpFile(tmpDir string) (*os.File, error) {
+	var retErr error
+	for i := 0; i < 100; i++ {
```
kolyshkin, line 176: nit: can use `for range 100` here, especially since `i` is not used.

pkg/chunked/compression_linux.go
```diff
@@ -157,10 +161,44 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
...
+	file, err := os.OpenFile(tmpDir, unix.O_TMPFILE|unix.O_RDWR|unix.O_CLOEXEC|unix.O_EXCL, 0o600)
+	if err == nil {
+		return file, nil
+	}
+	return openTmpFileNoTmpFile(tmpDir)
+}
+
+// openTmpFileNoTmpFile is a fallback used by openTmpFile when the underlying file system does not
+// support O_TMPFILE.
+func openTmpFileNoTmpFile(tmpDir string) (*os.File, error) {
```
kolyshkin, line 174: Is this code better than using `os.CreateTemp`? I.e. something like this:

```go
func openTmpFileNoTmpFile(tmpDir string) (*os.File, error) {
	file, err := os.CreateTemp(tmpDir, ".tmpfile")
	if err != nil {
		return nil, err
	}
	// Unlink the file immediately so that only the open fd refers to it.
	_ = os.Remove(file.Name())
	return file, nil
}
```

drivers/driver.go
liamstask: just noting that since an `*os.File` is being handed as an `io.Writer` or `io.Reader` to a variety of compressors, encoders, decoders, each of those cases will generally result in lots of internal write/read syscalls directly on the file descriptor as opposed to batching them via `bufio`

mtrmac: That would really need benchmarking; there’s a lot of invisible `ReadFrom`/`WriteTo` optimization in various libraries. And it’s at least possible to imagine that adding buffers and other layers of indirection _breaks that_ because the optimizations rely on specific type checks.

Either way, I don’t think this is a blocker _at this point_ — the purpose of these PRs is to make large pulls _possible_, and that’s far more important than making them ideally performant.

pkg/chunked/compression_linux.go
giuseppe: If the file is accessed multiple times, we need to call seek(0, 0) each time to read it fully from the beginning.  So we can't just call it once

pkg/chunked/compression_linux.go
giuseppe: much better.  I'll fix it

pkg/chunked/compression_linux.go
Luap99: ack, maybe add a comment to this function that the returned file was already read and the caller must seek to the beginning if it tries to read it again.
Also I guess `tar-split blob or nil` should be changed to `tar-split file or nil`

pkg/chunked/storage_linux.go
```diff
@@ -333,13 +339,16 @@ func makeConvertFromRawDiffer(store storage.Store, blobDigest digest.Digest, blo
 // makeZstdChunkedDiffer sets up a chunkedDiffer for a zstd:chunked layer.
 // It may return an error matching ErrFallbackToOrdinaryLayerDownload / errFallbackCanConvert.
 func makeZstdChunkedDiffer(store storage.Store, blobSize int64, tocDigest digest.Digest, annotations map[string]string, iss ImageSourceSeekable, pullOptions pullOptions) (*chunkedDiffer, error) {
-	manifest, toc, tarSplit, tocOffset, err := readZstdChunkedManifest(iss, tocDigest, annotations)
+	manifest, toc, tarSplit, tocOffset, err := readZstdChunkedManifest(store.RunRoot(), iss, tocDigest, annotations)
```
mtrmac, line 342: There are several error code paths; `tarSplit` doesn’t get closed?

mtrmac (commented):
I’m afraid I’m way behind, just a quick note.

I’m also generally confused about the lifetime of `DriverWithDifferOutput.TarSplit`.

pkg/chunked/storage_linux.go
giuseppe: follow-up in https://github.com/containers/storage/pull/2323

mtrmac, line 342: `RunRoot` is, by default, in `/run` … so the net total on this code path is that it moves the uncompressed tar-split from process memory (which can be swapped out, and is limited by swap size) to a `tmpfs` (_also_ limited by swap size).

It’s still a bit of an improvement, because we have a bit of a more direct control over allocation / uses / lifetime of the memory — but it does it substantially change the actual limit?

mtrmac (commented):
Super late — I’ll file a PR for the nits.

I’d love a discussion about placing the files in `/run`, though.

layers.go
```diff
@@ -2550,10 +2550,14 @@ func (r *layerStore) applyDiffFromStagingDirectory(id string, diffOutput *driver
 		if err != nil {
 			compressor = pgzip.NewWriter(&tsdata)
 		}
+		if _, err := diffOutput.TarSplit.Seek(0, 0); err != nil {
```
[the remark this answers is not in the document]
mtrmac: In https://github.com/containers/storage/pull/2328 .

pkg/chunked/compression_linux.go
```diff
@@ -157,10 +159,33 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
...
+	_ = os.Remove(file.Name())
+	return file, nil
+}
+
 // readZstdChunkedManifest reads the zstd:chunked manifest from the seekable stream blobStream.
-// Returns (manifest blob, parsed manifest, tar-split blob or nil, manifest offset).
+// tmpDir is a directory where the tar-split temporary file is written to.  The file is opened with
+// O_TMPFILE so that it is automatically removed when it is closed.
+// Returns (manifest blob, parsed manifest, tar-split file or nil, manifest offset).  The opened tar-split file
+// points to the end of the file (equivalent to Seek(0, 2)).
```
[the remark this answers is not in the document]
mtrmac: https://github.com/containers/storage/pull/2328 removes that instead.

pkg/chunked/compression_linux.go
```diff
@@ -457,3 +498,18 @@ func decodeAndValidateBlob(blob []byte, lengthUncompressed uint64, expectedCompr
 	b := make([]byte, 0, lengthUncompressed)
 	return decoder.DecodeAll(blob, b)
 }
+
+func decodeAndValidateBlobToStream(blob []byte, w *os.File, expectedCompressedChecksum string) error {
+	if err := validateBlob(blob, expectedCompressedChecksum); err != nil {
+		return err
+	}
+
+	decoder, err := zstd.NewReader(bytes.NewReader(blob)) //nolint:contextcheck
```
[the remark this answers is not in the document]
mtrmac: In https://github.com/containers/storage/pull/2328 .

pkg/chunked/compression_linux.go
```diff
@@ -157,10 +159,33 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
 	return manifestUncompressed, tocOffset, nil
 }
 
+func openTmpFile(tmpDir string) (*os.File, error) {
+	file, err := os.OpenFile(tmpDir, unix.O_TMPFILE|unix.O_RDWR|unix.O_CLOEXEC|unix.O_EXCL, 0o600)
```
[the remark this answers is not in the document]
mtrmac: In https://github.com/containers/storage/pull/2328 .

PR #2318, opened 2025-04-14
types: Always apply rootless path defaults

giuseppe:
When running as non-root, the code previously checked if usePerUserStorage() before applying default paths for RunRoot and GraphRoot if they were missing from the configuration file. This check prevented defaults from being applied if the STORAGE_DRIVER environment variable was set.

Do not merge before https://github.com/containers/common/pull/2420


Luap99 (approved):
LGTM

giuseppe:
@containers/storage-maintainers ready to merge, PTAL

PR #2323, opened 2025-04-29
fix tarsplit file leak

giuseppe:
follow-up for https://github.com/containers/storage/pull/2312

### DIFF
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -1873,6 +1873,9 @@ func (c *chunkedDiffer) ApplyDiff(dest string, options *archive.TarOptions, diff
 
 	output.Artifacts[fsVerityDigestsKey] = c.fsVerityDigests
 
+	// on success steal the reference to the tarSplit file
+	c.tarSplit = nil
+
 	return output, nil
 }
 
--- a/pkg/chunked/storage_linux.go
+++ b/pkg/chunked/storage_linux.go
@@ -80,7 +80,7 @@ type chunkedDiffer struct {
 	convertToZstdChunked bool
 
 	// Chunked metadata
-	// This is usually set in GetDiffer, but if convertToZstdChunked, it is only computed in chunkedDiffer.ApplyDiff
+	// This is usually set in NewDiffer, but if convertToZstdChunked, it is only computed in chunkedDiffer.ApplyDiff
 	// ==========
 	// tocDigest is the digest of the TOC document when the layer
 	// is partially pulled, or "" if not relevant to consumers.
@@ -95,7 +95,7 @@ type chunkedDiffer struct {
 	skipValidation bool
 
 	// Long-term caches
-	// This is set in GetDiffer, when the caller must not hold any storage locks, and later consumed in .ApplyDiff()
+	// This is set in NewDiffer, when the caller must not hold any storage locks, and later consumed in .ApplyDiff()
 	// ==========
 	layersCache     *layersCache
 	copyBuffer      []byte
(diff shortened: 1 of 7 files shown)


pkg/chunked/storage_linux.go
```diff
@@ -338,11 +338,18 @@ func makeConvertFromRawDiffer(store storage.Store, blobDigest digest.Digest, blo
 
 // makeZstdChunkedDiffer sets up a chunkedDiffer for a zstd:chunked layer.
 // It may return an error matching ErrFallbackToOrdinaryLayerDownload / errFallbackCanConvert.
-func makeZstdChunkedDiffer(store storage.Store, blobSize int64, tocDigest digest.Digest, annotations map[string]string, iss ImageSourceSeekable, pullOptions pullOptions) (*chunkedDiffer, error) {
+func makeZstdChunkedDiffer(store storage.Store, blobSize int64, tocDigest digest.Digest, annotations map[string]string, iss ImageSourceSeekable, pullOptions pullOptions) (_ *chunkedDiffer, retErr error) {
 	manifest, toc, tarSplit, tocOffset, err := readZstdChunkedManifest(store.RunRoot(), iss, tocDigest, annotations)
 	if err != nil { // May be ErrFallbackToOrdinaryLayerDownload / errFallbackCanConvert
 		return nil, fmt.Errorf("read zstd:chunked manifest: %w", err)
 	}
+	if tarSplit != nil {
```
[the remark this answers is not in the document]
giuseppe: not really sure what code is generated but I assumed avoiding a defer was better when not needed.  Probably not worth the effort, I'll fix it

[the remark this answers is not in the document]
kolyshkin: I mean, there's already another `if tarSplit != nil {` a few lines below, so you can merge yours into that one.

pkg/chunked/storage_linux.go
```diff
@@ -199,10 +200,10 @@ func (c *chunkedDiffer) Close() error {
 	return nil
 }
 
-// GetDiffer returns a differ than can be used with ApplyDiffWithDiffer.
+// NewDiffer returns a differ than can be used with ApplyDiffWithDiffer.
 // If it returns an error that matches ErrFallbackToOrdinaryLayerDownload, the caller can
 // retry the operation with a different method.
```
mtrmac, line 205: The caller must call `Close`

pkg/chunked/storage_unsupported.go
```diff
@@ -11,7 +11,7 @@ import (
 	digest "github.com/opencontainers/go-digest"
 )
 
-// GetDiffer returns a differ than can be used with ApplyDiffWithDiffer.
-func GetDiffer(ctx context.Context, store storage.Store, blobDigest digest.Digest, blobSize int64, annotations map[string]string, iss ImageSourceSeekable) (graphdriver.Differ, error) {
+// NewDiffer returns a differ than can be used with ApplyDiffWithDiffer.
```
mtrmac, line 14: The caller must call `Close`

store.go
```diff
@@ -3163,6 +3163,9 @@ func (s *store) ApplyStagedLayer(args ApplyStagedLayerOptions) (*Layer, error) {
 }
 
 func (s *store) CleanupStagedLayer(diffOutput *drivers.DriverWithDifferOutput) error {
+	if diffOutput.TarSplit != nil {
```
mtrmac, line 3166: Many functions that return `drivers.DriverWithDifferOutput` should document that the caller should ensure `Cleanup…` is called eventually … or whatever the rules are.

mtrmac (changes requested):
I think this fixes the part where closing `chunkedDiffer` closes / deletes the temporary file, by “stealing” the file handle. That’s fine.

But the current rules as documented in `PrepareStagedLayer` are that the caller only should only call `CleanupStagedLayer` _if `ApplyStagedLayer` is not called / does not succeed_; so, closing the file only in `Cleanup…` looks wrong. (Arguably the API is inconvenient, and should be changed in a way that callers always call “cleanup”, even on success of `ApplyStagedLayer`.)

store.go
giuseppe: I've added a patch to finally drop `ApplyDiffWithDiffer` and documented the other place where we use it

store.go
```diff
@@ -365,6 +365,8 @@ type Store interface {
 	// PrepareStagedLayer applies a diff to a layer.
 	// It is the caller responsibility to clean the staging directory if it is not
 	// successfully applied with ApplyStagedLayer.
+	// The caller must ensure ApplyStagedLayer or CleanupStagedLayer is called eventually with
```
kolyshkin, line 368: nit: we can start using references to other functions/methods, i.e.
```suggestion
// The caller must ensure [Store.ApplyStagedLayer] or [Store.CleanupStagedLayer] is called eventually with
```

giuseppe: thanks, I've fixed only this instance for now and the new comments I've added

store.go
```diff
@@ -3163,6 +3165,10 @@ func (s *store) ApplyStagedLayer(args ApplyStagedLayerOptions) (*Layer, error) {
 }
 
 func (s *store) CleanupStagedLayer(diffOutput *drivers.DriverWithDifferOutput) error {
+	if diffOutput.TarSplit != nil {
```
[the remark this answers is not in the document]
giuseppe: added, thanks

mtrmac (approved):
/lgtm

Thanks!

PR #2325, opened 2025-05-05
[PROTOTYPE] Temporary directory

Honny1:
This PR contains a prototype of a temporary directory that could be used to delete files outside the global storage lock to reduce the global lock time.

Design doc: https://docs.google.com/document/d/196S3MWJkuqoKDhkvtnbT6z9LFZzwibIFmcEM9EyJu8Q/edit?usp=sharing


cgwalters:
> Design doc: https://docs.google.com/document/d/196S3MWJkuqoKDhkvtnbT6z9LFZzwibIFmcEM9EyJu8Q/edit?usp=sharing

I think we should avoid using private/company-specific Google docs to design public FOSS in general especially as part of the movement of projects in this space to the CNCF.

Honny1:
@cgwalters ok

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+		rootDir:      rootDir,
+		tempDirPath:  tempDirPath,
+		lockFilePath: lockFilePath,
+		lock:         lock,
+	}
+	return td, nil
+}
+
+// Add moves the specified file to the temporary directory.
+func (td *TempDir) Add(path string, prefix string) error {
```
mtrmac, line 100: Every caller copy&pastes a time format for `prefix`; can that be moved inside this function instead?

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+	td.lock.AssertLocked()
+
+	base := filepath.Base(path)
```
mtrmac, line 103: This assumes that `filepath.Base` of all relevant paths will be unique. Is that safe? Can that assumption be removed?

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if prefix != "" {
+		base = prefix + "-" + base
+	}
+	dest := filepath.Join(td.tempDirPath, base)
+	return os.Rename(path, dest)
```
mtrmac, line 109: (This is fine, but eventually it should be documented that the submitted `path` must be on the same filesystem as the `TempDir`.)

pkg/tempdir/tempdir.go
mtrmac, line 100: (`Add` is probably not the best name of this function, but that’s something for much later.)

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	return errors.Join(recoveryErrors...)
+}
+
+// NewTempDir creates a new temporary directory in the specified root directory.
+// Root directory must exist and be writable. Root directory is cleared of stale
+// temporary directories and lock files before creating a new temporary directory.
+//
+// Required global storage lock must be held before calling this function.
```
mtrmac, line 68: Which “global storage lock”? (In particular, `store.graphLock` is the “most top-level” lock and usually not held at the time of `startWriting`.)

Eventually, this will need to be precise about what the requirements are.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// TODO: implement alternative for: system.EnsureRemoveAll(dir)
+
+// TempDir represents a temporary directory that is created in a specified root directory.
+type TempDir struct {
+	rootDir      string
+	tempDirPath  string
+	lockFilePath string
+	lock         *lockfile.LockFile
```
mtrmac, line 20: (Eventually, the semantics of `lock` will need to be precisely documented. That can wait until the last phases of the prototype.)

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+
+// Note: Cleanup must be deferred before differing unlocking global storage locks
+// to be executed after the global storage lock is released to benefit from
+// not performing slow removal files from disk during global lock.
+func (td *TempDir) Cleanup() (err error) {
+	td.lock.Unlock()
+
+	if err := os.RemoveAll(td.tempDirPath); err != nil {
+		return fmt.Errorf("removing temp dir failed: %w", err)
+	}
```
mtrmac, line 120: `lockFilePath` is never removed?

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+// and the function attempts to acquire a lock on each directory before
+// removing it. If a lock cannot be acquired, the directory is skipped, because other
+// processes may still be using it.
+//
+// It is important to note that this function should be called before
+// creating a new temporary directory to ensure a clean state.
+//
+// Required global storage lock must be held before calling this function.
+func recover(rootDir string) error {
+	tempDirPattern := filepath.Join(rootDir, "temp-dir-*")
```
mtrmac, line 35: Even within this package, the `temp-dir-` prefix should be a named constant, not copy&pasted.

Outside of this package, callers should not know the prefix (or that there even _is_ a prefix) at all. Either provide a method to check whether a directory is `TempDir`-owned, or maybe design things such that the callers don’t need to know that — maybe define a _stable_ “parent of staging directories” for the exclusive use of `TempDir` (and then call, e.g. `NewTempDir(filepath.Join(runDir, "staging"))` instead of `NewTempDir(runDir))`)

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+		return fmt.Errorf("error finding tmp dirs in %s: %w", rootDir, err)
+	}
+
+	var recoveryErrors []error
+
+	for _, tempDirPath := range tempDirs {
+		lockFilePath := filepath.Join(tempDirPath, ".lock")
+		lock, err := lockfile.GetLockFile(lockFilePath)
+		if err != nil {
+			recoveryErrors = append(recoveryErrors, fmt.Errorf("error getting lock file %s: %w", lockFilePath, err))
```
mtrmac, line 47: It might have been deleted by this time, that’s not an error.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+			continue
+		}
+
+		if err := lock.TryLock(); err == nil {
+			lock.Unlock()
```
mtrmac, line 52: Why the unlock _here_? That way a concurrent `recover` starting at this point will _also_ try to remove things, that seems unnecessary.

“The process that holds the lock is responsible for deleting the related subdirectory” might be a good invariant to hold, but I’m not sure what design you are working towards.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+	dest := filepath.Join(td.tempDirPath, base)
+	return os.Rename(path, dest)
+}
+
+// Note: Cleanup must be deferred before differing unlocking global storage locks
+// to be executed after the global storage lock is released to benefit from
+// not performing slow removal files from disk during global lock.
+func (td *TempDir) Cleanup() (err error) {
+	td.lock.Unlock()
+
```
mtrmac, line 117: At this point, the calling process has full intention of deleting the data, but a concurrent `recover` from another process would see an unlocked lock and (unnecessarily) try to delete things as well.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+// removing it. If a lock cannot be acquired, the directory is skipped, because other
+// processes may still be using it.
+//
+// It is important to note that this function should be called before
+// creating a new temporary directory to ensure a clean state.
+//
+// Required global storage lock must be held before calling this function.
+func recover(rootDir string) error {
+	tempDirPattern := filepath.Join(rootDir, "temp-dir-*")
+	tempDirs, err := filepath.Glob(tempDirPattern)
```
mtrmac, line 36: As is, this would add a `.Glob` to every RW lock acquisition. It would be interesting to try to optimize that.

drivers/vfs/driver.go
```diff
@@ -241,6 +245,9 @@ func (d *Driver) dir(id string) string {
 
 // Remove deletes the content from the directory for a given id.
 func (d *Driver) Remove(id string) error {
+	if d.tempDirectory != nil {
+		return d.tempDirectory.Add(d.dir(id), time.Now().Format("20060102-150405"))
```
mtrmac, line 249: `d.dir` might be in `d.imageStore` or `d.additionalHomes`, potentially on different filesystems; a `Rename` is not guaranteed to work.

(Applies also in other places, and it’s almost certainly out of scope of the prototype.)

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
+			recoveryErrors = append(recoveryErrors, fmt.Errorf("error getting lock file %s: %w", lockFilePath, err))
+			continue
+		}
+
+		if err := lock.TryLock(); err == nil {
+			lock.Unlock()
+			if rmErr := os.RemoveAll(tempDirPath); rmErr != nil && !os.IsNotExist(rmErr) {
+				recoveryErrors = append(recoveryErrors, fmt.Errorf("error removing stale temp dir %s: %w", tempDirPath, rmErr))
+			}
```
mtrmac, line 55: `lockFilePath` is never removed?

mtrmac (commented):
A very preliminary first look — **feel free to ignore if feedback is not yet wanted**.

At this point I’m mostly focused on the `TempDir` mechanism’s design/API; callers are interesting to the extent that they represent use cases that need to be handled, but

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,126 @@
...
+//
+// Required global storage lock must be held before calling this function.
+func NewTempDir(rootDir string) (*TempDir, error) {
+	if err := recover(rootDir); err != nil {
+		return nil, fmt.Errorf("recovering from stale temp dirs failed: %w", err)
+	}
+
+	pid := os.Getpid()
+
+	tempDirPath := filepath.Join(rootDir, fmt.Sprintf("temp-dir-%d", pid))
```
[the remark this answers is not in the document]
mtrmac: … actually, a single process can contain multiple threads using c/storage, and multiple goroutines (some goroutines possibly allocated to the same OS thread, and that allocation changes over time!). So a PID is not a unique identifier of the current caller at all.

This is not theoretical, c/image pulls 6 layers concurrently within a single process, and that _can_ concurrently obtain RW layer locks, at least via `CleanupStagedLayer`.

pkg/tempdir/tempdir.go
giuseppe: isn't `lockFilePath` under the temporary directory?

And I think we need to delete it after the data, so probably we need the temp dir to have another component:

```
/foo/bar/temp-dir/
/foo/bar/temp-dir/.lock
/foo/bar/temp-dir/data/
```

so we first do `rm -rf /foo/bar/temp-dir/data/` and after that succeeds (or the `_data dir doesn't exist), we can delete `temp-dir`, otherwise we risk to delete the lock file too early and confuse other callers

mtrmac: Ah, my mistake, indeed it is a file within `tempDirPath`.

I’m not sure that works:
- Process A creates `${tempDirPath}`
- $HERE
- Process A creates `${tempDirPath}/.lock`

If a cleanup starts runs $HERE, it sees a staging directory with lock removed, so it should remove that staging directory — and it will break process A, which is still running.

(We could fix this by pre-creating the directory structure in a private location and moving it into the staging directory, but that seems more expensive than using `${tempDirPath}.lock` as a peer path.)

pkg/tempdir/tempdir.go
mtrmac: (I’m wondering about even more optimization — we could have an “might have a staging directory” bool in `layers.json`, and use the `layerStore.lockfile.RecordWrite` notification mechanism to make all of this ~zero-cost to users of `layerStore.startReading`. But, one step at a time.)

Honny1:
@mtrmac, @giuseppe PTAL,  I've updated the prototype.

mtrmac (commented):
Thanks! Good progress, `TempDir` now looks safe, and the documentation is great.

For Podman service, and CRI-O, we do need layer deletion to happen _very soon_, not just at store shutdown. And, for those, we might also want the recovery to happen much more frequently (e.g. if someone runs `podman rmi` as an one-time operation on an OpenShift node, that crashes, and the only other c/storage user is a long-running CRI-O process).

That might, OTOH, make the serialization implied by `tempDirGlobalLock` more frequent = more costly and less desirable, again. I don’t think we absolutely _need_ that lock?
- Split instance state and `TempDir` state. Then there will be no need to do any locking in `TempDir` itself, for in-process purposes, it would just contain read-only fields. And we can make the instance single-goroutine-owned = again, no need for in-process locks.
- Move `instanceLock` out of the instance directory, and change lifetimes so that the lock is created before, and deleted after, the instance directory. Then I think we don’t need `tempDirGlobalLock` for cross-process purposes, either (? apart from coordinating recovery).

cgwalters:
I only tangentially skimmed this PR but just in case of interest, from the very start ostree has implemented a vaguely similar scheme in https://github.com/ostreedev/ostree/blob/7a32b86ee4c0232bfde97a3cd87c4a82b6e73218/src/libostree/ostree-repo.c#L6292
An important concern for ostree is keeping track of which files have been `fsync()`'d and to allow resuming.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,259 @@
...
+	rootDir/
+	  .global-tmp-lock  (tempDirGlobalLock)
+	  temp-dir-ABC/
+	    .lock           (instanceLock for temp-dir-ABC)
+	    file1
+	    file3
+	  temp-dir-XYZ/
+	    .lock           (instanceLock for temp-dir-XYZ)
+	    file2
+*/
```
giuseppe, line 73: https://github.com/containers/storage/pull/2325#discussion_r2074979127 still holds.

This structure is subject to a race condition when the `.lock` file is deleted before the other files in the directory.

So you have process A that holds the lock while it runs `rm -rf temp-dir-XYZ`, what a new process B should do when it attempts to lock the directory but doesn't find the lock file?  1: Assume the directory is being deleted (process A could crash), or 2: delete it as well?

I think it is safer to have:

temp-dir-XZY/data/
temp-dir-XZY/.lock

first `temp-dir-XZY/data/` and if that succeeded proceed to remove `temp-dir-XZY`.

When process B tries to lock the directory, it will find the lock file until `data/` exists, and if it doesn't find `lock` then it can do `rmdir("temp-dir-XZY")` because that could happen if the process A crashed after removing the lock file but before removing the directory (or after creating the directory but before the lock file was created)

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,259 @@
...
+	// instanceLock is a lock file (e.g., rootDir/temp-dir-XYZ/.lock) specific to this
+	// TempDir instance, indicating it's in active use.
+	instanceLock *lockfile.LockFile
+}
+
+// listPotentialStaleDirs scans the rootDir for directories that might be stale temporary directories.
+// It acquires and releases the tempDirGlobalLock to ensure safe concurrent access to the directory listing.
+func (td *TempDir) listPotentialStaleDirs(rootDirPath string) ([]string, error) {
+	td.tempDirGlobalLock.Lock()
+	defer td.tempDirGlobalLock.Unlock()
```
giuseppe, line 101: we don't need the lock here since it is released when the function exits, so for the caller it doesn't matter whether the code below runs with the lock or not.

As @mtrmac pointed out, we need to use `os.ReadDir` not Walk it recursively

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("getting global temp dir lock file %s: %w", globalLockPath, err)
+	}
+
+	td := &TempDir{
+		rootDir:           rootDir,
+		tempDirGlobalLock: tempDirGlobalLock,
+	}
+
+	if err := td.recoverStaleDirs(rootDir); err != nil {
```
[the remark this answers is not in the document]
giuseppe: in what cases do you think it would be better to check it multiple times?  Another process crashing while the current one runs?  Would it matter except long running servers (e.g. `podman system service`)?

giuseppe (changes requested):

pkg/tempdir/tempdir.go
giuseppe: what if the content we are trying to move to the temp-dir has a file/directory named `.lock`?

mtrmac: > This structure is subject to a race condition when the .lock file is deleted before the other files in the directory.

I think that part is good enough — things like `os.RemoveAll` really shouldn’t crash if a parent is removed in the meantime.

The problem with moving the lock inside the instance directory is at _creation_ time, where we have the instance directory without any lock, so we need to protect against cleanup/recovery immediately deleting it. The current implementation does solve that, by holding the global lock — but that also means that the cleanup/recovery needs to obtain the global lock every time just to check whether anything needs to be cleaned up.

Together with https://github.com/containers/storage/pull/2325#pullrequestreview-2837512226 arguing that we should do the cleanups frequently, not just at process shutdown, that implies obtaining an extra lock file on both `layerStore.startReading` and `layerStore.startWriting`. I’d rather prefer to avoid that cost. (I’d even like to avoid the `Readdir` on every `start…ing`, but that will require more cooperation with `layerStore`.)

pkg/tempdir/tempdir.go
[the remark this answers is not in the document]
mtrmac: Yes, long-running servers. Consider CRI-O, if an admin manually runs `podman rmi` on the node, and the `rmi` is interrupted.

I was defaulting to thinking that doing this on every store lock/unlock, just like we currently do for layers’ `incompleteFlag`, makes sense. On second thought, it probably wouldn’t be too bad to only do that (within a process) every 10 seconds or so, _and_ on shutdown; then efficiency would not matter _that_ much. 

---

Also… at least Skopeo never calls `store.Shutdown`; the store’s documentation doesn’t say it’s required (and the operation documentation reads scary enough that I now spent some time checking that it actually happens, outside of a `podman system reset`).

drivers/overlay/overlay.go
```diff
@@ -2027,7 +2042,7 @@ func (d *Driver) ListLayers() ([]string, error) {
 	for _, entry := range entries {
 		id := entry.Name()
 		switch id {
-		case linkDir, stagingDir, quota.BackingFsBlockDeviceLink, mountProgramFlagFile:
+		case linkDir, stagingDir, tempDir, quota.BackingFsBlockDeviceLink, mountProgramFlagFile:
```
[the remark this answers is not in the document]
mtrmac: > I'm not sure, but I moved it to `stagingDir`.

That might make things _worse_, see it would interfere with `pruneStagingDirectories`.

---

> But I'm not sure how it is with `recreateSymlinks()`. I think it should ignore `stagingDir`.

Yes. It probably works well enough now, because `$stagingDir/link` doesn’t exist, so the directory is ignored - but it seems attractive to clean this up, going forward.

The very last resort which always works would be to leave `d.home` for layer directories, and have the caller of the driver provide a separate `d.metadataHome` where we have no prior commitments.

But, maybe, the answer is that we can leave `tempDir` in `d.home`, as long as `$tempDir/link` is never created?? And should we _now_ at least reserve a new path prefix, e.g. defining that `${d.home}/meta*` is never a layer directory?

Or, alternatively, we could define a storage version number, and refuse to work with future versions. That would be drastic but clearly correct, OTOH we would be signing up for very precisely managing this to keep CRI-O and Podman in sync (and for not forgetting to update the version number…)

All of this forward-compatibility design seems valuable, but also somewhat out of scope of this “unlocked deletes” effort — or at least clearly out of scope of this prototype.

Cc: @giuseppe

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+		out = append(out, staleDir{
+			tempDirPath: path,
+			lockPath:    lockPath,
+		})
+	}
+
+	for path := range locks {
+		out = append(out, staleDir{
+			tempDirPath: "",
```
mtrmac, line 148: (I’m not sure we need to specifically detect “have data, missing lock” vs. “have lock, missing data” situations here. If there is no global lock, `Recover…` will need to handle situations where `list…` did find a file but it went missing afterwards, anyway.

So just returning a list of instances (by the varying ID??) might be sufficient.)

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+
+	if len(potentialStaleDirs) == 0 {
+		return nil
+	}
+
+	var recoveryErrors []error
+
+	for _, staleDir := range potentialStaleDirs {
+		if staleDir.lockPath != "" {
+			instanceLock, err := lockfile.GetLockFile(staleDir.lockPath)
```
mtrmac, line 174: `ErrNoExist` here is expected — we raced against the owner cleaning up.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+			recoveryErrors = append(recoveryErrors, fmt.Errorf("error getting lock file %s: %w", staleDir.lockPath, err))
+			if err := instanceLock.TryLock(); err != nil {
```
mtrmac, line 176: ⚠️ A problem for the no-global-lock design:

The owner of a new instance creates the lock first, and _then_ locks it; `Recover` can find the lock in the meantime, lock it, and start deleting things.

That’s fixable in the instance creation, see there.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+				logrus.Debugf("Skipping cleanup for locked directory corresponding to %s", staleDir.lockPath)
+				continue
+			}
+		}
+		if staleDir.tempDirPath != "" {
```
mtrmac, line 181: This is racy:

- Instance owner: create lock
- Recovery: `list…` finds a lock and no data dir
- Instance owner: create data dir
- Instance owner: crash
- Recovery: obtain lock
- Recovery: does not try to delete data

`list…` was called with no locks held, so it’s directory state is not authoritative. Only after `TryLock` succeeds we are the “owner” of the data, and we can check whether the data exists / try to delete it.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+			if rmErr := os.RemoveAll(staleDir.tempDirPath); rmErr != nil && !os.IsNotExist(rmErr) {
+				recoveryErrors = append(recoveryErrors, fmt.Errorf("error removing stale temp dir %s: %w", staleDir.tempDirPath, rmErr))
+			}
+		}
+
+		if staleDir.lockPath != "" {
+			if err := os.Remove(staleDir.lockPath); err != nil && !os.IsNotExist(err) {
```
mtrmac, line 188: Unlock the lock (first?), otherwise we are leaking `instanceLock`, including a file descriptor.

---

⚠️ actually, `pkg/lockfile.lockFiles` is a per-process global state (necessary due to semantics of the locks :| ) and we currently never remove entries from that map. So, repeated creation of new instance locks (incl. repeated `Recover…` calls just _listing_ locks owned by other processes and calling `TryLock` on them!) would leak memory.

The same applies to overlay’s `stagingDirBase`.

But we probably do need the `lockFiles` per-process map, because we can easily have two separate goroutines in a single process trying to access the same locks, and IIRC two separate `layerStore` objects over the same directory, due to the `modified` code path in `startUsingGraphDriver`, and closing _any_ file descriptor unlocks the locks, so we must have a global map.

My head hurts… I guess we would need something different from the current global `lockFiles` map = an entirely separate alternative to` pkg/lockfile.LockFile`, with operations “TryLockPath” and “UnlockAndRemovePath”, which are both happening under a process-global lock (and for that, we want to only support `TryLock` and not a blocking `Lock`), so that we can safely remove things from the lock file map?

Cc: @giuseppe

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+// It acquires the tempDirGlobalLock during its operation to ensure atomicity
+// with other temp dir management operations.
+// The caller should ensure .Cleanup() is called eventually.
+func (td *TempDir) initInstanceTempDir() error {
+	if td.tempDirPath != "" && td.tempDirLock != nil {
+		logrus.Debugf("Temp dir %s already initialized", td.tempDirPath)
+		return nil
+	}
+	randomSuffix := stringid.TruncateID(stringid.GenerateRandomID())
+	tempDirName := TempDirPrefix + randomSuffix
```
mtrmac, line 207: We would still need to worry about collisions, and probably retry…

I think using `os.CreateTemp` to decide on the lock file path might work well. But, also, see the “leaking memory for staging locks” discussion.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+
+	tempDirLockPath := filepath.Join(td.rootDir, TempdirLockPrefix+tempDirName)
+	actualTempDirPath := filepath.Join(td.rootDir, tempDirName)
+
+	tempDirLock, err := lockfile.GetLockFile(tempDirLockPath)
+	if err != nil {
+		return fmt.Errorf("getting instance lock file %s for new temp dir %s: %w", tempDirLockPath, actualTempDirPath, err)
+	}
+
```
mtrmac, line 216: At _this_ point, `Recover` can find the lock, lock it, and trigger cleanup.

In that case, currently `TryLock` fails (… but that’s guaranteed _only_ when `Recover` never unlocks the lock, which we need to fix, see elsewhere), so we can detect that and we know we have to retry.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+	if err := tempDirLock.TryLock(); err != nil {
+		if errRm := os.RemoveAll(actualTempDirPath); errRm != nil && !os.IsNotExist(errRm) {
```
mtrmac, line 218: We didn’t create anything there yet, no need to do this.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+// The caller MUST ensure .Cleanup() is called.
+func (td *TempDir) Add(path string) error {
+	if td.tempDirLock == nil {
+		if err := td.initInstanceTempDir(); err != nil {
+			return fmt.Errorf("initializing instance temp dir failed: %w", err)
+		}
+	}
+	td.tempDirLock.AssertLocked()
+
+	fileName := time.Now().Format("20060102-150405") + "-" + filepath.Base(path)
```
mtrmac, line 267: One second-level timestamps don’t guarantee uniqueness, computers are fast enough.

If this should manage unique names, maybe just increment a (per-instance) counter?

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	td.tempDirPath = ""
+	td.tempDirLock = nil
+	td.tempDirLockPath = ""
+	return nil
+}
+
+// Path returns the path to the instance-specific temporary directory.
+// It returns an empty string if the directory has not been initialized or has been cleaned up.
+func (td *TempDir) Path() string {
```
mtrmac, line 304: This has no users and can be removed.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	return out, err
+}
+
+// RecoverStaleDirs identifies and removes stale temporary directories in the root directory.
+// A directory is considered stale if its lock file can be acquired (indicating no active use).
+// The function attempts to remove both the directory and its lock file.
+// Any errors encountered during the recovery process are collected and returned as a single error.
+func (td *TempDir) RecoverStaleDirs() error {
```
mtrmac, line 160: This is not related to the mutable (“instance”) fields of the object.

I’d truly find it much cleaner to have a separate handling of the “temp dir root”, as one object, and “instance”, as another. (Maybe we don’t need a root _object_, and just pass a root path. Anyway.) The two have different lifetimes, and different concurrency regimes.

mtrmac (commented):
Ratholes inside ratholes …

If I’m reading things right, using `pkg/lockfile.LockFile` for staging directories would leak memory? Cc: @giuseppe to keep me honest.

If so, we need a new idea, or a new primitive. I was thinking a new `internal/staginglock` that supports
- `CreateAndTryLock`
- `TryLockExistingPath`
- `UnlockAndDelete`

without an ever-growing lock map. If so, can we do that in yet another separate PR, fix overlay’s `stagingDirBase`, and then return to this one?

But maybe there’s some much simpler approach I’m now missing.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,306 @@
...
+  - If directory is not found, the lock file is considered stale and removed.
+
+3. TempDir Usage:
+  - TempDir.Add() requires the instance lock to be held (enforced by AssertLocked).
+  - If the instance's temporary directory hasn't been initialized yet, Add will
+    call initInstanceTempDir to create and lock it.
+  - Files moved into the temporary directory are renamed with a timestamp prefix
+    to ensure uniqueness.
+  - If TempDir was cleaned up, Add() will reinitialize the new instance lock and
+    new temporary directory.
```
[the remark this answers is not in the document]
Honny1: I was referring to the same `temp-dir-XYZ` directory. Because it was removed. I should rephrase that.

Honny1:
I've edited the `RecoverStaleDirs`, `initInstanceTempDir`, and `Cleanup` so they're less racy and marked racy parts. I think that would be solved with `internal/staginglock`.

Honny1:
I would stick to this layout of the root Directory if `internal/staginglock` would solve racy parts:
```
	RootDir/
	    lock-<ID>          (instance lock for temp-dir-<ID>)
	    temp-dir-<ID>/
```

The `<ID>` is generated by `os.CreateTemp`.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+		// ----
+
+		if rmErr := os.RemoveAll(tempDirPath); rmErr != nil && !os.IsNotExist(rmErr) {
+			recoveryErrors = append(recoveryErrors, fmt.Errorf("error removing stale temp dir %s: %w", tempDirPath, rmErr))
+		}
+
+		// --- This is racy: Needs atomic unlock and remove.
+		instanceLock.Unlock()
+		if err := os.Remove(lockPath); err != nil && !os.IsNotExist(err) {
```
mtrmac, line 165: (I don’t think this is racy _as is_: we can `os.Remove()` and then `instanceLock.Unlock()`, and I think that works (well, on Unix or recent Windows where removing open files is allowed). Alternatively, we _can_ unlock + remove — future recovery calls will find the lock, get it, and remove it. Am I missing a race?)

The problem I’m thinking about is that to avoid leaks, we want to remove the lock file entry from `pkg/lockfile.lockFiles`, and _that_ would race against concurrent `GetLockFile`.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,286 @@
...
+// It acquires the tempDirGlobalLock during its operation to ensure atomicity
+// with other temp dir management operations.
+// The caller should ensure .Cleanup() is called eventually.
+func (td *TempDir) initInstanceTempDir() error {
+	if td.tempDirPath != "" && td.tempDirLock != nil {
+		logrus.Debugf("Temp dir %s already initialized", td.tempDirPath)
+		return nil
+	}
+
+	// ---- This is racy: Needs Atomic way to create uncinate file and lock it.
```
mtrmac, line 185: (I don’t know whether we can do that easily … we would need to create the lock file elsewhere, lock it, and then `rename()` it into the destination to “publish” it when it is already locked. Maybe with something modern like `O_TMPFILE`. _And_ it would need to integrate with a “generate a random name” functionality similar to `CreateTemp`.

I think using `CreateTemp`+`TryLock` — and if we lose the race to a recovery, abandoning that ID and retrying the `CreateTemp` — would work well enough, and it would be less expensive in the typical no-contention case.)

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,286 @@
...
+	RootDir string
+
+	tempDirPath string
+	// tempDirLock is a lock file (e.g., RootDir/lock-XYZ) specific to this
+	// TempDir instance, indicating it's in active use.
+	tempDirLock     *lockfile.LockFile
+	tempDirLockPath string
+
+	// Required tempDirLock to be held when accessing counter.
+	counter uint64
```
mtrmac, line 99: Speculating about users: All of the mutable fields (everything but `RootDir`) assumes that this object is only is only accessed by one goroutine at a time — even if the process as a whole holds the instance lock.

E.g. if `layerStore` held a `TempDir` instance, it would need to be accessed _only_ when holding a RW layer lock, or some other lock.

That’s almost certainly fine for the layer store’s delete operation. For the graph drivers, probably fine? Just something to document and verify.

Or, alternatively, `layerStore.Delete` might create its own private `TempDir[Instance]` object, with no need to synchronize with other goroutines. That might actually be cleaner.

pkg/tempdir/tempdir.go
Honny1: > I think using `CreateTemp`+`TryLock` — and if we lose the race to a recovery, abandoning that ID and retrying the `CreateTemp` — would work well enough, and it would be less expensive in the typical no-contention case.)

I wasn't sure if that was the right solution. I'll implement it.

pkg/tempdir/tempdir.go
Honny1: > Or, alternatively, `layerStore.Delete` might create its own private `TempDir[Instance]` object, with no need to synchronize with other goroutines. That might actually be cleaner.

I've tried to implement this, but I'm not sure how much I can change the API of `driver` or `layerStore`.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,296 @@
...
+		// Reserve name for lock file in the RootDir.
+		tempLockFile, err := os.CreateTemp(td.RootDir, TempdirLockPrefix)
+		if err != nil {
+			return fmt.Errorf("creating temp directory in %s failed: %w", td.RootDir, err)
+		}
+		tempLockFile.Close()
+
+		tempDirLockPath := tempLockFile.Name()
+		tempDirLock, err := lockfile.GetLockFile(tempDirLockPath)
+		if err != nil {
```
mtrmac, line 205: This might fail with `ENOENT` if a concurrent recovery found, locked, and deleted the lock. In that case, retry.

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,296 @@
...
+	RootDir/
+	    lock-ABC           (instance lock for temp-dir-ABC)
+	    temp-dir-ABC/
+	        0-file1
+	        1-file3
+	    lock-XYZ           (instance lock for temp-dir-XYZ)
+	    temp-dir-XYZ/
+	        0-file2
+*/
+const (
```
mtrmac, line 77: (These can all be private.)

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,296 @@
...
+			continue
+		}
+
+		// Remove the directory and lock file.
+		if rmErr := os.RemoveAll(tempDirPath); rmErr != nil && !os.IsNotExist(rmErr) {
+			recoveryErrors = append(recoveryErrors, fmt.Errorf("error removing stale temp dir %s: %w", tempDirPath, rmErr))
+		}
+
+		defer instanceLock.Unlock()
+		if err := os.Remove(lockPath); err != nil && !os.IsNotExist(err) {
```
mtrmac, line 174: (Remove and then unlock, preferably, so that other recoveries ignore this ID. Also in `Cleanup`. But this will change with the new locks anyway.)

pkg/tempdir/tempdir.go
```diff
@@ -0,0 +1,296 @@
...
+// If the instance's temporary directory doesn't exist yet, Add will create and lock it.
+// Files are renamed with a counter-based prefix (e.g., "0-filename", "1-filename") to ensure uniqueness.
+// Note: 'path' must be on the same filesystem as the TempDir for os.Rename to work.
+// The caller MUST ensure .Cleanup() is called.
+func (td *TempDir) Add(path string) error {
+	if td.tempDirLock == nil {
+		if err := td.initInstanceTempDir(); err != nil {
+			return fmt.Errorf("initializing instance temp dir failed: %w", err)
+		}
+	}
```
mtrmac, line 261: (If `NewTempDir` is called only by code that intends to delete something, this initialization can happen in there, as an ~ordinary constructor. But let’s settle the API shape first.)

store.go
```diff
@@ -2550,6 +2551,14 @@ func (s *store) Lookup(name string) (string, error) {
 }
 
 func (s *store) DeleteLayer(id string) error {
+	var cleanUp tempdir.CleanupTempDirFunc
+	defer func() {
+		if cleanUp != nil {
+			if err := cleanUp(); err != nil {
+				logrus.Errorf("Error cleaning up after deleting layer %q: %v", id, err)
```
mtrmac, line 2558: - This error should definitely be reported to the caller.
- (The `Error` prefix will probably be unnecessary if this is wrapped.)
- Probably `%w`

layers.go
```diff
@@ -1158,7 +1174,8 @@ func (s *store) newLayerStore(rundir, layerdir, imagedir string, driver drivers.
 			layersImageDir,
 			filepath.Join(volatileDir, "volatile-layers.json"),
 		},
-		layerdir: layerdir,
+		layerdir:        layerdir,
+		tempDirRootPath: s.graphRoot,
```
mtrmac, line 1178: Is this correct? Shouldn’t that be a dedicated directory?
Also, it needs to precisely match the directories of data which we are deleting (like `tspath`/`datadir`). `graphRoot` seems to amount to the correct filesystem now, with the current only caller, but tying it more precisely to the `layerStore` logic would be easier to maintain.

mtrmac (commented):
Overall, the “successful deletion” path now looks pretty good.

I’m not sure about the cleanup locations / timing / locking yet, but, anyway, we need a reliable staging lock as a first step.

Honny1:
I have integrated the staging lock into the temporary directory, added tests, and addressed feedback.

layers.go
```diff
@@ -2006,12 +2042,42 @@ func (r *layerStore) Delete(id string) error {
 			return err
 		}
 	}
-	if err := r.deleteInternal(id); err != nil {
-		return err
+	cleanFunc, err := r.deleteInternal(id)
+	rmErr := cleanFunc()
```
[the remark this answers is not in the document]
mtrmac: > For now, maybe just renaming it to an explicit `deleteWhileHoldingLock` could be enough.

Here I meant that the current `layerStore.Delete` (the one that does deferred + immediately calls the cleanups = the slow one) would be renamed to `deleteWhileHoldingLock`, to emphasize that it’s not ideal and we should migrate away from it.

mtrmac (commented):
A very quick first skim: I think the feature set is fine, and the safety is almost certainly fine as well (relying on the previous PR).

I think the recovery timing is not ideal / consistent: after we recover in `layerStore.load` (assuming, FIXME: double-check, that this happens after ~every time we obtain the RW lock), `layerStore.delete[DeferringTheCleanup]` should not need to trigger recovery again — we are holding the lock, so there should be nothing new that could possibly be recovered.

So, I’m thinking `NewTempDir` does not need to trigger a recovery. [And then `NewTempDir`+`initInstanceTempDir` could be ~merged because the object is created only immediately before adding items. This is not essential.]

OTOH we don’t currently have a recovery trigger when obtaining a lock in the graph drivers. So that’s inconsistent.

I’m leaning towards thinking that the `layerStore` model, where recovery is an explicit call, and `NewTempDir` (imitating a deletion) does not do that, is cleaner — OTOH we would need to trigger the recovery frequently enough (on every RW lock?) and I’m not absolutely sure we can stomach the cost of doing that so often.

Honny1:
I changed `NewTempDir` so that it doesn't call recovery, and merged it with `initInstanceTempDir`. I also implemented triggering recovery on every RW lock.

Honny1:
PTAL @mtrmac

drivers/overlay/overlay.go
```diff
@@ -1332,6 +1334,35 @@ func (d *Driver) Remove(id string) error {
 	return nil
 }
 
+func (d *Driver) DeferredRemove(id string) (tempdir.CleanupTempDirFunc, error) {
```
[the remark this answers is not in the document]
giuseppe: A function call is not not serializable, so how will we call the CleanupTempDirFunc from another process?

The idea is that we can run the cleanup function in a different process.

drivers/overlay/overlay.go
```diff
@@ -1353,8 +1379,8 @@ func (d *Driver) recreateSymlinks() error {
 		// Check that for each layer, there's a link in "l" with the name in
 		// the layer's "link" file that points to the layer's "diff" directory.
 		for _, dir := range dirs {
-			// Skip over the linkDir and anything that is not a directory
-			if dir.Name() == linkDir || !dir.IsDir() {
+			// Skip over the linkDir and anything that is not a directory or tempDir
```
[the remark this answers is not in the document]
mtrmac: … but see elsewhere first, the directory structure probably needs to be changed anyway.

layers.go
```diff
@@ -1988,10 +2037,20 @@ func (r *layerStore) deleteInDigestMap(id string) {
...
-func (r *layerStore) Delete(id string) error {
+// This is soft-deprecated and should not have any new callers; use deferredDelete instead.
+func (r *layerStore) deleteWhileHoldingLock(id string) error {
+	cleanupFunctions, deferErr := r.deferredDelete(id)
+	cleanupErr := tempdir.CleanupTemporaryDirectories(cleanupFunctions...)
+	return errors.Join(deferErr, cleanupErr)
+}
+
+// Requires startWriting.
+// Caller MUST run all returned cleanup functions after this. Ideally outside of
```
mtrmac, line 2048: > // Caller MUST run all returned cleanup functions after this.

Reading how we nave `cleanFunctions` set, and then we can fail in `saveFor`, I think this should explicitly document “EVEN IF the function returns an error”. Some of the callers already do this, I didn’t check whether all do.

mtrmac (commented):
Just a brief look, I didn’t review the driver code changes.

drivers/driver.go
```diff
@@ -123,7 +124,17 @@ type ProtoDriver interface {
 	// and parent, with contents identical to the specified template layer.
 	CreateFromTemplate(id, template string, templateIDMappings *idtools.IDMappings, parent string, parentIDMappings *idtools.IDMappings, opts *CreateOpts, readWrite bool) error
 	// Remove attempts to remove the filesystem layer with this id.
+	// This is soft-deprecated and should not get any new callers; use DeferredRemove.
 	Remove(id string) error
+	// DeferredRemove is used to remove the filesystem layer with this id.
+	// This removal happen immediately (the layer is no longer usable),
+	// but physically deleting the files may be deferred.
+	// Caller MUST call returned Cleanup function.
```
mtrmac, line 132: This also has the …EVEN [on error]  design.

drivers/vfs/driver.go
```diff
@@ -244,6 +248,41 @@ func (d *Driver) Remove(id string) error {
 	return system.EnsureRemoveAll(d.dir(id))
 }
 
+func (d *Driver) GetTempDirRootDirs() []string {
+	tempDirs := []string{filepath.Join(d.home, tempDirName)}
+	// VFS can have layers in different locations through additionalHomes
+	// Add temp directories for each additional home
+	for _, additionalHome := range d.additionalHomes {
+		tempDirs = append(tempDirs, filepath.Join(additionalHome, tempDirName))
+	}
```
mtrmac, line 257: It’s very non-obvious looking at the driver _only_, but this is not the right set of stores; I think (@giuseppe please keep me honest) only `d.home` and `d.imageStore` are relevant.

Yes, `d.dir2` can potentially return any of `d.additionalHomes`, but if you look at the top-level `store.DeleteLayer`, that only works on layers which can be found via the RW primary `store.layerStoreUseGetters` — and that, ultimately, refers to the graph driver’s `drivers.Options.{Root,ImageStore}` only, not the other additional image stores.

drivers/vfs/driver.go
```diff
@@ -244,6 +248,41 @@ func (d *Driver) Remove(id string) error {
...
+	return tempDirs
+}
+
+func (d *Driver) DeferredRemove(id string) (tempdir.CleanupTempDirFunc, error) {
+	layerDir := d.dir(id)
+	tempDirRoot := ""
+	tempDirRoots := d.GetTempDirRootDirs()
+	for _, potentialTempDirRoot := range tempDirRoots {
+		homeDir := filepath.Dir(potentialTempDirRoot)
+		if strings.HasPrefix(layerDir, homeDir) {
```
mtrmac, line 267: `HasPrefix` is _very_ imprecise for paths.

Two possible ways to model this:
- `d.dir2` knows exactly where it found the data, so it can return the “top-level home” it found the layer in.
- This function could hard-code the knowledge that layers exist in `${home}/dir/${id}`.

drivers/overlay/overlay.go
```diff
@@ -1327,6 +1334,46 @@ func (d *Driver) Remove(id string) error {
 	return nil
 }
 
+func (d *Driver) GetTempDirRootDirs() []string {
+	tempDirs := []string{filepath.Join(d.home, tempDirName)}
+	// If imageStore is configured and different from home, add it as well
+	// since layers can be stored in either location
+	if d.options.imageStores != nil {
```
mtrmac, line 1341: The VFS comments apply similarly to overlay.

drivers/overlay/overlay.go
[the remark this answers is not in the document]
mtrmac: Explicitly skipping `tempDirName` here would cut down on the noise.

mtrmac (commented):
I’m afraid the additional image store implementation is confusing , the way it is mixed over `Store` and drivers. (Exhibit N+1 why I refuse to work on adding more options to graph drivers.)

Honny1:
@mtrmac PTAL

giuseppe (approved):
LGTM

drivers/overlay/overlay.go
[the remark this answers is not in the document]
mtrmac: @giuseppe With this, old versions’ `GarbageCollect` == `podman system prune --external` can delete all of `tempDir`.

I think that’s good enough, it’s supposed to be run at boot at quiet state, but I’m highlighting this in case I’m missing something.

drivers/vfs/driver.go
```diff
@@ -244,6 +248,37 @@ func (d *Driver) Remove(id string) error {
 	return system.EnsureRemoveAll(d.dir(id))
 }
 
+func (d *Driver) GetTempDirRootDirs() []string {
+	tempDirs := []string{filepath.Join(d.home, tempDirName)}
+	// Only include imageStore if it's different from home
+	// Writable layers can only be in d.home or d.imageStore, not in additionalHomes (which are read-only)
+	if d.imageStore != "" && d.imageStore != d.home {
```
mtrmac, line 255: The `d.imageStore != d.home` condition seems unnecessary, and inconsistent with other users of `d.imageStore`.

 `loadStoreOptionsFromConfFile` already refuses options of ~that kind; but, also, `d.imageStore == d.home` is not _exactly_ the right condition _anyway_, because the layouts are different (the `d.String()` part).

Also in `DeferredRemove`.

drivers/overlay/overlay.go
```diff
@@ -1327,6 +1334,31 @@ func (d *Driver) Remove(id string) error {
 	return nil
 }
 
+func (d *Driver) GetTempDirRootDirs() []string {
+	tempDirs := []string{filepath.Join(d.home, tempDirName)}
+	// Only include imageStore if it's different from home
+	// Writable layers can only be in d.home or d.imageStore, not in additional image stores
+	if d.imageStore != "" && d.imageStore != d.home {
```
mtrmac, line 1341: Same as in VFS, the `!= d.home` condition seems unnecessary.

drivers/overlay/overlay.go
```diff
@@ -1327,6 +1334,31 @@ func (d *Driver) Remove(id string) error {
...
+		tempDirs = append(tempDirs, filepath.Join(d.imageStore, d.name, tempDirName))
+	}
+	return tempDirs
+}
+
+func (d *Driver) DeferredRemove(id string) (tempdir.CleanupTempDirFunc, error) {
+	_, homeDir, _ := d.dir2(id, false)
```
mtrmac, line 1348: I think this `homeDir` is [undocumented :( and] “where to create layers”; it does _not_, AFAICS, consistently match the directory where `id` is _found_ — so this does not return the right directory.

---

Really I think having `dir2`, in both graph drivers, return the parent(?) directory would be cleaner (centralizing knowledge/policy, and allowing to clearly track where data comes from) than hard-coding any knowledge in the `DeferredRemove` callers.

Admittedly, yes, that would imply documenting what `dir2` does, and that’s a bit of a puzzle. But, as this PR shows, that would be valuable.

Honny1: The rework and documentation of `dir2` should be done in another PR. Should I create an issue for this?

mtrmac (commented):
LGTM, thanks! A culmination of a lot of great work.

@giuseppe PTAL.

drivers/overlay/overlay.go
mtrmac: I think that’s ultimately up to you.

I think after this merges you are likely to be pinged if this behaves in a surprising way, and _right now_ you understand the code better than most people do most of the time. So, I think it would help your future self to document it very soon. But that’s a guess about the future, I don’t think I’m 100% justified in _requiring_ you to do that investment now.

giuseppe (approved):
thanks for the great work!

/lgtm

PR #2328, opened 2025-05-06
Fix nits from #2312

mtrmac:
Addresses most of https://github.com/containers/storage/pull/2312#pullrequestreview-2819433654 . See individual commit messages for details.

### DIFF
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -323,7 +323,7 @@ func ensureTOCMatchesTarSplit(toc *minimal.TOC, tarSplit *os.File) error {
 		}
 	}
 
-	if _, err := tarSplit.Seek(0, 0); err != nil {
+	if _, err := tarSplit.Seek(0, io.SeekStart); err != nil {
 		return err
 	}
 
--- a/pkg/chunked/compression_linux.go
+++ b/pkg/chunked/compression_linux.go
@@ -182,8 +182,8 @@ func openTmpFileNoTmpFile(tmpDir string) (*os.File, error) {
 // readZstdChunkedManifest reads the zstd:chunked manifest from the seekable stream blobStream.
 // tmpDir is a directory where the tar-split temporary file is written to.  The file is opened with
 // O_TMPFILE so that it is automatically removed when it is closed.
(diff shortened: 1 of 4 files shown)


pkg/chunked/compression_linux.go
```diff
@@ -160,9 +160,9 @@ func readEstargzChunkedManifest(blobStream ImageSourceSeekable, blobSize int64,
 }
 
 func openTmpFile(tmpDir string) (*os.File, error) {
-	file, err := os.OpenFile(tmpDir, unix.O_TMPFILE|unix.O_RDWR|unix.O_CLOEXEC|unix.O_EXCL, 0o600)
+	fd, err := unix.Open(tmpDir, unix.O_TMPFILE|unix.O_RDWR|unix.O_CLOEXEC|unix.O_EXCL, 0o600)
```
kolyshkin, line 163: Back when reviewing the original commit I checked that
 - os.OpenFile flags argument type is `int` (i.e. a generic one);
 - it is passed almost as is onto syscall.Open ("almost" since O_LARGEFILE and O_CLOEXEC are added);
 - `os.O_*`, `syscall.O_*`, and `unix.O_*` flag values should be equal (although there is no guarantee);
 - a lot of existing code uses `unix.O_*` flags with `os.OpenFile`;
 
and so I guesses it's OK as it is.

With this change, though
 - we lose the ability to retry on EINTR;
 - as far as I understand, `os.OpenFile` adds the file to the internal poller, while the new code doesn't (which probably doesn't matter here).
 
 All in all, I slightly prefer the old way.

giuseppe (commented):
LGTM after the `Don't mix os.OpenFile with x/sys/unix.O_* flags` patch is dropped

mtrmac:
> LGTM after the `Don't mix os.OpenFile with x/sys/unix.O_* flags` patch is dropped

I continue to think the mixing is incorrect, but there are several other pre-existing instances in the repo, and there is a unit test, so, fair enough. Dropped.

mtrmac:
Rebased, PTAL.

kolyshkin (approved):
LGTM

PR #2336, opened 2025-05-27
Fix: Introduce `StagingLockfile` to resolve overlay staging lock memory leak

Honny1:
This PR introduces `StagingLockfile`, a new internal lockfile designed to manage locking for temporary directories more effectively and to resolve a memory leak. The leak was caused by pruned lockfiles from overlay staging remaining in the global cache.

Key methods provided by `StagingLockfile`:

* `CreateAndLock`: Creates and locks a new unique file.
* `TryLockPath`: Attempts to lock an existing file. If the file does not exist, it will be created.
* `UnlockAndDelete`: Unlocks and deletes the file (fixes the memory leak).

An internal `filelock` component has been introduced to encapsulate common primitives for lock file operations. This allows `StagingLockfile` and `Lockfile` to share core locking mechanisms, reducing code duplication.


mtrmac:
> I think it’s very likely that the staging lock package doesn’t need all of the existing operations (no need for blocking `Lock` or plain `Unlock`, and maybe more such opportunities?); along with removal of the recursive read locking capability, that could allow simplifying things.

As a guess, consider the following invariant
> Unless `globalMapMutex` is currently locked, an entry for a file exists in `globalMap` iff the current process currently holds the lock for that file.

I _think_ that might suffice to do everything we need, at the cost of 1 in-process mutex + file lock; but it’s very possible I have overlooked something.

internal/staging_lockfile/staging_lockfile.go
```diff
@@ -0,0 +1,293 @@
...
+	if !l.locked {
+		// Panic when unlocking an unlocked lock.  That's a violation
+		// of the lock semantics and will reveal such.
+		panic("calling Unlock on unlocked lock")
+	}
+
+	l.locked = false
+
+	if err := os.Remove(l.file); err != nil && !os.IsNotExist(err) {
+		panic(fmt.Errorf("removing lock file %q: %w", l.file, err))
```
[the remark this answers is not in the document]
giuseppe: why?  Panicing the process could leave the storage in a dirty state, while here we are just dealing with a failure to remove a file.  The worst thing that happens is that we leak a file, but that is leaked anyway even if we panic.

I think a panic makes sense only when there is a clear error in the way locks are handled, and prevent data corruption.  I don't think it makes sense in any other case

internal/staging_lockfile/staging_lockfile.go
```diff
@@ -0,0 +1,173 @@
...
+// StagingLockFile represents a file lock used to coordinate access to staging areas.
+// Typical usage is via CreateAndLock (for a new file) or TryLockExisting (for an existing file),
+// both of which return a StagingLockFile that must eventually be released with UnlockAndDelete.
+// This ensures that access to the staging file is properly synchronized both within and across processes.
+//
+// WARNING: This struct MUST NOT be created manually. Use the provided helper functions instead.
+type StagingLockFile struct {
+	// The following fields are only set when constructing *StagingLockFile, and must never be modified afterwards.
+	// They are safe to access without any other locking.
+	file string
```
mtrmac, line 21: Non-blocking:
```suggestion
	file string // Also the key in stagingLockFiles
```
or an equivalent documentation in for `stagingLockFiles` itself, wouldn’t hurt

internal/staging_lockfile/staging_lockfile.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/containers/storage/internal/rawfilelock"
+)
+
+// StagingLockFile represents a file lock used to coordinate access to staging areas.
+// Typical usage is via CreateAndLock (for a new file) or TryLockExisting (for an existing file),
+// both of which return a StagingLockFile that must eventually be released with UnlockAndDelete.
+// This ensures that access to the staging file is properly synchronized both within and across processes.
+//
+// WARNING: This struct MUST NOT be created manually. Use the provided helper functions instead.
+type StagingLockFile struct {
```
mtrmac, line 18: Something like
```suggestion
type StagingLockFile struct {
    // Locking invariant: If stagingLockFileLock is not locked, a StagingLockFile for a particular
    // path exists iff the current process currently owns the lock for that file, and it is recorded in stagingLockFiles.
```

to help future readers understand the design?

internal/staging_lockfile/staging_lockfile_test.go
```diff
@@ -0,0 +1,170 @@
...
+	l, path, err := CreateAndLock(t.TempDir(), "locktest")
+	require.NoError(t, err)
+	defer os.Remove(path)
+
+	l.AssertLocked()
+
+	_, err = TryLockPath(path)
+	require.Error(t, err)
+
+	require.NoError(t, l.UnlockAndDelete())
```
mtrmac, line 100: Also `require.Len(t, stagingLockFiles, 0)` after this, to ensure that the failing `TryLockPath` did not allocate anything.

internal/staging_lockfile/staging_lockfile.go
```diff
@@ -0,0 +1,173 @@
...
+		// Panic when unlocking an unlocked lock. That's a violation
+		// of the lock semantics and will reveal such.
+		panic("calling Unlock on unlocked lock")
+	}
+
+	delete(stagingLockFiles, l.file)
+	l.locked = false
+
+	defer func() {
+		rawfilelock.UnlockAndCloseHandle(l.fd)
```
mtrmac, line 110:
```suggestion
		// It’s important that this happens while we are still holding stagingLockFileLock, to ensure
		// that no other goroutine has l.file open = that this close is not unlocking the lock under any
		// other goroutine.
		rawfilelock.UnlockAndCloseHandle(l.fd)
```
per https://github.com/containers/storage/pull/2336#discussion_r2110139525 .

internal/staging_lockfile/staging_lockfile.go
```diff
@@ -0,0 +1,173 @@
...
+		l.file = ""
```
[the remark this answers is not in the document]
Honny1: I removed `locked`.

internal/staging_lockfile/staging_lockfile.go
```diff
@@ -0,0 +1,145 @@
...
+		file.Close()
+
+		l, err := tryAcquireLockForFile(file.Name())
+		if err != nil {
+			if try < maxRetries {
+				continue // Retry if the lock cannot be acquired
+			}
+			return nil, "", fmt.Errorf(
+				"failed to allocate lock in %q after %d attempts; last failure on %q: %w",
+				dir, try, l.file, err,
```
mtrmac, line 129: `if err != nil`, `l == nil` and this crashes.

internal/staging_lockfile/staging_lockfile.go
```diff
@@ -0,0 +1,173 @@
...
+		l.file = ""
+	}()
+	if err := os.Remove(l.file); err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	return nil
+}
+
+// CreateAndLock creates a new temporary file in the specified directory with the given pattern,
+// then creates and locks a StagingLockFile for it. The file is created using os.CreateTemp.
```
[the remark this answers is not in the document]
mtrmac: I was thinking that it should return _only_ the base name relative to `dir`.

But we can tune that in #2325 .

internal/staging_lockfile/staging_lockfile.go
```diff
@@ -0,0 +1,145 @@
...
+// that must eventually be released with UnlockAndDelete. This ensures that access
+// to the staging file is properly synchronized both within and across processes.
+//
+// WARNING: This struct MUST NOT be created manually. Use the provided helper functions instead.
+type StagingLockFile struct {
+	// Locking invariant: If stagingLockFileLock is not locked, a StagingLockFile for a particular
+	// path exists if the current process currently owns the lock for that file, and it is recorded in stagingLockFiles.
+	//
+	// The following fields are only set when constructing *StagingLockFile, and must never be modified afterwards.
+	// They are safe to access without any other locking.
```
mtrmac, line 23: This is incorrect:
- `file` is modified by `UnlockAndDelete`
- (Pedantic `fd` is not modified, but we close the handle, so it eventually becomes invalid.)

Actually nowadays we never access these fields from a goroutine that doesn’t own the lock (if we find an existing entry in `tryAcquireLockForFile`, we don’t even read it); so “these fields can only be accessed by the goroutine owning the lock” seems fine.

internal/staging_lockfile/staging_lockfile.go
```diff
@@ -0,0 +1,145 @@
...
+	file string // Also the key in stagingLockFiles
```
mtrmac, line 24: Now that the `""` special case (~consistently) exists, it should be noted here.

mtrmac (commented):
Close to done I think.

@giuseppe PTANL.

mtrmac (commented):
LGTM, nice work.

(Hoping for another review due to how critical this is.)

giuseppe (approved):
great work!

/lgtm

PR #2341, opened 2025-06-02
overlay: drop unused fsMagic from support checks

giuseppe:
The `fsMagic` parameter was not used within the
`checkAndRecordOverlaySupport` and `supportsOverlay` functions.

PR #2342, opened 2025-06-02
Fix `git-validation` on PRs behind `main`

mtrmac:
`git-validation` fails on this repo unless the PR is rebased because the `CIRRUS_BASE_SHA` commit is not fetched into the build environment by default, so fetch it.

Also update `git-validation` from a 6-year-old version, although that’s probably not strictly necessary.


mtrmac:
https://github.com/containers/storage/pull/2342/commits/2c354c06a5a319d7a6cd912cb36ac1495eb4636c fails …

tests/tools/go.mod
```diff
@@ -1,16 +1,19 @@
 module github.com/containers/storage/tests/tools
 
-go 1.18
+go 1.23.0
 
 require (
 	github.com/cpuguy83/go-md2man v1.0.10
```
mtrmac, line 6: BTW this version is also 6 years old, but I’ll leave that update to someone who wants to investigate differences in the generated `nroff` syntax.

mtrmac:
@giuseppe PTAL, this should help with failures like #2341 in the future.

mtrmac:
(As a demo, this PR is intentionally behind `main`; it might be necessary to rebase it later…)

giuseppe (approved):
LGTM

mtrmac:
… marked as “ready for review” now.

Luap99 (approved):
/lgtm

PR #2343, opened 2025-06-05
[WIP] chunked: allow conversion without zstd compression

giuseppe:
This commit introduces the capability to convert tar layers to the zstd:chunked format, without performing any zstd compression.

This allows using zstd:chunked without the cost of compression and decompression.

Closes: https://issues.redhat.com/browse/RUN-3056


pkg/chunked/internal/minimal/compression.go
```diff
@@ -244,7 +244,39 @@ func WriteZstdChunkedManifest(dest io.Writer, outMetadata map[string]string, off
...
+
+func (n *noCompression) Flush() error {
+	return nil
+}
+
+func (n *noCompression) Reset(dest io.Writer) {
+	n.dest = dest
+}
+
+func NoCompression(dest io.Writer) (ZstdWriter, error) {
```
mtrmac, line 275: Could this be moved to `chunked/compressor`? `internal/minimal` is used even by `chunked/toc`, so including any more code here is not ideal, and there is already an interface . (OTOH, admittedly, for c/image, it would be included either way…)

… Actually can’t most of the writer code here be moved into `chunked/compressor` wholesale, leaving `minimal` to be mostly type definitions? The readers in `pkg/chunked`, separate from the writers, anyway.

pkg/chunked/internal/minimal/compression.go
```diff
@@ -244,7 +244,39 @@ func WriteZstdChunkedManifest(dest io.Writer, outMetadata map[string]string, off
...
+
+type noCompression struct {
+	dest io.Writer
+}
+
+func (n *noCompression) Write(p []byte) (int, error) {
+	return n.dest.Write(p)
+}
+
+func (n *noCompression) Close() error {
```
mtrmac, line 263: The way `restartCompression` uses `Close()`+`Reset` is not really fulfilling the `io.WriteCloser` API.

Maybe make this interface a bit higher-level, turning some variant of `restartCompression` into a method on the interface?

mtrmac (commented):
Just an extremely brief skim…

This might well make sense as an intermediate step. For _full_ performance, wouldn't we want to entirely avoid `copyAllBlobToFile` + `convertTarToZstdChunked`?

“Just” stream the incoming tar file as it comes; for every regular file, store it somewhere inside the staging destination, checksum it, and either hardlink/reflink it with a pre-existing one, or move it in place if it is entirely new. (Yes, that would be a _third_ implementation of setting file metadata, along with `pkg/archive` (maybe there are even more??), and we’d really want to consolidate them.)

Alternatively, after #2325 , wouldn’t it be _fairly easy_ to add chunked-like pre-staging of layer contents without holding locks to the regular pull path? And if we did that, could we teach `pkg/archive` to build the composefs structures ~directly? I think that (“just”, again) requires computing `MeasureVerity`, ordinary digests for `RegularFilePathForValidatedDigest`, and the tar header metadata.

Unlike truly-chunked pulls, on the “convert” path (and, also, for https://github.com/containers/image/issues/2792 ), there’s not that much benefit in trying not to read files, so the hole / roll-sum data computation can, I think (?) be entirely avoided; we _only_ want to match entire files, by digest, for reflinks/hardlinks.

I know, “just”…

giuseppe:
> This might well make sense as an intermediate step. For _full_ performance, wouldn't we want to entirely avoid `copyAllBlobToFile` + `convertTarToZstdChunked`?

I've done some benchmarking and the cost of writing the tarball is minimal compared to computing the checksum of each file/chunk.

It is still useful to compute the chunks because we fill the cache with this information and future downloads can use these chunks.

I agree it would look cleaner to work directly on the tar stream, but the implementation cost is much higher because we will need to duplicate a lot of code paths.

pkg/chunked/internal/minimal/compression.go
```diff
@@ -20,6 +20,16 @@ import (
 	"github.com/vbatts/tar-split/archive/tar"
 )
 
+// ZstdWriter is an interface that wraps standard io.WriteCloser and Reset() to reuse the compressor with a new writer.
+type ZstdWriter interface {
+	io.WriteCloser
+	Reset(dest io.Writer)
+	Flush() error
```
mtrmac, line 27: (BTW `Flush` is only called immediately before `Close`, and both code comparison and the `Close` documentation _suggests_ that `Close` already is a superset — but I didn’t test that nor review the code differences carefully.)

pkg/chunked/storage_linux.go
```diff
@@ -948,7 +943,17 @@ func (c *chunkedDiffer) storeMissingFiles(streams chan io.ReadCloser, errs chan
...
-			compression, err := c.prepareCompressedStreamToFile(partCompression, part, &mf)
+			if part != nil {
+				limit := mf.CompressedSize
+				// If we are reading from a source file, use the uncompressed size to limit the reader, because
+				// the compressed size refers to the original layer stream.
+				if missingPart.OriginFile != nil && partCompression == fileTypeNoCompression {
+					limit = mf.UncompressedSize
+				}
+				c.rawReader = io.LimitReader(part, limit)
+			}
```
mtrmac, line 954: I’m _generally_ unhappy with the complexity. I feel that there must be some way to express this simpler.

I think that conflating `compressedFileType` to mean _both_ “format of the file” and “origin+format of this chunk” is confusing, although it seems somewhat convenient for this PR.

We have the `if c.rawReader != nil` code to discard remaining data even for `OriginFile` chunks, where we don’t actually need to do that.

This part that duplicates the `OriginFile` condition could certainly be done without duplicating the condition, e.g. setting an extra `readingFromAlLocalFile` boolean in the `case missingPart.OriginFile != nil` section.

---

As a possible starting point, is there any need to split the `prepareCompressedStreamToFile` / `appendCompressedStreamToFile` code into two parts, and to interleave the `openDestinationFile` _in the middle_? I am quite possibly missing something, but I don’t see why that is necessary. And if it is not, combining the two `…StreamToFile` into a single function with a single `switch` would be simpler.

I’m afraid I can’t predict the full scope how the refactoring could be done, I can only see one or two steps ahead.

mtrmac:
> I've done some benchmarking and the cost of writing the tarball is minimal compared to computing the checksum of each file/chunk.

Thanks, that’s interesting.

> It is still useful to compute the chunks because we fill the cache with this information and future downloads can use these chunks.

Wait … we are writing this not-really-zstd data to durable storage as `bigDataKey`? … Ultimately I don’t see that it breaks anything, but I didn’t realize that’s one of the consequences.

pkg/chunked/internal/minimal/compression.go
```diff
@@ -20,6 +20,16 @@ import (
 	"github.com/vbatts/tar-split/archive/tar"
 )
 
+// ZstdWriter is an interface that wraps standard io.WriteCloser and Reset() to reuse the compressor with a new writer.
+type ZstdWriter interface {
```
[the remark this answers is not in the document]
mtrmac: This is all `internal` anyway… the `WithLevel` is more confusing.

pkg/chunked/internal/minimal/compression.go
giuseppe: I'll drop the calls to Flush and simplify the interface

pkg/chunked/storage_linux.go
```diff
@@ -666,20 +665,17 @@ func (o *originFile) OpenFile() (io.ReadCloser, error) {
 	return srcFile, nil
 }
 
-func (c *chunkedDiffer) prepareCompressedStreamToFile(partCompression compressedFileType, from io.Reader, mf *missingFileChunk) (compressedFileType, error) {
+func (c *chunkedDiffer) prepareCompressedStreamToFile(partCompression compressedFileType, mf *missingFileChunk) (compressedFileType, error) {
 	switch {
 	case partCompression == fileTypeHole:
 		// The entire part is a hole.  Do not need to read from a file.
-		c.rawReader = nil
```
mtrmac, line 673: I don’t think this is correct as is — if we set `c.rawReader` in a previous iteration, it will remain set, and in this iteration we will proceed to the trailing `io.CopyBuffer(io.Discard, c.rawReader, c.copyBuffer)` code again.

Admittedly, that seems to have ~no ill effects, because in the previous iteration of that code, we read `c.rawReader` until EOF, but it is imprecise and a risk for future refactoring.

`c.rawReader` is really private state of the `for _, mf := range missingPart.Chunks` loop, and I think it could now be made a local variable of that loop.

mtrmac (commented):
I’d really like `storeMissingFiles` to be simplified further, but I guess with the `c.rawReader = io.LimitReader` consolidation this PR, is, on net, not making things worse.

pkg/chunked/storage_linux.go
giuseppe: yes let's not risk.  I've added it again

mtrmac (commented):
LGTM, with a bit of heavy heart about the complexity of `storeMissingFiles`.

PR #2346, opened 2025-06-06
scope contiguity checks for idmap to mountProgram presence

giuseppe:
move the check for a contiguous mapping to overlay when a mount program is specified, since this is required only for fuse-overlayfs, not native idmapped mounts.

Closes: https://issues.redhat.com/browse/RHEL-94967
Closes: https://github.com/containers/storage/issues/2345

### DIFF
--- a/drivers/driver.go
+++ b/drivers/driver.go
@@ -193,8 +193,9 @@ type LayerIDMapUpdater interface {
 	UpdateLayerIDMap(id string, toContainer, toHost *idtools.IDMappings, mountLabel string) error
 
 	// SupportsShifting tells whether the driver support shifting of the UIDs/GIDs in a
-	// image and it is not required to Chown the files when running in an user namespace.
-	SupportsShifting() bool
+	// image to the provided mapping and it is not required to Chown the files when running in
+	// an user namespace.
+	SupportsShifting(uidmap, gidmap []idtools.IDMap) bool
 }
 
 // Driver is the interface for layered/snapshot file system drivers.
--- a/drivers/overlay/overlay.go
+++ b/drivers/overlay/overlay.go
@@ -2555,6 +2555,12 @@ func (d *Driver) supportsIDmappedMounts() bool {
 
 // SupportsShifting tells whether the driver support shifting of the UIDs/GIDs to the provided mapping in an userNS
 func (d *Driver) SupportsShifting(uidmap, gidmap []idtools.IDMap) bool {
+	if !idtools.IsContiguous(uidmap) {
+		return false
+	}
+	if !idtools.IsContiguous(gidmap) {
+		return false
+	}
 	if os.Getenv("_CONTAINERS_OVERLAY_DISABLE_IDMAP") == "yes" {
 		return false
 	}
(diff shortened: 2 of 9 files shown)


drivers/overlay/overlay.go
```diff
@@ -2554,11 +2554,17 @@ func (d *Driver) supportsIDmappedMounts() bool {
...
 	if os.Getenv("_CONTAINERS_OVERLAY_DISABLE_IDMAP") == "yes" {
 		return false
 	}
 	if d.options.mountProgram != "" {
+		if uidmap != nil && !idtools.IsContiguous(uidmap) {
+			return false
+		}
+		if gidmap != nil && !idtools.IsContiguous(gidmap) {
+			return false
+		}
```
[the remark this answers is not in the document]
giuseppe: thanks, fixed both

drivers/overlay/overlay_test.go
```diff
@@ -52,6 +53,47 @@ func TestContainersOverlayXattr(t *testing.T) {
...
+		{
+			ContainerID: 2,
+			HostID:      2,
+			Size:        1,
+		},
+	}
+
+	t.Run("no mount program", func(t *testing.T) {
+		driver := graphtest.GetDriver(t, driverName)
+		defer graphtest.PutDriver(t)
```
mtrmac, line 79: `GetDriver` uses `t.Cleanup()` to ensure `PutDriver` happens automatically.

(2 instances)

drivers/driver.go
```diff
@@ -194,7 +194,7 @@ type LayerIDMapUpdater interface {
 
 	// SupportsShifting tells whether the driver support shifting of the UIDs/GIDs in a
 	// image and it is not required to Chown the files when running in an user namespace.
```
mtrmac, line 196:
```suggestion
	// image … to the provided mapping, and it is not required to Chown the files when running in an user namespace.
```
? If so, please update all such comments.

drivers/overlay/overlay.go
```diff
@@ -2554,11 +2554,18 @@ func (d *Driver) supportsIDmappedMounts() bool {
 }
 
 // SupportsShifting tells whether the driver support shifting of the UIDs/GIDs in an userNS
-func (d *Driver) SupportsShifting() bool {
+func (d *Driver) SupportsShifting(uidmap, gidmap []idtools.IDMap) bool {
 	if os.Getenv("_CONTAINERS_OVERLAY_DISABLE_IDMAP") == "yes" {
 		return false
 	}
 	if d.options.mountProgram != "" {
+		// fuse-overlayfs needs contiguous mappings to avoid https://github.com/containers/podman/issues/10272
```
mtrmac, line 2562: I have absolutely no idea how that record relates to `fuse-overlays` or contiguous mappings.

The comment there that is closest to an explanation talks about whether the invoking users’ host UID does, or does not, map into the container’s namespace …

mtrmac (commented):
I 100% don’t understand the logic, but I also 100% support making the logic consistent in both locations.

So… ACK?

mtrmac:
/lgtm

Thanks!

PR #2347, opened 2025-06-24
fix: Ignore ENOENT error when calling unmount

ninja-quokka:
If the file or directoy doesn't exist then we can assume it is unmounted.

JIRA: https://issues.redhat.com/browse/RHEL-99488

### DIFF
--- a/pkg/mount/unmount_unix.go
+++ b/pkg/mount/unmount_unix.go
@@ -21,6 +21,9 @@ func unmount(target string, flags int) error {
 			// can be returned if flags are invalid, so this code
 			// assumes that the flags value is always correct.
 			return nil
+		case unix.ENOENT:
+			// Ignore "no such file or directory" error.
+			return nil
 		}
 		break
 	}


ninja-quokka:
Maybe I should handle this sooner, for example checking if `d.home` exists before calling `mount.Unmount()` https://github.com/containers/storage/blob/e1679c1222b3250e8eb512539ebd2e1241c0f929/drivers/overlay/overlay.go#L859-L865

mtrmac:
Looking a _tiny bit_ further, `libpod/Runtime.Reset` shuts down c/storage at least twice: Once via direct `r.store.Shutdown`, once via `r.Shutdown`→`r.libimageRuntime.Shutdown`.

I’m not saying that _definitely_ means that c/storage is not ultimately the best place to fix this — `reset` is rather a special case and fundamentally awkward to an extent — but it does, I think, strongly suggest that Podman’s idea of the states and lifetimes of `Runtime` and its components is not precisely modeled / thought out.

mtrmac:
> At that point in time, the store hasn't been initialized and the /var/lib/containers/storage/ directory has not been created yet.

Is the situation really that none of the directories were created _yet_? or is it that `podman system reset` _initializes everything_ (e.g. the `registry.ContainerEngine().Info` call, or `PodmanConfig.IsReset` is certainly suggestive), shuts down and deletes everything, _and then_ tries to shut things down the second time — all within a single process?

---

> What do you think about the following ideas?

I’m afraid this might be infuriating, but at this point I don’t quite care how the message is suppressed, or this particular unmount is made conditional; if we decide that that is the right thing to do, we can figure out a way. (If we know nothing else, adding targeted conditionals to `.Cleanup()` — in all the 6 graph drivers — would be least disruptive.)

I’m much more concerned that, from a quick and very ignorant look at the reset code, the `libpod.Runtime.Reset` code seems to be confused about its internal state, cleaning up things after deleting them, and the like. And I think that if that is cleaned up to be precise, it is quite possible that no c/storage changes will be necessary.

That should probably be discussed in the Podman issue with the Podman experts — _if_ my very cursory reading of the code is confirmed to be approximately correct.

ninja-quokka:
Thanks for the advise @mtrmac, I'll close this PR as we won't go this direction regardless if the root issue.

PR #2354, opened 2025-07-20
homedir: check ConfigHome is writable by current user

HastD:
This replaces the check in `GetConfigHome` that the user config directory is owned by the current user with a check that the directory is writable by the current user. Usually these go together, but in the case that they differ, presumably the important thing is whether the current user can read from and write to the directory, regardless of whether they own it.

(I also updated the URL in the nearby comment; the old URL gives a 404 error due to changes on the freedesktop.org site.)

This isn't just a theoretical issue; there is a change [under consideration by secureblue](https://github.com/secureblue/secureblue/pull/1190) that would lock down permissions on certain directories (including `~/.config`) as a security measure, and this is implemented in part by changing the ownership of these directories to `root:user` with permission mode 770. However, currently this causes podman to break, as `GetConfigHome` fails if the user doesn't own their own config directory (even if they still have full read-write-execute access to it).

### DIFF
--- a/pkg/homedir/homedir_unix.go
+++ b/pkg/homedir/homedir_unix.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/containers/storage/pkg/unshare"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 // Key returns the env var name for the user's home dir based on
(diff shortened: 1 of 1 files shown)


mtrmac (changes requested):
No, this check is very intentional. It is to protect against users running things like `su podman`, inheriting `XDG_CONFIG_HOME` from the unprivileged user to a root-owned process, and root overwriting the unprivileged accounts’ data (and potentially making it unreadable or even unremovable).

(That proposal in https://github.com/secureblue/secureblue/pull/1190 … I’m not now going to study the whole software, but at a glance, I don’t understand what is the security feature supposed to be — where is the privilege boundary?? If, as in here, users can (maybe not always but _frequently_) set `XDG_CONFIG_HOME` to any other path, what good does restricting writability of that one do? At best, it seems to require a very unusually modified Linux system… and then, I suppose, that can extend to patching this function.)

PR #2357, opened 2025-07-25
pkg/archive.tarWriter.addFile(): don't change hardlink targets

nalind:
When examining a file that we're adding to an archive to see if it's hard linked to another file that's already been written, or to one that might be written later, only set an entry in the inode->name map that points to the entry we're writing for the first file.  This keeps us from writing tar headers with Typeflag==TypeLink and a Linkname that points to another entry with Typeflag==TypeLink in cases where the inode has more than two names.

Fixes https://github.com/containers/buildah/issues/6297.

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -697,7 +697,10 @@ func (ta *tarWriter) addFile(headers *addFileData) error {
 	}
 
 	if !headers.fi.IsDir() && hasHardlinks(headers.fi) {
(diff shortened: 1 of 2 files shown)


nalind:
This reverts what I think was an unintentional part of #2289.

nalind:
> (Should this have a test? I’d merge immediately with a unit test.)

Extended a test to include the case from the original report.

PR #2371, opened 2025-08-27
Auto-close new issues.

jankaluza:


cgwalters:
IMO there's a much simpler way to handle this: Just [archive the repo](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories) - then no one can create issues in the first place.

This *used* to be a one way operation which meant it was a Big Deal, but now that's reversible there's basically zero downside to using it.

cgwalters:
> No we cannot do that, the old branches will stay active here as people may need backports for older releases and of course making the switch to the new repo seems totally unrealistic so we have decided to not archive.

OK, that makes sense except then...don't we want the ability to track issues for older branches and so we shouldn't auto-close issues?

mtrmac (commented):
/lgtm

